PR #255, opened 2015-01-08
Show all related committees if all fields are asked for

LindsayYoung:
the designation filtering was too tight so joint committees etc were not getting through

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -723,7 +723,7 @@ class Candidate(object):
         'party_fields': 'party_affiliation_desc,party_affiliation',
         'status_fields': 'election_yr,cand_status,ici_code',
         'properties_fields': 'cand_nm',
-        'cmte_fields': "'P'",
+        'cmte_fields': 'P',
     }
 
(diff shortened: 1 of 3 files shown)


webservices/rest.py
```diff
@@ -732,7 +732,15 @@ class Candidate(object):
 
     def query_text(self, show_fields):
         if show_fields['cmte_fields'] != '':
-            com_query = "/dimlinkages?cmte_dsgn={%s}{*, /dimcmte{/dimcmteproperties{cmte_nm}}}  :as affiliated_committees," % (show_fields['cmte_fields'])
+            cmte = show_fields['cmte_fields']
+            if cmte == 'P':
+                cmte_des_qry = "?cmte_dsgn={'P'}"
```
catherinedevlin: {'P"} - for future reference, the curly braces aren't necessary when you're comparing to a single value; their only purpose is to let you do an OR-type comparison to multiple values.  Still, I don't think they do any harm so there's no need to remove them.

PR #288, opened 2015-01-20
Part 2 Splitting Candidate

cmc333333:
Builds on #287 by also splitting out the two base resource types (SingleResource and Search). Also moves db initialization to a separate file and uses dynamic dispatch (i.e. method override) to determine format.

### DIFF
--- a/webservices/db.py
+++ b/webservices/db.py
@@ -1,0 +1,28 @@
+import os
+
+from flask import g
+from htsql import HTSQL
+import sqlalchemy as sa
+
+
+def sqla_conn_string():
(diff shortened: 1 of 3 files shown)


cmc333333:
Note that, because this "reaches back" across packages, it refers to the "webservices" module. This means that your `PYTHONPATH` will need to include this application's root. Generally,

```
export PYTHONPATH=$PYTHONPATH:.
```

works, if you're used to running the application from the root. Let me know if there's another strategy you'd prefer.

cmc333333:
Closing as this needs to be rejiggered to avoid pythonpath stuffs

PR #290, opened 2015-01-26
Revert 287

cmc333333:
Temporarily revert #287 until path/module issues are resolved


LindsayYoung:
Thanks for jumping on this!

PR #292, opened 2015-01-26
Split out candidate, take 2

cmc333333:
Same as #287, but doesn't use a relative import.


LindsayYoung:
I think this will work nicely. Thanks @cmc333333

PR #294, opened 2015-01-26
Separate out base resources

cmc333333:
* Places a `run_api.py` file in the repo root. This allows `webservices` to be a package (as the top-level module is the repo root, rather than within `webservices`). This is actually needed for this PR, but will be with the next one (which separates candidate resources)
* Split out database connection info
* Split out SingleResource and Searchable
* Format is now determined by the resource (rather than `assign_formatting`)
* Flake8 for new files

### DIFF
--- a/run_api.py
+++ b/run_api.py
@@ -1,0 +1,14 @@
+import doctest
+import os
+import sys
+
+
+from webservices.rest import app
+
+
+if __name__ == '__main__':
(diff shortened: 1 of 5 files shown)


LindsayYoung:
Just to reiterate, this changes the startup script to `python run_api.py` @theresaanna @catherinedevlin @micahsaul @noahmanger 

I updated the restart script on the server.

Thanks again @cmc333333

Issue #291, opened 2015-01-26
Broaden search capabilites


LindsayYoung:
Ideally, these things should work in the search:
- zipcode  (returns candidates that represent that zipcode)
- candidate_id 
- committee_id
(I haven't tried an id search but I think it would be easy to deal with three cases at once)

rjmajma:
The purpose of this is to allow users to search for elections, candidates, and committees that affect them locally. Our goal is to provide users with the most flexibility and ease of search for things that matter to them.

jmcarp:
Looks like this isn't relevant anymore; closing.

PR #298, opened 2015-01-27
Ids in query

LindsayYoung:
Will give candidate or committee if someone searches by id.

@theresaanna I wanted to make sure the 'Name' fields is populating `q=` and not using name so this provides the right functionality.


LindsayYoung:
This also makes sure you can pas in multiple fields at a time.

LindsayYoung:
I am planning to merge @cmc333333 outstanding PRs and then remerge this.

LindsayYoung:
this tackles 2/3s of #291

cmc333333:
Side note: an alternative to commas is to repeat a query param, e.g.
`?myvar=val1&myvar=val2&myvar=val3` rather than `?myvar=val1,val2,val3`. Obviously a preference thing.

webservices/rest.py
```diff
@@ -639,6 +648,18 @@ def get(self):
             args['year'] = default_year()
         year = args['year']
 
+        # Checking for special cases
+        if 'q' in args and args['q'] is not None:
+            # use committee id
```
catherinedevlin: It wouldn't be hard to add committee_id and candidate_id to the fulltext search index, so they would automatically be found by the regular q= queries, and I think that would accomplish the same thing - would that be a better solution?

(sorry, I feel like I would have been able to suggest this before you did the work if I'd been paying better attention)

PR #331, opened 2015-01-28
``/name?q=`` endpoint returns one result per cand_id or cmte_id

### LINKED ISSUE #328: Avoid duplicate results in /name REST endpoint (for typeahead)
Typeahead is less useful when one candidate or committee appears multiple times in the search results.

For each candidate or committee, ensure that only one typeahead value is returned.  Most recent name provided should be used.

catherinedevlin:
last loaded wins in case of conflicts
Closes #328 

Is presently in place on production db, can view at https://fec.18f.us/rest/name?q=boehner


catherinedevlin:
no tests!

PR #348, opened 2015-02-03
Creating indexes for coverage dates on fact* tables; cleaning up obsolete Blaze references

catherinedevlin:


arowla:
Looks good. Can we add to this PR a comment at the top of each file summarizing where and when it should be run? Like "this runs on box X to do Y before Z can happen," or something like that.

rjmajma:
Hey @catherinedevlin, just checking status on @arowla's comment above.

catherinedevlin:
Did these for the files in data/synch, but not yet for those in data/ddl...

arowla:
Ok, just going ahead with the merge.

PR #349, opened 2015-02-03
Silly hack for /eregs/ paths

catherinedevlin:


micahsaul:
We're horrible people. We'll revisit this when we know what's actually happening with eregs for FEC.

Issue #337, opened 2015-01-30
keep api results consistent when there are no results returned


LindsayYoung:
Needs the page information etc. Right now if it doesn't find a query it just returns an empty list.

PR #353, opened 2015-02-04
Totals use fewer tables when possible

LindsayYoung:
This make the call a bit faster by automatically reducing the amount of tables if the call does not need them all. I also reduced the amount of fields that show up automatically.

### DIFF
--- a/webservices/db.py
+++ b/webservices/db.py
@@ -1,7 +1,9 @@
 import os
+from datetime import datetime
+from psycopg2._range import DateTimeRange
 
 from flask import g
-from htsql import HTSQL
+import htsql
 import sqlalchemy as sa
 
 
@@ -24,7 +26,7 @@ def db_conn():
 def htsql_conn():
     if not hasattr(g, '_htsql_conn'):
         htsql_conn_string = sqla_conn_string().replace('postgresql', 'pgsql')
-        g._htsql_conn = HTSQL(htsql_conn_string)
+        g._htsql_conn = htsql.HTSQL(htsql_conn_string)
     return g._htsql_conn
 
 def as_dicts(data):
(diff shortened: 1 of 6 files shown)


webservices/db.py
```diff
@@ -26,3 +26,18 @@ def htsql_conn():
         htsql_conn_string = sqla_conn_string().replace('postgresql', 'pgsql')
         g._htsql_conn = HTSQL(htsql_conn_string)
     return g._htsql_conn
+
+def as_dicts(data):
```
cmc333333: If you're moving this here (seems like it makes more sense), remember to remove the version in `webservices/resources.py`

webservices/totals/models.py
```diff
@@ -48,13 +40,48 @@ def query_text(self, show_fields):
...
+                    ?cmte_id={'%s'}"%(
+                        "','".join(ids)
+                )
+            data = htsql_conn().produce(pre_qry)
+            types = []
+            # looking for the codes in the htsql string
+            for d in str(data):
+                if d.isalpha() == True:
+                    if d not in types:
+                        types.append(d)
```
cmc333333: Don't know if you'd like this one-liner better:

```python
types = set(d for d in str(data) if d.isalpha())
```

cmc333333:
I don't understand the underlying data well enough to validate, but it looks like this divides what type of query it makes based on the request, which no doubt removes a lot of unneeded work. :+1:

webservices/totals/models.py
LindsayYoung: Good point. I will do that.

webservices/db.py
LindsayYoung: Good catch, I am doing that now!

LindsayYoung:
@cmc333333 thanks for the suggestions. They should all be incorporated now.

I also fixed #337 while I was in there and added a test. The API will return a sane format when no search results are found.

LindsayYoung:
just merged this in to the older pr about totals

PR #355, opened 2015-02-04
Enable CORS and set utf-8

### LINKED ISSUE #267: format API headers
Add the heading requirements outlined in- https://github.com/18F/api-standards
create proper headers including:
- add UTF-8 to content type
- enable CORS

It would be good to get this done before API testing

micahsaul:
Pretty sure this is how to do this. Can someone who is more comfortable in nginx config land verify? Fixes #267.

### DIFF
--- a/webservices/setup/nginx_conf_default
+++ b/webservices/setup/nginx_conf_default
@@ -18,6 +18,8 @@ server {
 			auth_basic "Restricted";
 			auth_basic_user_file /etc/nginx/.htpasswd;
       add_header 'Access-Control-Allow-Origin' '*';
+      add_header 'Access-Control-Allow-Headers' "X-Requested-With";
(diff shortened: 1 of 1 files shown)


konklone:
@micahsaul I believe you'll need more than this. Here's [what I did](https://github.com/konklone/cdns-to-https/blob/gh-pages/nginx/cdn.cors) for a recent project:

```
add_header Access-Control-Allow-Headers "X-Requested-With";
add_header Access-Control-Allow-Methods "GET, HEAD, OPTIONS, POST";
add_header Access-Control-Allow-Origin "*";
```

You may wish to remove `POST` from the mix, if this is read-only.

micahsaul:
Thanks @konklone, I've added those as well (removing POST as this is a read-only API).

PR #360, opened 2015-02-06
Initial candidate list (`/candidate`) endpoint refactor

arowla:
NOTE: this will be an open pull request, not ready to merge to master.

- Created `ofec_candidates_vw` to summarize candidate metadata. This
  view exists in the live data, but you will need to run the following
  file for any local databases.
	- new file:   data/sql_updates/create_candidates_vw.sql
- Created `Candidate` model to access and filter on the view
	- new file:   webservices/resources/candidates.py
- Created common module to hold utility functions and initialize
  database access
	- new file:   webservices/common/__init__.py
	- new file:   webservices/common/models.py
	- new file:   webservices/common/util.py
- resources.py has been renamed to not clash with the new resources/
  directory, but can still be imported from in the old way,
  as resources/__init__.py does an `import *` from resources_old
	- renamed:    webservices/resources.py -> webservices/resources_old.py
	- new file:   webservices/resources/__init__.py
- Using new CandidatesList resource, adding some sqlalchemy support
	- modified:   webservices/rest.py
- Many tests WILL NOT PASS, pending discussion of some API interface
  changes
	- modified:   webservices/test_api.py
- Changed imports
	- modified:   webservices/candidates/resources.py
	- modified:   webservices/totals/resources.py

PR #365, opened 2015-02-06
Candidate tests

cmc333333:
Adds a regression test for querying all fields on a single candidate + comments on where these fields are coming from. This will need to be rewritten for #360, but may help debug that, too.

Also fixed a bug with "other_names".


arowla:
Thanks!

PR #413, opened 2015-02-18
Changed candidate list view election years behavior

arowla:
- `active_through` now shows most recent year
- `election_years` will show all election years candidate participated
  regardless of filters
- ofec_candidates_vw will now only return one line per candidate
- all status details will be according to the most recent record

### DIFF
--- a/data/sql_updates/create_candidates_vw.sql
+++ b/data/sql_updates/create_candidates_vw.sql
@@ -1,33 +1,38 @@
 drop view if exists ofec_candidates_vw;
 create view ofec_candidates_vw as
-select distinct
+select
     dimcand.cand_sk as candidate_key,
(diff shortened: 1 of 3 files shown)


LindsayYoung:
That is much faster @arowla 

Also, @noahmanger @theresaanna I think the naming on the browse elections page should also be "active through" instead of "election year" It will make more sense once we add proper date filtering on the front end. I can explain more if you are interested.

theresaanna:
@LindsayYoung @noahmanger Do you want to talk about the "active through" thing and decide if we should do it? If so, put an issue in the backlog?

LindsayYoung:
Yes I would like to talk about it. It shouldn't take too long.

PR #414, opened 2015-02-19
Fixed committee date handling for single years -- I think

arowla:
I think this may fix your date handling problem. Simplified it by extracting the year for comparison from the datetimes, and that made it clear that the operators were switched.


LindsayYoung:
Thank you!

PR #419, opened 2015-02-20
Cand lookup committee

### LINKED ISSUE #418: some committees are not displaying in the new /committee branch
Most committees work but some are not showing up. I am not sure why.

These committees don't work on refactored /committees but work in /committees/<id>
C00484188, C00000422

LindsayYoung:
Adds `candidate/<id>/committees` 

You can also do `candidate/<id>/committees?designation=P` to only see primary committees or `P,J` to see primary and joint etc.  

This also fixes #418 sql join error.


webservices/resources/committees.py
```diff
@@ -127,6 +127,96 @@ def get_committees(self, args, page_num, per_page):
...
+                'count': count,
+                'pages': int(count / per_page),
+            },
+            'results': committees
+        }
+
+
+        return data
+
+    def get_committees(self, args, page_num, per_page, **kwargs):
```
arowla: Seems like the `get_committees` method from `CommitteeList` could be pretty readily modified to take a candidate_id and give back the results we want, eliminating the need for this method (and perhaps even this whole Resource, but I'm not sure on that yet...)

LindsayYoung: I thought that was how you wanted it. I like having it in one place. I can make it work that way, taking one candidate_id.

theresaanna:
@LindsayYoung Does it make sense for me to work on implementing this on the front end today? We could then merge this PR and the front end one at the same time.

LindsayYoung:
I think this would be good to implement. The final implementation was simplified so that 
`/committee?candidate_id=<id>` takes one candidate id and returns all the relevant committee information. You can also add filters to just see joint and principal committees. Though, would this be a good time to change the name to `/committees` since you are making changes anyway?

theresaanna:
I don't think its particularly easier now to change the name to /committees. If I'm going to implement it today, I think we should keep it as is and tackle that one in a subsequent changeset.

PR #424, opened 2015-02-25
Got committee/candidate relationships working

arowla:
- Had to move models into a common file so that they could reference
  each other without it being an error-prone PITA
- Simplified querying
- Altered dynamic filters to work with the join table condition
- Modified one of the underlying views to make sure it is not dropping rows (especially for much older candidate data that doesn't have as much relational data associated with it)


LindsayYoung:
These all sound like brilliant things!

PR #449, opened 2015-03-05
Reports refactor

arowla:
Outstanding (known) issues:

* pagination is not working properly, only showing two results per page
* a few tests are failing, but some failures seem unrelated to this change


arowla:
The pagination issue is fixed as of 32bdf89. We need to make a note to check for this bug on all other endpoints.

webservices/resources/reports.py
```diff
@@ -0,0 +1,592 @@
...
+    def get_reports(self, committee_id, reports_class, args, page_num, per_page):
+
+        reports = reports_class.query.filter_by(committee_id=committee_id)
+
+        if args['cycle'] != '*':
+            reports = reports.filter(reports_class.cycle.in_(args['cycle'].split(',')))
+
+        reports = reports.order_by(reports_class.cycle)
+        count = reports.count()
+        return count, reports.order_by(desc(reports_class.report_year), desc(reports_class.report_type)).paginate(page_num, per_page, True).items
```
LindsayYoung: I think we want to order by the coverage_end date, since that is more precise.

LindsayYoung:
@theresaanna @noahmanger This is the report refactor, which will break reports but now we have coverage dates!!!!!!!!! Isn't @arowla the best?!

### Changes
 - reports will be found at `committee/<committee_id>/reports`
 - totals will be at `committee/<committee_id>/totals` 
 - reports have `coverage_start_date` and `coverage_end_date` and are sorted most recent first.

PR #451, opened 2015-03-05
Committee cand cleanup

LindsayYoung:


data/sql_updates/create_linkage_names_vw.sql
```diff
@@ -5,7 +5,8 @@ select
     dimlinkages.cand_sk as candidate_key,
     dimlinkages.cmte_sk as committee_key,
     dimlinkages.cand_id as candidate_id,
-    dimlinkages.cand_election_yr as active_through,
+    max(dimlinkages.cand_election_yr) as active_through,
```
arowla: I don't think this is going to work the way you want. With a `group by` on the table's primary key, this will mean that `max(cand_election_yr)` will always equal `cand_election_year`, since grouping by the primary key won't actually be aggregating any rows.

webservices/tests/candidate_tests.py
```diff
@@ -1,96 +1,154 @@
...
+        self.assertResultsEqual(result['name'], 'MORAN, JAMES P. JR.')
+        #address
+        self.assertEqual(result['address_city'], 'ALEXANDRIA')
+        self.assertEqual(result['address_state'], 'VA')
+        self.assertEqual(result['address_street_1'], '311 NORTH WASHINGTON STREET')
+        self.assertEqual(result['address_street_2'], 'SUITE 200L')
+        self.assertEqual(result['address_zip'], '22314')
+        self.assertEqual(16, len(result['committees']))
+        # office
+        self.assertResultsEqual(result['office'], 'H')
```
arowla: `assertResultsEqual` should be used on the full dict-- not broken out line by line. It exists for that reason: to make testing complex data structures more clear and simple. If there's a problem with the test as it was, let's fix it in the format it was in.

LindsayYoung: The page is much flatter so there can't be tests on the "address" sub section in the same way. Should we just test the whole page?

arowla: Let me dig in here a bit and see what we can do. In the version of `assertResultsEqual` that I originally wrote for 18F/hourglass, it could test a subset of keys in the dict, IIRC. Not sure the version that @cmc333333 ported over here does that or not, but we can make it if it doesn't.

arowla: I've worked on this a bit, and I need to port some more assert features over from Hourglass, but since the sprint is ending, let's go ahead and get this merged. We need to make a separate pass on the tests. 

And I missed answering your question last time, @LindsayYoung ... yes, basically, we should just test the entire JSON structure as a whole, unless we are doing something specific. It's way more straightforward that way.

data/sql_updates/create_linkage_names_vw.sql
```diff
@@ -40,4 +41,9 @@ select
     dimlinkages.expire_date as expire_date,
     (select cand_nm from dimcandproperties where dimcandproperties.cand_sk = dimlinkages.cand_sk order by candproperties_sk desc limit 1) as candidate_name,
     (select cmte_nm from dimcmteproperties where dimcmteproperties.cmte_sk = dimlinkages.cmte_sk order by cmteproperties_sk desc limit 1) as committee_name
-from dimlinkages;
\ No newline at end of file
+from dimlinkages
+    group by
+        dimlinkages.linkages_sk
```
arowla: Don't group by a primary key. That is redundant with it being primary key.

PR #465, opened 2015-03-11
fix base url so there is no 500

LindsayYoung:
It was referencing old classes so it was throwing a 500


arowla:
Is there a reason to take this out, rather than updating with the new classes?

LindsayYoung:
I am fine with putting in the new classes, but we will have nice documentation so I would rather point people to that when it goes live. Right now, I just want it not to error out while we are debugging cloud foundry.

PR #483, opened 2015-03-18
Added a 'refresh_db' command to manage.py

arowla:
This is a simpler system than "migrations", but we don't need real migrations when all of our tables are (materialized) views, and updating them simply means dropping and recreating them, no matter the change. We also don't have views that rely on one another, at least at this point, so ordering is not a problem.

What this provides is a one-step command that runs all the SQL scripts, and can be easily called by either a dev or another process, to keep the database up to date.

```
python webservices/manage.py refresh_db
```


arowla:
Note that anyone who has trouble with imports after this should try creating a fresh virtualenv.

LindsayYoung:
Looks good to me!

PR #493, opened 2015-03-23
This fixes #447 Reid bug.

### LINKED ISSUE #447: Match up story candidates and priorities doc
Add cards for items in the priorities doc that don't exist and should, and reorder story candidates according to priorities doc.

LindsayYoung:
Just added future candidates to the default for better results. This will fix Harry Reid not showing up because while he has not ran for election in the last 4 years. Now, he will show up because he is on the 2016 ticket.

### DIFF
--- a/webservices/common/util.py
+++ b/webservices/common/util.py
@@ -18,7 +18,7 @@ def merge_dicts(x, y):
 
 def default_year():
     year = datetime.now().year
-    years = [str(y) for y in range(year, year-4, -1)]
+    years = [str(y) for y in range(year, year-4, -1)] + [str(year +1)]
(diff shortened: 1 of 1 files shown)


webservices/common/util.py
```diff
@@ -18,7 +18,7 @@ def merge_dicts(x, y):
 
 def default_year():
     year = datetime.now().year
-    years = [str(y) for y in range(year, year-4, -1)]
+    years = [str(y) for y in range(year, year-4, -1)] + [str(year +1)]
```
arowla: The `range()` call can just be changed to `range(year+1, year-4, -1)`

PR #501, opened 2015-03-25
Restore candidate detail

LindsayYoung:
This makes sure there is only one candidate returned by candidate/id

### DIFF
--- a/data/sql_updates/create_candidate_detail_vw.sql
+++ b/data/sql_updates/create_candidate_detail_vw.sql
@@ -1,54 +1,57 @@
 drop materialized view if exists ofec_candidate_detail_vw;
-create materialized view ofec_candidate_detail_vw as
+create view ofec_candidate_detail_vw as
 select
-    dimcandproperties.candproperties_sk as properties_key,
     dimcand.cand_sk as candidate_key,
     dimcand.cand_id as candidate_id,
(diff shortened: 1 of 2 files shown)


data/sql_updates/create_candidate_detail_vw.sql
```diff
@@ -1,54 +1,57 @@
...
-    dimcandproperties.cand_ici_desc as incumbent_challenge_full,
-    dimcandproperties.cand_ici_cd as incumbent_challenge,
-    dimcandproperties.cand_status_cd as candidate_status,
-    dimcandproperties.cand_status_desc as candidate_status_full,
-    dimcandproperties.form_tp as form_type
-from dimcandproperties
-    left join dimcand using (cand_sk)
-    left join dimcandstatusici using (cand_sk)
-    left join dimcandoffice co using (cand_sk)
+    -- This seems like an awful way to do this but the left join was not giving the right data
```
arowla: Let's not do this in an awful way. ;) These columns can be selected from our aliased table `cand_p_most_recent` to achieve the same result.

rjmajma:
Aww, Travis.

data/sql_updates/create_candidate_detail_vw.sql
LindsayYoung: I tried, but I could not get the join to work correctly.

Issue #460, opened 2015-03-09
Too many candidates in committee page


LindsayYoung:
Listing the candidate multiple times on the committee page.
The API is showing each linkage so a candidate is listed with a unique election year each time. It would probably be more desirable to just show the candidate once.
https://fec.18f.us/committees/C00213512

LindsayYoung:
In the models, the 'ofec_name_linkage_vw' needs an primary_key set to the linkage_key for the candidate endpoints. For the committee endpoints, using the linkage_key as the primary key produces too many results, I would like to use candidate_key as the primary key. I can't have the same table with 2 primary keys, even if I make them separate models.I have not been able to specify to select distinct by canidate_id in the committee endpoints. 

1) I feel like I could copy the table so that I could specify different foreign keys, but that would be hacky.
2) I could try to make the committee endpoints more responsive to time so that they work like the candidate endpoints. This was a less pressing concern because the important thing for committees that can change is the designation. Committee designation is not tracked over time by dimcmteproperties, it is linked by foreign key to dimcmtetpdsgn and only the most recent designation is available. Though it would be good to have the treasurers to be responsive over time, so it might be good to do anyway.

arowla:
This was terribly hard to fix. We are going to remove candidates from being nested in committee results instead.

LindsayYoung:
I am still seeing candidate name repeated-
committee/C00431445

arowla:
This will be taken care of in #511.

PR #516, opened 2015-03-26
[WIP] Remove embedded committees and candidates

LindsayYoung:
NOT READY TO MERGE

This will un-nest committees.


webservices/test_api.py
```diff
@@ -276,10 +276,6 @@ def test_committee_by_candidate(self):
         results =  self._results('http://localhost:5000/candidate/P60000247/committees?year=*')
         self.assertEquals(3, len(results))
 
-    def test_candidates_by_committee(self):
```
arowla: Why is this test being removed?

webservices/tests/candidate_tests.py
```diff
@@ -96,22 +95,11 @@ def test_fields(self):
...
     def test_extra_fields(self):
         response = self._results('/candidate/P80003338?year=2008')
-        self.assertIn('committees', response[0])
         self.assertIn('PO BOX 8102', response[0]['address_street_1'])
         self.assertIn('60680',response[0]['address_zip'])
         # testing for year sensitivity
         self.assertIn('O', response[0]['incumbent_challenge'])
 
-    def test_candidate_committes(self):
-        response = self._results('/candidate/P80003338?year=*')
```
arowla: A couple questions:
1. Is this territory well-covered by a test for `candidate/P80003338/committees`?
2. If not, can this test be morphed into one for that^^ without too many changes?

If it's already covered or it would be too much work, we can leave it for when we work on the other endpoint, in another issue.

webservices/tests/committee_tests.py
```diff
@@ -166,6 +166,7 @@ def test_committee_by_cand(self):
         results =  self._results('http://localhost:5000/candidate/P60000247/committees?year=*')
         self.assertEquals(3, len(results))
 
+    @unittest.skip('removing this functionality form this endpoint now')
```
arowla: I don't think this test should be skipped, it's for a different endpoint.

webservices/test_api.py
LindsayYoung: `committee/id/candidates` was the same function as `candidate/id` and since we were un-nesting I took it out.  I would think that in the next step, when we add the candidate-committee linkages, they will live at `committee/id/candidates`. I thought that was going to be different enough that it will require a new test.

webservices/resources/candidates.py
```diff
@@ -194,14 +176,12 @@ def get(self, **kwargs):
 
         return data
 
-    def get_candidate(self, args, page_num, per_page, candidate_id, committee_id):
```
arowla: We need to put back in support for passing in `committee_id` since this still powers `/committee/<id>/candidates`

LindsayYoung: Ok, this is were the confusion was coming from, I thought `/committee/<id>/candidates` was changing. I understand what you are saying now. Thanks for explaining.

arowla: It may end up changing somewhat, but we'll get there when we get there. :) Glad it's clear now.

webservices/tests/candidate_tests.py
LindsayYoung: I removed this test because it was checking the embedded committees that were in the candidate information.

arowla: OK, having looked into it a bit more, it looks like these should be covered by tests in committee_tests.py. Could you also take out the other `test_candidate_committees` from the same candidate_tests.py file? (It has a skip test decorator on it.)

LindsayYoung: sure

LindsayYoung:
dependency is refactoring web app noted in issue #527

LindsayYoung:
We will have to merge this or figure out another way to solve the "too many cooks" problem where the candidate names show up too many times on the committee pages. #460

arowla:
I think perhaps we can re-prioritize the front-end task for this sprint. I am going to try and clean up what I have so far.

rjmajma:
@arowla, how do you want to handle this? First half or back half of the sprint?

arowla:
I think I'd better work on it in the first half so I can put a bow on what
I did so far, and then either a) finish the rest, or b) pass it off.

On Wed, Apr 22, 2015 at 5:48 PM, Raphael Majma <notifications@github.com>
wrote:

> @arowla <https://github.com/arowla>, how do you want to handle this?
> First half or back half of the sprint?
>
> —
> <https://github.com/18F/openFEC/pull/516#issuecomment-95348050>.
>



-- 
Alison Rowland
202-317-0124

rjmajma:
:+1:

LindsayYoung:
replaced by #648

PR #523, opened 2015-03-31
Update search

### LINKED ISSUE #491: Add full text tables to the refresh db commands
Right now the tables are made in data/ddl/fulltext.sql those need to be in the db command.

LindsayYoung:
This narrows the full text search so that fields like address are not searchable. However, candidate_id and committee_ids should be searchable and are not for some reason. I tried a few different variations on the search implementation, but it is would not search the ids. I will file that as a bug for now. 

fixes #491


arowla:
Which ticket is the new bug report in? Also, the commit to narrow the search seems to be this one, which is not included in the PR f9d4b3c694a0d886, so I was confused for a bit about how this fixes that issue.

LindsayYoung:
This moves that script and which made the search more narrow.  Now, that table will be recreated locally by just running the management command. The id searches don't seem to have worked, which is noted in issue #524

arowla:
Ok thanks. Merging!

Issue #502, opened 2015-03-25
Factor pagination into a common class


arowla:
Pagination from each endpoint should be factored into a common class.

- [x] Create a `Pagination` class in `webservices/common/util`
- [x] The `__init__` should take the request object
- [x] The class should set page_num and per_page as attributes
- [x] The class should have an `as_json()` method to return pagination info for the API
- [ ] Each endpoint should initialize that class in an instance variable
- [ ] Each endpoint should no longer deal with pagination directly, but delegate to the class
- [x] Check/write automated tests of pagination behavior

Please let me know if anything is unclear or something in this plan doesn't work out.

LindsayYoung:
This has been pretty straight-forward except for `/names`. That endpoint doesn't produce a count of possible matches it always limits to 20. I was wondering if that was done on purpose to make that endpoint faster.

LindsayYoung:
In `test_api.py` the following tests check pagination:

test_header_info
test_per_page_defaults_to_20
test_per_page_param
test_invalid_per_page_param
test_page_param

PR #530, opened 2015-04-03
Page class

LindsayYoung:
This adds a Pagination class to `util.py`.  The init takes `page_num`, `per_page` and `count`. It has an `as_json()` method that returns the pagination information.

I did not convert `/names` to use it because I think it might be purposefully not counting possible matches in order to speed up the typeahead. I can add a correct count if we are not worried about slowing it down.


openfec/entities/annotate.py
```diff
@@ -1,132 +0,0 @@
-"""
-Tools for annotating records
```
LindsayYoung: I took these out because I thought we had already removed them from master and that I was accidentally adding things back in. This was Sean's earlier work. I can put it back in if you want, @arowla .

arowla:
Regarding the `/names` question, it's fine to hardcode the count to 20, because the query `LIMIT` is set to 20, too.

webservices/rest.py
```diff
@@ -114,6 +115,7 @@ def get(self):
         findme = ' & '.join(args['q'].split())
         data = db_conn().execute(qry, findme=findme).fetchall()
 
+        ### The count is wrong if more than 20 results are returned.
```
arowla: Can we still use the Pagination class, but just pass in 20? No need to say "the count is wrong." It won't be unless someone purposely changes the SQL query. :)

openfec/entities/annotate.py
arowla: In the future, I'd like to see PR's kept to just what the original task is about, but since this stuff ought to be taken out anyway, and we're short on time, just leave it as-is.

arowla:
Looks good! Just the one thing to change on `/names` (see line note).

webservices/rest.py
LindsayYoung: Ok, will do that now

LindsayYoung:
Ok, I made the change and the tests pass. 

Fixes #502

PR #534, opened 2015-04-03
Candidate history

### LINKED ISSUE #505: Make candidate history endpoint
Make a new view, ofec_candidate_history_mv. Remove active_through and election_years. Replace with just “election_year”. Route should be `/candidate/<id>/history`.

### LINKED ISSUE #506: Make candidate `/recent` endpoint
Depends: #505 

This should return just the most recent record from the candidate history.

LindsayYoung:
This endpoint will take a year and give the candidate information for that year.

The url will look like:
` /candidate/<id>/history/<year>`

This will close #505 and close #506


arowla:
This shouldn't get rid of #506 just yet... we still need to implement `/candidate/<id>/recent` to pull back just the top record from `/candidate/<id>/history` and not require specifying a year.

We could also call this `/candidate/<id>/history/latest`. Which would perhaps be more consistent with the pattern we've established with `/candidate/<id>/history/<year>`.

LindsayYoung:
Currently, `/candidate/<id>/history` defaults to the most recent year if no year is provided. I can easily add another url like `/candidate/<id>/history/latest` that does the same thing if you think there is additional value to that.

arowla:
The history should default to showing all years.

On Mon, Apr 6, 2015 at 10:04 AM, Lindsay Young <notifications@github.com>
wrote:

> Currently, /candidate/<id>/history defaults to the most recent year if no
> year is provided. I can easily add another url like
> /candidate/<id>/history/latest that does the same thing if you think
> there is additional value to that.
>
> —
> <https://github.com/18F/openFEC/pull/534#issuecomment-90075753>.
>



-- 
Alison Rowland
202-317-0124

LindsayYoung:
I reverted to an earlier version that worked. I will re-implement to make it faster but I think that will leave gaps in old data. I just don't think I can justify spending more time to try and get gapless data. It should only affect old records. Perhaps we could do something on the web app side that defaulted to the most recent name, if the record requested was not in the web app.

LindsayYoung:
When this is merged people will need to drop and recreate cfdm_test to run things locally.

data/cfdm_test.pgdump.sql
```diff
@@ -26755,10 +26755,13 @@ COPY dimreporttype (reporttype_sk, rpt_tp, rpt_tp_desc, load_date, expire_date)
...
 COPY dimyears (year_sk, year, load_date) FROM stdin;
-2040	2040	2014-09-03 11:27:56
-2003	2003	2014-09-03 11:27:56
-2052	2052	2014-09-03 11:27:56
-1999	1999	2014-09-03 11:27:56
+2040	2040	2014-09-03 11:27:56 \N
+2012    2012    2014-09-03 11:27:56 \N
+2013	2013	2014-09-03 11:27:56 \N
+2014    2014    2014-09-03 11:27:56 \N
+2008    2008    2014-09-03 11:27:56 \N
```
arowla: Are the years included here meant to be incomplete? We're missing 2009 and 2011...

arowla: BTW, this diff is super hard to read with all the whitespace changes. Any way you can take those out?

data/cfdm_test.pgdump.sql
```diff
@@ -65456,6 +65459,12 @@ ALTER TABLE ONLY sched_i
 ALTER TABLE ONLY sched_i
     ADD CONSTRAINT sched_i_filer_cmte_id_fkey1 FOREIGN KEY (filer_cmte_id) REFERENCES dimcmte(cmte_id);
 
+insert into dimyears values(2012, 2012, CAST('2014-09-03' as timestamp));
```
arowla: Are you sure these inserts are necessary? The additions in the dump data above should be enough, I think.

data/cfdm_test.pgdump.sql
arowla: Actually, don't spend time on the whitespace. I've located the relevant portions.

Just try to avoid a lot of whitespace changes mixed with real changes in the future. :)

data/sql_updates/create_candidate_history_view.sql
```diff
@@ -0,0 +1,44 @@
+drop materialized view if exists ofec_candidate_history_mv;
+create materialized view ofec_candidate_history_mv as
+select
+    dcp_recent.candproperties_sk as properties_key,
```
arowla: Compared to how we usually use "recent," I feel like it is a bit misleading in this context, since it will go back through time. Can we alias the table to something like "dcp_by_period"?

webservices/resources/candidates.py
```diff
@@ -213,4 +237,58 @@ def get_candidate(self, args, page_num, per_page, candidate_id, committee_id):
         return count, candidates.order_by(CandidateDetail.expire_date.desc()).paginate(page_num, per_page, False).items
 
 
+class CandidateHistoryView(Resource):
+
+    parser = reqparse.RequestParser()
+    parser.add_argument('page', type=inputs.natural, default=1, help='For paginating through results, starting at page 1')
+    parser.add_argument('per_page', type=inputs.natural, default=20, help='The number of results returned per page. Defaults to 20.')
+    # We are standardizing on year
+    parser.add_argument('year', type=str, dest='year', help="See most recent record for a two-year cycle.")
```
arowla: No need to have `dest` if it is the same.

webservices/resources/candidates.py
```diff
@@ -213,4 +237,58 @@ def get_candidate(self, args, page_num, per_page, candidate_id, committee_id):
...
+    # Would like symmetry between results and query terms
+    parser.add_argument('two_year_period', type=str, dest='year', help="Synonym for year.")
```
arowla: Is it worth it to add this if we're going to move to primarily putting the year in the url, ala `/candidate/<id>/history/<year>`?

arowla:
I'm a little torn on whether we should implement a two-step sql_update process. On the one hand, it's not necessary for this task in isolation. The years view could be placed in the beginning of the candidate history SQL file. On the other hand, the two-step process could come in handy if we try to implement history for committees or something. I think that going forward, we may in fact need a "real" incremental migrations system that understands the dependencies between the different models, but this could behave as a stop-gap. So that said, I guess I'm for leaving as-is... Just wanted you to understand where I'm coming from on this (wish we didn't have to move to a two-step), and where I think we may be going (incremental migrations).

data/cfdm_test.pgdump.sql
LindsayYoung: The other changes didn't work so I added the insert statement. Adding the inserts worked so I kept them.

webservices/resources/candidates.py
LindsayYoung: The information is not responsive but it would be helpful for filtering api calls. For example, if you wanted a list of incumbents that ran in 04 you could use the year and incumbent/challenger  = I to get those names.

LindsayYoung:
I agree it is not ideal, I couldn't get it to run in one step.  `db.engine.execute()` won't create 2 views in one call and the strategies I used to try and enforce an order that they would be run in did not work in Travis.

webservices/resources/candidates.py
LindsayYoung: Yea, those filters are redundant. I will take those out.

Do you not like 'two_year_period` as a param or do you not like the that name at all?  I like `two_year_period` because it is very literal.

webservices/resources/candidates.py
LindsayYoung: Never mind- I misread what you wrote.

LindsayYoung:
I just merged master but now I need to update the mangement commands since this has 2 steps.

LindsayYoung:
@jmcarp could you double check that my changes to manage.py don't mess with your work?

manage.py
```diff
@@ -13,6 +15,25 @@
 manager = Manager(app)
 
 
+def excute_sql_folder(files):
```
jmcarp: `excute` -> `execute`

manage.py
```diff
@@ -13,6 +15,25 @@
...
+    files = glob.glob(sql_dir + '*.sql')
+    for sql_file in files:
+        print(("Running {}".format(sql_file)))
+        with open(sql_file, 'r') as sql_fh:
+            sql = '\n'.join(sql_fh.readlines())
+            db.engine.execute(sqla_text(sql))
+
+
+@manager.command
+def update_schemas():
```
jmcarp: @LindsayYoung LGTM :+1:

manage.py
LindsayYoung: Thanks, I am sure you have figured out by now that spelling is not my strength. :)

arowla:
Ok, imma call this good to go!

PR #539, opened 2015-04-06
Fix up linkage view file

arowla:
The SQL as it was was not compatible with the snapshot available to our CloudFoundry instances, and this file was failing as part of the `refresh_db` command on CF. This fix is to make it backwards compatible, as well as to fix up the naming a little bit.


LindsayYoung:
Is this still needed?

arowla:
This has been pushed into obsolescence.

PR #544, opened 2015-04-08
Query candidate ids

jmcarp:
The `CommitteeList` resource may attempt to query on `candidate_ids`, but that field isn't mapped in the `Committee` model. This patch adds the missing column and tests on querying by candidate ID.


LindsayYoung:
:+1:

PR #541, opened 2015-04-07
Materialize ephemeral views.

jmcarp:
Thanks for help with this @arowla @LindsayYoung 
[Closes #540]

PR #545, opened 2015-04-08
Index queried fields

jmcarp:
Columns that are going to be queried should probably have an index. This patch adds indices on all columns that are filtered on in `resources`. Depends on #541.

### DIFF
--- a/data/sql_updates/create_candidate_detail_view.sql
+++ b/data/sql_updates/create_candidate_detail_view.sql
@@ -1,4 +1,3 @@
-drop view if exists ofec_candidate_detail_vw;
 drop materialized view if exists ofec_candidate_detail_mv;
 create materialized view ofec_candidate_detail_mv as
 select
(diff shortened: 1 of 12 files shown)


data/sql_updates/create_candidate_detail_view.sql
```diff
@@ -1,5 +1,6 @@
-drop materialized view if exists ofec_candidate_detail_vw;
-create materialized view ofec_candidate_detail_vw as
+drop view if exists ofec_candidate_detail_vw;
```
LindsayYoung: This is already a materialized view so this line will error out. This was the weird problem we were having last week.

jmcarp: Should be fixed now. Thanks for catching this!

data/sql_updates/create_totals_presidential_view.sql
```diff
@@ -1,3 +1,4 @@
+drop view if exists ofec_totals_presidential_vw;
```
LindsayYoung: This hasn't been a view before so we don't need this line. It is the same case for create_totals_pacs_parties_view.sql and create_totals_house_senate_view.sql

PR #562, opened 2015-04-08
 Abstract models

jmcarp:
Use abstract base classes to repeat less code cross similar models. Low priority--I just wanted to get this filed before I forgot about it.


arowla:
If I try to access:
http://localhost:5000/committee/C00550715/reports
or
http://localhost:5000/committee/C00550715/totals

It works on master, but reports a 500 error running from this branch. (I am not sure why I am only getting a 500 error reported and not a stack trace-- must be my local config.)

jmcarp:
Interesting, will check this out.

jmcarp:
@arowla this should be fixed now. Thanks for catching. FWIW, #571 adds more tests for the reports and totals resources, so it should be easier to catch these issues going forward.

Issue #518, opened 2015-03-27
Whitelist IPs for API access, including data.gov ones


theresaanna:
We need to restrict access to the API instances by whitelisting IPs. We should include data.gov addresses.

Per Kaitlin, we should do this in Python so that it works on Cloud Foundry and a note from her taken from Slack:

"depending on how flask interprets the headers, it may put the forwarded IP (the actual client) and the data.gov IP on the header
so you might need to investigate and make sure you're checking the right one"

LindsayYoung:
- [ ] Add data.api.gov whitelist
 - [ ] Make sure it works

PR #564, opened 2015-04-09
White listing the api.data.gov ip when PRODUCTION=true

LindsayYoung:
This is for security and use by api.data.gov. We could add more ips later, if we wanted to. 
issue #518

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -46,12 +46,11 @@ db.init_app(app)
 # api.data.gov
 trusted_proxies = ('192.168.114.100',)
 PRODUCTION = os.getenv('PRODUCTION', False)
-print(PRODUCTION)
+
 
 @app.before_request
 def limit_remote_addr():
-    if PRODUCTION == 'true':
-        print('running')
+    if PRODUCTION != False:
         remote = request.remote_addr
         route = list(request.access_route)
         while remote in trusted_proxies:
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -45,13 +45,13 @@ db.init_app(app)
 
 # api.data.gov
 trusted_proxies = ('192.168.114.100',)
-PRODUCTION = os.getenv('PRODUCTION', False)
+WHITE_LIST = os.getenv('WHITE_LIST', False)
 
 
 @app.before_request
 def limit_remote_addr():
     falses = (False, 'False', 'false', 'f')
-    if PRODUCTION not in falses :
+    if WHITE_LIST not in falses :
         remote = request.remote_addr
         route = list(request.access_route)
         while remote in trusted_proxies:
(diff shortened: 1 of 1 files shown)


LindsayYoung:
Once this is done we want to point the web app to api.data.gov

webservices/rest.py
```diff
@@ -43,6 +43,24 @@ def sqla_conn_string():
 api = restful.Api(app)
 db.init_app(app)
 
+# api.data.gov
+trusted_proxies = ('192.168.114.100')
```
jmcarp: It looks like `trusted_proxies` is expected to be a tuple, but here it's going to be a string. For a 1-tuple, use
```python
trusted_proxies = ('192.168.114.100', )
```

LindsayYoung: Good catch, thanks!

webservices/rest.py
```diff
@@ -43,6 +43,24 @@ def sqla_conn_string():
...
 db.init_app(app)
 
+# api.data.gov
+trusted_proxies = ('192.168.114.100',)
+PRODUCTION = os.getenv('PRODUCTION', False)
+print(PRODUCTION)
+
+@app.before_request
+def limit_remote_addr():
+    if PRODUCTION == 'true':
```
arowla: Can we convert PRODUCTION to a boolean or at least a "truthy" value (like 1) instead of just checking for string equivalence? Even if the PRODUCTION value comes through as a string, it is likely to be set to "True", in which case this check would fail on the case difference, as well.

webservices/rest.py
```diff
@@ -43,6 +43,24 @@ def sqla_conn_string():
...
+        print('running')
```
arowla: Do we need this `print()` and the one on line 59?

LindsayYoung: No, I'll get rid of it

webservices/rest.py
```diff
@@ -43,6 +43,23 @@ def sqla_conn_string():
...
 db.init_app(app)
 
+# api.data.gov
+trusted_proxies = ('192.168.114.100',)
+PRODUCTION = os.getenv('PRODUCTION', False)
+
+
+@app.before_request
+def limit_remote_addr():
+    if PRODUCTION != False:
```
arowla: Unfortunately this will still fail in case somebody sets PRODUCTION=False, because False will come in from `os.getenv()` as the string 'False', and 'False' evaluates to True in a boolean context.

    >>> bool('False')
    True

I'd recommend doing something like this:

http://stackoverflow.com/a/715468

arowla:
Sorry, can we change the var to a more qualified name to fit with the pattern we have going, like `FEC_API_WHITELIST_IPS`?

LindsayYoung:
This is updated and ready for review.

Issue #547, opened 2015-04-08
CF Database Migrations


arowla:
The DB migrations need a standard way of being done, preferably not totally manually via someone using `cf-ssh` as I've done now. Figure out the best way to do this reliably on future deploys.

arowla:
The `refresh_db` command does double-duty as a data refresh and as a schema change command. The primary thing we need to figure out is the best way to run this command on our Cloud Foundry instances in response to schema changes. So there are two possibilities:

1. run whenever something in the `/data/sql_updates` directory changes;
2. or in conjunction with every code deploy (`cf push`).

(In the future, we'll also need to run this command nightly to keep the data up to date, which is something to consider, but that shouldn't be part of this task.)

You can refer to the [CF migration docs](http://docs.cloudfoundry.org/devguide/services/migrate-db.html), the catch being that our `refresh_db` command almost certainly takes longer than their minute-long timeout allowed by default for application startup under `cf push`. I am not certain whether it can fit within the maximum three minutes, or whether we can expect it to as we start to receive more data from the client.

arowla:
Also, in case it matters for this task at all, we are not using a CF "service" to host the databases, but rather a standalone AWS RDS instance located within the CF VPC.

jmcarp:
Can this be closed @arowla?

PR #566, opened 2015-04-13
[wip] Run migration in subprocess on cf push.

jmcarp:
Here's a very simple proposed solution for #547. The database migration script is run automatically on pushing the application to Cloud Foundry. To avoid Cloud Foundry stopping the migration after its timeout period, the migration is run in a subprocess that continues to run after its parent exits. I verified that `subprocess` works as expected on Cloud Foundry in a separate testing project, but haven't been able to get this to build yet. I'll update here once I verify that this works.

@arowla @LindsayYoung does this seem reasonable?

### DIFF
--- a/run_api.py
+++ b/run_api.py
@@ -1,14 +1,21 @@
-import doctest
 import os
 import sys
+import doctest
+import subprocess
 
 
 from webservices.rest import app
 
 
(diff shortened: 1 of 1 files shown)


jmcarp:
Superseded by #574.

PR #568, opened 2015-04-13
Removed the year filering on singe resources.

### LINKED ISSUE #565: Empty committee result set that should be filled
A committee is returning both totals and reports but the actual committee is returning empty.

Reproduction:
On both my localhost database, and on staging site:
`GET /committee/C00000422/reports` returns a full result set
`GET /committee/C00000422/totals` returns a full result set
`GET /committee/C00000422` returns an empty result set.

Expected:
I would expect `GET /committee/C00000422` to return a full result set with committee information such as name, etc.

May potentially be blocking me from completing #560 as I will need the correct data for it.

### LINKED ISSUE #567: No year filter on single resources
We are using year defaults on singular and plural endpoints. It makes sense to filter multiple resources to a manageable default but not the singe resource endpoints, since the user is specifying that particular resource. 

For instance you can see  /committee/C00202176/?year=* but not /committee/C00202176

LindsayYoung:
This removed the year defaulting on the singe resources, they can sill be filtered by year if the user requests a year.

fix #565
fix #567

### DIFF
--- a/webservices/resources/committees.py
+++ b/webservices/resources/committees.py
@@ -247,7 +247,7 @@ class CommitteeView(Resource):
                     committees = committees.filter(getattr(Committee, argname)==args[argname])
 
         # Should this handle a list of years to make it consistent with /candidate ?
-        if args.get('year') and args['year'] != '*':
+        if args.get('year'):
             # before expiration
             committees = committees.filter(or_(extract('year', CommitteeDetail.expire_date) >= int(args['year']), CommitteeDetail.expire_date == None))
             # after origination
--- a/webservices/resources/committees.py
+++ b/webservices/resources/committees.py
@@ -246,14 +246,8 @@ class CommitteeView(Resource):
                 else:
                     committees = committees.filter(getattr(Committee, argname)==args[argname])
 
-        # default year filtering
-        if args.get('year') is None:
-            earliest_year = int(sorted(default_year().split(','))[0])
-            # still going or expired after the earliest year we are looking for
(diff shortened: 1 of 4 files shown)


webservices/resources/committees.py
```diff
@@ -246,14 +246,8 @@ def get_committee(self, args, page_num, per_page, committee_id, candidate_id):
...
 
-        # default year filtering
-        if args.get('year') is None:
-            earliest_year = int(sorted(default_year().split(','))[0])
-            # still going or expired after the earliest year we are looking for
-            committees = committees.filter(or_(extract('year', CommitteeDetail.expire_date) >= earliest_year, CommitteeDetail.expire_date == None))
-
         # Should this handle a list of years to make it consistent with /candidate ?
-        elif args.get('year') and args['year'] != '*':
+        if args.get('year') and args['year'] != '*':
```
arowla: We can remove the part about `*` since that's just to counteract default year filtering.

LindsayYoung: That makes sense I will do that now

LindsayYoung: Oh, I remember my thought process on that. Since other branches worked like that, I did not want the user to get errors if they tried, year=* , even though it was unnecessary. I can still take it out though.

arowla: Ok, that makes sense. I'm fine with leaving it in, then, but I'd recommend removing the comment/question that's above that and replacing it with something to the effect of wanting to support '*' across all endpoints.

LindsayYoung: Ok I will add that back in with the comment.

Issue #569, opened 2015-04-13
Update API documentation


jmcarp:
API documentation currently lives in markdown files that have to be manually updated when behavior changes. This leads to extra work and the potential for stale docs. Where possible, documentation should be automatically built from code. I'm working on a proof of concept using https://github.com/marshmallow-code/smore--will present for review by 4/14.

LindsayYoung:
Sounds awesome, Josh!

LindsayYoung:
There is a lot of stuff going on here so I will defer to Alison, but it looks like good work to me.

LindsayYoung:
If possible people in API testing requested a dropdown menu of possible options for things like fec codes

#383

PR #571, opened 2015-04-14
[WIP] Automatic API documentation

jmcarp:
This is the start of automatic API documentation for the OpenFEC API. The goal is to generate the complete API specification, in Swagger format, from code, such that documentation won't need to be updated manually going forward. At the moment, this patch generates and serves Swagger JSON and YAML, but it would also be straightforward to add a web-facing API console with Swagger-UI in the future.

So far, this is just a prototype, and only affects views for the `Candidate` model. The patch makes a lot of changes--it's using smore to generate the Swagger specification, which depends on webargs for request parsing and marshmallow for marshalling results. If this is too big a change or conflicts with other ongoing work, I'm happy to implement differently, but based on my research so far, the smore-webargs-marshmallow stack is the most promising path in the python world for automatic documentation. Flask-Restful also looks to be adopting both webargs and marshmallow to replace its current request parsing and marshalling utilities, so this would put us ahead of the curve for upcoming changes there.

To run locally:
* pip install -r requirements.txt --upgrade
* python manage.py runserver

Check out the auto-generated documentation in different formats:
```
curl http://localhost:5000/swagger/ -H "Accept:application/json"
curl http://localhost:5000/swagger/ -H "Accept:application/yaml"
```

Addresses #569

Todo:
- [x] candidate views
- [x] committee views
- [x] reports views
- [x] totals views
- [x] add missing descriptions


jmcarp:
This is now ready for review, time permitting. #569 isn't part of a milestone, so I'm guessing this isn't slated for the current sprint--I'll keep this updated with `master` so that it doesn't rot.

Here's how the proposed API works:

* Request arguments (from the query string, form data, headers, etc) are collected using `webargs`. Decorate a view function with `register_kwargs` to capture arguments *and* to auto-document the parameters that the view accepts
* Response outputs are marshalled with `marshmallow`. Decorate a view function with `marshal_with` to marshall the return value *and* to auto-document the response type of the view
* Request arguments that are part of the path (e.g. `/candidate/{candidate_id}`) are documented using the `doc` decorator:

    ```python
    @spec.doc(path_params=[
        {'name': 'candidate_id', 'in': 'path', 'type': 'string'},
        {'name': 'committee_id', 'in': 'path', 'type': 'string'},
    ])
    class CandidateView(Resource):
        pass
    ```
* Flask-Restful `Resource` classes are registered to the API specification in `rest.py`:
    `register_resource(CandidateView)`

The auto-generated Swagger specification is then served at `/swagger/`. For a browsable API console, use `/swagger/ui/`.

Sorry this patch touches so much code, but the changes aren't as extensive as they might look. Almost all the diffs are find-and-replaces converting from `reqparse` and `marshal` to `webargs` and `marshmallow`, e.g.

`parser.add_argument('q', type=str, help='Text to search all fields for')`
to
`q = Arg(str, description='Text to search all fields for')`

I'm happy to revise / refactor / document as needed @arowla @LindsayYoung

webservices/resources/reports.py
```diff
@@ -1,592 +1,37 @@
...
-    'per_page': fields.Integer,
-    'page': fields.Integer,
-    'count': fields.Integer,
-    'pages': fields.Integer,
+
+reports_schema_map = {
+    'P': (models.CommitteeReportsPresidential, schemas.ReportsPresidentialPageSchema),
+    'H': (models.CommitteeReportsHouseOrSenate, schemas.ReportsHouseSenatePageSchema),
+    'S': (models.CommitteeReportsHouseOrSenate, schemas.ReportsHouseSenatePageSchema),
+    None: (models.CommitteeReportsPacOrParty, schemas.ReportsPacPartyPageSchema),
```
LindsayYoung: Do we want to be more precise here? PACs and Parties are committee types: N, O, Q, V, W, X, Y and Z.

We currently don't have types- I, C, D, U because their information is not in these tables. These tables are populated by the three versions of form 3.

And if we do want to be more precise, I can confirm the type-codes with Paul or Jeff.

jmcarp: TBH, I'm not sure what all the committee types mean. The idea here is to preserve the current API behavior, which uses `CommitteeReportsPacOrParty` as the default model. If we use committee types N, O, Q, etc., for PAC or party, do we need to specify a default option?

LindsayYoung: Yea, I am jumping the gun on this one. You don't have to change anything. When we do add data for other types of committees we will want to make sure that we get the right tables for them instead of pointing them here. 

Also, here is the explanations for the type codes if you are curious- http://www.fec.gov/finance/disclosure/metadata/CommitteeTypeCodes.shtml

jmcarp:
Replaced by #662.

PR #574, opened 2015-04-16
Cron refresh

jmcarp:
Add cron-like task based on Celery Beat to refresh materialized views nightly. Subsumes #566.

This has been updated and should now actually work with the usual CF workflow. Note that the celery beat task isn't debounced at the moment, so invocations could stack up if it were scheduled too frequently--but this shouldn't be an issue for running nightly.

Also, postgres 9.4 added a `REFRESH MATERIALIZED VIEW CONCURRENTLY` function, which might make the refresh faster. @arowla: if we upgrade, we might want to give that a try (we're using 9.3, right?).

* Add cron-like task to run `refresh_materialized` every night at midnight
* Rename management command `refresh_db` to `update_schemas`
* Add management command `refresh_materialized`; refreshes all materialized views
* Add management commands `start_beat` and `stop_beat` to start and stop celery beat workers
* On `run_api`, run `update_schemas` in a subprocess
* On `run_api`, restart celery beat workers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ before_script:
   - psql -f data/cfdm_test.pgdump.sql -U postgres cfdm_test
   - psql -c "SELECT count(*) FROM dimcand" -U postgres cfdm_test
   - pip install -r requirements.txt
-  - python manage.py refresh_db
+  - python manage.py update_schemas
 script: nosetests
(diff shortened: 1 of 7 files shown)


arowla:
Interesting! Sorry if I'm missing something, but how do we kick off the worker in cron.py?

jmcarp:
My mistake--I meant to label this [WIP] until I got that part figured out. I've been having trouble with CF, but I *think* it should be something like `cf push --command "python manage.py start_beat"`, which will start the celery worker in a subprocess and run it indefinitely.

PR #580, opened 2015-04-20
Materialize fulltext tables.

jmcarp:
Based on discussion in #fec with @arowla and @LindsayYoung 

* Use materialized views instead of tables for fulltext search
* Drop fulltext search tables


jmcarp:
After this change, all reshaped data would live in a materialized view, so running `refresh_materialized` should update all the data without the need to drop and recreate tables.

PR #581, opened 2015-04-21
Changed PRODUCTION env parsing

arowla:
- This version drives the DEBUG setting
- Fixes issue where having PRODUCTION defined at all would always make
  DEBUG = False

### DIFF
--- a/run_api.py
+++ b/run_api.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import ast
 import os
 import sys
 import doctest
@@ -23,5 +24,5 @@ if __name__ == '__main__':
     if len(sys.argv) > 1 and sys.argv[1].lower().startswith('test'):
         doctest.testmod()
     else:
-        debug = not os.getenv('PRODUCTION')
+        debug = not ast.literal_eval(os.getenv('PRODUCTION', False))
         app.run(debug=debug, port=int(port), host='0.0.0.0')


run_api.py
```diff
@@ -23,5 +24,5 @@
     if len(sys.argv) > 1 and sys.argv[1].lower().startswith('test'):
         doctest.testmod()
     else:
-        debug = not os.getenv('PRODUCTION')
+        debug = not ast.literal_eval(os.getenv('PRODUCTION', False))
```
jmcarp: Looks like this will fail if the `PRODUCTION` environment variable isn't set: `ast.literal_eval(False)` raises a `ValueError`. Could change `False` to `'False'` to fix.

jmcarp: But parsing environment variables to booleans is brittle however we do it--if a developer sets `PRODUCTION` to `"false"` instead of `"False"`, they'll get a potentially confusing error. What about renaming `PRODUCTION` to something like `FEC_ENVIRONMENT`, with values of `"dev"`, `"staging"`, or `"production"`? Then if a developer has an invalid value, we can default to `"dev"`.

LindsayYoung: Sorry, I merges before I saw this. Just reverted.

arowla: I like that idea, @jmcarp. We've run into this issue before. Can you put in a ticket to do that? We can forget this PR and I'll just use it locally for debugging work until we can make that change, which should be relatively quick.

jmcarp: Filed in #582.

PR #583, opened 2015-04-21
Use FEC_API_DEBUG flag to control debug mode.

### LINKED ISSUE #582: Configure environment (dev, stage, prod) with FEC_ENVIRONMENT
Deprecate boolean-valued `PRODUCTION` flag and replace with string-valued `FEC_ENVIRONMENT` flag. Namespaced environment variable names are less prone to conflict, and parsing boolean environment variables can be a pain.

jmcarp:
Make environment configuration consistent with webapp, which uses
FEC_WEB_DEBUG flag for debug mode.

[resolves #582]

I noticed that openFEC-web-app handles this issue with a `FEC_WEB_DEBUG` flag, so figured it would be simplest to use that approach here for consistency. The `PRODUCTION` flag was only used to set debug mode on the app, so I think it's reasonable to use the more specific name.

### DIFF
--- a/run_api.py
+++ b/run_api.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 
-import ast
 import os
 import sys
 import doctest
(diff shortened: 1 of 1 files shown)


arowla:
One thing is that PR #564 uses `PRODUCTION`... That could be a situation where it makes sense for the environment flag to be decoupled from the debug flag.

jmcarp:
Good catch. I didn't realize `PRODUCTION` was being used for whitelisting too. Agreed that decoupling flags seems reasonable--it's easy to forget exactly which behaviors are controlled by broad flags like `PRODUCTION`. How about using a flag like `FEC_API_WHITELIST` for #564, just for maximum clarity?  I'm also fine with the earlier proposal of a more general flag like `FEC_ENVIRONMENT` if you all prefer.

PR #586, opened 2015-04-21
[WIP] Add zero-downtime schema migrations.

jmcarp:
On migrating each materialized view:
* Delete *_mv_tmp if exists
* Create new view as *_mv_tmp
* Optionally create indices in *_mv_tmp
* Within transaction, delete *_mv and rename *_mv_tmp to *_mv

This is the new proposed setup for zero-downtime schema migrations, based on discussion in #fec earlier. Each new materialized view is built to a temporary view, so the previous version is still accessible while the new version is building. Then each temporary view is renamed, which should be almost instant. We might go one step further and run all the renames in a single transaction at the end of the migration, so that users never get old data from one view and new data from another.

- [x] build materialized views as *_mv_tmp
- [x] rename all new views in a single transaction
- [x] parallelize view creation

Thoughts @arowla @LindsayYoung @seanherron?


arowla:
Sounds great. And I like the idea of doing  all the renames in a single
step at the end.

jmcarp:
Ready for review @arowla @LindsayYoung. The proposed changes will run schema updates in parallel, defaulting to two concurrent processes. Concurrency is controlled with an optional argument to `update_schemas`, e.g. `python manage.py update_schemas --processes 4`.

Running locally, using three or four processes is fastest. If this patch gets merged, we should probably experiment with different concurrency levels on CF to make a more informed choice.

data/sql_updates/alter_read_only_web_role.sql
```diff
@@ -3,8 +3,8 @@ $body$
 BEGIN
    IF NOT EXISTS (
       SELECT *
-      FROM   pg_catalog.pg_user
-      WHERE  usename = 'webro') THEN
+      FROM   pg_roles
+      WHERE  rolname = 'webro') THEN
```
jmcarp: @arowla I had to make this change to get `alter_read_only_web_role` to run locally, but I haven't figured out why it wasn't working for me before. Does this change look all right to you?

arowla: Perhaps something changed about the pg_catalog tables between minor versions? I know Postgres has been moving to the concept of roles subsuming the concept of users and groups.

That change looks fine to me.

PR #595, opened 2015-04-22
Refresh concurrently

jmcarp:
This patch changes the refresh materialize views script so that views are refreshed concurrently, and views aren't locked during refresh. This requires postgres 9.4, so should *not* be merged until / unless the rds instances are using the latest version. Concurrent refresh also requires that each view has at least one unique index, so I added a dummy row counter to each view.


arowla:
I am going to assign this to myself as a reminder to change over our RDS instances to 9.4.

arowla:
@jmcarp this is reporting conflicts... mind updating your branch?

jmcarp:
Conflict resolved @arowla.

PR #601, opened 2015-04-23
Cloud foundry and New Relic

arowla:
Getting all to play nice. This is what's currently deployed in production and it seems to work. See commit messages for further description.


manifest_prod.yml
```diff
@@ -1,12 +1,16 @@
...
+env:
+  NEW_RELIC_CONFIG_FILE: newrelic.ini
+  NEW_RELIC_ENV: staging
 applications:
 - name: api
   path: .
   host: fec-prod-api
   env:
-    PRODUCTION: True
+    FEC_API_DEBUG: False
```
jmcarp: If environment variables are cast to strings, this will set `FEC_API_DEBUG` to `"false"`, which is truthy.

jmcarp: Also, maybe it's worth thinking about a more generic approach to parsing environment variables. `configargparse` looks decent: https://pypi.python.org/pypi/ConfigArgParse

arowla: Thanks for catching that. Agree we should work on a standard solution. `configargparse` looks like it has a lot of command-line oriented stuff we probably don't need-- what do you think about https://github.com/mbr/flask-appconfig, especially the "ENV-only" mode?

arowla: Oh, and I particularly like that the aforementioned lib enforces an app-name based prefix, which we have started to move toward anyway.

webservices/rest.py
```diff
@@ -52,13 +52,8 @@ def sqla_conn_string():
...
     if FEC_API_WHITELIST_IPS not in falses :
-        remote = request.remote_addr
-        route = list(request.access_route)
-        while remote in trusted_proxies:
-            remote = route.pop()
-        print(route.pop())
-
-        if remote not in trusted_proxies:
+        route = request.access_route
+        if route[-2] not in trusted_proxies:
```
jmcarp: Is this because `route[-1]` is the CF router, `route[-2]` is the api.data.gov IP, and `route[:-2]` is stuff we don't care about? If that's right, it would be helpful to clarify what's going on, and possibly handle the case where `access_route` is too short:

```python
try:
    *_, api_data_route, cf_route = request.access_route
except ValueError:  # Not enough routes
    abort(403)
else:
    if api_data_route not in trusted_proxies:
       abort(403)
```

arowla: That's exactly right. (`route[:-2]` is the end-user IP, which we can't whitelist for anyway.) Great suggestion.

manifest_prod.yml
jmcarp: Agreed that enforcing a standard prefix should happen. Flask-appconfig looks good. The only thing I'd like to see is a more declarative method for listing expected config variables, types, defaults, etc., which is what I like about argparse-like libraries (although we can store default values in a config file with flask-appconfig too). People also seem to like Click: http://click.pocoo.org/4/options/#values-from-environment-variables.

Anyway, I think that basically any approach to standardizing environment parsing is a good idea, so let's go with whichever looks best to you @arowla.

PR #603, opened 2015-04-23
Optimize queries to build materialized views.

jmcarp:
A handful of materialized view queries used per-row SELECT statements,
so that a separate SELECT was emitted for each row of the view. This
patch replaces per-row SELECTs with JOINs, which makes queries
noticeably faster on the production data.

Once this gets merged, we should see faster schema migrations and view refreshes. On the local data, these changes took migration time from ~7s to ~3s. On GoldenGate, queries that took an unacceptably long time (I got impatient and cancelled anything that ran for 10 minutes without finishing) actually finished.

### DIFF
--- a/data/sql_updates/create_candidates_view.sql
+++ b/data/sql_updates/create_candidates_view.sql
@@ -34,7 +34,7 @@ from dimcand
     inner join dimparty using (party_sk)
     left join (
         select distinct on (cand_sk) cand_sk, cand_nm from dimcandproperties
-            order by cand_sk desc
+            order by cand_sk, candproperties_sk desc
     ) candprops on dimcand.cand_sk = candprops.cand_sk
 group by
     dimcand.cand_sk,
--- a/data/sql_updates/create_candidates_view.sql
+++ b/data/sql_updates/create_candidates_view.sql
@@ -33,12 +33,8 @@ from dimcand
     inner join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
     left join (
-        select props.cand_sk, cand_nm from (
-            select cand_sk, max(candproperties_sk) as candproperties_sk from dimcand
-                join dimcandproperties using (cand_sk)
-                group by cand_sk
-        ) props
-            join dimcandproperties using (candproperties_sk)
+        select distinct on (cand_sk) cand_sk, cand_nm from dimcandproperties
+            order by cand_sk desc
     ) candprops on dimcand.cand_sk = candprops.cand_sk
 group by
     dimcand.cand_sk,
--- a/data/sql_updates/create_candidates_view.sql
+++ b/data/sql_updates/create_candidates_view.sql
@@ -25,13 +25,21 @@ select
     max(dimparty.party_affiliation) as party,
     max(dimparty.party_affiliation_desc) as party_full,
     max(dimoffice.office_state) as state,
-    (select cand_nm from dimcandproperties cp where cp.cand_sk = dimcand.cand_sk order by candproperties_sk desc limit 1) as name
+    max(candprops.cand_nm) as name
 from dimcand
     left join (select distinct on (cand_sk) cand_sk, election_yr, cand_status, ici_code from dimcandstatusici order by cand_sk, election_yr desc) csi_recent using (cand_sk)
     left join dimcandstatusici csi_all using (cand_sk)
     inner join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)  -- only joined to get to dimoffice
     inner join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
+    left join (
+        select props.cand_sk, cand_nm from (
+            select cand_sk, max(candproperties_sk) as candproperties_sk from dimcand
+                join dimcandproperties using (cand_sk)
+                group by cand_sk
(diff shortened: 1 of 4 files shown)


data/sql_updates/create_candidates_view.sql
```diff
@@ -25,13 +25,21 @@ select
...
     max(dimoffice.office_state) as state,
-    (select cand_nm from dimcandproperties cp where cp.cand_sk = dimcand.cand_sk order by candproperties_sk desc limit 1) as name
+    max(candprops.cand_nm) as name
 from dimcand
     left join (select distinct on (cand_sk) cand_sk, election_yr, cand_status, ici_code from dimcandstatusici order by cand_sk, election_yr desc) csi_recent using (cand_sk)
     left join dimcandstatusici csi_all using (cand_sk)
     inner join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)  -- only joined to get to dimoffice
     inner join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
+    left join (
```
arowla: This subquery should be rephrased with a DISTINCT ON, which will make it far more efficient (and easier to read).
```
select distinct on (cand_sk) cand_sk, cand_nm from dimcandproperties order by cand_sk desc
```

data/sql_updates/create_linkage_names_view.sql
```diff
@@ -39,9 +39,31 @@ select
...
     l.load_date as load_date,
     l.expire_date as expire_date,
-    (select cand_nm from dimcandproperties where dimcandproperties.cand_sk = l.cand_sk order by candproperties_sk desc limit 1) as candidate_name,
-    (select cmte_nm from dimcmteproperties where dimcmteproperties.cmte_sk = l.cmte_sk order by cmteproperties_sk desc limit 1) as committee_name
-from dimlinkages l;
+    candprops.cand_nm as candidate_name,
+    cmteprops.cmte_nm as committee_name
+from dimlinkages l
+    left join (
+        select props.cand_sk, cand_nm from (
```
arowla: This and the following subquery should also use DISTINCT ON.

data/sql_updates/create_candidates_view.sql
jmcarp: Thanks for catching--this is much cleaner. It's *slightly* different from the original query, which sorted by `candproperties_sk` rather than `cand_sk`, but that difference may not be meaningful. For the bundled data, one row changes as a result of this change, such that `"MORAN, JAMES P. JR."` becomes `"MORAN, JAMES P  JR."`. On the assumption that this isn't an important change, I'm going to update the PR with your revisions, but let me know if otherwise.

LindsayYoung: I am a little worried that things look different. We should be consistently going by the most recent name of the candidate. We do need that information to be consistent across the site, but even more important, the other details need to be up to date.  We can't show incumbents as challengers and things like that.

jmcarp: @LindsayYoung OK, we can revert that last change or think of a more efficient way to write the query without changing the results. Can you confirm that sorting on `candproperties_sk` instead of `cand_sk` gives us the most recent name for each candidate?

arowla: @jmcarp ah, thanks, I overlooked that. Then I think the answer is just to make the ORDER BY clause read `cand_sk, candproperties_sk desc`.

Is there something we've overlooked which changes it to not use the most recent status, @LindsayYoung ?

data/sql_updates/create_linkage_names_view.sql
```diff
@@ -39,9 +39,23 @@ select
...
     l.expire_date as expire_date,
-    (select cand_nm from dimcandproperties where dimcandproperties.cand_sk = l.cand_sk order by candproperties_sk desc limit 1) as candidate_name,
-    (select cmte_nm from dimcmteproperties where dimcmteproperties.cmte_sk = l.cmte_sk order by cmteproperties_sk desc limit 1) as committee_name
-from dimlinkages l;
+    candprops.cand_nm as candidate_name,
+    cmteprops.cmte_nm as committee_name
+from dimlinkages l
+    left join (
+        select distinct on (cand_sk) cand_sk, cand_nm from dimcandproperties
+            order by cand_sk desc
```
arowla: This and the clause below probably need the fix to add the properties key as the last ordering column, too.

data/sql_updates/create_candidates_view.sql
jmcarp: @arowla thanks, that works. Candidate and committee names are now identical pre- and post-patch again. @LindsayYoung, I think this should resolve any issues about changing the data.

LindsayYoung: Looking at Moran `cand_sk=10143946` in dimcand properties, the most recent record I see is 2013-04-24, on that record his name is "MORAN, JAMES P. JR." Not showing the most recent record could lead to other worse errors, like showing an out-dated congressional district for a member.

LindsayYoung: :+1: That is great to hear! I think this will resolve the too many cook issue as well.

rjmajma: You guys are so smart (sorry if this isn't appropriate to write here).

Issue #605, opened 2015-04-24
On the API, the totals route is failing with an error


msecret:
The URL http://fec-dev-api.cf.18f.us/committee/C00000422/totals

With the following error:

`AttributeError: 'NoneType' object has no attribute 'filter_by'`

Here is the stacktrace:
```
Traceback (most recent call last)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/app.py", line 1836, in __call__
return self.wsgi_app(environ, start_response)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/newrelic-2.50.0.39/newrelic/api/web_transaction.py", line 1329, in _nr_wsgi_application_wrapper_
result = wrapped(*args, **kwargs)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/app.py", line 1820, in wsgi_app
response = self.make_response(self.handle_exception(e))
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask_restful/__init__.py", line 265, in error_router
return original_handler(e)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/newrelic-2.50.0.39/newrelic/hooks/framework_flask.py", line 98, in _nr_wrapper_Flask_handle_exception_
return wrapped(*args, **kwargs)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/app.py", line 1403, in handle_exception
reraise(exc_type, exc_value, tb)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/_compat.py", line 32, in reraise
raise value.with_traceback(tb)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/app.py", line 1817, in wsgi_app
response = self.full_dispatch_request()
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask/app.py", line 1477, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/flask_restful/__init__.py", line 265, in error_router
return original_handler(e)
File "/home/vcap/app/.heroku/python/lib/python3.4/site-packages/newrelic-2.50.0.39/newrelic/hooks/framework_flask.py", line 98, in _nr_wrapper_Flask_handle_exception_
return wrapped(*args, **kwargs)
[...]

PR #606, opened 2015-04-24
Fix default totals model.

jmcarp:
The default totals model was using the abstract `CommitteeTotals` model, not the concrete `CommitteeTotalsPacOrParty` model. My bad.


jmcarp:
Fixes #605.

PR #607, opened 2015-04-24
Change New Relic config

### LINKED ISSUE #602: New Relic Config Tweaks
New Relic is currently showing two apps, neither of which seem to be correctly reporting... We need to make sure that:

- [ ] The web and API apps are clearly named in New Relic config
- [ ] The different environments (dev/stage/prod) of each are also clearly delineated
- [ ] Each environment is reporting correctly:
  - [ ] dev-api
  - [ ] dev-web
  - [ ] stage-api
  - [ ] stage-web
  - [ ] prod-api
  - [ ] prod-web
- [ ] Any cruft is cleaned up (if applicable) in the NR dashboard

dlapiduz:
Removes some settings from `newrelic.ini` and moving them to env variables.
Removes double reporting by only using the `Procfile` command instead of the `run_api` instantiation. 

It requires a `NEW_RELIC_LICENSE_KEY` to be set.

Fixes #602
/cc @arowla 

Also, use Gunicorn for better performance and better reporting.

### DIFF
--- a/manifest_dev.yml
+++ b/manifest_dev.yml
@@ -5,6 +5,9 @@ applications:
 - name: api
   path: .
   host: fec-dev-api
+  env:
+    NEW_RELIC_APP_NAME: OpenFEC (development)
+    NEW_RELIC_LOG: stdout
 - name: web
   path: ../openFEC-web-app
   buildpack: python_buildpack
--- a/manifest_dev.yml
+++ b/manifest_dev.yml
@@ -6,7 +6,7 @@ applications:
   path: .
   host: fec-dev-api
   env:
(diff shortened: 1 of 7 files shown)


manifest_dev.yml
```diff
@@ -5,6 +5,9 @@ applications:
 - name: api
   path: .
   host: fec-dev-api
+  env:
+    NEW_RELIC_APP_NAME: OpenFEC (development)
```
arowla: Can we change this (and the others) to "OpenFEC API (env)"? Since we'll also be reporting from "OpenFEC Web (env)".

dlapiduz: Done!

PR #613, opened 2015-04-24
Update import scripts for golden gate.

jmcarp:
Note: This patch will fail on Travis, since the data in the repo have
diverged from the data on golden gate. This patch should be updated with
new sample data before merge.

* Disambiguate `candidate_id` column
    * Column now exists on multiple tables in join, leading to ambiguous reference
* Pull `filing_freq` from `dimcmtetpdsgn` rather than `dimcmteproperties`
* Remove `link_date` column

### DIFF
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -3,7 +3,7 @@ create materialized view ofec_candidate_history_mv as
 select
     dcp_by_period.candproperties_sk as properties_key,
     dcp_by_period.cand_sk as candidate_key,
-    dcp_by_period.cand_id as candidate_id,
+    dcp_by_period.candidate_id,
     dcp_by_period.cand_nm as name,
     dcp_by_period.record_expire_date as expire_date,
     dcp_by_period.record_load_date as load_date,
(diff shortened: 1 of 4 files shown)


arowla:
Are you working on the test dump? I believe the current test db dump was made with https://github.com/18F/rdbms-subsetter

jmcarp:
This patch worked against the production gg instance but not for the local data. I'll check out the subsetter and generate a new test set.

LindsayYoung:
The tests are tied to some particular examples that need to make it back into the test subset for the tests to pass. 

candidates- H0VA08040, P80003338, P80003338, H2CA37213, H2CA38088, P60000247 
committees- C00048587, C00048587,C00116574, C00484188, C00000422, C00111245 

We could also use mocking or another solution, but wanted to let you know.

arowla:
Thanks @LindsayYoung. If we could extract the needed data (we'd only really
need the data from the views, not all the underlying tables) for these and
model them with object factories, so that the tests aren't tied to a DB
dump, that would be A+++++.

On Mon, Apr 27, 2015 at 9:46 AM, Lindsay Young <notifications@github.com>
wrote:

> The tests are tied to some particular examples that need to make it back
> into the test subset for the tests to pass.
>
> candidates- H0VA08040, P80003338, P80003338, H2CA37213, H2CA38088,
> P60000247
> committees- C00048587, C00048587,C00116574, C00484188, C00000422,
> C00111245
>
> We could also use mocking or another solution, but wanted to let you know.
>
> —
> <https://github.com/18F/openFEC/pull/613#issuecomment-96657151>.
>



-- 
Alison Rowland
202-317-0124

jmcarp:
Agreed that mocks or factories would be much more straightforward. The only hitch is integrating factories with materialized views. Ordinarily, we would use something like `factory_boy` to insert fake rows for testing, then drop after tests are done. But we can't insert into materialized views.

What do you all think about having the tests use tables instead of views? It should be as simple as throwing a `Metadata.create_all`() into the base `setUpClass` method. It would make the tests subtly different from the application, but likely not in ways that would matter for what we're doing.

arowla:
Interesting... I'd be in favor of having the tests use tables.

What do you all think about merging this PR (with Travis not passing,
because we know it won't) to get the API working again, and then creating
another task to fix the tests?

On Mon, Apr 27, 2015 at 10:21 AM, Joshua Carp <notifications@github.com>
wrote:

> Agreed that mocks or factories would be much more straightforward. The
> only hitch is integrating factories with materialized views. Ordinarily, we
> would use something like factory_boy to insert fake rows for testing,
> then drop after tests are done. But we can't insert into materialized views.
>
> What do you all think about having the tests use tables instead of views?
> It should be as simple as throwing a Metadata.create_all() into the base
> setUpClass method. It would make the tests subtly different from the
> application, but likely not in ways that would matter for what we're doing.
>
> —
> <https://github.com/18F/openFEC/pull/613#issuecomment-96672132>.
>



-- 
Alison Rowland
202-317-0124

LindsayYoung:
I think we should not wait for tests to past this once so we can do a successful demo tomorrow.

jmcarp:
Sounds good to me. I can take a look at revising the tests.

PR #629, opened 2015-04-28
Remove extra db setup

arowla:


jmcarp:
Just to double check: should this patch include #516?

arowla:
Gaah. No. Let me cherry-pick out dc7d3a8.

PR #630, opened 2015-04-28
Changed CommitteeDetail to filter with CommitteeDetail

arowla:
Queries were being constructed with faulty where clause that tried to filter on ofec_committees_mv instead of ofec_committee_detail_mv. Fliters like 'designation' were not working at all.


jmcarp:
LGTM. We should add tests here to prevent regressions--I'll add this to my test refactor later today.

PR #635, opened 2015-04-28
Test factories

### LINKED ISSUE #614: Use factories for unit tests
The unit tests have hard-coded dependencies against the test data. This means that we need to build a new test data set whenever the underlying data change, and makes it hard to generate test cases that meet specific requirements (e.g. unicode in names). Tests should use data from factories instead of production data for greater flexibility.

jmcarp:
Refactor tests to use factories rather than real data.

* Use tables rather than materialized views for testing. Otherwise, we can't insert fake data, since views are read-only
* Wrap each test in a transaction and rollback on `tearDown`
* Clear test database after each test suite
* Use separate database for testing; defaults to `cfdm-unit-test` but can be controlled with the `SQLA_TEST_CONN` environment variable

[Resolves #614]


webservices/resources/reports.py
```diff
@@ -531,5 +532,6 @@ def get_reports(self, committee_id, reports_class, args, page_num, per_page):
             reports = reports.filter(reports_class.cycle.in_(args['cycle'].split(',')))
 
         count = reports.count()
+        import ipdb; ipdb.set_trace()
```
arowla: Looks like this got cherry-picked in here? Or needs a merge?

jmcarp: Was reconciling branches and forgot to take debugger out here. This is fixed, and I just merged #642 here to deal with merge conflicts.

arowla:
Was this intended to also contain the `idx` primary key fixes? Having a little trouble following what's going on here.

jmcarp:
@arowla I did merge in the primary key fix because the two branches kept diverging and giving me merge headaches. I'll separate if it would be better for review.

jmcarp:
Replaced by #651.

PR #651, opened 2015-04-28
Feature/test factories

jmcarp:
Replaces #635.


arowla:
This still targets master-- shouldn't it target develop?

jmcarp:
Replaced by #654.

PR #655, opened 2015-04-29
Add Travis CI and Waffle badges to docs.

jmcarp:
Add status badges to show new people that our builds are (normally) passing and which issues are up for grabs.


noahmanger:
Can we add these to the web app repo too?

PR #662, opened 2015-04-29
Feature/apidoc

jmcarp:
Replacement for #571.


webservices/resources/candidates.py
```diff
@@ -1,290 +1,131 @@
...
+
+
+filter_fields = {
+    'candidate_id',
+    'candidate_status',
+    'district',
+    'incumbent_challenge',
+    'office',
+    'party',
+    'state',
```
LindsayYoung: We need a year filter for `/candidates` is this list just for detail and history?

jmcarp: This logic is for generic filtering, meaning we want an exact match between the query parameter and the column in the database. Year filtering is more complicated and varies more across endpoints, so I left it in the `get_candidate` and `get_candidates` methods. For example, https://github.com/jmcarp/openFEC/blob/feature/apidoc/webservices/resources/candidates.py#L59-L64

LindsayYoung: Ok, that makes sense. Thanks for explaining!

LindsayYoung:
This looks good to me.

webservices/rest.py
```diff
@@ -4,23 +4,37 @@
...
+import smore.apispec
+from flask import abort
+from flask import request
+from flask import jsonify
+from flask import url_for
+from flask import render_template
+from flask import Flask
+from flask import Blueprint
 
 from flask import Flask, Blueprint, abort, request
```
arowla: There are some duplicated imports here.

jmcarp: Merge conflicts :confounded:. Fixing now.

PR #663, opened 2015-04-29
Update first and last filing dates for committees.

### LINKED ISSUE #624: Committee start date issue
One of the things I am looping back on now is using the `committee_origination_ date` in `create_committee_view.sql` and `create_committe_view` we are using the most recent load date, when we would want the oldest load date for that committee in dimcmteproperties. `` and start date of the committee.

In theory, they should be giving us the `committee_origination_ date` in `dimcmteproperties` now. Is this something you can see in Golden Gate @jmcarp? (The older version of the data we got from the vpn had that column with nothing populated so we were calculating it ourselves.)

If we keep using the wrong origination date, the result filtering by year will be incorrect.

### LINKED ISSUE #649: Year filtering on /committees is broken
Year filtering for /committees works for me locally but not on http://api.open.fec.gov/?year=2012&api_key=xxx

This is needed for filtering the committee results in the browse view.

jmcarp:
Note: Based on the recommendation of FEC, this patch only considers
receipt dates from F1 filings. Since we're doing an `INNER JOIN` to combine
`dimcmte` and `dimcmteproperties`, committees that have never filed an F1
are excluded. We could switch to a `LEFT JOIN`, but then we'd have a bunch
of committees with `NULL` dates and values from `dimcmteproperties`.

Thoughts @lindsayyoung @arowla?

[Resolves #624]
[Resolves #649]

### DIFF
--- a/data/sql_updates/create_committee_detail_view.sql
+++ b/data/sql_updates/create_committee_detail_view.sql
@@ -94,7 +94,7 @@ select distinct
 from dimcmte
     inner join dimcmtetpdsgn dd using (cmte_sk)
     -- do a DISTINCT ON subselect to get the most recent properties for a committee
-    inner join (
+    left join (
         select distinct on (cmte_sk) * from dimcmteproperties
             where form_tp = 'F1'
             order by cmte_sk, receipt_dt desc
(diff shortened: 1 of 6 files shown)


arowla:
My inclination would be to go with the LEFT JOIN, so that we aren't silently eliminating a bunch of committees at the database level, and then deal with the ones with NULL dates in code. @LindsayYoung, what do you think?

jmcarp:
I went ahead and made that change--we can always revert if we think better of it later.

LindsayYoung:
I agree that we don't want to leave out filers! Thanks!

PR #648, opened 2015-04-28
[WIP]Feature/remove nesting refactor

LindsayYoung:
this replaces pr #516 

I am also testing gitflow.

PR #665, opened 2015-04-30
Feature/api version

### LINKED ISSUE #44: Add version to API urls
As we near deploying our API into the public, we should add the version into the URL so that we won't cause regressions in users' applications if we make significant changes.

jmcarp:
Prefix API URLs with the current API version, starting with `v1`.

Note: Depends on https://github.com/18F/openFEC-web-app/pull/108

[Resolves #44].

### DIFF
--- a/webservices/common/models.py
+++ b/webservices/common/models.py
@@ -43,6 +43,7 @@ class Candidate(BaseModel):
     party_full = db.Column(db.String(255))
     state = db.Column(db.String(2))
     name = db.Column(db.String(100))
+    committees = db.relationship('CandidateCommitteeLink', backref='candidates')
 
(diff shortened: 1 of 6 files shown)


webservices/common/models.py
```diff
@@ -43,6 +43,7 @@ class Candidate(BaseModel):
     party_full = db.Column(db.String(255))
     state = db.Column(db.String(2))
     name = db.Column(db.String(100))
+    committees = db.relationship('CandidateCommitteeLink', backref='candidates')
```
arowla: What are these additions about?

jmcarp: I actually deleted these lines by accident in an earlier PR--this commit is reverting to the earlier version. They're going to go away with #648, but we need them for now.

PR #674, opened 2015-05-01
Use `election_yr` instead of `load_date`.

jmcarp:
After the switch to Golden Gate, the meaning of the `load_date` column
no longer indicates the election year correctly. Currently, all load dates
in the `dimcandproperties` table fall within 2015. This patch uses the
`election_yr` column instead.

@LindsayYoung, this resolves the issue you caught earlier today about candidate histories only being found for the 2016 cycle. I wasn't sure which column to use instead of `load_date`, which now basically uninformative, but `election_yr` seemed like the only reasonable option in `dimcandproperties`. @LindsayYoung @arowla let me know if you have any concerns.


LindsayYoung:
Thanks for fixing that. Though, I do want the correct load dates because multiple reports can be turned in during a year and we want the most recent. Did you talk to Paul about it or did you just fix it?

PR #677, opened 2015-05-01
Fix duplicates in committees views.

jmcarp:
The `dimcmtetpdsgn` table sometimes contains multiple entries per
`cand_sk`, possibly a change since the switch to Golden Gate. This patch
uses only the most recent entry in `dimcmtetpdsgn` when constructing the
materialized views for committees.

This should fix the duplicate committees issue that @LindsayYoung noticed earlier today.
attn @LindsayYoung @arowla


LindsayYoung:
Sounds good. We would want a table with most recent info for this purpose
and when we add committee history we would want multiples since they can
change over time.

PR #678, opened 2015-05-01
Use most recent candidate properties in materialized views.

jmcarp:
Continuing in the same vein as #674, this patch sorts by
`candproperties_sk` instead of the now-uninformative `load_date` when
building views that pull from the `dimcandproperties` table.

@LindsayYoung I think this resolves the issue you brought up in conversation on #674 about getting the most recent candidate properties for the candidate history view.


LindsayYoung:
Thank you, Josh!

PR #680, opened 2015-05-01
Use buildpack-multi for webapp.

jmcarp:
This is necessary for a parallel change to the webapp that builds assets
remotely. The webapp needs to use both the python buildpack (to run the
app) and the node buildpack (to compile the assets).


jmcarp:
Depends on https://github.com/18F/openFEC-web-app/pull/115.

PR #684, opened 2015-05-01
Harmonize committee names across views.

### LINKED ISSUE #676: Typeahead search showing different committees than committee page
When I do a typeahead search for Brownback, a committee named "Sam Brownback for U S Senate" is listed as one of the options.  However, when I search for Brownback on the Committees page, this is not a listed option.  If I select "Sam Brownback for U S Senate" from the dropdown, it takes me to the page for Friends of Brownback (C00320051).

screen shot 2015-05-01 at 10 31 51 am

Chrome 42.0.2311.90 (64-bit)

jmcarp:
We currently build committee names different in the committee and
committee detail views versus the full-text search view. This means that
it's possible to find committee names in the typeahead search, click on
a name, and not find that name in the committee results page. This patch
updates the full-text search view to use the same logic to build
committee names as used in the other committee views. This also replaces
some indirection using `load_order` with `select distinct` plus `order
by`, which is more consistent with the rest of the import scripts.

Thanks @rdpiazza reporting!
attn @LindsayYoung @arowla

[Resolves #676]

### DIFF
--- a/data/sql_updates/create_fulltext_table.sql
+++ b/data/sql_updates/create_fulltext_table.sql
@@ -45,31 +45,28 @@ drop materialized view if exists name_search_fulltext_mv_tmp;
 create materialized view name_search_fulltext_mv_tmp as
 with
     ranked_cand as (
-        select
+        select distinct on (c.cand_sk)
             p.cand_nm as name,
             to_tsvector(p.cand_nm) as name_vec,
             c.cand_id,
-            row_number() over
-                (partition by c.cand_id order by p.load_date desc)
-                as load_order,
             null::text as cmte_id,
             o.office_tp as office_sought
         from dimcand c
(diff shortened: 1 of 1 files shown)


noahmanger:
Does this deal with the issue of the date range on the typeahead results being different from the date range on the browse committee page?

jmcarp:
@noahmanger this patch handles differences in names (e.g. "Friends of Brownback" in the typeahead and "Brownback for Senate" in the detail page) but not years. Is the date range issue described somewhere else, or was it part of #676? The typeahead results don't include date information AFAIK.

noahmanger:
Ok cool. That's what I thought but was just checking. Yeah, the typeahead results don't include it, but there's some logic behind it that is different from the default years on the browse pages which leads to different results.

Issue #618, opened 2015-04-27
Missing data


khandelwal:
The fields being empty is confusing. Is the data missing? Or is that an error in the app I'm using?

![screen shot 2015-04-27 at 12 32 07 pm](https://cloud.githubusercontent.com/assets/636752/7352616/d3d5b7b4-ecda-11e4-8eb5-d08bbaab8ae7.png)

LindsayYoung:
We don't have any reports for him. 
https://api.open.fec.gov/committee/C00168492?api_key=xxx
https://api.open.fec.gov/committee/C00168492/reports?api_key=xxx
https://api.open.fec.gov/committee/C00168492/totals?api_key=xxx

khandelwal:
Right. However, what's more important here is to display that fact to the user. It being empty as displayed above leads to confusion.

Maybe put 'No Data Available' in the right places?

LindsayYoung:
Agreed, I was trying to confirm that it was not a data error, since we have a few bugs like that, which I am trying to hunt now.

This guy in particular has only filed a pre-convention report and right now, we are only showing data from "F3s"

noahmanger:
So is it possible for us to show the "No Data Available" message?

On Mon, Apr 27, 2015 at 12:09 PM, Lindsay Young <notifications@github.com>
wrote:

> Agreed, I was trying to confirm that it was not a data error, since we
> have a few bugs like that, which I am trying to hunt now.
>
> This guy in particular has only filed a pre-convention report and right
> now, we are only showing data from "F3s"
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/openFEC/issues/618#issuecomment-96783031>.
>



-- 
Noah Manger
18F <http://18f.gsa.gov> | Work: 415-696-6146

LindsayYoung:
I think this is actually brings up a larger issue too, I would like it if the "browse" candidates or committee views with no filters should return something more interesting than the A's of the alphabet. 

Maybe we could the browse view ordered by the top 100 spenders for the cycle or something interesting? That would make it less likely for people to stumble on random outliers.

This is not necessary for launch, but I think it would be nice.

LindsayYoung:
I think we could triggered a no results template when `results=[]` but, I spend less time on the front so it might be more complicated than that for some reason.

noahmanger:
+1 to having a better set of candidates that show up initially on Browse.

We do have a "No Results" message that shows up in certain situations like
[...]

Issue #592, opened 2015-04-22
Manual QA Test Pages on Production Instance


rjmajma:
The FEC Team and 18F team will each do manual tests on the production instance. We'll note any thing that doesn't work as the user would expect it to.

The 18F team may identify some selenium tests that would be worth running as a separate ticket.

LindsayYoung:
Dependency is getting the committee and candidate pages in a final-ish state.

LindsayYoung:
So far this has resulted in issues:
#620 #621 #623

PR #688, opened 2015-05-02
Feature/reports by cycle

LindsayYoung:
This adds 3 endpoints that will be useful for testing.
   `/reports/house-senate/cycle`
   `/reports/pac-party/cycle`
   `/reports/presidential/cycle`

This will allow us to page through reports by date and compare results with the bulk data. 

I feel like there might be a better way to do the url routing so please make suggestions if you think of something.

This is related to issues #618 and #592


jmcarp:
Could we switch the `committee_type` and `cycle` arguments from path parameters to query parameters? Typically, only collection names and entity IDs get included in the path, and optional search arguments live in the query:
`/reports`
`/reports?committee_type=P`
`/reports?cycle=2014`
`/reports?committee_type=P&cycle=2014`

LindsayYoung:
I wanted something that I could quickly page through the API results for reports. I thought it would be quicker to make one for each underlying table. Also, since the reports are different that would make the results more consistent. But, if you want to tweak it so it works like you described, that would be great.  That would be a useful endpoint. 

Also, I did not mean for anyone to review this on the weekend. I hope I didn't disturb you.

Issue #619, opened 2015-04-27
Force HTTPS by default


seanherron:
Dependent on https://trello.com/c/mKH3vlfk/34-add-hsts-and-redirect-http-https-in-gorouter

dlapiduz:
You can use [djangosecure](https://pypi.python.org/pypi/django-secure) like Foia: https://github.com/18F/foia-hub/blob/master/foia_hub/settings/production.py#L15.

We are not doing the router stuff for a few weeks at least.

LindsayYoung:
Ok, we are using flask so maybe something like this -
https://github.com/kennethreitz/flask-sslify

PR #691, opened 2015-05-04
Force HTTPS in production.

jmcarp:
Enable HTTPS redirect implemented in https://github.com/18F/openFEC-web-app/pull/117 in prod space. Related to #619.

### DIFF
--- a/manifest_prod.yml
+++ b/manifest_prod.yml
@@ -21,6 +21,7 @@ applications:
   env:
     NEW_RELIC_APP_NAME: OpenFEC Web (production)
     FEC_WEB_API_URL: https://api.open.fec.gov/
+    FEC_FORCE_HTTPS: true
     # These must be set manually via `cf set-env`
     # FEC_WEB_PASSWORD
     # FEC_WEB_USERNAME


konklone:
@jmcarp Is openFEC using the same system as openFEC-web-app? Just setting the environment variable in a separate app wouldn't be enough.

I ask because http://api.open.fec.gov/developers does not force a redirect to HTTPS.

```
$ curl --head http://api.open.fec.gov/developers

HTTP/1.1 200 OK
```

I'm looking over the code now and may file a PR soon.

konklone:
Ah ha, I see in https://github.com/18F/openFEC/commit/eacb26d0962bdb6ee18ed8c59a0c0c0e49fe856a that this PR was always intended to keep the production web app environment variables deployed correctly, not to trigger a forced HTTPS for the API specifically.

One thing that likely kept the dev team from noticing the issue is that https://github.com/18F/openFEC-web-app/pull/117 added an `includeSubDomains` policy to `open.fec.gov`, which would cause requests to `api.open.fec.gov` to automatically redirect to HTTPS -- *if* the user had visited `https://open.fec.gov` once before.

konklone:
I've been working at it for a while, but I've not been able to figure out the correct place to put this, or it's conflicting with something else. I tried doing the same work I did in https://github.com/18F/openFEC-web-app/pull/117 in `webservices/rest.py`, but it wasn't triggering. Something y'all can take a stab at?

jmcarp:
So SSL termination and redirection for the API is handled by the API Umbrella. Our Umbrella configuration was supposed to use the redirect to HTTPS option, but for some reason it wasn't. I just updated the configuration and verified that the redirect to HTTPS is in place.

konklone:
Oh, that's so much simpler! Thank you!

PR #693, opened 2015-05-04
Feature/better election years#657

LindsayYoung:
Getting election year for candidates from dimcandoffice election year, which seems like a more reliable place to get that.

### DIFF
--- a/data/sql_updates/create_candidate_detail_view.sql
+++ b/data/sql_updates/create_candidate_detail_view.sql
@@ -51,7 +51,7 @@ from dimcand
             order by cand_sk, candproperties_sk desc
     ) cand_p_most_recent using(cand_sk, cand_id)
     left join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)
-    inner join dimoffice using (office_sk)
+    left join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
 group by
     dimcand.cand_sk,
--- a/data/sql_updates/create_candidate_detail_view.sql
+++ b/data/sql_updates/create_candidate_detail_view.sql
@@ -50,7 +50,7 @@ from dimcand
         select distinct on (cand_sk) * from dimcandproperties
             order by cand_sk, candproperties_sk desc
     ) cand_p_most_recent using(cand_sk, cand_id)
-    left join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)
+    left join dimcandoffice co on co.cand_sk = dimcand.cand_sk
     left join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
 group by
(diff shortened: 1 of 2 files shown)


LindsayYoung:
This is woking locally for me using the dev rds. I think the dev api needs to  be redeployed after this merge.

data/sql_updates/create_candidates_view.sql
```diff
@@ -29,8 +29,7 @@ select
     max(candprops.cand_nm) as name
 from dimcand
     left join (select distinct on (cand_sk) cand_sk, election_yr, cand_status, ici_code from dimcandstatusici order by cand_sk, election_yr desc) csi_recent using (cand_sk)
-    left join dimcandstatusici csi_all using (cand_sk)
-    inner join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)  -- only joined to get to dimoffice
+    inner join dimcandoffice co on co.cand_sk = dimcand.cand_sk --and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)
```
LindsayYoung: This seemed to work, but tell me @jmcarp please confirm this still gives us the most recent result. I want to make sure I am not being rash.

jmcarp: This isn't a new change in this PR, but is it expected to join `dimcand` and `dimcandoffice` with a `LEFT JOIN` in `ofec_candidates_mv` but an `INNER JOIN` in `ofec_candidate_detail_mv`? My possibly naive expectation is that we would use the same joins so that every candidate in one view also shows up in the other.

jmcarp: Very similar to the last question, but how come we're filtering on `csi_recent.election_yr is null or co.cand_election_yr = csi.election_yr` in the detail view but not in the list view?

data/sql_updates/create_candidates_view.sql
```diff
@@ -39,8 +38,8 @@ from dimcand
     ) candprops on dimcand.cand_sk = candprops.cand_sk
 group by
     dimcand.cand_sk,
-    dimcand.cand_id,
-    csi_recent.election_yr
+    dimcand.cand_id
+    -- csi_recent.election_yr
```
jmcarp: Could you delete any commented-out code?

LindsayYoung: Yes, sorry I missed that

data/sql_updates/create_candidate_detail_view.sql
```diff
@@ -50,9 +50,8 @@ from dimcand
         select distinct on (cand_sk) * from dimcandproperties
             order by cand_sk, candproperties_sk desc
     ) cand_p_most_recent using(cand_sk, cand_id)
-    left join dimcandstatusici csi_all using (cand_sk)
-    left join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)  -- only joined to get to dimoffice
-    inner join dimoffice using (office_sk)
+    left join dimcandoffice co on co.cand_sk = dimcand.cand_sk and (csi_recent.election_yr is null or co.cand_election_yr = csi_recent.election_yr)
```
jmcarp: Could we drop the `and...` here?
`left join dimcandoffice co on co.cand_sk = dimcand.cand_sk`

PR #694, opened 2015-05-04
Move `year` param from path to query.

jmcarp:
* Move `year` param from path to query
* DRY up `ReportsView` helper methods
* Add tests for `/reports` endpoint

After thinking about this more, it seemed like `committee_type` should live in the path because it's a required argument (for now), and `cycle` should live in the query because it's optional. This patch also makes the behavior of the `cycle` argument consistent for the two routes to the reports endpoint (i.e., both routes now accept a comma-separated list of cycle years).

@LindsayYoung: This PR is targeting your feature branch, not develop, so it's safe to merge if you're all right with the changes.


LindsayYoung:
This looks good. I notice that it can still only look at one report type at a time. I wanted to confirm you still wanted to do that change.

jmcarp:
@LindsayYoung after looking at this more, I think it's fine to leave the report type argument as it is. I'd like to look into returning all report types from one endpoint at some point, but it's a little tricker than I realized, so I think we should defer for now.

LindsayYoung:
Sounds good. I did not see a quick and easy way to merge those either.

PR #703, opened 2015-05-06
Move form filter to correct join.

jmcarp:
Based on correspondence with FEC, we should only use F1 filings to
determine original filing dates, but we can use the most recent filings
regardless of form type for other committee properties. I got this
almost right in #678 but applied the F1 filter to the most recent
committee properties rather than the original properties. This patch
moves the filter to the right join.

Thanks @lindsayyoung and @arowla for identifying the issue.

### DIFF
--- a/data/sql_updates/create_committee_detail_view.sql
+++ b/data/sql_updates/create_committee_detail_view.sql
@@ -99,11 +99,11 @@ from dimcmte
     -- do a DISTINCT ON subselect to get the most recent properties for a committee
     left join (
         select distinct on (cmte_sk) * from dimcmteproperties
-            where form_tp = 'F1'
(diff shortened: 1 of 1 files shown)


LindsayYoung:
This looks right to me. 

I will document the logic here- 

The F1 is a statement of organization for committees, F1Z is an amended F1.

We want so we want all the amendments in the data we return so we don't want to filter it out from the query. We know that we want the receipt date of the earliest filing to use as a start date for committee filtering. 

Also, if I understand correctly, there may be M1/M1z in that table that are labeled as F1/F1Z but that is an error that is going to be corrected. An M1 is the qualifying paperwork for a PAC that gives it multi-candidate status.

PR #714, opened 2015-05-06
Make subset work outside my environment.

jmcarp:
Fix problems discovered by @arowla such that subsets can be created and imported regardless of environment. @arowla would you mind checking this out and verifying that it works on your machine when you have time? Thanks / sorry.

"But it worked on my laptop™"

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -11,8 +11,11 @@ FULL_TABLES = [
     'dimreporttype',
 ]
 EXCLUDE_TABLES = [
+    '*_mv',
+    '*_tmp',
     'sched_a',
     'sched_b',
+    'ofec_two_year_periods',
 ]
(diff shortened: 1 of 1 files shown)


arowla:
Just tried it again on a fresh DB.

```
[snip]
CREATE INDEX
Time: 2.945 ms
REVOKE
Time: 0.303 ms
ERROR:  role "fec" does not exist
Time: 17.116 ms
ERROR:  current transaction is aborted, commands ignored until end of transaction block
Time: 0.194 ms
[snip]```

jmcarp:
Updated based on discussion with @arowla in 4ab692dc68135447bf93ddc0f60428d782388a02. Schemas are now imported using `pg_restore`, so that imports will work with auto-commit off.

Issue #701, opened 2015-05-05
Make sure all endpoints are filterable by 2-year-period


LindsayYoung:
This will make sure all of our data is consistent on the web app, the bulk of those changes will be done in #698  but we will also need to add two_year_period to committees from cand_election_yr in dimcmteproperties

PR #718, opened 2015-05-08
Feature/restrict dates

### LINKED ISSUE #698: Date Limit all views to recent data
Filter out psychedelic data from the 70's and cover 1980-present.

This will fix data quality issues from old ids that were reused. 

@arowla There is an email about this yesterday, but this seems like the most straight forward fix.

jmcarp:
Limit views to recent data

* Filter views by a configurable `START_YEAR` as described in #698.
* Add array of `cycles` to candidate and committee views for filtering in preparation for #701.
* Use GIN indices for array columns

[Resolves #698]


jmcarp:
Note: Not ready for merge. Need to determine whether this patch uses all the appropriate date columns for filtering and election cycles.

data/sql_updates/create_totals_presidential_view.sql
```diff
@@ -42,7 +42,8 @@ from
     left join dimdates start_date on cvg_start_dt_sk = start_date.date_sk and cvg_start_dt_sk != 1
     left join dimdates end_date on cvg_end_dt_sk = end_date.date_sk and cvg_end_dt_sk != 1
 where
-    p.expire_date is null or p.expire_date > date_trunc('day', now())
+    (p.expire_date is null or p.expire_date > date_trunc('day', now()))
```
LindsayYoung: For the fact tables, If something has an expire date at all, that means that it has been amended. Since we are looking for totals, we don't want to add up the expired records. 

I don't think this is wrong, but you could just check if the expire_date is null, if you wanted to. This would apply to all the totals tables.

data/sql_updates/create_candidate_detail_view.sql
```diff
@@ -50,12 +51,16 @@ from dimcand
         select distinct on (cand_sk) * from dimcandproperties
             order by cand_sk, candproperties_sk desc
     ) cand_p_most_recent using(cand_sk, cand_id)
-    left join dimcandoffice co on co.cand_sk = dimcand.cand_sk
+    left join dimcandoffice co using (cand_sk)
     left join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
+    left join dimcandproperties dcp using (cand_sk)
+    where dcp.election_yr >= :START_YEAR
```
LindsayYoung: We want start year - 1 because we are looking at election year and if there was a special election in the first year of the cycle this wouldn't catch it.

data/sql_updates/create_candidates_view.sql
```diff
@@ -28,17 +29,25 @@ select
...
+    left join dimcandoffice co using (cand_sk)
     inner join dimoffice using (office_sk)
     inner join dimparty using (party_sk)
     left join (
         select distinct on (cand_sk) cand_sk, cand_nm from dimcandproperties
             order by cand_sk, candproperties_sk desc
-    ) candprops on dimcand.cand_sk = candprops.cand_sk
+    ) candprops using (cand_sk)
+    left join dimcandproperties dcp using (cand_sk)
+    where dcp.election_yr >= :START_YEAR
```
LindsayYoung: When comparing start year to an election year, make sure the beginning of the cycle is included

data/sql_updates/create_candidate_detail_view.sql
arowla: Since this is configurable, we could put a comment in the config to always start with the actual minimum year (not even-numbered cycle) you want, to keep the SQL simple.

jmcarp: :+1:

data/sql_updates/create_totals_presidential_view.sql
jmcarp: @arowla do you have a preference here? Based on @LindsayYoung's description, it sounds like we don't need to check for expired dates that are in the future, and I confirmed that none of the facts tables have expired dates that haven't already happened:
```
fec=> select count(*) from factpresidential_f3p where expire_date > date_trunc('day', now());
 count
-------
     0
```
Any reservations about dropping the `expire_date` part of the query, or should we leave it in?

webservices/tests/test_integration.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class TestViews(common.IntegrationTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestViews, cls).setUpClass()
+        manage.update_schemas(processes=1)
+
+    def test_update_schemas(self):
```
jmcarp: Seemed like an opportune time to sneak in an extra test. This will import the test subset, build the materialized views, and verify that the views have been populated (or at least contain more than zero rows). Practically, this means that the build will fail if anybody checks in a SQL script that doesn't run.

@arowla @LindsayYoung this should make it harder for me to check in SQL errors :smiley:

LindsayYoung: nice addition!

PR #719, opened 2015-05-11
Feature/cycle filter

### LINKED ISSUE #701: Make sure all endpoints are filterable by 2-year-period

### LINKED ISSUE #706: Use repeated query parameters for list arguments
Several API endpoints accept query parameters that can either be scalars (`year=2015`) or lists (`year=2015,2016`). This means developers have to remember which parameters can be list-like and split on commas as appropriate. We also miss out on simple type validation with flask-restful's reqparse (or with webargs if #662) is merged in its current state): we're forced to validate list-like arguments as strings, and then manually validated each item after comma-splitting later on.

I propose using repeated query parameters for list-like arguments (`/candidates?year=2015&year=2016`). It gives us a simpler API to work with internally (just add `action='append'` for reqparse or `list=True` to webargs), and we get validation as normal, without having to write and debug extra code. This is also, in my limited experience, the more "standard" way of handling list-like fields for GET requests.

If folks are all right with the proposed change, I'd be happy to implement it--I don't see this taking much time. I'm also open to being convinced that the current approach is better :smiley:!

jmcarp:
Make all resources filterable by election cycle, and year where appropriate.

* Make reports filterable by cycle and year
* Make totals filterable by cycle
* Add cycle filters to candidate and committee resources
* Remove year filter on candidates until load and expire dates are valid
* Move default cycle filtering to web app
* Use repeated query parameters for list-like arguments
* Update unit tests

[Resolves #701]
[Resolves #706]

### DIFF
--- a/webservices/resources/committees.py
+++ b/webservices/resources/committees.py
@@ -1,7 +1,7 @@
 from flask.ext.restful import Resource, reqparse, fields, marshal_with, inputs, marshal
 from webservices.common.models import db, Candidate, Committee, CandidateCommitteeLink, CommitteeDetail
-from webservices.common.util import default_year, Pagination
-from sqlalchemy.sql import text, or_
+from webservices.common.util import Pagination
+from sqlalchemy.sql import text, or_, and_
 from sqlalchemy import extract
 from datetime import date
 
@@ -103,6 +103,21 @@ committee_list_fields = {
 }
 
 
+def filter_year(model, query, years):
+    return query.filter(
+        or_(*[
+            and_(
+                or_(
+                    extract('year', model.last_file_date) >= year,
+                    model.last_file_date == None,
+                ),
+                extract('year', model.first_file_date) <= year,
+            )
+            for year in years
+        ])
+    )  # noqa
+
+
 class CommitteeList(Resource):
     parser = reqparse.RequestParser()
     parser.add_argument('q', type=str, help='Text to search all fields for')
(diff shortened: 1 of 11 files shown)


jmcarp:
Note: Parallel changes to webapp in https://github.com/18F/openFEC-web-app/pull/156.

webservices/resources/committees.py
```diff
@@ -103,35 +103,50 @@
...
+
 class CommitteeList(Resource):
     parser = reqparse.RequestParser()
     parser.add_argument('q', type=str, help='Text to search all fields for')
-    parser.add_argument('committee_id', type=str, help="Committee's FEC ID")
-    parser.add_argument('candidate_id', type=str, help="Candidate's FEC ID")
-    parser.add_argument('state', type=str, help='Two digit U.S. State committee is registered in')
+    parser.add_argument('committee_id', type=str, action='append', help="Committee's FEC ID")
+    parser.add_argument('candidate_id', type=str, action='append', help="Candidate's FEC ID")
+    parser.add_argument('state', type=str, action='append', help='Two digit U.S. State committee is registered in')
```
arowla: Nitpick, not related to this task: the help text should say "two character", not "two digit".

jmcarp: Updated. Once the API docs are ready, I think we should have a thorough review of the help text.

PR #723, opened 2015-05-11
Adding fields that were missing from reports and totals

LindsayYoung:
Making sure we have everything


arowla:
Should this still be WIP?

LindsayYoung:
Planning on finishing presidential tomorrow and giving it a once over

LindsayYoung:
I added missing fields and did some modest name standardization. The tests aren't passing but I think things might be missing from the sample data.

LindsayYoung:
No, the tests are really failing. Looking at that now.

jmcarp:
LGTM :+1:

PR #733, opened 2015-05-12
Proposed implementation of continuous deployment.

jmcarp:
* Store encrypted Cloud Foundry credentials in `.travis.yml`
* Push to Cloud Foundry on successful builds, ignoring pull requests
* Auto-detect Cloud Foundry space from branch name if not provided

If this implementation makes sense, I'll add a corresponding `deploy` task to the webapp, and split out the CF manifest files between the repos. Paging @arowla :bell:


tasks.py
```diff
@@ -80,3 +81,54 @@ def add_hooks():
...
+        ('--u', '$FEC_CF_USERNAME'),
+        ('--p', '$FEC_CF_PASSWORD'),
+        ('--o', 'fec'),
+        ('--s', space),
+    )
+    login = 'cf login {0}'.format(' '.join(' '.join(arg) for arg in args))
+    run(login, echo=True)
+
+    # Push
+    push = 'cf push -f manifest_{0}.yml'.format(space)
```
arowla: Are we missing `cf target` somewhere in here?

jmcarp: The work that `cf target` would do got subsumed into `cf login`--we're passing `--o` and `--s` there. I can split into more commands (`api`, `login`, `target`, `push`) if that would be clearer, but this implementation does work.

arowla:
Sounds good.

tasks.py
arowla: Just saw that the space selection happens in the login step. Nevermind.

PR #748, opened 2015-05-13
Trying out codecov.io integration

arowla:
Analagous to 18F/openFEC-web-app#171

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
 language: python
+
 python:
   - "3.4"
+
 addons:
   postgresql: "9.4"
+
+before_install:
+    pip install codecov
+
+after_success:
+    codecov
+
 before_script:
   - psql -c 'create database "cfdm-unit-test";' -U postgres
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,19 @@
...
+    pip install codecov
+
+after_success:
+    codecov
+
 before_script:
   - psql -c 'create database "cfdm-unit-test";' -U postgres
   - pip install -r requirements.txt
+
 script: nosetests
```
jmcarp: I think we need to tell nose to output coverage metrics for this to work, e.g. `nosetests --with-coverage --cover-package webservices`. Unless codecov is just magic.

arowla: In the codecov repo I see a `nose.cfg`... maybe it uses that somehow? https://github.com/codecov/codecov-python/blob/master/nose.cfg

arowla: Nevermind, you are probably right. https://travis-ci.org/18F/openFEC/builds/62447295

PR #752, opened 2015-05-14
Fix name of committee fulltext view.

jmcarp:
* Fix typo in view name
* Add regression test

This patch fixes a three-character typo ("dimcmte_fulltext" becomes "dimcmte_fulltext_mv") and adds a regression test to make sure we're covering fulltext search on committees.

### DIFF
--- a/tests/committee_tests.py
+++ b/tests/committee_tests.py
@@ -2,6 +2,7 @@ import urllib
 import datetime
 import functools
 
+import sqlalchemy as sa
 from marshmallow.utils import isoformat
 
 from tests import factories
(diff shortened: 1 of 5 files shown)


tests/committee_tests.py
```diff
@@ -156,13 +169,13 @@ def test_committee_filters(self):
 
         # checking one example from each field
         filter_fields = (
-            # ('committee_id', ['bartlet', 'ritchie']),
-            # ('state', ['CA', 'DC']),
+            ('committee_id', ['bartlet', 'ritchie']),
```
LindsayYoung: I think that the test works fine, but having the committee_id be a name makes it a little confusing for me to read.

jmcarp: Good point, will change now. It's just harder to talk about the West Wing without using names!

PR #758, opened 2015-05-14
Feature/round 2 adding reports totals

LindsayYoung:
These have been tricky to find and fix because everything needs to be right in 3 places to get the information through the app.


webservices/schemas.py
```diff
@@ -529,12 +533,14 @@ class TotalsPresidentialSchema(TotalsSchema):
     individual_itemized_contributions = ma.fields.Integer()
     loan_repayments_made = ma.fields.Integer()
     loans_received_from_candidate = ma.fields.Integer()
+    loand_recieved = ma.fields.Integer()
```
jmcarp: Should this be `loans_received`?

LindsayYoung: Thank you!!!!! fixing now

PR #767, opened 2015-05-14
Fixing incumbent

LindsayYoung:
there are some holes in this data so we want the lat time it was updated

The example we have been using for testing is Nancy Pelosi

PR #769, opened 2015-05-14
Harmonize incumbent status across views.

jmcarp:
Some rows in `dimcandproperties` have non-null values for `cand_ici_cd`
and null values for `cand_ici_desc`. We already build incumbent
description values manually in the candidate list view, but not in the
candidate detail view. The two views also pull incumbent status values
differently. This patch updates both views to use the correct logic.

@LindsayYoung, I completely forgot about this when I merged #767. My mistake! In other news, this appears to finally fix the incumbent status for Nancy Pelosi.


LindsayYoung:
Thanks for remembering that. This week has left me totally brain dead!

PR #774, opened 2015-05-15
Feature/narrative docs

jmcarp:
Add narrative descriptions to API docs. This patch largely borrows the description text from the markdown files in `documentation`, which are also deleted here, since we shouldn't need them anymore. Folks can check out these changes to see them in action, but the updates look something like this:

screen shot 2015-05-15 at 1 13 31 pm


webservices/docs.py
```diff
@@ -0,0 +1,83 @@
...
+COMMITTEE_DETAIL = '''
+Fetch detailed information about a committee.
+
+For further description, see `/v1/committees`.
+'''
+
+REPORTS = '''
+Fetch key information about a committee's financial reports.
+
+As part of each financial report, committees must provide a summary of their financial
```
LindsayYoung: It  should be "some committees" must provide a summary of their financial activity. Because not FEC filers have to file form 3.

jmcarp:
Merged revisions from @LindsayYoung. Anybody else want to review before we call this good?

PR #778, opened 2015-05-15
Want a list of years, not just most recent.

LindsayYoung:
Modeling this off the election cycles behavior, which is working.

### DIFF
--- a/data/sql_updates/create_candidates_view.sql
+++ b/data/sql_updates/create_candidates_view.sql
@@ -14,7 +14,7 @@ select
         else 'Unknown' end as candidate_status_full,
     max(dimoffice.office_district) as district,
     max(co.cand_election_yr) as active_through,
-    array_agg(distinct dcp.cand_election_yr)::int[] as election_years,
+    array_agg(distinct dcp.election_yr)::int[] as election_years,
     array_agg(distinct(dcp.election_yr + dcp.election_yr % 2))::int[] as cycles,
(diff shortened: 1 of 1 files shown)


data/sql_updates/create_candidates_view.sql
```diff
@@ -14,7 +14,7 @@ select
         else 'Unknown' end as candidate_status_full,
     max(dimoffice.office_district) as district,
     max(co.cand_election_yr) as active_through,
-    array_agg(distinct co.cand_election_yr)::int[] as election_years,
+    array_agg(distinct dcp.cand_election_yr)::int[] as election_years,
```
jmcarp: That column doesn't exist on `dimcandproperties`.

LindsayYoung: Yea, it is just election_year. sorry

jmcarp:
A few thoughts about this:

* In some sense, this was fixed in #770, which restored multiple values for `election_years` for both candidate views
* If `dimcandproperties.election_yr` is a better choice that `dimcandoffice.cand_election_year`, then I'm in favor of this patch, in which case we should also do the following:
    * Also use `dimcandproperties.election_yr` to calculate `active_through`
    * Remove the now-unnecessary join on `dimcandoffice`
    * Apply all of the above changes to the candidate detail view as well

LindsayYoung:
I just refreshed on dev and got the right results. was the problem that the fix in #770 just hadn't deployed yet?

PR #783, opened 2015-05-18
Changed staging branch from 'master' to 'release'

arowla:
Realized that git flow uses a branch called 'release' to prepare releases before it gets merged into master AND tagged... so we should probably host the release branch on staging.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -136,7 +136,7 @@ def _detect_space(branch=None, yes=False):
 
 DEPLOY_RULES = (
     ('prod', _detect_prod),
-    ('stage', lambda _, branch: branch == 'master'),
+    ('stage', lambda _, branch: branch == 'release'),
     ('dev', lambda _, branch: branch == 'develop'),
 )
 


tasks.py
```diff
@@ -136,7 +136,7 @@ def _detect_space(branch=None, yes=False):
 
 DEPLOY_RULES = (
     ('prod', _detect_prod),
-    ('stage', lambda _, branch: branch == 'master'),
+    ('stage', lambda _, branch: branch == 'release'),
```
jmcarp: The release branches created by git-flow are prefixed with "release" but also include more text: e.g., "release/v0.1.1". This should probably be `lambda _, branch: branch.startswith('release')`.

jmcarp: Actually, thinking about this a bit, release branches aren't the only way for code to get to master--we also have hotfixes to deal with. We could add hotfixes to the `lambda` here, but I'm not sure if that would capture the workflow correctly. For example:

* I submit a hotfix from my own fork
* The build passes, but since I'm not pushing to 18F, no deploy is triggered
* The fix gets merged into master with a release tag and is deployed to production

arowla: Won't there be a hotfix branch on the upstream that hotfixes would get merged into? Or is that not how they traditionally work?

We'll also need to consider whether we want to have hotfix branches deployed onto staging first... or do you think hotfixes are usually enough of an emergency situation to merit going straight to production?

Can it work like this? Most recent commit to `[hotfix|master]', untagged, goes to staging. Most recent commit to '[hotfix|master]', tagged for a release, goes to production. And when I say "most recent," it's just to imply how this works over time, but in reality, it just means "a commit". So stable commits supersede hotfix commits by virtue of existing after/getting deployed after a hotfix commit. And vice versa.

jmcarp: Here's what confuses me about using git-flow with continuous deployment:
* *Every* update to master is tagged, either via hotfix or release
* If devs push features and hotfixes to their forks and not the main repo, Travis never builds those branches with `TRAVIS_PULL_REQUEST = 'false'`

One solution would be to have devs push feature and hotfix branches to origin and not to use forks. This is a little inefficient, since it triggers two Travis builds (one for the push to 18f/hotfix/<...>, and another for the pull request to 18f/master), but builds are free. Also, I think this is what everyone on the team except for me is already doing, so it wouldn't require anyone else to change workflows. Is this compatible with (or the same as) what you're proposing @arowla?

arowla: Coming back from vacation and reading this convo, I feel like I'm losing sight of what we were trying to accomplish, so I'd like to step back a bit. After reading up some more on the nitty-gritty of git-flow (I didn't realize that every change to master got tagged!), I'm beginning to think that our CD mapping should be basically what we would have had after your first line comment here:

* develop -> dev
* `startswith('release') -> staging
* master -> prod

If we wanted to "stage" a hotfix, then we could pull it into a release branch, but then it wouldn't be very "hot", would it? (The idea being that there's an urgent fix that needs to go out.) But under normal circumstances, a hotfix would only get deployed once it's merged to master (and develop). Does that seem too rash, or is it ok?

arowla: @jmcarp ^^

jmcarp: I wrote up the associated deploy workflow in #942 but still have a few questions. If we're happy with that workflow (assuming I described it correctly), then I agree with your revision. Let's resolve #942 and then make and merge that change.

PR #784, opened 2015-05-18
Blue-green deploys.

jmcarp:
* Make blue and green copies of apps in manifests.
* Detect active version of app on push.
* Add route un/mapping after push.

### DIFF
--- a/manifest_prod.yml
+++ b/manifest_prod.yml
@@ -6,7 +6,16 @@ env:
   NEW_RELIC_CONFIG_FILE: newrelic.ini
   NEW_RELIC_ENV: production
 applications:
-- name: api
+- name: api-a
+  path: .
+  buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
+  host: fec-prod-api
+  env:
+    PRODUCTION: True
+    NEW_RELIC_APP_NAME: OpenFEC API (production)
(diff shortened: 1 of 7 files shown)


jmcarp:
Needs some discussion before deploy (see https://github.com/18F/openFEC-web-app/pull/191#issuecomment-103269504 for details).

arowla:
See #796

arowla:
I'm assuming we want to wait on #796 so we ensure our blue-green deploys go off without an env-var hitch.

jmcarp:
Agree.

manifest_prod.yml
```diff
@@ -6,10 +6,23 @@ env:
...
   path: .
   buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
   host: fec-prod-api
+  services:
+    - fec-creds-prod
+  env:
+    PRODUCTION: True
+    NEW_RELIC_APP_NAME: OpenFEC API (production)
+    NEW_RELIC_LOG: stdout
+    WEB_CONCURRENCY: 4
```
arowla: Can we also list somewhere here that we need the creds service to have (manually) set:

```
# NEW_RELIC_LICENSE_KEY
# SQLA_CONN
```

PR #829, opened 2015-05-22
ttl_receipts_per_i, ttl_receipts_ii act as one data point

LindsayYoung:
We might want to wait for Jeff to confirm this is the case he should get back to us today.

For these variables, either could be blank, so this combines them for greater accuracy.

This should solve some of the problems /committee/C00042366 
(I think there is also a typo in the web app, it is not reporting everything on the API)


LindsayYoung:
My tests on this have been passing locally and failing with Travis. Not sure what the problem is yet.

LindsayYoung:
I am going to check we are not using any of deleted data types on the web app

LindsayYoung:
This is fine for the webapp. I am not sure why the tests are not failing for me locally.

PR #830, opened 2015-05-22
Feature/committee history

jmcarp:
Add committee history endpoint. To be used for committee cycle filtering in the webapp.

* Add committee history view, model, schema, resource
* Move variables to base classes for better code reuse
* Move repeated logic to SQL functions


LindsayYoung:
I will start looking at this now

PR #843, opened 2015-05-26
Calculate committee cycles from fact tables.

jmcarp:
We currently calculate committee cycles from committee statements of
organization (F1 filings), but committees may file financial reports
(e.g. F3s) in cycles during which they have not filed F1s. This patch
uses F3s instead of F1s to calculate active cycles per committee. Among
other issues, this resolves the missing data for the 2015-2016 cycle
reported by FEC.


LindsayYoung:
This does look better than what we had. I think there are still a few outliers, but we can deal with them later.

PR #844, opened 2015-05-27
Feature/harmonize financials columns

jmcarp:
Harmonize column names related to refunds that are shared across reports and totals types.

Note: Don't merge until corresponding patch is applied to the webapp!


jmcarp:
@LindsayYoung if you approve of the revised column names, I'll go ahead and write the patch for the webapp.

LindsayYoung:
Yea, I noticed that before. Thanks for digging into that!

LindsayYoung:
Yes, I think it will be a better API user experience to standardized names across the endpoints where we can. 

Do we have any tests to make sure we have all the fields hooked up properly? I am little nervous since there are a few points of possible failure.

LindsayYoung:
I am happy to help write additional tests for this too.

PR #845, opened 2015-05-27
Auto-generate schemas using marshmallow_sqlalchemy.

jmcarp:
Auto-generate schemas using marshmallow_sqlalchemy. This deletes about 400 lines of brittle code without changing the behavior of the application.

* Auto-generate schemas.
* Harmonize model and schema names.
* Delete duplicated `extend` helper function.


LindsayYoung:
This will show ids that we are using for internal filtering. Like report key and idx. We weren't showing those before because if they are based on _sk values, those are generated by the datawharehouse and we have seen them change when they do schema changes or reload the data. Perhaps we can hide those or it might be worth exposing those values if it improves data integrety. What do you think, @jmcarp ?

PR #867, opened 2015-05-28
Adding drop-downs and explanations to the API documentaion

LindsayYoung:
People from API testing asked for drop-downs. I am also adding some expanded explanations, those need to be formatted.


LindsayYoung:
It is a little off for `committee_type` in `/reports/{committee_type}` but overall it is an improvement.

webservices/args.py
```diff
@@ -64,37 +64,50 @@ def one_of(value, options):
...
 committee = {
-    'year': Arg(int, multiple=True, description='A year that the committee was active- (after original registration date but before expiration date.)'),
-    'cycle': Arg(int, multiple=True, description='An election cycle that the committee was active- (after original registration date but before expiration date.)'),
-    'designation': Arg(str, multiple=True, description='The one-letter designation code of the organization'),
-    'organization_type': Arg(str, multiple=True, description='The one-letter code for the kind for organization'),
-    'committee_type': Arg(str, multiple=True, description='The one-letter type code of the organization'),
+    'year': Arg(int, multiple=True, description='A year that the committee was active- (After original registration date but before expiration date.)'),
+    'cycle': Arg(int, multiple=True, description='A 2-year election cycle that the committee was active- (after original registration date but before expiration date.)'),
+    'designation': Arg(str, multiple=True, enum=['', 'A', 'J', 'P', 'U', 'B', 'D'],
+        description='The one-letter designation code of the organization: \
```
jmcarp: If markdown works here, these might look better as lists:
* A: Authorized by a committee
* J: Joint fundraising committee
etc.

webservices/args.py
```diff
@@ -64,37 +64,50 @@ def one_of(value, options):
 
 candidate_detail = {
     'cycle': Arg(int, multiple=True, description='Filter records to only those that were applicable to a given election cycle'),
-    'office': Arg(str, multiple=True, description='Governmental office candidate runs for'),
-    'state': Arg(str, multiple=True, description='U.S. State candidate is registered in'),
+    'office': Arg(str, multiple=True, enum=['', 'H', 'S', 'P'], description='Governmental office candidate runs for: House, Senate or President.'),
```
jmcarp: Let's pick one style for enumerating options (markdown lists, comma-separated lists, etc.) and use it consistently for fields like this.

webservices/args.py
LindsayYoung: It wasn't rendering markdown, but now we have our own fork and I think I can just add the markdown class.

webservices/args.py
LindsayYoung: :+1:

jmcarp:
This is looking good! I made a few nitpick-y suggestions inline and pushed a one-character typo fix that was breaking npm.

LindsayYoung:
I like your suggestions. 

Can the formatting be done in a separate request? I am going see if I can add markdown to those table cells. If I can, I will update the fork, and then reformat.

PR #873, opened 2015-05-29
Add pdf-url to schema

LindsayYoung:
Not working yet but it should give the urls for all reports that should have them.

### DIFF
--- a/webservices/common/models.py
+++ b/webservices/common/models.py
@@ -224,7 +224,7 @@ class CommitteeReports(BaseModel):
     committee_id = db.Column(db.String(10))
     cycle = db.Column(db.Integer)
 
-    beginning_image_number = db.Column(db.Integer)
+    beginning_image_number = db.Column(db.BigInteger)
     cash_on_hand_beginning_period = db.Column(db.Integer)
     cash_on_hand_end_period = db.Column(db.Integer)
     coverage_end_date = db.Column(db.DateTime)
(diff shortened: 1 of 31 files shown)


webservices/common/models.py
```diff
@@ -263,6 +263,16 @@ class CommitteeReports(BaseModel):
     offsets_to_operating_expenditures_period = db.Column(db.Integer)
     offsets_to_operating_expenditures_ytd = db.Column(db.Integer)
 
+    @property
+    def pdf_url(self):
+        if self.report_year is not None and self.report_year > 1994:
```
jmcarp: There are at least a handful of reports from 1994 that have PDFs on the FEC site (here's one: http://docquery.fec.gov/pdf/626/95039632626/95039632626.pdf). Should this be `>= 1994`, or does FEC have only partial PDF coverage before 1994?

LindsayYoung: I was going off of a comment from the FEC but, maybe I misinterpreted it. I will double check now.

LindsayYoung:
I looks like the 94 results are incomplete. 

example http://fec-dev-web.cf.18f.us/committee/C00237198?cycle=1994

The oldest filings in for that committee on the current website are 1996.

I think we will have a more accurate idea of what pdfs and html forms have links when we get the filings view from FEC. 

I think this is the safest course for now.

Issue #881, opened 2015-06-01
Termination reports included in reports endpoints


jmcarp:
The reports endpoints currently return all report types, including termination reports. But since termination reports don't include a coverage start date, or most other fields, this can lead to bugs in charts, including 500s.

Possible solutions:
* Always exclude termination reports from reports endpoints
* Add a query parameter (e.g. `exclude_types`) to endpoints to allow exclusion of termination reports

Thoughts @LindsayYoung?

jmcarp:
Example: http://fec-dev-api.cf.18f.us/v1/committee/C00486647/reports?cycle=2012

LindsayYoung:
I think it is an amendment problem that frequently shows up in Termination reports.

If an amendment doesn't have all the previous pages and just the part that is amended, we are getting null values for most of the fields. 

My instinct is to show termination reports as the most recent. Since this is really an amendment problem. If it is too much trouble, we could think of something else. 

I think the real fix would be to look at the form type and if it is adding or replacing a previous record and then, do the transformation. However, I don't see that any information like form type where we could look for "F3z" or an is amended flag etc.  I currently don't see anything that would tell us it this amendment is meant to add to or replace the form it is amending- just the expire date. I will follow up on the with the FEC. Hopefully, I am just missing something. 

Additionally, I think we might want the web-app to not break completely when there are null values. 

example of amended form:
http://docquery.fec.gov/pdf/333/12030812333/12030812333.pdf

LindsayYoung:
Just reread the DW documentation and in 3.3 it looks like amendments in CDFM are supposed to be complete, and we are not supposed to have to fix them.

PR #887, opened 2015-06-01
Allow specification of included and excluded report types.

jmcarp:
Allow API consumers to specify report types in reports endpoints. This can be used to either include or exclude reports by type:

* `/committee/<committee_id>/reports?type=Q1&type=Q2`
* `/committee/<committee_id>/reports?type=-TER`

Partial fix for #881 based on feedback from @dcpcc1967 and @jwchumley.


LindsayYoung:
This looks good. One small thing, do you mind if we change the arg form `type` to `report_type` so the argument matches the json output?

jmcarp:
Sounds good, making the change now.

PR #890, opened 2015-06-01
Use BigInteger to avoid overflow on image numbers.

jmcarp:
@LindsayYoung: This was kind of a subtle problem. In the tests, we're using `Integer` fields to store `beginning_image_number`, but the realistic number we provided here was too big. Switching to `BigInteger` fixes the issue. For some reason, Flask swallowed the `integer out of range` error we *should* have been getting and threw an interpretable error instead. And the issue would never come up outside of testing because the SQL scripts will intelligently detect the correct data type for us when we're importing real data.

`¯\_(ツ)_/¯`


LindsayYoung:
Thanks for helping me with that!

PR #912, opened 2015-06-04
Delete two year period table.

jmcarp:
The cycles table isn't necessary, since we can compute cycles from year
variables in the candidate, committee, and report tables on the fly when
building materialized views. This patch also resolves a migration issue
in which candidate views were unavailable during data import, causing
500s on all candidate views: dropping the cycles table also dropped all
dependent views, which currently means all candidate views. Last, this
change makes the construction of the candidate and committee history
views more similar, which is helpful for readability.


data/sql_updates/create_candidate_history_view.sql
```diff
@@ -1,81 +1,86 @@
...
-    left join (
+with
+    years as (
         select
             cand_sk,
             array_agg(distinct(cand_election_yr))::int[] as election_years
         from dimcandoffice
         group by cand_sk
-    ) year_agg using (cand_sk)
-    left join (
```
LindsayYoung: I think we originally used a join to make duplicate records during gaps in the records. The gaps are less common in recent records but, I would rather be over cautious.

arowla: Agreed-- we might change this subquery to something like:

```sql
select cand_sk, array_agg(election_year)
from (
    select cand_sk, generate_series(min_cand_elec_yr::int, max_cand_elec_yr::int, 2) as election_year
    from (
        select cand_sk, min(cand_election_yr) as min_cand_elec_yr, max(cand_election_yr) as max_cand_elec_yr
        from dimcandoffice group by cand_sk
    ) x 
) y group by cand_sk
```

jmcarp: * This patch has the same behavior as develop, so we may want to talk about this change in a separate issue / patch.
* Would this change have implications about the way we make history records? Right now, senate candidates get a history entry for each election year they're active in--meaning candidates who run for multiple elections have six-year gaps between history rows. Do we want history records for every two-year period for these candidates? What about a candidate who runs for office in two cycles a decade apart--should she get a history row for every cycle between?

jmcarp: Added intermediate two-year cycles in a0f35c9ae46b4a4525d4c306b970cb165f10a266 and 3a40e226db1c19661aeb7141342ccbc32caff138. Thanks @arowla for `generate_series`.

LindsayYoung:
LGTM Do you want me to add a test for it?

arowla:
A test would be great...

@jmcarp where are we now on preserving functionality vs not? I didn't mean to lead you astray with my code sample earlier-- if that was the case. I am having trouble remembering how the cycles table started and what problem it was meant to solve.

jmcarp:
@arowla the latest commits add functionality. As of develop, candidate history rows are distinct by election year, and committee history rows are distinct by F1/F3 report year. With the latest changes, we would be computing a history row for each candidate and committee from the earliest to the latest years on record, such that there aren't gaps for Senate and Presidential candidates, or for candidates or committees that are inactive between elections. The `cycles` column indicates all two-year cycles for which we have candidate or committee history.

arowla:
Thanks for the clarification!

PR #919, opened 2015-06-05
Update deploy script.

jmcarp:
* Use postgres 9.4.
* Use sudo to install tmuxinator.
* Turn off password protection.

### DIFF
--- a/scripts/bootstrap/Vagrantfile
+++ b/scripts/bootstrap/Vagrantfile
@@ -1,6 +1,9 @@
-Vagrant.configure(2) do |config|
+Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/trusty64"
   config.vm.provision :shell, path: "provision.sh"
   config.vm.network "forwarded_port", guest: 3000, host: 3001
   config.vm.network "forwarded_port", guest: 5000, host: 5001
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 1024
+  end
 end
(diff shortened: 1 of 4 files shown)


jmcarp:
As of f5ecd1f, this is working for me locally as it's supposed to. @arowla or @LindsayYoung: could you verify that the vagrant setup works on your machines and look this over when you have time? Not urgent, but it would be great if new people / open source contributors could get the project running locally a little easier.

arowla:
I went to the directory with the Vagrantfile and ran `vagrant up`. Should it have started the API and webapp? I get nothing when I go to localhost:3001 or localhost:5001.

It would be helpful to have a quick primer on how to use/what to expect.

jmcarp:
No, as currently implemented, it takes more work to start the services. Check out the docs at https://github.com/18f/openfec#vagrant--you have to ssh into the vagrant box, copy and execute the bootstrap script, and start tmux. I didn't write the vagrant setup and am open to changing any of it--the goal of this patch is just to get it working based on the instructions in the README.

arowla:
So I got through the whole bootstrapping process and was able to start the API and website, but the website seems to not be receiving any data (while the API does have data). Is this expected and/or in scope?

jmcarp:
That part should work. Will need more information to debug. Were there any errors during bootstrap? Does the webapp load at all, or just fail to get data from the API? Happy to jump on screenhero (which we apparently have for free through slack now) to debug if that would be useful.

arowla:
I've looked at it again, but couldn't see why there was no data getting to the web app. Let me take one more look and then maybe we can debug further later today or tomorrow?

LindsayYoung:
Are we waiting for something on this?

arowla:
I just tried it again and found that I had missed the instruction to `source ~/.fec_vars` in the vagrant user's `~/.bashrc`. I restarted tmuxinator and then it worked. :+1:

PR #921, opened 2015-06-08
Feature/improve performance

jmcarp:
Add indices on sorted and filtered columns; speed up count queries.

@LindsayYoung: the tiny change to the count query in ea510c8 should help with the speed issues on committee reports that we were seeing last week.


arowla:
:+1:

PR #923, opened 2015-06-08
List of options now look like formatted code

LindsayYoung:
It looks better. I could not get the markdown to render lists in the table at all.


arowla:
Hopefully this isn't just a yak shaving request, but I'm not crazy about having formatting embedded in Python. Since this is clearly a common need for us, what about specifying the options and their definitions in a dict (ala `{ 'O': 'Option 1', 'S', 'Option Definition 2' }`), and then using the docs to format it in a standard way? There is also a lot of duplication between `enum` and this. We could subclass `Arg` and then build `enum` from `options.keys()` or something.

LindsayYoung:
I think that is a good long-term solution, our fork is [here](https://github.com/18F/swagger-ui), so you can add whatever you like. I came up with this solution because it was better than what we were doing but I don't like it either.

arowla:
Ok let's save that for later, then.

Issue #901, opened 2015-06-03
better PDF filtering 


LindsayYoung:
Add more complex date filtering 

Right now we are just filtering since 1994, this will make some bad links for Senators. 

As Paul said:
```
Older paper reports are availabe with the following limitations:

PAC and party reports since 1993 
House committee reports since May 1996
Senate filings since May 2000

http://www.fec.gov/finance/disclosure/help_candidate_committee_viewer.shtml

```
@dcpcc1967  what about Presidential?

dcpcc1967:
1993 for presidential filings

jmcarp:
Fixed in #928.

PaulClark2:
I am out of the office until 06/16/2015.

I'm on leave and will respond to your e-mail when I return.  If you need
immediate assistance please contact Jeff Chumley, jchumley@fec.gov, or Wei
Luo, wluo@fec.gov.



Note: This is an automated response to your message  "Re: [openFEC] better
PDF filtering  (#901)" sent on 6/15/2015 11:00:02 AM.

This is the only notification you will receive while this person is away.

PR #928, opened 2015-06-08
Feature/pdf link fix#901

LindsayYoung:
This creates the links with the logic Paul outlined in #901


LindsayYoung:
h/t to @jmcarp!

webservices/common/models.py
```diff
@@ -437,11 +458,22 @@ class CommitteeReportsPresidential(CommitteeReports):
     transfer_to_other_authorized_committee_period = db.Column(db.Integer)
     transfer_to_other_authorized_committee_ytd = db.Column(db.Integer)
 
+    @property
+    # PAC, Party and Presidential records start May 1993
+    def pdf_url(self):
+        if self.report_year is None or self.report_year < 1993:
+            return None
+        else:
```
jmcarp: Lines 466-470 can be replaced with `return utils.make_pdf_url(self.beginning_image_number)`, as in other reports classes.

arowla: :+1:

LindsayYoung:
Ready for review

webservices/common/util.py
```diff
@@ -16,3 +16,9 @@ def filter_query(model, query, fields, kwargs):
         column = getattr(model, field)
         query = query.filter(column.in_(value))
     return query
+
+def make_pdf_url(imagenumber):
```
jmcarp: This looks like a duplicate from `webservices/utils.py`. It also doesn't work, since `self` isn't defined in this context. I think we can just delete this function.

jmcarp:
@arowla can you look this over? I made a few comments inline, but I also wrote some of the code, so could use a more objective review.

webservices/common/util.py
arowla: +1. cc @LindsayYoung 

Not for this task, but we might want to consolidate the two util files to avoid this sort of mistake in the future.

LindsayYoung:
made the fix. back to you @arowla

LindsayYoung:
Since Alison is out do you think you go do the honors, @jmcarp ?

PR #930, opened 2015-06-09
Updating Develop

LindsayYoung:
All the work since May 14


arowla:
Should this be called "Updating Master"? And do we want to make a release branch first to deploy to staging? (Depends on getting #783 and 18f/openfec-web-app#186 resolved.)

LindsayYoung:
Doing this with gitflow instead

PR #959, opened 2015-06-11
Document database snapshots.

### LINKED ISSUE #835: Document / automate database snapshots
IIRC, the staging and development environments are using snapshots of the production database built a few weeks to a month ago. We should document, and ideally automate, the process of making those snapshots while @seanherron is still around.

jmcarp:
Note: Everything in these instructions can be automated using Boto. But since we shouldn't need to create new snapshots often, and may pick a different approach for copying data to dev and stage, I don't think automating is worth the time just now. The next person to update the snapshots may want to revisit this and write scripts as they go along.

[Resolves #835]

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,49 @@ through Travis CI accordingly.
 
 * [auto] Master is deployed to prod
 
+#### Data for development and staging environments
+
+Note: The following can be automated using Boto or the AWS CLI if we continue on with this model and need
+to update snapshots frequently.
+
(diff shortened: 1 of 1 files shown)


arowla:
We should add a note on properly client-tagging the new instances (for billing purposes-- see the current tag on any of our existing resources).

jmcarp:
Done. I noticed that the client tags were automatically copied to the new instances from the snapshot, but I couldn't find that behavior in the docs, so it's best not to rely on it.

PR #990, opened 2015-06-17
Update caching and deployment docs.

jmcarp:
Self-explanatory, hopefully. @LindsayYoung and @arowla: the goal of the docs is to make sure that the team understands the changes I made to deployment in the event that I get hit by a bus or become a Jesuit or something, so please take a look and let me know if I should elaborate.


LindsayYoung:
Thank you and please say away from the front of busses.

Issue #938, opened 2015-06-09
Create summary endpoint for ie only spenders


LindsayYoung:
This will get rid of the remaining missing summary data.  

This is pretty straight forward use of `factindpexpcontb_f5`

LindsayYoung:
Starting to look at this while I wait on some other things

LindsayYoung:
- [ ] report endpoint
 - [ ] summary endpoint

PR #991, opened 2015-06-18
Feature/ie reports

LindsayYoung:
Got the reports for form-5 only filers going in http://localhost:5000/v1/reports/ie-only


webservices/schemas.py
```diff
@@ -213,6 +213,16 @@ class CommitteeSearchListSchema(ApiSchema):
 register_schema(CommitteeReportsSchema)
 register_schema(CommitteeReportsPageSchema)
 
+make_ie_reports_schema = functools.partial(
```
jmcarp: We're only making one schema for `CommitteeReportsIEOnly`, so the functional partial doesn't buy us anything here. Also, since this model doesn't have `committee` or `report_key` fields, we don't need to `exclude` them:

```python
CommitteeReportsIEOnlySchema = make_schema(
    models.CommitteeReportsIEOnly,
    fields={
        'pdf_url': ma.fields.Str(),
    },
    options={'exclude': ('idx', )},
)
```

webservices/resources/reports.py
```diff
@@ -46,25 +50,35 @@ def parse_types(types):
...
-            'description': 'House, Senate or presidential',
-            'enum': ['presidential', 'pac-party', 'house-senate'],
+            'description': 'House, Senate, presidential, independent expenditure only',
+            'enum': ['presidential', 'pac-party', 'house-senate', 'ie-only'],
         },
     ],
 )
 class ReportsView(Resource):
 
+    @schemas.marshal_with(schemas.CommitteeReportsPageSchema(), wrap=False)
```
jmcarp: What would you think about putting independent expenditures under a separate endpoint: `/committee/{committee_id}/expenditures` and `/expenditures`? This class gets more complicated if we want it to handle data from both F3 and F5 filings--see `if self.committee_type == 'ie-only'` checks and these extra methods. We also lose the ability to describe the output of the resource in Swagger, since Swagger doesn't have the ability to specify that the output will belong to one of several different schemas.

tests/report_tests.py
```diff
@@ -230,3 +230,31 @@ def test_report_type_exclude(self):
         results = self._results(api.url_for(ReportsView, committee_id=committee_id, report_type=['-M3']))
         self.assertTrue(all(each['report_type'] in ['Q2', 'TER'] for each in results))
 
+    def test_ie_only(self):
+        factories.ReportsIEOnlyFactory(
```
jmcarp: What's the purpose of the first call to `ReportsIEOnlyFactory` and the `flush` afterwards?

tests/report_tests.py
```diff
@@ -230,3 +230,31 @@ def test_report_type_exclude(self):
...
+            committee_id='C90013723',
+            total_independent_contributions=200,
+            total_independent_expenditures=100,
+        )
+        db.session.flush()
+        number = 12345678902
+        factories.ReportsIEOnlyFactory(
```
jmcarp: If we assign the result of this factory to a variable, we can use it later below:

```python
report = factories.ReportsIEOnlyFactory(
    beginning_image_number=12345678902,
    total_independent_contributions=200,
    total_independent_expenditures=100,
)
...
self.assertEqual(results[0]['total_independent_contributions'], report.total_independent_contributions)
```
That way, constants are only defined once.

webservices/resources/reports.py
LindsayYoung: This is what I am thinking-

There are different types of form 5- the ones that I am dealing with here are the Quarterly and Year-end reports, similar in concept to the form 3, 3x and 3p Because they cover a period of time rather than transactions. 

Here is the [form 5](http://www.fec.gov/pdf/forms/fecfrm5.pdf) for reference.

This endpoint doesn't have all the form 5s, The 24 and 42 hour notices will go in a separate independent expenditures endpoint, like you suggested, like the Form 3 details that are in the schedules, will go in a separate endpoint. 

That way, all the summary reports are covered by a /reports endpoint and you can find them with committee/id/reports without having to look up what the committee type is.

We could try and put them all into one schema with nulls  or n/a for fields that don't apply so it works in Swagger, but  I don't want to burden users with having to understand the different kinds of summary reports and where to look for each kind because, the independent expenditure totals for the other groups is on the form 3x.

tests/report_tests.py
LindsayYoung: sounds good

tests/report_tests.py
LindsayYoung: I will take it out, thanks

webservices/schemas.py
LindsayYoung: Good point.

webservices/resources/reports.py
LindsayYoung: To document the conversation we had- we decided it would be best to document have all the reports share a schema and an endpoint. 

We will improve the documentation by displaying all the fields in swagger, instead of just the shared reports fields. (I think that can be a different PR.)

Also, I am going to add what form the information is coming from,  to make things slightly more clear.

jmcarp: :+1: :+1:

LindsayYoung:
This takes care of most of #938

LindsayYoung:
I am going to add _period after the names so they match the other reports

PR #996, opened 2015-06-19
Feature/schedule a

jmcarp:
Add Schedule A filings to API.

The Schedule A table is much larger than the other tables used by the API so far, so the approach we're taking here is different in several ways:
* Instead of a materialized view, we're sticking with the original table provided by FEC (`sched_a`). We're adding indexes to this table for performance, but not creating a new table or view.
* To support full-text search on contributor name and employer, we're creating an auxiliary table (`ofec_sched_a_fulltext`) with full-text representations of the relevant fields. This table is kept in sync with `sched_a` using a trigger in postgres.
* Counting all rows in a cursor is prohibitively expensive on tables of this size. We're compromising by using an approximate count for large result sets, based on the results of a much cheaper `EXPLAIN` query. This means that counts will *not* be exactly correct.
* Pagination using `OFFSET` is also too expensive for large tables. Instead, this patch supports pagination by passing the index of the last row in the previous result set. This means we'll be able to fetch the next page in a query, but not seek to an arbitrary page number.

Still in progress:
* [x] Restrict sort options to avoid sorting on unindexed columns
* [x] Restrict data to recent years to stay within our current hardware capabilities
* [x] Discuss and benchmark proposed changes with @arowla 
* [x] Write unit and integration tests


webservices/common/models.py
```diff
@@ -556,3 +556,79 @@ class CommitteeTotalsHouseSenate(CommitteeTotals):
...
+    election_type_full = db.Column('election_tp_desc', db.String)
+    contributor_employer = db.Column('contbr_employer', db.String)
+    contributor_occupation = db.Column('contbr_occupation', db.String)
+    contributor_aggregate_ytd = db.Column('contb_aggregate_ytd', db.Float)
+    contributor_receipt_date = db.Column('contb_receipt_dt', db.DateTime)
+    contributor_receipt_amount = db.Column('contb_receipt_amt', db.Float)
+    receipt_type = db.Column('receipt_tp', db.String)
+    receipt_type_full = db.Column('receipt_desc', db.String)
+    candidate_id = db.Column('cand_id', db.String)
+    candidate_name = db.Column('cand_nm', db.String)
```
LindsayYoung: Did you check if the candidate _prefix, _first_name, _middle_name, _last_name and _suffix are ever populated? On other tables, the existed but were unused/barely used. The forms are not designed for that so I don't think it gets populated like that. (Maybe it doesn't matter if there are empty columns, but I thought I would mention it. )

jmcarp: They're not *always* empty, but they're often null or filled with some dummy value.

LindsayYoung: It sounds similar to dimcandproperties, we decided not to include it at all in the API output because we did not want to give people data that was so inconsistent that it would not be useful. I don't have strong feelings on this. But if it makes the data slightly smaller or faster or cheaper, I would consider taking it out. If it doesn't make a difference, let's not waste your time taking it out.

jmcarp: Sounds reasonable--might as well delete columns that are rarely or never useful.

LindsayYoung:
This looks pretty amazing. I have a few things I would like to talk over, nothing that is a barrier to publishing this. (Discussing this stuff can totally wait till Monday.) 

- Can we put these under `/schedules/` instead of `/filings/`? Granted most people won't know the difference or care, but schedules are line items rather than whole documents. If you don't want too many tags, I would be more comfortable to group reports with filings rather than schedules with filings.  We will also be adding schedule E eventually if you were thinking that there was not enough types for schedules to have their own grouping. (I am for adding this as is and renaming it in a different pr, if that is what we decide to do.)

- I would really like to make memoed items more approachable for the uninitiated. Putting them in a different column, might be too extreme and cumbersome. Maybe we could rename `memo_code` , `memoed_subtotal` and make it a boolean? I am up for other suggestions. I just think that making it clear what numbers are included in totals will make a huge impact on making the data more approachable. (At Sunlight when we did reporter training, the most useful piece of "insider" information that we would teach. I feel like you shouldn't have to attend a workshop to use this information.) We can figure this out later too.

- I am making a separate pr with documentation updates. I have some helpful hints for people who want to use this data that I would like to share.

jmcarp:
* Changing the endpoint URLs sounds reasonable.
* Changing the structure of the Schedule A table is a bit of a pain, but we can get a boolean column using a property. I'll make that change as well.

Also, the latest changes are all in #1008, which includes Schedule A and B endpoints, plus aggregates on both.

LindsayYoung:
That all sounds awesome!

PR #1003, opened 2015-06-22
Update docs.py

emileighoutlaw:
Style edits. 

Re-wrote a few sentences to pluralize humans. Government writing hack: I'm using plural people whenever possible to move away from the "he or she" construct, which is not gender inclusive. 

*Note:* Don't want to change any legal definitions, but I would love (some day) to see a less tautological definition of committee than "committees include the committees and organizations that file with the FEC. "

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -2,7 +2,7 @@
 
 CANDIDATE_ID = '''
 A unique identifier assigned to each candidate registered with the FEC.
-If a person runs for several offices, they will have separate candidate IDs for each office.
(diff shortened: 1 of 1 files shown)


jmcarp:
@emileighoutlaw thanks! I forgot to include this earlier, but the text describing inputs to each endpoint are stored here: https://github.com/18F/openFEC/blob/develop/webservices/args.py, if you want to take a look.

emileighoutlaw:
Will do now! Thanks @jmcarp

PR #1007, opened 2015-06-24
Feature/filings

LindsayYoung:
Creating an API endpoint that has all of the filings from the FEC.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -1,9 +1,9 @@
 import logging
 import functools
-import time
 
 from smore import swagger
 from flask.ext.restful import abort
+from dateutil.parser import parse as parse_date
 
 import webargs
 from webargs import Arg
@@ -41,9 +41,6 @@ def _validate_natural(value):
     if value < 0:
         raise webargs.ValidationError('Must be a natural number')
 Natural = functools.partial(Arg, int, validate=_validate_natural)
-
-def make_date(datestr):
-    return time.strptime(datestr, '%Y-%m-%d')
 
 
 paging = {
@@ -148,11 +145,11 @@ filings = {
     'beginning_image_number': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
     'report_type': Arg(str, multiple=True, description='Report type'),
     'report_year': Arg(int, multiple=True, description='Report year'),
-    'begin_image_numeric': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
+    'beginning_image_number': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
     'report_year': Arg(str, multiple=True, description='Year that the report applies to'),
-    'recipt_date': Arg(str, multiple=True, description='The day the filing was received by the FEC', type=make_date ),
+    'recipt_date': Arg(parse_date, multiple=True, description='The day the filing was received by the FEC'),
     'form_type': Arg(str, multiple=True, description='Form type'),
-    'report_pgi': Arg(str, multiple=True, description='Primary Gereral or Special election indicator.'),
+    'primary_general_indicator': Arg(str, multiple=True, description='Primary Gereral or Special election indicator.'),
     'amendment_indicator': Arg(str, multiple=True, description='''
         -N   new\n\
         -A   amendment\n\
--- a/webservices/common/models.py
+++ b/webservices/common/models.py
@@ -647,11 +647,10 @@ class Filings(db.Model):
 
     @property
     def pdf_url(self):
-        if self.report_year and self.report_year >=2000:
+        if self.report_year and self.report_year >= 2000:
             return utils.make_pdf_url(self.beginning_image_number)
-        elif self.form_type in ['F3X', 'F3P'] and self.report_year > 1993:
+        if self.form_type in ['F3X', 'F3P'] and self.report_year > 1993:
             return utils.make_pdf_url(self.beginning_image_number)
-        elif self.form_type == 'F3' and self.committee.committee_type == 'H' and self.report_year > 1996:
+        if self.form_type == 'F3' and self.committee.committee_type == 'H' and self.report_year > 1996:
             return utils.make_pdf_url(self.beginning_image_number)
-        else:
-            return None
+        return None
--- a/webservices/resources/filings.py
+++ b/webservices/resources/filings.py
@@ -10,17 +10,17 @@ from webservices.common.util import filter_query
 
 fields = {
     'committee_id',
-    'begin_image_numeric',
+    'beginning_image_number',
     'report_type',
     'report_year',
     'recipt_date',
     'form_type',
-    'report_pgi',
+    'primary_general_indicator',
     'amendment_indicator',
 }
 
 @spec.doc(
-tags=['filings'],
+    tags=['filings'],
     description=docs.FILINGS,
     path_params=[
         {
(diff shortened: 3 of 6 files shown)


webservices/resources/filings.py
```diff
@@ -0,0 +1,47 @@
...
+            'description': docs.COMMITTEE_ID,
+            'type': 'string',
+        },
+    ],
+)
+class FilingsView(Resource):
+
+    @args.register_kwargs(args.paging)
+    @args.register_kwargs(args.reports)
+    @args.register_kwargs(args.make_sort_args(default=['-coverage_end_date']))
```
jmcarp: We should make sure any fields we're querying or sorted on are indexed.

webservices/resources/filings.py
```diff
@@ -0,0 +1,47 @@
...
+    @schemas.marshal_with(schemas.FilingsPageSchema(), wrap=False)
```
jmcarp: We probably don't want to set `wrap=False` here, since that directs `marshal_with` not to use the schema to serialize the return value.

webservices/resources/filings.py
```diff
@@ -0,0 +1,47 @@
...
+    def get(self, **kwargs):
+        query = self.get_filings(kwargs)
+        return utils.fetch_page(query, kwargs, model=models.Filings)
+
+    def get_filings(self, kwargs, committee_id=None):
```
jmcarp: We can probably combine `get` and `get_filings` into a single method.

webservices/resources/filings.py
```diff
@@ -0,0 +1,47 @@
+import sqlalchemy as sa
```
jmcarp: Unused import.

webservices/common/models.py
```diff
@@ -595,3 +595,49 @@ class CommitteeTotalsIEOnly(BaseModel):
...
+    committee_id = db.Column(db.String)
+    sub_id = db.Column(db.BigInteger, primary_key=True)
+    coverage_start_date = db.Column(db.DateTime)
+    coverage_end_date = db.Column(db.DateTime)
+    receipt_date = db.Column(db.DateTime)
+    election_year = db.Column(db.Integer)
+    form_type = db.Column(db.String)
+    report_year = db.Column(db.Integer)
+    report_type = db.Column(db.String)
+    # document_type = db.Column(db.String)
```
jmcarp: Why are some columns commented out?

webservices/resources/filings.py
LindsayYoung: I started a doc of indexes for it but, what should it be named?

create_filings_vw or should it not be called a view since it is just indexes?

webservices/args.py
```diff
@@ -139,6 +139,13 @@ def one_of(value, options):
     'party': Arg(str, multiple=True, description='Three letter code for the party. For example: DEM=Democrat REP=Republican'),
 }
 
+filings = {
+    'committee_id': Arg(str, multiple=True, description=docs.COMMITTEE_ID),
+    'beginning_image_number': Arg(int, multiple=True, description='Unique identifier for the electronic or paper report. If report is amended, it will show the most recent report.'),
+    'report_type': Arg(str, multiple=True, description='Report type'),
+    'report_year': Arg(str, multiple=True, description='Report year'),
```
jmcarp: Should the type here be `int`?

webservices/args.py
```diff
@@ -139,6 +139,13 @@ def one_of(value, options):
     'party': Arg(str, multiple=True, description='Three letter code for the party. For example: DEM=Democrat REP=Republican'),
 }
 
+filings = {
+    'committee_id': Arg(str, multiple=True, description=docs.COMMITTEE_ID),
+    'beginning_image_number': Arg(int, multiple=True, description='Unique identifier for the electronic or paper report. If report is amended, it will show the most recent report.'),
```
jmcarp: Can this text move to `docs.py`?

webservices/args.py
LindsayYoung: yes!

webservices/common/models.py
LindsayYoung: After the email, they are making the suggested changes, so when they do we want to uncomment the new models and delete the old model below. They made the change on their end but I don't see it on our side yet.

webservices/resources/filings.py
LindsayYoung: Thank you! that was throwing me off.

webservices/resources/filings.py
LindsayYoung: I am thinking we might want to join with the committee names, at least in some views, so that would make it more complicated. But will merge and break them later if necessary.

webservices/args.py
LindsayYoung: Sure!

LindsayYoung:
Just to document it here, vw_filing_history is only on production right now

LindsayYoung:
It is kind of slow and it is not showing up on swagger

jmcarp:
My fault, I forgot to document the swagger test. Try running `npm install -g swagger-tools`.

jmcarp:
What specifically is slow?

LindsayYoung:
Thanks the install worked.

/committee/C00008078/filings takes 1.9 seconds

LindsayYoung:
Tests are failing because it does not see vw_filing_history, but I can see that table in the dump.

LindsayYoung:
Tests pass for me locally- not sure why it is not working for travis.

webservices/resources/filings.py
```diff
@@ -0,0 +1,67 @@
...
+    ],
+)
+class FilingsView(Resource):
+
+    @args.register_kwargs(args.paging)
+    @args.register_kwargs(args.make_sort_args(default=['-receipt_date']))
+    @schemas.marshal_with(schemas.FilingsPageSchema())
+    def get(self, committee_id=None, **kwargs):
+        filings = models.Filings.query
+        if hasattr(filings, 'committee'):
```
jmcarp: Can this line ever be reached? Query objects like `filings` don't have the same attributes as models, so the `hasattr` check should always be false. We could check `hasattr(models.Filings, 'committee')` instead, but based on the code in `models.py`, that check should also always be false.

Unless we want to nest committees in filings, I don't think we need this check.

data/sql_setup/create_filings_indexes.sql
```diff
@@ -0,0 +1,8 @@
+create index on public.vw_filing_history(committee_id);
```
jmcarp: I think we can drop this file now that we're using materialized views.

webservices/args.py
```diff
@@ -139,15 +140,37 @@ def one_of(value, options):
...
 }
 
+filings = {
+    'committee_id': Arg(str, multiple=True, description=docs.COMMITTEE_ID),
+    'beginning_image_number': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
+    'report_type': Arg(str, multiple=True, description='Report type'),
+    'report_year': Arg(int, multiple=True, description='Report year'),
+    'beginning_image_number': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
+    'report_year': Arg(str, multiple=True, description='Year that the report applies to'),
+    'recipt_date': Arg(parse_date, multiple=True, description='The day the filing was received by the FEC'),
```
jmcarp: Should be `receipt_date`. Also, in the schedules views, we're using `min_date` and `max_date` instead. Probably best to be consistent.

LindsayYoung: :+1:

webservices/resources/filings.py
LindsayYoung: Ok, I thought we were going to have to nest it for something- if we are not using that at all let's get rid of it.

PR #1008, opened 2015-06-25
Feature/incremental aggregates

jmcarp:
Add queued incremental aggregates.

Aggregates over Schedule A and B receipts can be expensive to build from
scratch. This patch implements queued incremental aggregates, such that
updates to the `sched_a` table are pushed into a queue, and aggregates
are incrementally updated on demand:

* Create initial aggregates (slow!)
* Store all changes to `sched_a` in queue tables
* As needed, update all aggregates and clear queues

This approach adds the overhead of copying added and deleted rows to the
queue tables, but this is much less expensive than updating all
aggregates in real time or rebuilding aggregates from scratch.

Many thanks to @arowla for suggesting this approach.

Note: This patch depends on the Schedule A and B pull requests; I'm filing this separately to flag for review.


LindsayYoung:
I think we can merge.

The image number lookup is not working for me for schedule A and B locally. Is it working for you? 

Also, what does last_contributor_receipt_amount mean?

jmcarp:
Image number lookups should be fixed now. `last_contributor_receipt_amount` is used for pagination--it's the receipt amount of the last result of the previous query.

PR #1021, opened 2015-06-30
Feature/update subset

jmcarp:
* Create new test subset with filings; should help #1007 pass
* Update subset creation
* Update subset docs


LindsayYoung:
Thank you!

PR #1023, opened 2015-07-02
Add filters on committee filing dates.

jmcarp:
Also update tests, fix PEP8 issues.

Necessary for https://github.com/18F/openFEC-web-app/issues/289.


webservices/resources/committees.py
```diff
@@ -74,6 +74,11 @@ def get_committees(self, kwargs):
         if kwargs['cycle']:
             committees = committees.filter(models.Committee.cycles.overlap(kwargs['cycle']))
 
+        if kwargs['start_date']:
+            committees = committees.filter(models.Committee.first_file_date >= kwargs['start_date'])
+        if kwargs['end_date']:
```
LindsayYoung: Calling this end_date is kind of confusing.

jmcarp: We could go for something more verbose, like `max_first_file_date`. What do think?

LindsayYoung: That is a great solution!

PR #1024, opened 2015-07-02
Feature/filings

jmcarp:
Test and swagger fixes.


data/sql_updates/create_filings_view.sql
```diff
@@ -42,8 +42,8 @@ select
     amendment_indicator,
     update_date
 from vw_filing_history fh
-left join ofec_committee_history_mv com on fh.committee_id = com.committee_id and fh.report_year + fh.report_year % 2 = com.cycle
-left join ofec_candidate_history_mv cand on fh.committee_id = cand.candidate_id and fh.report_year + fh.report_year % 2 = cand.cycle
+left join ofec_committee_history_mv_tmp com on fh.committee_id = com.committee_id and fh.report_year + fh.report_year % 2 = com.cycle
```
LindsayYoung: I think I accidentally committed this, it shouldn't be tmp, right?

jmcarp: It actually has to be, otherwise the view won't exist at this point. We build all the views with _tmp appended to their names, then rename them all at once when they're done.

LindsayYoung: OK, that makes sense. thanks for explaining

PR #1036, opened 2015-07-07
[WIP] Begin documenting model fields.

jmcarp:
These will appear in the Swagger documentation.

@LindsayYoung: Want to help with this one? I don't think it's urgent, but it would be nice to document our models, especially the ambiguous fields.


LindsayYoung:
Yea I would love to and I think adding where things originate on forms would be good too

PR #1039, opened 2015-07-07
adding info & link about FEC data use restrictions

leahbannon:
cc @emileighoutlaw @konklone

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -4,6 +4,8 @@ public domain within the United States.
 Additionally, we waive copyright and related rights in the work
 worldwide through the CC0 1.0 Universal public domain dedication.
 
+A few restrictions limit the way you can use FEC data. For example, you can’t use contributor lists for commercial purposes or to solicit donations. [Learn more on FEC.gov.](http://www.fec.gov/pages/brochures/saleuse.shtml)
+
 ## CC0 1.0 Universal Summary
 
 This is a human-readable summary of the [Legal Code (read the full text)](https://creativecommons.org/publicdomain/zero/1.0/legalcode).


konklone:
I personally think this belongs in the README and the published API documentation, but not the LICENSE, since the LICENSE describes how people can use the software and not the data.

leahbannon:
@konklone It will be in the API docs and I'm up for adding it to the README as well. 

This should also be in the license file. Users should be warned about this, in part because the license says "this project" not "the code in this repo only," but also because I think it's helpful to provide a heads up in the place that users expect to find restrictions on use.

konklone:
> This should also be in the license file. Users should be warned about this, in part because the license says "this project" not "the code in this repo only," but also because I think it's helpful to provide a heads up in the place that users expect to find restrictions on use.

I don't agree -- I think putting data restriction information intermingled in the software license file could give users of the software the mistaken impression that the use restrictions on data apply to the software as well, which is not the case. People are free to fork the openFEC codebase and use it, or pieces of it, to (somehow) solicit donations or do things that are otherwise not allowed with the FEC's campaign finance data.

As a user of open data, I wouldn't expect the software license to address these issues. I would expect them to be addressed where the data is published and/or documented. Adding it to this project's README is even a bit superfluous, but on balance could work as a courtesy and extra layer of diligence, as it's more clearly detached from software licensing information.

leahbannon:
Those are reasonable points, but the language in the license that says "this project" is ambiguous and the language in my pull request is clear that it applies to "the data."

konklone:
That could well be worth a pull request to [our default LICENSE file](https://github.com/18F/open-source-policy/blob/master/LICENSE.md) if you think that "this project" is ambiguous phrasing for that file's purpose.

LindsayYoung:
How about we add it here for now and think of a better solution later. I think it would be nice to have that as a separate file.

PR #1046, opened 2015-07-08
Allow hiding null values on sorted columns.

jmcarp:
And improve testing and documentation on sort utility.

This will be useful for hiding null values on sorted columns in the webapp--as @noahmanger discovered, sorting by fields with nulls can yield pages and pages of unwanted results. This is also a useful option for the API in general.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,45 @@
+from tests import factories
+from tests.common import ApiBaseTest
+
+from webservices import sorting
+from webservices.common import models
+
+
+class TestSort(ApiBaseTest):
+
+    def test_single_column(self):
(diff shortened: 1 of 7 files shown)


LindsayYoung:
Is the count going to change depending on sorting? I am a little hesitant to return a different set of results based on sorting. I am not 100% against it but would like to talk it over.

jmcarp:
Right, counts would potentially change here. This patch adds an option to exclude nulls on sorted columns, which defaults to false. As implemented, API users would have to specifically pass `sort_hide_null=true` for the results to change.

This issue originally came up in discussion of https://github.com/18F/openFEC-web-app/pull/312.

Issue #1006, opened 2015-06-24
Aggregate Schedule A receipts


jmcarp:
* [ ] By contribution size
    * Cut points?

@noahmanger which other aggregates (if any) are we going to want?

LindsayYoung:
Amount is not necessary because that is on the summary pages.  
 
For committees,  especially candidate committees it is good to figure out from schedule As
 -  Totals by state
 -  Totals by zipcode
 -  Totals given to other groups with FEC ids
 -  Donations by size - totaling donation in a given range, the under $200 bucket comes from unitemized summary filings. 
   The FEC's map currently uses:
      $200 and Under
      $200.01 - $499
      $500 - $999
      $1000 - $1999
      $2000 and Over

 - For contributions by party, it makes more sense to use aggregates.

jmcarp:
Amount = size. I'll revise the issue.

noahmanger:
Great. This all makes sense to me. Though point of clarification, @LindsayYoung do you mean "Totals *received from* other groups with FEC ids"? Might we want to total those up by certain type / designation / org type? Not sure what the distinctions would be, but I could see it being helpful.

LindsayYoung:
For the website, I was thinking listing the top 10 groups that fund a committee. It is possible to then group them by type, but there are a lot of other features I think we should spend our time on first. But tell me if you have a really good use case in mind. 

For, the API would want to have more detailed PAC to PAC transfer information, not just top 10. 

It would be much more difficult to do top individual donors within a committee, we could talk about that, but that is a separate, longer conversation.

jmcarp:
Done a while back.

PR #1054, opened 2015-07-10
Filter contributor aggregates by contributor type.

jmcarp:
Part of #1006.


LindsayYoung:
This is fantastic!

I can add the options so that they are a drop down in the documentation and write up what they mean I found explanations on the instructions for the form so I can add those to the documentation too.

Maybe each line-number should have a human readable code?

http://www.fec.gov/pdf/forms/fecfrm3pposti.pdf
http://www.fec.gov/pdf/forms/fecfrm3i.pdf
http://www.fec.gov/pdf/forms/fecfrm3xi.pdf

jmcarp:
Sounds good on both counts.

PR #1058, opened 2015-07-14
Add indexes on filings columns.

jmcarp:
We only permit sorting on indexed columns, since sorting on non-indexed
columns can be painfully slow. This patch adds indexes on columns on
which the webapp tries to sort filings data.


LindsayYoung:
Thanks for catching this!

PR #1059, opened 2015-07-14
Use <= 200 for first contribution size bucket.

jmcarp:
To match FEC reporting rules.

@LindsayYoung: Think we should keep the rest of the buckets as they are, or also change to `<=` for consistency?


LindsayYoung:
Let's mirror their logic, which on the Presidential Page is: 
 
$200 and under
 $200.01 - $499	
 $500 - $999
 $1000 - $1999
 $2000 +

PR #1074, opened 2015-07-20
Feature/recipient aggregate

jmcarp:
This patch adds two new aggregates on the Schedule B data:
* Aggregate disbursements by recipient name (raw text)
* Aggregate disbursements by recipient committee ID (if not null)

Question: The aggregate by recipient committee ID (obviously) only includes disbursements for which a recipient committee ID is provided. Should the aggregate by recipient name exclude disbursements that *do* have a committee ID? I'm asking because if we include all disbursements in both aggregates, we'll likely see disbursements with recipient committee IDs showing up in both aggregates, which could make the two tables look a little redundant when viewed side by side.

Ping @LindsayYoung @noahmanger

### DIFF
--- a/data/functions/aggregates.sql
+++ b/data/functions/aggregates.sql
@@ -7,6 +7,9 @@ begin
     perform ofec_sched_a_update_aggregate_employer();
     perform ofec_sched_a_update_aggregate_occupation();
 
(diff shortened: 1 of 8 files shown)


LindsayYoung:
I am not sure I remember the current state of designs to answer your previous question.

PR #1078, opened 2015-07-21
Feature/pretty report names

LindsayYoung:
Uses report name or document name and filing year, that way, all filings have a name and if not, it will say document.

This also adds filtering by doc type, I think that would be useful to people

### DIFF
--- a/webservices/common/models.py
+++ b/webservices/common/models.py
@@ -1,3 +1,4 @@
+import re
 from flask.ext.sqlalchemy import SQLAlchemy
 from sqlalchemy.dialects.postgresql import ARRAY, TSVECTOR
 from sqlalchemy.ext.declarative import declared_attr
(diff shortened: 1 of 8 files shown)


webservices/schemas.py
```diff
@@ -297,6 +297,7 @@ class CommitteeSearchListSchema(ApiSchema):
     models.Filings,
     fields={
         'pdf_url': ma.fields.Str(),
+        'document_discription': ma.fields.Str(),
```
jmcarp: Should be `document_description`.

LindsayYoung: Thank you!

webservices/common/models.py
```diff
@@ -855,6 +856,17 @@ class Filings(db.Model):
     update_date = db.Column(db.Date)
 
     @property
+    def document_description(self):
+        if self.report_type_full and self.report_type_full != '':
```
jmcarp: We don't need to check both `self.report_type_full` and `self.report_type full != ''`, since the empty string is falsy in Python. We can just say `if self.report_type_full:` here.

webservices/common/models.py
```diff
@@ -855,6 +856,17 @@ class Filings(db.Model):
...
+            clean_report = report_typere.sub(r'\{[^)]*\}', '', self.report_type_full)
```
jmcarp: I don't see a variable called `reportype_re`. Should this be `re.sub(...)`? Also, like most regex logic, it would be helpful to include a docstring and unit test(s) here.

jmcarp:
LGTM! I made some minor suggestions inline.

webservices/common/models.py
LindsayYoung: :+1:

LindsayYoung:
Ok, fixed the errors and added a test.

PR #1081, opened 2015-07-23
[dontmerge] Hotfix/cap per page

jmcarp:
Don't let users request unlimited records per page, which can lead to out of memory errors. h/t @LindsayYoung.


LindsayYoung:
LGTM!

PR #1083, opened 2015-07-24
Feature/sum election filings

jmcarp:
* Sum each candidate's most recent reports across committees
* Add indexes and filters for tolerable performance
* Update tests


jmcarp:
Note: Includes #1082 from @LindsayYoung.

PR #1085, opened 2015-07-27
WIP Feature/reporting dates

LindsayYoung:
Grabbing the reporting dates that are currently only on staging. 

Filtering is not working yet


LindsayYoung:
@jmcarp 

- `upcoming = true` works on reporting dates but not election dates
- I was trying to use sql alchemy to add report type full. It should be easy enough, but it wasn't working for me.

jmcarp:
Merge conflicts fixed. @LindsayYoung, if this is behaving as expected, I think we're good to merge.

LindsayYoung:
This looks good, though I notice the trc_ tables are on on dev.

PR #1087, opened 2015-07-29
Feature/aggregate contributor type

jmcarp:
Add aggregates by contributor type for committee and for candidate. Required for election pages.


jmcarp:
Yep, agreed. Patch updated @LindsayYoung.

PR #1090, opened 2015-07-30
Implement incremental aggregates using SQLAlchemy.

### LINKED ISSUE #1077: Write aggregate updates using SQLAlchemy
Incremental updates for our large aggregates (Schedule A and B) are currently written as SQL stored procedures. This means that adding another aggregate requires copying and pasting an existing aggregate, then replacing a few column names in the stored procedure (compare https://github.com/18F/openFEC/blob/develop/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_zip.sql and https://github.com/18F/openFEC/blob/develop/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_state.sql). Not that this is news to anyone, but copy-and-pasted code is hard to maintain and test. I propose that we refactor the stored procedures to a Python / SQLAlchemy helper that we can more easily parameterize, so that we don't have to copy code between aggregates anymore.

@arowla: does this seem right to you?

jmcarp:
Run incremental updates using SQLAlchemy Core rather than plan SQL to make it easier to reuse functionality across aggregates. The next step will be to create the initial tables with SQLAlchemy as well.

Not urgent to review @LindsayYoung, just something I was writing in my free time.
cc @arowla 

[Resolves #1077]


jmcarp:
I forgot this still existed. I'm not convinced this is actually a good idea--closing.

PR #1093, opened 2015-07-31
Restrict election filter to running candidates.

jmcarp:
Since we interpolate cycles between election years, we need to filter
candidate history records on both (interpolated) cycle and (raw)
election year.


noahmanger:
Is this going to solve the issue of having multiple senate candidates who aren't running in a particular year showing up?

jmcarp:
That's the idea, yeah.

jmcarp:
Tests passing @LindsayYoung.

PR #1104, opened 2015-08-05
Miscellaneous updates for election pages.

jmcarp:
* Rename "presidential" to "president"
* Un-cap page size on election view
* Update tests


jmcarp:
Since these are very minor changes and @LindsayYoung is going to be out for a few days, I'm going to merge myself. @LindsayYoung, feel free to review later--we can make any changes you recommend in a separate patch.

PR #1107, opened 2015-08-06
Join designation by form ID.

jmcarp:
Previously, the committee history view was built by joining
`dimcmteproperties` on `dimcmtetpdsgn` using date columns. Recently, the
`receipt_date` column on `dimcmtetpdsgn` appears to have been changed
such that nearly all values are `NULL`. This patch uses `form_sk` to
join the tables and preserve designation information.


jmcarp:
This is another tiny patch that I'm going to merge myself. As usual, @LindsayYoung is welcome to review post-merge.

PR #1109, opened 2015-08-07
Add election search view.

jmcarp:
Add election search. To be used in https://github.com/18F/openFEC-web-app/issues/354.


jmcarp:
Ready for review @LindsayYoung. I was looking at using the `trc_election` table to look up elections instead of the candidate history view, but IMO the election table is missing too much data to be useful here. If we do get a complete table of elections later on, we can switch at that point.

PR #1114, opened 2015-08-10
Feature/documention page updates

LindsayYoung:
- added the per_page max to the documentation
 - added the methodology update to the aggregates

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -254,7 +254,9 @@ When comparing the totals from reports to line items. the totals will not match 
 take out items where `"memoed_subtotal":true`. Memoed items are subtotals of receipts
 that are already accounted for in another schedule a line item.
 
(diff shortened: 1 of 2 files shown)


webservices/docs.py
```diff
@@ -254,7 +254,9 @@
 take out items where `"memoed_subtotal":true`. Memoed items are subtotals of receipts
 that are already accounted for in another schedule a line item.
 
-For the Schedule A aggregates, "memoed" items are not included to avoid double counting.
+For the Schedule A aggregates such as `by_occupation` and `by_state` reflect individual
```
jmcarp: How about "Schedule A aggregates, such as `by_occupation` and `by_state`, include only individual contributions, i.e. receipts from line numbers 11AI and 17A."

LindsayYoung: sounds good

PR #1116, opened 2015-08-11
Include committee IDs in election view endpoint.

jmcarp:
Necessary for https://github.com/18F/openFEC-web-app/issues/449.


LindsayYoung:
When I run this locally, I am not getting any data-  in the `elections/search` endpoint, I see a results for president, but no results come up when I try to query `/elections` specifying president and the cycle that shows up in the other end point.

jmcarp:
@LindsayYoung can you send URLs for anything that's not working as expected? I haven't been able to replicate so far (FWIW, this patch doesn't touch election search).

PR #1117, opened 2015-08-11
Use most recent designation row.

jmcarp:
In #1107, we changed the join between dimcmteproperties and
dimcmtetpdsgn to use the form_sk column. This works for most records but
causes problems with missing designation records. This patch joins
dimcmteproperties on the most recent tuple from a separate join of
dimcmteproperties and dimcmtetpdsgn.

Note: If the receipt_date column is repopulated in dimcmtetpdsgn, we can
revert to the logic we were using pre-#1107.


jmcarp:
Replaced by #1122.

PR #1121, opened 2015-08-12
Use Flask-CORS for more robust CORS configuration.

jmcarp:
The API currently uses the `Access-Control-Allow-Origin` header, but not
the `Access-Control-Allow-Headers` header. This leads to CORS errors
when using Safari, which sends the `Accept-Encoding` header with its
`OPTIONS` requests. Rather than allowing the `Accept-Encoding` header,
this patch switches to Flask-CORS for CORS configuration, which has
better defaults and is easier to configure than settings CORS headers
manually.

Thanks @onezerojeremy and @noahmanger for help with debugging.

[Resolves https://github.com/18F/openFEC-web-app/issues/462]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 psycopg2==2.6
+Flask-Cors==2.1.0
 Flask-RESTful==0.3.3
 Flask-SQLAlchemy==2.0
 newrelic==2.52.0.40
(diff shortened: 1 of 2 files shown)


LindsayYoung:
Thanks for figuring this out!

PR #1123, opened 2015-08-12
Add script to output district count per state.

jmcarp:
To be used with the election lookup tool in the webapp. This will help us show the right number of district options for the selected state, as discussed with @noahmanger.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -101,6 +101,12 @@ def load_districts(source=None):
 
 
 @manager.command
+def build_district_counts(outname='districts.json'):
+    import utils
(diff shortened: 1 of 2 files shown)


LindsayYoung:
I think this is fine for now but the FIPS state code don't include territories and the zip codes we have don't include DC, or the territories. I will add that as an issue so we don't forget.

jmcarp:
That would be a nice improvement. For now, we should be all right, since I'm assuming that entities that aren't in the FIPS table have exactly one district (which should be true for Wyoming, American Samoa, DC, etc.).

PR #1136, opened 2015-08-17
Fix filtering on filings view.

jmcarp:
Currently, the filings view mounted as a sub-resource of a candidate or
committee does no filtering, which breaks the filter widget on the
committee and candidate filings pages. This patch adds filtering on
filings sub-resources and merges the two filings resources into a single
class.

Note: To make this play nicely with Swagger, the path parameters for
committee and candidate IDs are prefixed with "path_". This is a hack
but has no effect on the behavior of the API, so I think it's
acceptable.


jmcarp:
Merging this one too; also needs a review post-merge.

Issue #1147, opened 2015-08-24
Inconsistent dates in `vw_filing_history`


jmcarp:
The `vw_filing_history` view includes several columns with date information, but no single column can be used across all form types. Examples: the `report_year` and `election_year` columns are always set to `0` for F1 filings; `receipt_date` is always `null` for RFAI filings. When we filter filings from this view by year, it's unclear which column(s) we should use to determine the year.

It also looks like some date information that's available in FEC Viewer isn't included in `vw_filing_history`. For example, RFAI filings have a "Filed On" date in FEC Viewer, but I'm not seeing that information in `vw_filing_history`.

@jwchumley and @paulclark2: Would it be feasible to populate at least one date column consistently for this view? For example, it would be helpful if `report_year` were always populated, so that we could use a single column for date filtering. If not, can you suggest a method for us to identify the year a given filing was filed? As a (possibly temporary) fix, I'm using `report_year` if present, else the year from `receipt_date`.

PaulClark2:
@jmcarp 
VW_FILING_HISTORY
Form 1 - election year null
	This is correct.  Election years are not part of the Form 1 filing.  

Form 1 - report year null
	This is an error.  I've asked our DBA team to update the script that generates VW_FILING_HISTORY.

RFAI - receipt date null
	This is an error.  The RFAI Date will be used to populate receipt date in VW_FILING_HISTORY. I've asked our DBA team to update the script that generates VW_FILING_HISTORY. 

I'll let you know when the script changes and table updates are completed.

jmcarp:
Thanks! I'll leave the issue open for now, and close once we have updated data and I've reverted the temporary workaround I added to the API to deal with missing data.

jmcarp:
@PaulClark2: Has the DBA team gotten back to you about this?

PaulClark2:
No  ... not yet

LindsayYoung:
So following up on this. There are now no null report_years, so that is great. But I am still seeing 26 nulls in receipt_date.

PaulClark2:
I can fix these two of these and will do that tomorrow.  

FORM_TP | SUB_ID | CAND_CMTE_ID | BEGIN_IMAGE_NUM | END_IMAGE_NUM | PAGES | RPT_YR | RPT_TP | RECEIPT_DT
------------ | --------------------------- | --------------------- | --------------------------- | ------------------------ | --------- | ------------ | ------------ | ----------------
[...]

PR #1148, opened 2015-08-24
Include F1 filings in filings endpoint.

jmcarp:
The filings view is currently filtered on `report_year`, but this column
is always set to 0 for F1 filings. This patch checks both the
`report_year` and `receipt_date` columns, such that F1 filings have
dates and can be included.

Partial workaround for #1147.


arowla:
LGTM pending conflict resolution.

jmcarp:
@arowla conflict resolved. I also added a fix for the individual contribution filter that hadn't made it into the last patch.

PR #1162, opened 2015-08-27
Use detailed query statistics.

jmcarp:
By default, the histograms used by the query planner are too coarse to
capture rarely-occurring values in large tables. This means that queries
by state on rare states (AS, PR) can be extremely slow. This patch
insructs the database to use finer histogram bins for state columns on
the itemized tables and explicitly re-analyzes the tables after creating
indexes for better performance on these queries.


arowla:
+:100:

PR #1164, opened 2015-08-31
Move swagger logic to smore.

jmcarp:
After a few recent updates to smore, custom logic for converting
Marshmallow schemas to JSON Schema dicts and forcing path parameters to
be required is no longer necessary. Depends on the latest version of
smore.


jmcarp:
Note: This is expected to fail until https://github.com/marshmallow-code/smore/pull/24 is merged. I'll make a note here at that point.

jmcarp:
Tests passing, ready for review.

PR #1165, opened 2015-08-31
Feature/election search incumbent

jmcarp:
* Add incumbent officeholders to elections search results
* Add win / lose status to election detail results


LindsayYoung:
I am a little confused because we already have the incumbency data in candidate history. If it is wrong there, we should fix it. I am not seeing how using this view is different from what we have. 

I am a little concerned that we will be inconsistent if we are pulling the same data from 2 different sources.

jmcarp:
It looks like the existing candidate history view would be usable for Senate and House races, but not for Presidential races. As far as I can tell, Senate and House candidates are automatically entered as incumbents for future races after they win an election, which means we can use incumbency data to infer who won an election. But this doesn't seem to happen for Presidential winners:

```
fec=> select * from ofec_candidate_history_mv where candidate_id = 'P00003335';
  idx   | two_year_period | properties_key | candidate_key | candidate_id |      name      |
--------+-----------------+----------------+---------------+--------------+----------------+---
 420826 |            2004 |       10086092 |      10015798 | P00003335    | BUSH, GEORGE W |
 420823 |            2002 |       10086086 |      10015798 | P00003335    | BUSH, GEORGE W | 20
 420820 |            2000 |       10086086 |      10015798 | P00003335    | BUSH, GEORGE W | 20
(3 rows)
```

From this data, we can infer that Bush won in 2000, but we don't know whether he won or lost in 2004. But we do get that information from `cand_valid_fec_yr` (it's a little messy, but the materialized view I wrote cleans it up):

```
fec=> select * from cand_valid_fec_yr where cand_id = 'P00003335';
 cand_valid_yr_id |  cand_id  | fec_election_yr | cand_election_yr | cand_status | cand_ici | c
------------------+-----------+-----------------+------------------+-------------+----------+--
           221152 | P00003335 |            1998 |             2000 | N           | C        | P
           225680 | P00003335 |            2000 |             2000 | C           | O        | P
           229845 | P00003335 |            2002 |             2000 | P           | O        | P
           233587 | P00003335 |            2004 |             2004 | C           | I        | P
           237328 | P00003335 |            2006 |             2004 | P           | I        | P
           245416 | P00003335 |            2010 |             2004 | P           | C        | P
           241263 | P00003335 |            2008 |             2004 | P           | I        | P
           265357 | P00003335 |            2012 |             2004 | P           | C        | P
           297309 | P00003335 |            2014 |             2004 | P           | C        | P
           327331 | P00003335 |            2016 |             2004 | P           | C        | P
(10 rows)
```

Maybe @PaulClark2 can shed some light on the situation and explain where `cand_valid_fec_yr` comes from. But for now, I'm not sure how else we can get incumbents for all offices.

PaulClark2:
A little background about ICI (incumbent, challenger, open).  Frankly, it's a bad data element trying to describe two different things:  1) if the seat is open or not; 2) if the current occupant of the seat is running or not incumbent/challenger.   "Open" means the incumbent isn't running but really is information about the seat not the candidates.  

How ICI is populated at the beginning of a new Congress.  I'll use the beginning of the current, 114th, Congress:  In January 2015 I found candidate IDs for all House and Senate (all 3 classes) incumbents.  If they did not have fec_election_yr = 2016 rows already, I inserted Form 2Z rows for them.  The newly created 2016 row maintained the candidate_election_yr indicated on their most recent Form 2.   So for some this was 2014.  The 2016 row updates when the candidate files a Form 2 for 2016.   

Once I have 2016 rows for all incumbents for the 114th Congress I make sure their ici_status = "I"  For former members that lost elections, if they have 2016 rows, I change their ICI to "C"  

Second term presidents are excluded from this process because they can not be incumbents in the next election.

PaulClark2:
select cand_id, cand_name, cand_ici, cand_valid_yr_id, fec_election_yr, cand_election_yr
from CAND_VALID_FEC_YR where cand_id = 'P00003335' order by fec_election_yr,  cand_election_yr;

For GW Bush, we need to make two corrections to ICI in cand_valid.  

untitled

LindsayYoung:
Thank you Paul!

PaulClark2:
@jmcarp @LindsayYoung Where are we displaying win/loss information?   Is that coming from the TRC Election tables?   Doesn't being an incumbent imply you won?   I guess I'm just confused about what we are trying to do here.  

@jwchumley @akort

PaulClark2:
@jmcarp @LindsayYoung Is this ICI / Win information going the candidate pages?   How would it look there?

akort:
Please leave me out of the conversation, I don't know you and I'm not working on your project. Thanks and Regards, Andreas

AmyKort:
Hi, it's Amy. :)

jmcarp:
@PaulClark2 there are currently two places we're looking to add information about election outcomes: searching for elections, and showing election results. When a user searches for elections (e.g., all elections in the 2015-2016 cycle in zip code 19041), we'd like to show them the incumbent for each relevant election. Second, when we show an election detail page (e.g. VA Senate 2014), we want to indicate which candidate won the race, if the race has already happened.

Just to be clear: all of this is working with this pull request--we were just curious about exactly how the `cand_valid_fec_yr` table was generated.

emileighoutlaw:
The real Amy! Not the imposter! :)

PaulClark2:
Thanks for the clarification @jmcarp   Above I explain how the ICI is updated.  Do you want to know more about how the cand_valid_fec_yr table itself is generated?

LindsayYoung:
This sounds good to me. I'm going to merge. I think it would be a separate issue, but I would love to deal with the seat status and incumbency as separate things in the API.

PR #1172, opened 2015-09-01
Feature/document enhancement

LindsayYoung:
- adds better names
- adds cycle filtering

### DIFF
--- a/data/sql_updates/post_committees_create_filings_view.sql
+++ b/data/sql_updates/post_committees_create_filings_view.sql
@@ -22,6 +22,7 @@ select
     election_year,
     fh.form_type,
     filings_year(report_year, receipt_date) as report_year,
+    report_year + report_year % 2 as cycle,
     report_type,
     to_from_indicator as document_type,
     expand_document(to_from_indicator) as document_type_full,
--- a/webservices/schemas.py
+++ b/webservices/schemas.py
@@ -348,6 +348,7 @@ register_schema(ScheduleEPageSchema)
 FilingsSchema = make_schema(
     models.Filings,
     fields={
+        'cycle': ma.fields.Int(),
         'pdf_url': ma.fields.Str(),
         'document_description': ma.fields.Str(),
     },
--- a/webservices/utils.py
+++ b/webservices/utils.py
@@ -9,6 +9,7 @@ from marshmallow_pagination import paginators
 from webservices import docs
 from webservices import sorting
 from webservices import exceptions
+from webservices import decoders
 
 
 def check_cap(kwargs, cap):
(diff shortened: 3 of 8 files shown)


LindsayYoung:
closes 18F/openFEC-web-app#544

LindsayYoung:
Appended RFAI too much- fixing that now

LindsayYoung:
Ok, that should be good to go

webservices/utils.py
```diff
@@ -141,14 +142,19 @@ def related(cls):
...
         clean = document_type
+    elif form_type and form_type in decoders.form_types:
+        clean = decoders.form_types[form_type]
     else:
         clean = 'Document '
-    return '{0}{1}'.format(clean, report_year)
+
+    if form_type and form_type == 'RFAI':
+       clean = "RFAI: " + clean
+    return re.sub("\s\s+" , " ", '{0} {1}'.format(clean, report_year))
```
jmcarp: * Prefer raw strings for regular expressions: `re.sub(r'...', '...')`
* Prefer single quoted strings
* If the purpose of this is to replaces multiple spaces with a single space, we can use `re.sub(r'\s+', ' ', ...)`
* Else if the purpose is to replace two or more spaces, we can use `re.sub(r'\s{2,}', ...)`, although I'm guessing we want the previous version.
* But actually, if the only purpose of this regex is to remove trailing spaces after `clean`, we don't even need it: `return '{0} {1}'.format(clean.strip(), report_year)`.

data/sql_updates/post_committees_create_filings_view.sql
```diff
@@ -22,6 +22,7 @@ select
     election_year,
     fh.form_type,
     filings_year(report_year, receipt_date) as report_year,
+    report_year + report_year % 2 as cycle,
```
jmcarp: Need to add an index on cycle if we're going to filter on it.

webservices/schemas.py
```diff
@@ -348,6 +348,7 @@ class CommitteeSearchListSchema(ApiSchema):
 FilingsSchema = make_schema(
     models.Filings,
     fields={
+        'cycle': ma.fields.Int(),
```
jmcarp: Do we need to add `cycle` here? Model fields should be added to the schemas automatically--we only include `pdf_url` and `document_description` because they're property methods.

jmcarp:
Made a few notes on the diffs; otherwise LGTM :tada:

webservices/schemas.py
LindsayYoung: You are right, I was adding it as a property before but it was not sorting so I switched it and forgot to take that out.

webservices/utils.py
LindsayYoung: Yea, the multiple spaces happen within the string.

PR #1173, opened 2015-09-02
Use base resource class to avoid duplicated code.

jmcarp:
Clean up some duplicated code. One base resource class to rule them all.


LindsayYoung:
image

PR #1174, opened 2015-09-02
Fix last cash on hand.

### LINKED ISSUE #1110: Verify totals on first candidate comparison table on elections page are correct

jmcarp:
Fix a typo mixing up starting and ending cash on hand. After this
change, fundraising totals on the election page are identical to the
corresponding totals on each candidate's detail page.

[Resolves #1110]

### DIFF
--- a/data/sql_updates/create_totals_house_senate_view.sql
+++ b/data/sql_updates/create_totals_house_senate_view.sql
@@ -7,7 +7,7 @@ with last as (
     )
         cmte_sk,
         two_yr_period_sk,
-        coh_cop as cash_on_hand_end_period,
+        coalesce(coh_cop_i, coh_cop_ii) as cash_on_hand_end_period,
         begin_image_num as beginning_image_number,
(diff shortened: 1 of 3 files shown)


data/sql_updates/create_totals_house_senate_view.sql
```diff
@@ -7,7 +7,7 @@ with last as (
     )
         cmte_sk,
         two_yr_period_sk,
-        coh_bop as cash_on_hand_end_period,
+        coh_cop as cash_on_hand_end_period,
```
LindsayYoung: We check 2 fields for that because filers don't report that consistently 

Example:
openFEC/data/sql_updates/create_reports_house_senate_view.sql
```
coalesce(coh_cop_i, coh_cop_ii) as cash_on_hand_end_period,
```

jmcarp:
Fixed. Out of the tables changed here, only the House/Senate table uses the two cash on hand variables, so the others didn't have to change.

PR #1192, opened 2015-09-14
Move generic helpers and Swagger tools to flask-smore.

jmcarp:
Glue code for parsing request parameters with webargs and marshalling responses with marshmallow, plus generating swagger documentation, has moved over to https://github.com/jmcarp/flask-smore, which is more flexible and better tested than the code in this repo. This patch removes code that lives in flask-smore and takes advantage of improved APIs to reduce duplication.

h/t @sloria for help with flask-smore design


jmcarp:
Ready for review @LindsayYoung

Issue #646, opened 2015-04-28
Add max-width to chart bars


noahmanger:
There should be a `max-width` set on chart bars so that they don't expand so wide if there's only a few.

PR #1195, opened 2015-09-14
Show all candidates in election views.

jmcarp:
Currently, we use an `INNER JOIN` to join candidates on totals for the
election summary and detail views. This has the unwanted effect of
excluding candidates with no totals from these views. This patch uses
a `LEFT OUTER JOIN` and updates the queries accordingly, such that all
relevant candidates are included.

Related to #646.

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -286,7 +286,7 @@ def join_candidate_totals(query, kwargs, totals_model):
     ).join(
         CommitteeHistory,
         CandidateCommitteeLink.committee_key == CommitteeHistory.committee_key,
-    ).join(
+    ).outerjoin(
         totals_model,
         CommitteeHistory.committee_id == totals_model.committee_id,
     )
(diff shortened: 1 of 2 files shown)


LindsayYoung:
Is this affecting the election pages or just the top level aggregate?

jmcarp:
Both. Do you think we should only apply the change to the summary?

LindsayYoung:
Yea, I think that keeping the election pages as clean as possible is a good priority. We could add links on the election pages to the candidate viewer by race to see everyone who filed, if people are looking for candidates without any money.

Can we update the aggregate without disturbing the election pages?

LindsayYoung:
The discussion in #partners seemed to settle on only non-zero candidates on election pages.

jmcarp:
Circling back on this, we're currently excluding all candidates with no financial reports from the election page, and we've stopped showing candidate counts entirely on the homepage widget. If that's the desired behavior, we can close this patch and https://github.com/18F/openFEC-web-app/issues/646.

@PaulClark2 @jwchumley is that what you have in mind? Would you like to restore candidate count on the homepage, or is that a fraught question? We also talked about thresholding candidates at some non-zero amount--is that something we should pursue, or is the current behavior right?

jwchumley:
If you include candidate count it should have a clear label like "66 candidates have reported financial activity". 

Whether to include candidates with no financial activity is an interesting question. I think the only thing required to be considered a candidate is a form 2. We could include all candidates but sort in an order that shows those with financial reports first (in places that display amounts). Maybe?

jmcarp:
After some discussion (see #1408), it sounds like hiding candidates with no financial reports is the correct behavior after all. Closing.

PR #1207, opened 2015-09-18
Use more restrictive join between candidates and committees.

jmcarp:
Prevents double-counting some candidate aggregates on the elections
page.


LindsayYoung:
Good catch!

PR #1210, opened 2015-09-19
Made committee first file date with the filings table. 

### LINKED ISSUE #1209: First file date for form 5 only filers
It looks like we have no electioneering committees in the browse view, because we are automatically sorting by first file date on the F1 and not all committees have F1. We should look at the receipt date from the filings table.

LindsayYoung:
This should be more accurate for groups that do not file F1. It is looking only after the start date because the old ids got reused in some cases.

fixes #1209

### DIFF
--- a/data/sql_updates/create_committee_history.sql
+++ b/data/sql_updates/create_committee_history.sql
@@ -20,7 +20,7 @@ with
         from cycles
         group by committee_id
     ),
-    dcp_original as (
+    filings_original as (
         select committee_id, min(receipt_date) receipt_date from vw_filing_history
         where report_year >= :START_YEAR
         group by committee_id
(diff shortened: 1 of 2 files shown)


data/sql_updates/create_committee_history.sql
```diff
@@ -21,9 +21,9 @@ with
         group by committee_id
     ),
     dcp_original as (
```
jmcarp: Since this isn't pulling from `dimcmteproperties`, let's rename this alias to something that's more descriptive, like `filings` or `filings_original`.

data/sql_updates/create_committee_history.sql
```diff
@@ -21,9 +21,9 @@ with
...
-        select cmte_sk, min(receipt_dt) receipt_dt from dimcmteproperties
-        where form_tp = 'F1'
-        group by cmte_sk
+        select committee_id, min(receipt_date) receipt_date from vw_filing_history
+        where report_year >= :START_YEAR
```
jmcarp: Should we filter this by form type at all? If we don't filter, how about updating the documentation to indicate that `first_file_date` indicates the first form of any type filed by the committee?

LindsayYoung: Good idea

PR #1213, opened 2015-09-21
Feature/fix cumulative totals

### LINKED ISSUE #1212: Some totals variables are too high
A handful of year-to-date and period-to-date variables from reports are being summed in our totals views, leading to large overestimates of those values. Examples include net receipts and net operating expenditures. These values don't show up on the webapp but are available through the API.

jmcarp:
Fix net contributions for presidential data.

Unlike Congressional and PAC and Party reports, Presidential reports
don't include "net contributions" and "net expenditures" fields that are
distinct from from total contributions and expenditures. Currently,
we're treating cycle-to-date contributions and expenditures as net for
the reporting period, which leads to dramatically inflated estimates in
the API. This patch removes net contributions and expenditures from
presidential reports, since they don't seem to exist, and uses the most
recent cycle-to-date contributions and expenditures values for
presidential totals instead of summing cumulative totals from each
report.

[Resolves #1212]

Pinging @LindsayYoung for review.


data/sql_updates/create_reports_presidential_view.sql
```diff
@@ -33,8 +33,6 @@ select
     items_on_hand_liquidated as items_on_hand_liquidated,
     loans_received_from_cand_per as loans_received_from_candidate_period,
     loans_received_from_cand_ytd as loans_received_from_candidate_ytd,
-    net_contb_sum_page_per as net_contributions_period,
-    net_op_exp_sum_page_per as net_operating_expenditures_period,
```
LindsayYoung: We want to take the sums out of totals, but I think we should leave them in all the reports, since this is data reported by the committee and we do display it in the summaries.

PR #1220, opened 2015-09-22
Added contact info back

LindsayYoung:
@leahbannon Adding the google group and stackexchange info


leahbannon:
There were a lot more changes in the link you sent me earlier today - should I submit a separate PR with those? Or am I missing them?

LindsayYoung:
The limitations on data use is already in there. 

The key situation is different, you don't need to explicitly use DEMO_KEY we are adding it as a default if you don't provide one. 

You can make other changes in a separate pr or as a pr to this pr, if I missed something.

jmcarp:
Merging for now--if we need more text, let's add it in separate patches.

PR #1221, opened 2015-09-22
Add cycle filter on Schedule E resource.

jmcarp:
Used by candidate Schedule E filings tab in the webapp.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -483,6 +483,7 @@ entities = {
 }
 
 schedule_e = {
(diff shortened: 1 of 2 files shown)


LindsayYoung:
For the candidate page, can we show the candidate aggregates instead of the line items? So we would just show "$100  Fake PAC  oppose" on the candidate page. What you have here, I think, would be perfect for the committee page.  

Also @mahmanger @emileighoutlaw any ideas how to make it very clear that this money is not spent by the candidate?

Issue #1226, opened 2015-09-23
Identify amended Schedule E transactions


jmcarp:
Unlike the Schedule A and B tables, which are updated in place when amendments are filed, the Schedule E table may include multiple amended versions of the same transaction. That's why @noahmanger noticed the same PAC both supporting and opposing Hillary Clinton in #742. @paulclark2 explained that this happening because Schedule E transactions that are reported on 24- and 48-hour reports haven't been aggregated by unique transaction. In the long term, we'll want to figure out how to properly group those transactions, but this may be complicated and will probably have to be handled by @paulclark2 and @jwchumley (I'm happy to help but much less familiar with the details of the data). In the short term, @paulclark2 and I agreed that we should exclude Schedule E transactions from 24- and 48-hour reports from the aggregates. That means we'll be missing some data, but at least we won't be counting multiple versions of the same transaction.

I'll put in a quick patch to implement tonight. If we go with this approach for the time being, we'll need some explanatory text, probably from @emileighoutlaw. Also flagging @LindsayYoung and @noahmanger for thoughts on this.

dwillis:
I agree that handling Schedule E transactions is complicated, but I think excluding them from 24 and 48-hour reports makes the API much less useful from a timeliness standpoint. Imagine this scenario: multiple IE reports come in on the Monday before a Tuesday election, but don't get into the API until the following week.

jmcarp:
To be clear, this issue is only talking about aggregated Schedule E data by candidate. We'll continue to include all available Schedule E transactions in the itemized view.

LindsayYoung:
To re-iterate we are not filtering out any data from the api, this is just for a methodology to figure out how much outside money is used to oppose or support certain candidates.

Filings will be available here: 
   [https://api.open.fec.gov/v1/filings/?api_key=DEMO_KEY](https://api.open.fec.gov/v1/filings/?api_key=DEMO_KEY)
The line items will be available here:
   [https://api.open.fec.gov/v1/schedules/schedule_e/?api_key=DEMO_KEY](https://api.open.fec.gov/v1/schedules/schedule_e/?api_key=DEMO_KEY)

PR #1227, opened 2015-09-23
Restrict Schedule E aggregates to form F3X.

jmcarp:
This is a temporary fix for #1226. As described in the issue, Schedule E
transactions from 24- and 48-hour reports aren't currently grouped by
unique transaction, such that multiple amended versions of the same
transaction can exist in the database. In the long term, @paulclark2 and
@jwchumley will work on aggregating Schedule E transactions across
amendments, but as a short-term fix, this patch simply restricts
Schedule E aggregates to transactions filed on form F3X.

### DIFF
--- a/data/sql_updates/create_sched_e_aggregate_candidate.sql
+++ b/data/sql_updates/create_sched_e_aggregate_candidate.sql
@@ -11,7 +11,7 @@ select
 from sched_e
 where
     exp_amt is not null and
-    filing_form = 'F3X' and
+    filing_form in ('F3X', 'F5') and
     (memo_cd != 'X' or memo_cd is null)
 group by
     cmte_id,
--- a/data/sql_updates/create_sched_e_aggregate_candidate.sql
+++ b/data/sql_updates/create_sched_e_aggregate_candidate.sql
@@ -9,8 +9,10 @@ select
     sum(exp_amt) as total,
     count(exp_amt) as count
 from sched_e
-where exp_amt is not null
-and (memo_cd != 'X' or memo_cd is null)
+where
+    exp_amt is not null and
+    filing_form = 'F3X' and
+    (memo_cd != 'X' or memo_cd is null)
 group by
     cmte_id,
     cand_id,


noahmanger:
Just flagging that on the frontend we need to explain that's what this is doing.

data/sql_updates/create_sched_e_aggregate_candidate.sql
```diff
@@ -9,8 +9,10 @@ select
     sum(exp_amt) as total,
     count(exp_amt) as count
 from sched_e
-where exp_amt is not null
-and (memo_cd != 'X' or memo_cd is null)
+where
+    exp_amt is not null and
+    filing_form in ('F3X', 'F5') and
```
LindsayYoung: Maybe the filtering is going on somewhere else, but we only want form_tp="SE" from the sched_e table

PaulClark2:
+    filing_form in ('F3X', 'F5')

data/sql_updates/create_sched_e_aggregate_candidate.sql
LindsayYoung: @paulclark2 are you saying: 
- we should be using the `F57 table` instead of the `sched_e` table
- we should be using `filing_form` and not `form_tp`

PaulClark2: For F3X filers the IEs will be in the SCHED_E table.  For F5 filers the IEs will be in the FORM_57 table.  Does this help?  

select * from form_5 f, FORM_57 s where f.sub_id = s.link_id and f.rpt_tp not in ('24', '48') and f.rpt_yr = 2015;
--147 rows returned but doesn't deal with amendments but these behave like regular financial forms because we are excluding 24/48-hour reports.

select * from form_5 f, SCHED_E s where f.sub_id = s.link_id and f.rpt_tp not in ('24', '48') and f.rpt_yr = 2015;
--No rows returned (as expected)

LindsayYoung: Thanks for explaining the table structure, that makes more sense now.

jmcarp:
We've been talking about several ways to filter and combine the data. Let's hammer out a query here and then update the patch. Based on the discussion so far, my best guess looks like this, but please correct me @PaulClark2 and @LindsayYoung:

```

with records as (
	select
		cmte_id,
		s_o_cand_id as cand_id,
		s_o_ind as support_oppose_indicator,
		rpt_yr,
		rpt_tp,
		memo_cd,
		exp_amt
	from sched_e
	union all
	select
		filer_cmte_id as cmte_id,
		s_o_cand_id as cand_id,
		s_o_in as support_oppose_indicator,
		rpt_yr,
		rpt_tp,
		null as memo_cd,
		exp_amt
	from form_57
	join form_5
		on (form_5.sub_id = form_57.link_id)
)
select
    row_number() over () as idx,
    cmte_id,
    cand_id,
    support_oppose_indicator,
    rpt_yr + rpt_yr % 2 as cycle,
    sum(exp_amt) as total,
    count(exp_amt) as count
from records
where
    exp_amt is not null and
    rpt_tp not in ('24', '48') and
    (memo_cd != 'X' or memo_cd is null)
group by
    cmte_id,
    cand_id,
    support_oppose_indicator,
    cycle
;
```

LindsayYoung:
amndt_ind not equal to "A"

Though, what does "C" and "T" mean as an amendment indicator, @PaulClark2 ?

Also, does memo_code apply here?

jwchumley:
C-Continuation
T- Termination

The amendment indicator column unfortunately multipurpose.

jmcarp:
The amendment indicator tells whether a transaction is an amendment, not whether it's been amended. Also, if I'm understanding @PaulClark2 correctly, FEC has already aggregated amended transactions for these tables, so we shouldn't have to deal with this.

jwchumley:
My response to Lindsay about C and T as amendment indicators assumes we 
are talking about forms not transactions. Are you seeing C and T in 
transactions?

LindsayYoung:
Ok, so this sounds good to me. 

Paul, I just did a quick look at the tables we were pulling from, so I wasn't looking after the join.

Thanks @jmcarp @PaulClark2 and @jwchumley for answering my questions!

LindsayYoung:
Just to be clear, @PaulClark2 and @jwchumley does https://github.com/18F/openFEC/pull/1227#issuecomment-142998274 look right to you? 

If so, Josh will implement it and I will merge. 

Thanks!

jwchumley:
Looks right to me. Paul?

PaulClark2:
@jmcarp @LindsayYoung @jwchumley Yes this looks good and it's ready to be merged.

I manually checked a few F3X and F3 filers and Josh's aggregates match.   I did notice a few data issues that FEC will need to fix.  For example:

order by cycle desc, cmte_id; (I added order by .... to your SQL)  The rows below are missing candidate / other id in the postgres DB but have that data in FECP.  

164;"C90013145";"";"S";2016;229577.85;6
168;"C90015454";"";"S";2016;2700.00;2
169;"C90015504";"";"S";2016;719.82;1

PR #1237, opened 2015-09-26
Add 48 hr notice and standardise capitalization

LindsayYoung:
F6 was not listed in the form types section
standardized capitalization

### DIFF
--- a/webservices/decoders.py
+++ b/webservices/decoders.py
@@ -15,11 +15,12 @@ election_types = {
 
 form_types = {
     'F1': 'Statement of organization',
-    'F1M': 'Notification Of Multicandidate Status',
+    'F1M': 'Notification of multicandidate Status',
     'F2': 'Statement of candidacy',
     'F9': '24-hour notice of disbursements for electioneering communications',
     'F10': '24-hour notice of expenditure of personal funds',
     'F11': '24-hour notice of opposition personal funds amount',
     'F12': '24-hour notice of suspension of increased limits',
-    'F99': 'Miscellaneous Document',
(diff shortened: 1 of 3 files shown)


webservices/decoders.py
```diff
@@ -15,11 +15,12 @@
 
 form_types = {
     'F1': 'Statement of organization',
-    'F1M': 'Notification Of Multicandidate Status',
+    'F1M': 'Notification of multicandidate Status',
```
jmcarp: Should we also lower-case "Status"?

webservices/decoders.py
```diff
@@ -15,11 +15,12 @@
...
     'F2': 'Statement of candidacy',
     'F9': '24-hour notice of disbursements for electioneering communications',
     'F10': '24-hour notice of expenditure of personal funds',
     'F11': '24-hour notice of opposition personal funds amount',
     'F12': '24-hour notice of suspension of increased limits',
-    'F99': 'Miscellaneous Document',
+    'F99': 'Miscellaneous document',
+    'F6': '48-Hour notice contribution or loans',
```
jmcarp: For consistency, how about lower-casing "Hour", and using the full title of the form: "48-hour notice of contributions/loans received"?

webservices/decoders.py
LindsayYoung: Yea!

webservices/decoders.py
LindsayYoung: sure, I will update those two

LindsayYoung:
This looks like it already happened

Issue #964, opened 2015-06-13
Hillary Clinton is showing up as running for election in 2000 and 2004


LindsayYoung:
She first ran for president in 2008- 
http://fec-dev-web.cf.18f.us/candidate/P00003392

There is a F2z for Hillary for 2000 in dimcandproperties. This seems like a data error. 

I will email Paul and Jeff during working hours about this.

jmcarp:
Also weird: FEC Viewer seems to have the correct F2 filings for HRC, but it doesn't seem to include any of her financial reports from 2007-2008, which *do* show up in OpenFEC.

PaulClark2:
A general comment, we should not use F2Zs to determine which elections a candidate participates in.   F2Zs are entered for a variety of reasons.  F2s, filed by the candidate, should be use to make this determination.  

There's anything to fix here.  P00003392 was linked to C00342261 (People for American Leadership) a single candidate independent expenditure committee in 1999.    P00003392  was generated on 1/13/1999 .  The Form 1 filed by C00342261 on 1/12/1999 references "hillary4president.org."

noahmanger:
This is still showing up this way. Because this is a data issue for a high profile candidate. I'm moving to a current sprint thing we should try to figure out ASAP.

LindsayYoung:
This would be fixed if we remove f2z from the create cycle logic.

PaulClark2:
Using statutory candidate information (re:  https://github.com/18F/openFEC-web-app/issues/785) or using linked PCCs information are possible solutions.    H. Clinton's candidate ID doesn't not have a PCC linked to it in 2000.

LindsayYoung:
I think this might be resolved in Josh's next pr. might want to discuss if it is not

PR #1245, opened 2015-10-02
Improve the cycle selection for candidates

LindsayYoung:
Checking for an F2 for cycle selection, so that we don't have dropdown options on candidate pages for periods that the candidates didn't file to run for office. 

(I tried to add the same logic to the elections, but it was too aggressive and took out real elections. So, that can be a different PR.) 

This should help with #964


jmcarp:
LGTM

Issue #943, opened 2015-06-09
To help users find the most relevant results, develop a better method of weighting search and typeahead results


LindsayYoung:
Improve search

jmcarp:
Search results aren't sorted alphabetically now: they're ordered by text similarity rank (`ts_rank_cd`) with the input query. Based on conversation earlier today, maybe what we're looking for is a combined weighting by text similarity and some index of importance--maybe total funds raised or something similar.

LindsayYoung:
That it looks alphabetical on the web app

http://fec-dev-web.cf.18f.us/?search_type=committees&search=dccc

http://fec-dev-web.cf.18f.us/?search_type=committees&search=republican

jmcarp:
Oh, I was looking at typeahead search. If the ordering of the typeahead results is correct (or correct enough), it would be really easy to make the `search_type` results use the same order.

jmcarp:
After a bit of digging into this, it actually looks like our sorting on full-text columns has been kind of random all along. Our text vectors and queries are simple enough that text similarity values are mostly small integers that effectively count up the number of times the query tokens occur in the vector (example below). That means that searching for a query like "republican" more or less returns a randomly ordered list of committees that include "republican" in their names.

Postgres includes some normalization methods that weight by the inverse of the length of the string, which makes some intuitive sense--basically weighting by the proportion of words in the vector that belong to the query. But in practice, this just weights towards shorter names, which probably isn't a great proxy for relevance. Again, maybe weighting be recency, money raised, etc. would be helpful.

Thoughts @LindsayYoung @arowla @noahmanger? What are good metrics of search relevance for our users?

```
fec=> select name, ts_rank_cd(fulltxt, query) as rank from ofec_committee_fulltext_mv, to_tsquery('republican' || ':*') query where query @@ fulltxt order by rank desc;
                                                          name                                                          | rank
------------------------------------------------------------------------------------------------------------------------+------
 NEW ALLIANCE OF CALIFORNIA REPUBLICANS-FEDERAL                                                                         |    2
 REPUBLICAN CLUB OF DALLAS: FKA DALLAS COUNTY REPUBLICAN MEN'S CLUB PAC                                                 |    2
 ARIZONA FEDERATION OF COLLEGE REPUBLICANS AZ COLLEGE REPUBLICANS                                                       |    2
 FUND FOR A REPUBLICAN MAJORITY (FKA-REPUBLICAN WHIP LEADERSHIP FUND)                                                   |    2
[...]

Issue #854, opened 2015-05-27
To make search more flexible, enable searching by nicknames


LindsayYoung:
the `q=` kinds of searches are not catching nick names or slight misspellings. I am guessing we could make a more forgiving search if we use sounds like `~`, rather than`=` (I might not be thinking of postgres syntax but I mean whatever will do a more forgiving search)

jmcarp:
Looks like we have a few different options here.

* Postgres full-text search allows the use of custom synonym dictionaries, which could help with nicknames (liz => elizabeth, betsy => elizabeth, bob => robert).
* For a small number of idiosyncratic nicknames (John Ellis => Jeb), we could use the approach we're taking with PACronyms: make a table of name mappings, and join them to the candidate full text search view
* For typos and misspellings, Postgres also has a range of similarity queries (trigram similarity seems to be the most full-featured, but there are other options). This can be used to to suggest better queries on searches with no results ("Did you mean...?").

Thoughts @arowla @LindsayYoung? Have you run into this before @catherinedevlin?

arowla:
If we want to go the route of having a list of idiosyncratic nicknames-- which I think would be a good option in concert with perhaps some fuzzy matching in postgres, like levenshtein distance or soundex-- there are a couple lists of nickname mappings that I/we developed at Sunlight in response to data from the FEC:

https://github.com/sunlightlabs/name-cleaver/blob/master/name_cleaver/nicknames.py
https://github.com/sunlightlabs/datacommons/blob/a8b33cc5a24afe7b8ffcc690ab58b3c84fb0c348/dcentity/tools/nicknames.txt

jmcarp:
Great! So adding a synonym dictionary based on `nicknames.txt` might be a good place to start.

For similarity measures, I get the impression that different methods (soundex, trigram, etc.) have different performance characteristics and handle different use cases better or worse--this would probably require some experimentation to make a good choice.

noahmanger:
Not sure if this is covered with fuzzy matching, but it's also important to account for common misspellings (e.g. "hilary" with one L).

LindsayYoung:
I like where this is headed! Adding nicknames is a great idea and I don't have a preference to which kind of fuzzy matching, as long as we have something.

arowla:
One thing we'll need to find out is whether we can add custom dictionaries
in RDS.

noahmanger:
[...]

PR #1252, opened 2015-10-06
limiting typeahead

LindsayYoung:
This will decrease the chances of offensive names popping up in typeahead.

This is not perfect because there are some issues with dimcandproperties. there are candidates that should show up that don't.

fix 18F/openFEC-web-app#804

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -288,7 +288,7 @@ reporting_dates = {
 }
 
 election_dates = {
-    'election_state': fields.List(fields.Str, description='State of the office sought.'),
+    'election_state': fields.List(fields.Str, description='State or territory of the office sought.'),
     'election_district': fields.List(fields.Str, description='House district of the office sought, if applicable.'),
     'election_party': fields.List(fields.Str, description='Party, if applicable.'),
     'office_sought': fields.List(fields.Str(validate=validate.OneOf(['H', 'S', 'P'])), description='House, Senate or presidential office'),
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -172,7 +172,7 @@ candidate_detail = {
     'office': Arg(str, multiple=True, enum=['', 'H', 'S', 'P'], description='Governmental office candidate runs for: House, Senate or President.'),
     'state': IString(multiple=True, description='U.S. State candidate or territory where a candidate runs for office.'),
     'party': IString(multiple=True, description='Three letter code for the party under which a candidate ran for office'),
-    'year': Arg(str, dest='election_year', description='See records pertaining to a particular election year.'),
+    'year': Arg(str, dest='election_year', description='See records pertaining to a particular election year. The list of election years is based on a candidate filing a F2 for that year.'),
     'district': District(multiple=True),
     'candidate_status': IString(multiple=True, enum=['', 'C', 'F', 'N', 'P'], description='One letter code explaining if the candidate is:\n\
         - C present candidate\n\
@@ -269,7 +269,7 @@ filings = {
 
 
 reports = {
-    'year': Arg(int, multiple=True, description='Year in which a candidate runs for office'),
+    'year': Arg(int, multiple=True, description=docs.REPORT_YEAR),
     'cycle': Arg(int, multiple=True, description=docs.RECORD_CYCLE),
     'beginning_image_number': Arg(int, multiple=True, description=docs.BEGINNING_IMAGE_NUMBER),
     'report_type': Arg(str, multiple=True, description='Report type; prefix with "-" to exclude'),
@@ -306,7 +306,7 @@ reporting_dates = {
 }
 
 election_dates = {
-    'election_state': Arg(str, multiple=True, description='State of the office sought.'),
+    'election_state': Arg(str, multiple=True, description='State or territory of the office sought.'),
     'election_district': Arg(str, multiple=True,
         description='House district of the office sought, if applicable.'),
     'election_party': Arg(str, multiple=True, description='Party, if applicable.'),
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -146,6 +146,7 @@ select distinct on (candidate_id)
     end
 as fulltxt
 from ofec_candidate_detail_mv_tmp
+where candidate_status is not null
 ;
 
 create unique index on ofec_candidate_fulltext_mv_tmp(idx);


webservices/args.py
```diff
@@ -172,7 +172,7 @@ def one_of(value, options):
     'office': Arg(str, multiple=True, enum=['', 'H', 'S', 'P'], description='Governmental office candidate runs for: House, Senate or President.'),
     'state': IString(multiple=True, description='U.S. State candidate or territory where a candidate runs for office.'),
     'party': IString(multiple=True, description='Three letter code for the party under which a candidate ran for office'),
-    'year': Arg(str, dest='election_year', description='See records pertaining to a particular election year.'),
+    'year': Arg(str, dest='election_year', description='See records pertaining to a particular election year. The list of election years is based on a candidate filing a F2 for that year.'),
```
jmcarp: "...filing an F2..."?

jmcarp:
@LindsayYoung can you say more about the candidates that should show up but don't? Are there specific categories of candidates who won't show up, and are there many of them? Are there data errors that should be fixed related to this?

LindsayYoung:
There are two places for the ici status. We started out using the ici table, but @PaulClark2 suggested using dimcandproperties since there were multiple ways to update ICI and the data might be incomplete. 

I am thinking we might want to strategize on how to do a little extra clean up. 

Let's take Issa for an example:
```
=> select * from dimcandstatusici where cand_sk=10025283;
```

 candstatusici_sk | cand_sk  | election_yr | ici_code | cand_status | cand_inactive_flg |      load_date      | expire_date | pg_date 
------------------|----------|-------------|----------|-------------|-------------------|---------------------|-------------|---------
          1100948 | 10025283 |        2004 |          | C           |                   | 2015-06-24 18:55:17 |             | 
          1073269 | 10025283 |        2008 | I        | C           |                   | 2015-06-24 18:55:17 |             | 
          1073271 | 10025283 |        2012 | I        | P           |                   | 2015-06-24 18:55:17 |             | 
          1100953 | 10025283 |        2016 |          | C           |                   | 2015-06-24 18:55:17 |             | 
          1073268 | 10025283 |        2006 | I        | C           |                   | 2015-06-24 18:55:17 |             | 
          1073270 | 10025283 |        2010 | I        |             |                   | 2015-06-24 18:55:17 |             | 
          1100952 | 10025283 |        2014 |          | C           |                   | 2015-06-24 18:55:17 |             | 


```
c=> select election_yr, cand_ici_cd, cand_status_cd from dimcandproperties where cand_sk=10025283;
```

 election_yr | cand_ici_cd | cand_status_cd 
-------------|-------------|----------------
        2000 | O           | C
        2002 | I           | C
        2004 |             | C
        2004 |             | C
        2004 |             | C
        2006 |             | C
        2008 |             | P
        2008 | I           | P
        2008 | I           | C
        2010 |             | P
        2010 | I           | P
        2010 | I           | P
        2006 | I           | C
        2012 | I           | N
        2012 | I           | N
        2012 | I           | N
        2012 | I           | C
        2014 |             | N
        2014 |             | N
        2014 | I           | N
        2014 | I           | C
        2014 | I           | C
        2016 |             | N
        2016 | I           | N
        2016 | I           | F
        2016 | I           | C

Paul, do you think we could  "fill down" the ici status for each candidate in the dimcandproperties table?

LindsayYoung:
Actually, we might just want to add another variable that is basically-- has this person ever been a statuary candidate? We could then filter out those who haven't from typeahead etc.

webservices/args.py
LindsayYoung: Yea, that is jargon-y I will update.

LindsayYoung:
I changed `year` to `election_year` in candidate detail, because that is what we have on master and I don't want to break people's stuff if we can avoid it.

jmcarp:
@LindsayYoung @PaulClark2 what are your thoughts on this? Do you think this patch would hide many interesting candidates? Does the benefit of hiding irrelevant candidates outweigh the cost of potentially hiding some relevant candidates?

LindsayYoung:
I think this is a good idea, this just gets rid of them in the typeahead, and you can still search for them and they will come up. We might be able to improve it later too.

PaulClark2:
The more I think about this the less I like this idea.   Is the goal just to reduce the number of candidates that users see in the type-ahead?   Or is it to reduce seemingly offensive names popping up.   That happens on our website now.  Have we done any testing to confirm that users are actually confused by seeing more candidates?   @onezerojeremy  Does this cause more problems that it solves?   
Currently using the "statutory candidate" logic will exclude the following presidential candidates:  Lawrence Lessig, Jim Webb, Chris Christie, Jim Gilmore and John Kasich.

Examples from earlier this year (this would happen every cycle):  
 J. Bush filed his F2-2016 on 6/15/15 but would have been missing from the type-ahead until the end of July 2015. 
H. Clinton filed her F2-2016 on 4/13/15 but would have been missing from the type-ahead until the end of July 2015. 
T. Cruz filed his F2-2016 on 3/23/15 but would have been missing from the type-ahead until the end of July 2015.
B. Sanders filed his F2-2016 on 4/30/2015 but would have been missing from the type-ahead until the end of July 2015.

@jwchumley @amykort @amypike thoughts?

jwchumley:
I agree with Paul. It isn't worth losing real candidates just to try to keep offensive names out of the type ahead.

LindsayYoung:
It is a simple thing to roll back.

There was an error where candidates without committees would be eliminated on current fec.gov site and those changes were not being pushed through the data warehouse. Is that still the case, Paul? If so, might we might want to hold off until that is fixed and then go back to the broader search?

onezerojeremy:
Someone remind me what "statutory candidate" means? Sorry. 

We haven't tested specifically for this, but I have heard a smattering of complaints of users looking for "the real" candidates. I know the reality or unreality of the candidates is more complex than the users model, and we haven't tested it in the type-ahead specifically. My gut, which is not data, is that it's a pretty low risk place to experiment with a switch, given that it's a convenience feature (a convenience feature that people LOVE) that doesn't replace the full search. 

Also, it'll be easy to test for confusion: I can add a search for one of those you've mentioned ( Lessig or Webb or whoever) to the testing docket for next week and see if anyone gets tripped up.

noahmanger:
I believe "Statutory candidate" in this sense means someone who has reported $5,000+ raised. 

I think the biggest concern isn't the offensive names, but in posing a barrier to people getting to what they're looking for quickly. Considering that this change would mean that certain legitimate candidates are temporarily not included in the typeahead until they meet the threshold, I see how this could have the contrary effect we're looking for. 

But on the other hand, filing the typeahead with garbage results also poses a barrier. Which means we probably need to make a judgement call about what's better. 

I like the idea of adding a test along those lines.

jmcarp:
This change has been reverted. Unless testing shows that "fake" candidates are a significant barrier to users finding what they're looking for, I agree with Paul and Jeff that we shouldn't hide candidates who don't have status entries. FWIW, @leahbannon ran into this issue and reported a bug almost immediately after we hid the candidates.

An alternative approach is to get back to the question of ordering typeahead search results. They're not shown in any particular order right now, so sorting by funds raised, candidate status, etc. would help bring relevant candidates to the top of the list. This is something @LindsayYoung has brought up before--maybe it's time to implement.

onezerojeremy:
The alternative approach of sorting makes sense to me. 

Confusion for something missing is easier to test for in usability testing than confusion due to signal to noise ratio. Anyone have ideas on what are some particularly noisy searches?

LindsayYoung:
My favorite approach is still weighting by money raised in some way. Perhaps cumulative money raised or how much they raised in their last 2-year cycle. If that is not enough, we might want to add extra weight if a candidate is active.  

Here is an example of the search not doing that great of a job if surfacing the most likely candidates to the top. If you search for Clinton:
 1. Some guy who filed in 1995 and never raised any money
 2. Bill Clinton-- but not President Clinton, A 2016 candidate for the Democratic Farm Labor Party
 3. Hillary Clinton for President
 4. Hillary Clinton for Senate
 5. Former President Bill Clinton

screen shot 2015-10-16 at 10 03 42 am

LindsayYoung:
Here is a committee example. The result people are most likely looking for is listed 5th.

screen shot 2015-10-16 at 10 12 57 am

jwchumley:
what data elements do you have access to for this? Maybe you use max(years active) then name and possibly total receipts?

LindsayYoung:
Right now it is just going off of name last year active should be easy to add.

Here is a previous thread about this too #943 and fuzzy matching came up in #854

leahbannon:
@LindsayYoung I love that idea of weighting by money raised because I think it solves both problems of too much noise and risking hiding candidates too much

noahmanger:
Let's move this conversation over to the original issue for weighting search results: https://github.com/18F/openFEC/issues/943

PR #1253, opened 2015-10-07
Feature/upgrade deps

jmcarp:
Some of our core dependencies (marshmallow, webargs, smore) have added useful new features (but also breaking changes) in recent releases. This patch upgrades to the latest versions of the relevant dependencies. Lots of things changed, but IMO the code is simpler now than before. The biggest change is that the `Arg` object from webargs is gone--now both request parsing and response marshalling work with marshmallow `Field` objects.


LindsayYoung:
Just merged, but then looked and removing the args breaks our documentation.

PR #1254, opened 2015-10-07
Parse args from querystring.

jmcarp:
Now with args from the querystring.


LindsayYoung:
The tests pass for me locally

jmcarp:
My mistake here was leaving a few remaining references to smore in the code (it's been renamed to "apispec"). Tests passed locally for both of us because we still had smore installed. Tests are passing on travis now.

LindsayYoung:
Thanks!

PR #1259, opened 2015-10-08
Import election dates and Senate classes.

jmcarp:
* Add election dates and Senate classes from @PaulClark2
* Remove some unused code in manage.py

@LindsayYoung: please take a look and see if there's more information we should be loading here. We're going to need Senate classes for future enhancements to election pages and for summing financials over Senate terms, but there may be other use cases you have in mind that I haven't thought about.


jmcarp:
@PaulClark2, now that we're importing the election dates with Senate classes that you sent, I'm wondering if it would be possible to integrate this table with the `trc_election` table. Both tables seem to include a lot of overlapping information, but each has a few columns that the other doesn't. Ideally, we'd have exactly one place to look for information related to election dates. Do you think the tables could be merged in some way?

LindsayYoung:
@PaulClark2 is there a way to crosswalk those tables?

LindsayYoung:
Added the data meeting tag about the crosswalk. 

I am going to take another look at this for now and merge if it looks right to me.

PR #1260, opened 2015-10-08
Further simplify report year logic.

jmcarp:
I was looking over my patch yesterday and realized it should be even simpler. We don't need to extract the year from the receipt date at all now that we have reliable report years in the filings view.


LindsayYoung:
Yea, I like the idea of trying to catch data issues upstream, rather than keep putting in workarounds.

PR #1262, opened 2015-10-09
It was casting report year as a decimal for me locally 

LindsayYoung:
report year as a decimal.

This is the only way it works for me locally.

### DIFF
--- a/data/sql_updates/post_committees_create_filings_view.sql
+++ b/data/sql_updates/post_committees_create_filings_view.sql
@@ -12,7 +12,7 @@ select
     receipt_date,
     election_year,
     fh.form_type,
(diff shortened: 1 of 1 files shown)


arowla:
Sorry, drive-by here. ;)

Numeric can still be a decimal. You should cast to an integer (`smallint`, actually, for the greatest memory savings).

`date_part('year', receipt_date)::smallint as report_year,`

LindsayYoung:
Sorry, I think was badly translating the underlying problem. 

Thanks for the `smallint` memory saving tip! I am sure I will need to know that in the future. 

I will close this in favor of Josh's simplifying request #1260

PR #1264, opened 2015-10-09
Summarize itemized aggregates by committee and election.

jmcarp:
Create materialized views summing itemized aggregates by committee and
election year. Note: materialized views are created using SQLAlchemy
Core rather than the usual SQL scripts to simplify code reuse and
testing.

* [x] Summarize Schedule A aggregates
* [ ] Summarize Schedule B aggregates
* [ ] Add election summaries to flask-script tasks
* [ ] Write unit tests

Based on recent updates from FEC, this is now a lower priority, but I thought I'd get my work committed and solicit some feedback from @arowla before switching gears. I'd also like to discuss with @LindsayYoung and @PaulClark2 if Alison is convinced that the overall approach makes sense.

### DIFF
--- a/webservices/common/ddl.py
+++ b/webservices/common/ddl.py
@@ -1,0 +1,33 @@
+"""Utilities for creating and dropping (materialized) views using SQLAlchemy.
+"""
+
+from sqlalchemy.schema import DDLElement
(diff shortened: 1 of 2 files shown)


jmcarp:
Replaced by #1352.

PR #1268, opened 2015-10-13
Feature/change log

LindsayYoung:
Using https://github.com/skywinder/github-changelog-generator to auto-generate a change log. The change log begins at the last public release. In the docs, there is also a link to where the log will be when it is on master.


jmcarp:
This looks good, but it looks like the changelog includes some pretty old stuff that I'm thinking happened before the last release. It might help to pass the `--since-tag` option to `github_changelog_generator` to restrict the issues and patches more.

LindsayYoung:
To clarify, I ran the script, added the `.github_changelog_generator` and re-ran the command before committing.

LindsayYoung:
@PaulClark2 and @jwchumley we are adding a [change log](https://github.com/18F/openFEC/blob/feature/change-log/CHANGELOG.md) to the API do make it more clear what is being updated with each release.

Right now, the interactivity on the documentation at https://api.open.fec.gov/developers is broken, and there are some formatting errors. It would be nice to do another release so that the documentation was working for the quarterly deadline. I don't think it needs any fanfare.

There is one new endpoint: 
https://fec-dev-api.18f.gov/developers#!/financial/get_elections_summary

Everything else is bug fixes and methodology updates.

PaulClark2:
Sounds like a good idea to me.   Thanks for all your work on this Lindsay and Josh.

Issue #1271, opened 2015-10-15
bug: Incorrect office districts for various candidates


jmcarp:
While debugging https://github.com/18F/openFEC-web-app/issues/766, I noticed that many candidates seem to have incorrect office districts, with values of 99 instead of their true values. So far, I haven't seen a clear pattern here: both recent and older records are affected, and all the relevant records so far have come from F2 (and not F2Z) records. Also, these records all appear to be correct on FEC Viewer.

@PaulClark2 and @jwchumley: I've included a few sample queries and results below (for Dennis Heck, James Sensenbrenner, and David Giles). Please let me know if I'm using the wrong queries here.

cc @LindsayYoung 

```
fec=> select office_district, * from dimcand join dimcandoffice using (cand_sk) join dimoffice using (office_sk) where cand_id = 'H0WA03161';
 office_district | office_sk | cand_sk  |  cand_id  | form_sk | form_tp |      load_date      | expire_date |         pg_date
-----------------+-----------+----------+-----------+---------+---------+---------------------+-------------+--------------------
 03              |  10000509 | 10008223 | H0WA03161 | 1458283 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:20:42
 99              |  10000516 | 10008223 | H0WA03161 | 1458283 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:20:42
 99              |  10000516 | 10008223 | H0WA03161 | 1458283 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:20:42
 99              |  10000516 | 10008223 | H0WA03161 | 1458283 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:20:42
(4 rows)
```

```
fec=> select office_district, * from dimcand join dimcandoffice using (cand_sk) join dimoffice using (office_sk) where cand_id = 'H8WI09050';
 office_district | office_sk | cand_sk  |  cand_id  | form_sk | form_tp |      load_date      | expire_date |         pg_date
-----------------+-----------+----------+-----------+---------+---------+---------------------+-------------+--------------------
 99              |  10000526 | 10031976 | H8WI09050 | 1538510 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:25:12
 99              |  10000526 | 10031976 | H8WI09050 | 1538510 | F2      | 2015-06-24 00:00:00 |             | 2015-07-09 06:25:12
[...]

PR #1277, opened 2015-10-15
Hide district 99 from election search.

jmcarp:
Hide elections with district "99" from election search results.

Note: this is a workaround; a better solution would be to fix the
underlying data, as described in #1271.

Related to https://github.com/18F/openFEC-web-app/issues/766

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -89,6 +89,11 @@ class ElectionList(utils.Resource):
             CandidateHistory.two_year_period,
         ).filter(
             CandidateHistory.candidate_inactive == None,  # noqa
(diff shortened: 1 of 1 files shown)


webservices/resources/elections.py
```diff
@@ -89,6 +89,11 @@ def _get_elections(self, kwargs):
             CandidateHistory.two_year_period,
         ).filter(
             CandidateHistory.candidate_inactive == None,  # noqa
+            # TODO(jmcarp) Revert after #1271 is resolved
+            sa.or_(
+                CandidateHistory.district == None,  # noqa
+                CandidateHistory.district != '99',
```
LindsayYoung: Makes sense for now.

PR #1284, opened 2015-10-21
Handle duplicate candidate-committee links.

jmcarp:
Because some candidate-committee links occur more than once for a given
year in `dimlinkages`, some committee totals were multiple-counted for
the election summary endpoint. This patch adds the necessary DISTINCT ON
to the query and updates the unit tests to simulate duplicate linkage
records.

Related to https://github.com/18F/openFEC-web-app/issues/646.

PR #1285, opened 2015-10-21
Exclude expired linkage rows from views.

jmcarp:
This patch fixes double-counting in election summary totals. Some totals
were multiply counted because of repeated candidate-committee linkage
rows; this patch ignores these expired duplicate rows. Thanks
@LindsayYoung and @PaulClark2 for guidance.

* Exclude all linkage rows where `expire_date is not null`.
* Add missing index on linkage view.
* Use helpers to expand acronyms.

Replaces #1284.


LindsayYoung:
I like that you shifted the type decoding to the functions.

PR #1289, opened 2015-10-22
Included amended F3* records.

jmcarp:
* Include F3* records with non-null `expire_date`
* Add "amended" filter to committee reports views
* Index `expire_date`

Note: this requires a quick patch to the frontend so that only current
reports are requested for the committee detail view. This will be
submitted shortly.

[Resolves https://github.com/18F/openFEC-web-app/issues/874]

### DIFF
--- a/data/sql_updates/create_reports_house_senate_view.sql
+++ b/data/sql_updates/create_reports_house_senate_view.sql
@@ -97,12 +97,12 @@ from
     left join dimdates end_date on cvg_end_dt_sk = end_date.date_sk and cvg_end_dt_sk != 1
 where
     two_yr_period_sk >= :START_YEAR
-    and f3.expire_date is null
 ;
 
 create unique index on ofec_reports_house_senate_mv_tmp(idx);
 
 create index on ofec_reports_house_senate_mv_tmp(cycle);
+create index on ofec_reports_house_senate_mv_tmp(expire_date);
 create index on ofec_reports_house_senate_mv_tmp(report_type);
 create index on ofec_reports_house_senate_mv_tmp(report_year);
 create index on ofec_reports_house_senate_mv_tmp(committee_id);
--- a/data/sql_updates/create_reports_pacs_parties_view.sql
+++ b/data/sql_updates/create_reports_pacs_parties_view.sql
@@ -119,12 +119,12 @@ from
     left join dimdates end_date on cvg_end_dt_sk = end_date.date_sk and cvg_end_dt_sk != 1
 where
     two_yr_period_sk >= :START_YEAR
-    and f3x.expire_date is null
 ;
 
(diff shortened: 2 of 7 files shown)


jmcarp:
One point about this: we're adding a new query parameter to the filings endpoint called "amended". If "true", only show amended records; if "false", only show non-amended records; if not specified, show all records. This is a concise way to describe the filtering rules, but it occurs to me that it might be confusing. Happy to revise if folks have better ideas, but this does solve the immediate problem.

cc @LindsayYoung

LindsayYoung:
Yea the language on that is a bit confusing. 

How about: 
`is_amended=true` only records that have been amended
`is_amended=false' only records that have not been amended

noahmanger:
Will we also still have the `amendment_indicator` parameter? This could be confusing to have both.

LindsayYoung:
How about keeping this for now, so that there isn't breaking changes and doing this as part of some systematic clean up later?

noahmanger:
Could we keep it on the API but remove the filter on the /filings page and
replace it with the new one?

On Thu, Oct 22, 2015 at 12:00 PM, Lindsay Young <notifications@github.com>
wrote:

> How about keeping this for now, so that there isn't breaking changes and
> doing this as part of some systematic clean up later?
>
> —
> <https://github.com/18F/openFEC/pull/1289#issuecomment-150323346>.
>



-- 
Noah Manger
18F <http://18f.gsa.gov> | Work: 415-696-6146

LindsayYoung:
Absolutely, @noahmanger!

jmcarp:
> Could we keep it on the API but remove the filter on the /filings page and
replace it with the new one?

Well, no. The new filter ("amended", potentially to be renamed to "is_amended") applies to the reports view, not the filings view. We don't currently have that information for filings.

LindsayYoung:
I forgot we don't have that view. Thanks Josh

noahmanger:
Ah. Ok. So... this stays? 
image

How will the new filter be applied?

jmcarp:
@noahmanger this patch makes zero changes to the filings view, so that filter doesn't need to change. I'm not sure if it's clear or useful, but that's a separate question.

noahmanger:
Gotcha. So amended reports will still be distinguished by `amendment_indicator=N` and amendments will have `amendment_indicator=A`?

LindsayYoung:
We don't have a good way to show what has been amended in the filings, which is what is listed and filterable on that page. 

Once you click on this to look at the panel, we have the reports view, which will have the true/false for amendments.

is_amended true/false form the reports and N, A or null from filings don't correspond to each other because one table only knows if something has been amended and the other only knows if the record is an amendment.

LindsayYoung:
LGTM will merge when tests pass.

Issue #1247, opened 2015-10-05
primary/unique key in models such as filings & reports


pulkitnandan:
While trying to create a local cache for queries to api.open.fec I'm stuck to figure out a way to update my local database using the fec's schema which states no unique key for schema's such as filings. So how can my database be consistent with the fec's api one where I'm periodically querying fec's api for data and as there's no primary key, I can't update a record at my end, while I can stop duplicate insertion by query sql db that a particular row is present or not. Is there anything I'm missing in respect to my database or fec's api is not meant for such things.

LindsayYoung:
For filings,  beginning_image_number and ending_image_number are unique, as is the document's url, which is based on the beginning image number. 

Hope that is helpful!

pulkitnandan:
Oh Ok. Yes it solves my issue.

Thanks

LindsayYoung:
That shouldn't happen, we will look into the cause of that.

LindsayYoung:
That was due to a data error and the FEC has fixed it. Thanks for taking the time to report this and make the project better.

fedex1:
I agree with @pulkitnandan that it would be helpful to have a single UUID (https://en.wikipedia.org/wiki/Universally_unique_identifier)  and it would be helpful if the openfec API was consistent across all endpoints.

For example, add uuid field for each endpoint.

My question is specific to Schedule A

See http://www.fec.gov/finance/disclosure/metadata/DataDictionaryContributionsbyIndividuals.shtml

There seems to be a lot of references to "unique" but none of the fields are actually unique for the individual contribution.  Even when used in combination.

![screenshot from 2015-10-20 10 31 14](https://cloud.githubusercontent.com/assets/167614/10610532/4b031370-7716-11e5-93bc-e980dd02c084.png)


Can you tell me what the UUID is for Schedule A API https://api.open.fec.gov/v1/schedules/schedule_a

LindsayYoung:
For `schedule_a`:
`file_num` in the bulk data should translate to `file_number` in the API.
`sub_id` should translate directly to `sub_id` in the API.

Tell us if you are you getting non-unique results you getting different results than the bulk data.

Thanks!

fedex1:
Hi @LindsayYoung 

Thanks for answering.

It appears `file_number` is ONLY unique to the particular form submitted  (just a guess)

But the form can contain thousands of individual contributions.

And the `sub_id` does NOT uniquely describe the individual contribution.

[...]

PR #1290, opened 2015-10-22
Serialize sub_id values as strings.

jmcarp:
To avoid truncation of large numbers by some JSON parsers, serialize
`sub_id` and `original_sub_id` values on filings and itemized
transactions as strings.

Related to #1247.


jmcarp:
cc @LindsayYoung

PR #1291, opened 2015-10-22
Update LindsayYoung's case

mbland:
The Team API should probably be case-insensitive, but this fixes a breakage for now.

cc: @jmcarp @LindsayYoung


LindsayYoung:
you are always on my case @mbland

mbland:
This is very true, @LindsayYoung. :stuck_out_tongue:

PR #1292, opened 2015-10-22
Handle multiple representations of null districts.

jmcarp:
Following changes in the construction of the candidate history view,
representations of missing states and districts changed such that
missing districts may be encoded as "00" or `None`. Until this
inconsistency is resolved, handle both missing values in the election
search view.

h/t @noahmanger

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -136,7 +136,13 @@ class ElectionList(utils.Resource):
                 ),
                 # Senate and presidential races from matching states
                 sa.and_(
-                    CandidateHistory.district_number == None,  # noqa
+                    # Note: Missing districts may be represented as "00" or `None`.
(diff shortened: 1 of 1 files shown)


jmcarp:
cc @LindsayYoung

LindsayYoung:
Maybe I am over thinking, but district 00 is how an at-large district is represented, are we treating that differently from cases where district is not being reported?

jmcarp:
In this specific case, it shouldn't matter--here we're just trying to fix a bug on joining zip codes on districts. This seems to work fine for single-district states and territories. If you want to explore, try http://localhost:5000/v1/elections/search/?cycle=2016&state=AS, etc.

LindsayYoung:
Ok, makes sense to me.

PR #1300, opened 2015-10-26
Use `greatest` instead of `coalesce` for missing values.

### LINKED ISSUE #1298: Cash on hand issue
https://fec-proxy.18f.gov/data/elections/senate/PA/2016/

Only Ed Pawlowski shows cash on hand. Everyone else is Zero. 

Should be:
Toomey = $8,598,021
Sestak = $2,416,291
McGinty = $892,359
Fetterman = $143,923 
Vodvarka = $40,412

jmcarp:
Many columns are paired in the source tables (e.g. coh_cop_i, coh_cop_ii).
We currently use `coalesce` to choose the first non-null value for these
pairs, but in some cases, the missing value is zero rather than `NULL`.
To handle these cases, this patch uses `greater` instead of `coalesce`.

Thanks @jwchumley for reporting.

[Resolves #1298]


data/sql_updates/create_totals_house_senate_view.sql
```diff
@@ -45,7 +45,7 @@ select
     sum(hs.tranf_to_other_auth_cmte_per) as transfers_to_other_authorized_committee,
```
LindsayYoung: We want to replace the coalesce on line 40 too, right?

jmcarp: Yep, had that one in the buffer but forgot to save. Thanks for catching!

PR #1311, opened 2015-10-28
Made fec_bootstrap.sh exit if a command fails.

### LINKED ISSUE #1246: In order to make it easier to use fec_bootstrap.sh, don't ignores command failures
If any of the commands run by fec_bootstrap.sh fail, I think that the script should exit immediately.

- [ ] Add `set -e` to fec_bootstrap.sh.

http://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin

evankroske:
See [this blog entry][1] for why I didn't use `set -e`. Fixes #1246.

[1]: http://david.rothlis.net/shell-set-e/

### DIFF
--- a/scripts/bootstrap/fec_bootstrap.sh
+++ b/scripts/bootstrap/fec_bootstrap.sh
@@ -1,62 +1,62 @@
 #!/usr/bin/env bash
 
-source `which virtualenvwrapper.sh`
+source `which virtualenvwrapper.sh` &&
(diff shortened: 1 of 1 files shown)


LindsayYoung:
Thanks for taking your time to improve the openFEC repo! 

I also found the `set -e` blog entry interesting, I had not heard of that before. 

:+1:

PR #1314, opened 2015-10-28
Expose expiration date on IE reports.

jmcarp:
Expose expiration date on IE reports.

Fixes an uncaught exception on fetching IE reports with
`is_amended=false`. Thanks @PaulClark2 for reporting.

To verify, check a URL like localhost:5000/v1/committee/C00590398/reports/?is_amended=false (or likewise for any IE filer) and check that the response code is 200, not 500.

Attention @LindsayYoung


LindsayYoung:
LGTM

PR #1316, opened 2015-10-29
fixing milestones problem

leahbannon:

PR #1318, opened 2015-10-29
Milestone changes

gboone:
I think the missing dashes are causing the production errors. @mbland @jeremiak @davidebest I wonder if there's a way to make the validator pick up on a syntax error like this?


gboone:
I think this is a duplicate of #1316. I withdraw this PR.

mbland:
https://team-api.18f.gov/api/errors/. Will be priming that endpoint with more data soon.

PR #1320, opened 2015-10-30
Send email on nightly refresh.

### LINKED ISSUE #1200: In order to prevent users from seeing stale data, email results of nightly import / refresh to dev team on complete / fail.
At the moment, we aren't being notified when the nightly data import tasks complete (or, occasionally, fail). I propose sending an email with logs to the dev team after each import so we can more easily detect and fix errors.

jmcarp:
* Capture logs from refresh
* Log exceptions from refresh
* Send results and timing to team

Todo:
* [x] Unit tests
* [x] Documentation

[Resolves #1200]


jmcarp:
Ready for review @LindsayYoung. To verify, set the environment variables to reasonable values, then run the nightly refresh, and wait for an email to show up:

```
export MANDRILL_API_KEY=<api key>
export FEC_EMAIL_SENDER=<some email address>
export FEC_EMAIL_RECIPIENTS=<some email address>

python manage.py shell

> import cron
> cron.refresh()
```

A few other points:
* We'll want to add the new environment variables to our user-provided services after merge.
* This patch switches our test runner from nose to pytest. Pytest runs unittest-style tests, so we don't have to change any of our existing code. But going forward, we get cleaner assertions (`assert 1 < 2` vs `self.assertLess(1, 2)`), flexible test fixtures, monkey-patching, parameterized tests, etc. I like pytest.
* The new tests use the `mock` module to stub out calls to the Mandrill API. @LindsayYoung: we haven't used mocks elsewhere in the tests, so we can talk more about how they're being used here if it would be useful.

LindsayYoung:
This looks good to me. I tested it out a bit but I don't have a valid API key. Do you want me to merge or does in need to be in sync with updating the environment variables?

jmcarp:
I added the relevant credentials to our user-provided services, so this should work on CF. I can send you the credentials privately, or you can grab them from CF with `cf env web-a`.

LindsayYoung:
Works for me!

LindsayYoung:
There is a conflict in the readme I will resolve that and then merge

PR #1325, opened 2015-11-03
Feature/update committee data

### LINKED ISSUE #1287: Handle committees that file F3s before F1s
Some committees begin to file F3* forms before filing their first F1. For example, committee C00507384 filed its first F1 in 2011, but filed year-end F3X forms in 2007 and 2008. We're currently pulling basic committee information from `dimcmteproperties` (name, treasurer, party, etc.), and because there are no records in that table before 2011, the API returns empty results for the 2008 and 2010 cycles.

The current FEC site does show data for these cycles, and I'm guessing it comes from `cmte_valid_fec_yr`, since the relevant information isn't available from F1 filings or `dimcmteproperties` alone. If that's correct, then I propose that we switch from `dimcmteproperties` to cmte_valid_fec_yr` to fetch basic committee details. This would bring us in line with the current site, and resolve https://github.com/18F/openFEC-web-app/issues/875.

Does this sound right @PaulClark2 @LindsayYoung?

jmcarp:
Where possible, use `cmte_valid_fec_yr` instead of `dimcmteproperties`,
`dimparty`, and other data warehouse tables for committee materialized
views. This table is better vetted than the tables we're currently
using.

[Resolves #1287]
[Resolves https://github.com/18F/FEC/issues/109]

Paging @LindsayYoung for review. This should resolve a bunch of miscellaneous data errors (likely including some we haven't even noticed yet!), but it needs to be tested thoroughly on dev before it's ready for production.

Also, as implemented, this patch still pulls several fields from `dimcmteproperties` when they aren't available on `cmte_valid_fec_yr`. This includes treasurer address, custodian committee information, and fax number, for a few examples. As I understand it, we're going to want to stop using `dimcmteproperties` entirely, as with all tables that live in the data warehouse. If that's correct, it would be ideal if we could add those columns to `cmte_valid_fec_yr`, if they're useful. Thoughts @PaulClark2?

### DIFF
--- a/data/sql_updates/create_committee_history.sql
+++ b/data/sql_updates/create_committee_history.sql
@@ -3,54 +3,42 @@ create materialized view ofec_committee_history_mv_tmp as
 with
     cycles as (
         select
-            committee_id,
-            generate_series(
-                min(get_cycle(report_year)),
-                max(get_cycle(report_year)),
-                2
-            ) as cycle
-        from vw_filing_history
-        group by committee_id
-    ),
-    cycle_agg as (
-        select
-            committee_id,
-            array_agg(cycles.cycle)::int[] as cycles,
-            max(cycles.cycle) as max_cycle
-        from cycles
-        group by committee_id
+            cmte_id,
+            array_agg(fec_election_yr)::int[] as cycles,
+            max(fec_election_yr) as max_cycle
+        from cmte_valid_fec_yr
+        group by cmte_id
     ),
     filings_original as (
         select committee_id, min(receipt_date) receipt_date from vw_filing_history
         where report_year >= :START_YEAR
         group by committee_id
     ),
-    candidate_agg as (
+    candidates as (
         select
-            cmte_sk,
+            cmte_id,
             array_agg(distinct cand_id)::text[] as candidate_ids
         from dimlinkages dl
         where dl.expire_date is null
-        group by cmte_sk
+        group by cmte_id
     )
-select distinct on (dcp.cmte_sk, cycle)
+select distinct on (fec_yr.cmte_id, fec_yr.fec_election_yr)
     row_number() over () as idx,
-    cycles.cycle,
+    fec_yr.fec_election_yr as cycle,
     dcp.cmte_sk as committee_key,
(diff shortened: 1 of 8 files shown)


LindsayYoung:
What do you think about adding `date_entered` and `date_changed` as filterable columns while we make this change?

data/sql_updates/create_committee_history.sql
```diff
@@ -79,35 +67,33 @@ select distinct on (dcp.cmte_sk, cycle)
...
     dcp.cmte_custodian_zip as custodian_zip,
     -- Properties
-    dcp.cmte_email as email,
+    fec_yr.cmte_email as email,
     dcp.cmte_fax as fax,
-    dcp.cmte_web_url as website,
+    fec_yr.cmte_url as website,
     dcp.form_tp as form_type,
     dcp.leadership_pac as leadership_pac,
-    dcp.load_date as load_date,
```
LindsayYoung: If we have to get rid of `load_date` from dimcmteproperties, we can add `date_entered` and `date_changed`.

@jwchumley and @paulclark2 do you think that `date_entered` is comparable enough to `load_date` that we could call it that?

jwchumley: Context? Why do you have to get rid of load_date and what are you using 
the date to accomplish?

jmcarp: The `load_date` column comes from `dimcmteproperties`, which we're doing our best to stop using. It also looks like an field that captures internal FEC procedures (when a record was entered into the database) rather than information from the forms (when a report was filed). We can use `date_entered` and `date_changed`, but I'm not sure these fields add value.

LindsayYoung: I think that date changed could be valuable if you wanted to sort and see committees that had recent updates. I think there might be value to finding the new and changing content for advanced data users.

LindsayYoung:
OK, this looks good. It fixes the bugs and I spot checked a bunch of committees. 

We will want to make clear in the change log that this depreciates `load_date` 

I still think people would be interested in a sortable `date_changed` field but I don't think that discussion should hold up this pr.

Issue #1327, opened 2015-11-03
To improve data quality, migrate away from data warehouse tables


jmcarp:
In general, tables from the data warehouse (among others, tables named dim*) are less reliable than tables and views used to power FEC Viewer. As more of the latter tables are pushed to postgres, we should use more data from them and less from the data warehouse. Since some fields in the DW tables aren't available in the newly pushed views, we may want to change these views (e.g., `cand_valid_fec_yr`, `cmte_valid_fec_yr`) so that we don't need to rely on the DW.

Here's a partial list of DW tables that we'll eventually want to stop using:
* [x] dimcandproperties
* [ ] dimcmteproperties
* [ ] dimparty
* [x] dimcmtetpdsgn
* [x] dimlinkages
* [x] dimcandstatusici
* [ ] factpacsandparties_f3x
* [ ] factpresidential_f3p
* [ ] facthousesenate_f3
* [ ] factindpexpcontb_f5
* [ ] sched_a
* [ ] sched_b
* [ ] sched_e

@PaulClark2 and @LindsayYoung: Feel free to comment, change the list of tables, or otherwise improve this issue. We're also going to need to push over more tables before we can start on some of these--for example, at some point we'll want to get access to `F_REPORT`, and the FECP equivalent of `dimlinkages`.

Also, #1325 drops `dimparty` and `dimcmtetpdsgn`, and gets us pretty far toward dropping `dimcmteproperties`.

jmcarp:
Since we're currently getting the `candidate_inactive` flag from `dimcandstatusici`, which we're planning to stop using, it would also be useful to get the `cand_inactive` table pushed over, as described in https://github.com/18F/openFEC/issues/1193#issuecomment-140137067. It would also be useful to get the FECP equivalent of `dimparty`.

LindsayYoung:
One reason for the delay in totals getting to the API is that the data warehouse tables are taking longer to  show the updated data. This might make us want to prioritize moving away from the data warehouse tables.

For example:
Today is the 22, the filing deadline was the 20th, I can see the new records in `f_rpt_or_form_sub` but not in `factpresidential_f3p` I did a quick check using the Clinton and Rubio campaigns- C00575795, C00458844

FYI:
@PaulClark2 @jwchumley @noahmanger @wluoFEC

noahmanger:
You mentioned the new tables have less info. What info specifically are they lacking?

LindsayYoung:
[...]

PR #1331, opened 2015-11-03
Feature/deprecate dimlinkages

### LINKED ISSUE #1328: Bug: "Palouse Says Enough" listed as a principal committee for Bernie Sanders
Cross-posted from https://github.com/18F/openFEC/issues/1304#issuecomment-151236236:

A committee named "Palouse Says Enough" is listed as one of Bernie Sanders's principal committees in `dimlinkages`. @jwchumley and @PaulClark2, can you look into why this committee is showing up as one of Sanders's principal committees? I'm wondering if this is another issue that will be resolved for free when as we move away from the DW--it looks like the data are incorrect in `dimlinkages`, but hopefully they're correct in whatever the FECP equivalent of that table is. If that's the case, there's no need to investigate the issue in `dimlinkages`--we can just push over the FECP table and go from there.

```
fec=> select cmte_tp, cmte_dsgn, * from dimlinkages where cand_id = 'P60007168' and cmte_id = 'C00589432';
 cmte_tp | cmte_dsgn | linkages_sk | fec_linkage_id | cand_sk  | cmte_sk  |  cand_id  | cand_election_yr | fec_election_yr |  cmt
---------+-----------+-------------+----------------+----------+----------+-----------+------------------+-----------------+-----
 P       | P         |     4948611 |         210123 | 10009887 | 10059737 | P60007168 |             2016 |            2016 | C005
(1 row)
```

cc @LindsayYoung

jmcarp:
Use candidate-committee linkage data from FECP instead of the data
warehouse. Part of #1327.

[Resolves #1328]

Attention @LindsayYoung


jmcarp:
Merge conflicts resolved @LindsayYoung.

data/sql_updates/create_linkage_names_view.sql
```diff
@@ -1,42 +1,13 @@
-drop materialized view if exists ofec_name_linkage_mv_tmp;
-create materialized view ofec_name_linkage_mv_tmp as
+drop materialized view if exists ofec_cand_cmte_linkage_mv_tmp;
```
LindsayYoung: If we want to rename the table, do we want to rename the file `create_cand_cmte_linkage_view.sql`?

jmcarp: Good catch, will change :+1:

tests/test_reports.py
```diff
@@ -151,7 +151,7 @@ def test_reports_sort_default(self):
             datetime.datetime(2015, 7, 5),
         ]
         dates_formatted = [isoformat(each) for each in dates]
-        reports = [
+        [
```
LindsayYoung: This is super minor, since it works fine, but few lines up, in `test_reports_sort`, those reports are assigned to `reports`. It might be nice to keep that consistent.

jmcarp: Agreed. I'm taking out a bunch of these variable names, which makes `test_reports_sort` consistent with this style. This will also silence a bunch of linter errors around creating variables that are never referenced.

jmcarp:
Patch updated based on suggestions from review. Let me know if anything else needs revisions @LindsayYoung.

LindsayYoung:
:+1: looks good I notice it fixes a lot of 404s for committees that register but don't have spending info.

PR #1334, opened 2015-11-04
Cache compiled dependencies for faster builds.

jmcarp:
See https://github.com/nickstenning/travis-pip-cache.

To verify that this works, run the build more then once; expect faster install times after the first run.


jmcarp:
Nice! Builds got a few minutes faster.

cc @cmc333333

PR #1336, opened 2015-11-05
Add manually entered districts to data import.

jmcarp:
Because loading the district to zip code mapping from CSV can be slow,
we dump the related tables to `data/districts.dump` so that we can
quickly reload them using `./manage.py load_districts`. This patch fixes
a few miscellaneous bugs in the data import tasks and rebuilds
`districts.dump` with districts that have been manually added since we
originally added these tables.

[Resolves https://github.com/18F/FEC/issues/82]

I'm going to push the rebuilt district table to all our RDS instances. Ping @LindsayYoung for review.


manage.py
```diff
@@ -75,8 +75,8 @@ def load_table(frame, tablename, indexes=()):
 @manager.command
 def build_districts():
     import pandas as pd
-    load_table('ofec_fips_states', pd.read_csv('data/fips_states.csv'))
-    load_table('ofec_zips_districts', pd.read_csv('data/natl_zccd_delim.csv'), indexes=('zcta', ))
+    load_table(pd.read_csv('data/fips_states.csv'), 'ofec_fips_states')
```
LindsayYoung: Just curious, did the order change for a reason?

jmcarp: Yep, it was wrong before :confounded:

LindsayYoung:
fixes 18F/FEC#82

PR #1340, opened 2015-11-05
Censor duplicate committee IDs in itemized tables.

### LINKED ISSUE #1306: Recipient Name in disbursement browse
After Demo: 
There are four disbursements here. All of them are to Facebook and all of them show Pattie Curran Committee as the recipient instead of Facebook. 

https://fec-proxy.18f.gov/data/disbursements/?committee_id=C00574590&recipient_state=CA&disbursement_description=social+media&min_date=01-01-2015&max_date=12-31-2016

### LINKED ISSUE #1330: Revisit rules for individual contributions
Context: A few months back, we constructed a set of rules to identify individual contributions that aren't earmarked. Receipts must have a receipt type of 15, 15E, 15J, or 18J, or be less $200, appear on line number 11AI, 12, 17, 17A, or 18, and not have a word like "EARMARK" in their memo text to count as individual and non-earmarked; see https://github.com/18F/openFEC/issues/1141#issuecomment-132635114 for the original discussion and https://github.com/18F/openFEC/blob/develop/data/functions/individual.sql for our implementation.

These rules seem to work most of the time, but we're running into some exceptions:

* As pointed out by @jwchumley in #1308, some receipts that appear to be earmarks are classified as non-duplicates by our rules because they've been flagged with the internal 15E receipt type code. It's possible that the receipts in question were coded incorrectly, or that our rules should change to accommodate them.
* As reported by a user in https://github.com/18F/FEC/issues/126, the Right to Rise PAC has suspiciously few contributions--and all its contributions are < $200. Based on the F3 PDFs, it looks like Right to Rise has actually received a substantial amount in large-dollar contributions, but most of them have been coded with a `receipt_tp` of `10`, meaning that our rules exclude them. Again, I'm not sure if we should approach this by fixing the codes or changing our rules.

Since this category of issue affects at least one PAC that's likely to be in the news this cycle, I'm hoping we can make quick progress here. If we decide that we need to change our rules, I can implement this quickly.

cc @PaulClark2 @LindsayYoung @wluoFEC

jmcarp:
As described in #1306 and #1330, committees sometimes enter the same values for
`contbr_id` and `cmte_id` (for receipts) and for `recipient_cmte_id` and
`cmte_id` (for disbursements). In these cases, values of `contbr_id` and
`recipient_cmte_id` should be ignored for receipts and disbursements,
respectively. This patch creates `clean_contbr_id` and
`clean_recipient_cmte_id` columns, which are displayed in the API
instead of the original `contbr_id` and `recipient_cmte_id` columns.

[Resolves #1306]
[Resolves #1330]

To verify, we'll rebuild the itemized tables on the dev server and verify that the incorrect contributor and recipient committee IDs described in #1306 aren't displayed in the API (or on the data site).

### DIFF
--- a/data/functions/clean.sql
+++ b/data/functions/clean.sql
@@ -3,7 +3,18 @@
 -- * "Green Party Added)" becomes "Green Party"
 create or replace function clean_party(party text)
 returns text as $$
-    begin
-        return regexp_replace(party, '\s*(Added|Removed|\(.*?)\)$', '');
-    end
+begin
+    return regexp_replace(party, '\s*(Added|Removed|\(.*?)\)$', '');
+end
+$$ language plpgsql;
+
+-- Compare two values. If equal, return `NULL`, else return the first value.
+create or replace function clean_repeated(first anyelement, second anyelement)
+returns anyelement as $$
+begin
+    return case
+        when first = second then null
+        else first
+    end;
+end
 $$ language plpgsql;
--- a/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_contributor.sql
+++ b/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_contributor.sql
@@ -4,17 +4,17 @@ create table ofec_sched_a_aggregate_contributor as
 select
     cmte_id,
     rpt_yr + rpt_yr % 2 as cycle,
-    contbr_id,
+    clean_repeated(contbr_id, cmte_id) as contbr_id,
     max(contbr_nm) as contbr_nm,
     sum(contb_receipt_amt) as total,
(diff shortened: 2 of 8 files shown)


jmcarp:
For context, there aren't very many records that this would affect. Here are the numbers for the post-2011 receipts and disbursements (which include about 40 million and 20 million rows, respectively):

```
fec=> select count(*) from ofec_sched_a where cmte_id = contbr_id;
 count
-------
 18484
(1 row)

fec=> select count(*) from ofec_sched_b where cmte_id = recipient_cmte_id;
 count
-------
 88098
(1 row)
```

PR #1347, opened 2015-11-10
Pin node and buildpack versions.

jmcarp:
The Cloud Foundry team recently updated the node buildpack and the
default version of node installed. This causes deployments to fail on
buildpack compilation. This patch pins our node and buildpack versions
to the last known working releases.


LindsayYoung:
Oops I merged on GitHub instead of gitflow

PR #1352, opened 2015-11-13
Sum candidate aggregates by election cycle.

jmcarp:
This is part of the 2/4/6-year aggregation feature.

Add a `period` flag to candidate arguments. When true, aggregates are
summed over the two-year periods of the relevant election cycles. This
patch replaces #1264, since it's cheap to compute these aggregates on
the fly.

* [x] Sum totals over election cycles
* [x] Sum committee aggregates over election cycles
* [x] Sum candidate aggregates over election cycles
* [x] Sum election summaries over election cycles
* [x] Unit test all of the above

### DIFF
--- a/tests/test_aggregates.py
+++ b/tests/test_aggregates.py
@@ -1,5 +1,8 @@
 import functools
 
+from tests import factories
+from tests.common import ApiBaseTest
+
 from webservices import schemas
 from webservices.rest import db, api
 from webservices.resources.aggregates import (
@@ -12,9 +15,6 @@ from webservices.resources.candidate_aggregates import (
     ScheduleAByStateCandidateView,
     TotalsCandidateView,
 )
-
-from tests import factories
-from tests.common import ApiBaseTest
 
 
 class TestAggregates(ApiBaseTest):
@@ -290,9 +290,9 @@ class TestCandidateAggregates(ApiBaseTest):
         }
 
     def test_totals_full(self):
-        """Assert that all two-year totals for the given election period,
-        including the current two-year period and the two preceding periods
-        (since the test candidate is a Senate candidate).
+        """Assert that all two-year totals for the given election period are
+        aggregated by candidate, including the current two-year period and the
+        two preceding periods (since the test candidate is a Senate candidate).
         """
         totals = self.get_totals()
         last_totals = totals[:2]
--- a/tests/test_elections.py
+++ b/tests/test_elections.py
@@ -1,11 +1,11 @@
 import datetime
 import functools
 
-from webservices.rest import db, api
-from webservices.resources.elections import ElectionList, ElectionView, ElectionSummary
-
 from tests import factories
 from tests.common import ApiBaseTest
+
+from webservices.rest import db, api
+from webservices.resources.elections import ElectionList, ElectionView, ElectionSummary
 
 
 class TestElectionSearch(ApiBaseTest):
--- a/data/nicknames.csv
+++ b/data/nicknames.csv
@@ -1,0 +1,6 @@
+candidate_id,nickname
+P60007168,bernie sanders
+S4VT00033,bernie sanders
+H8VT01016,bernie sanders
+P00003392,hilary clinton
+S0NY00188,hilary clinton
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -135,6 +135,13 @@ create index on ofec_candidate_detail_mv_tmp using gin (election_years);
 drop table if exists dimcand_fulltext;
 drop materialized view if exists ofec_candidate_fulltext_mv_tmp;
 create materialized view ofec_candidate_fulltext_mv_tmp as
+with nicknames as (
+    select
+        candidate_id,
+        string_agg(nickname, ' ') as nicknames
+    from ofec_nicknames
+    group by candidate_id
+)
 select distinct on (candidate_id)
     row_number() over () as idx,
     candidate_id as id,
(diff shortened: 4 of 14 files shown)


jmcarp:
@arowla, would you have time to review over the next week?

arowla:
Yes. Putting it on my list.

On Sat, Nov 14, 2015 at 12:23 AM, Joshua Carp <notifications@github.com>
wrote:

> @arowla <https://github.com/arowla>, would you have time to review over
> the next week?
>
> —
> <https://github.com/18F/openFEC/pull/1352#issuecomment-156635920>.
>



-- 
Alison Rowland
202-317-0124

arowla:
Are you ready for me to review this, @jmcarp? Asking b/c of the WIP annotation.

jmcarp:
Yep! I still have a few unit tests on my list (should have those done later today), but I don't think that needs to hold up review.

jmcarp:
Realized I forgot to tag @arowla in my reply above--sorry for the possibly duplicate mention.

tests/test_elections.py
```diff
@@ -105,10 +106,7 @@ def setUp(self):
                 disbursements=75,
                 committee_id=self.committees[0].committee_id,
                 coverage_end_date=datetime.datetime(2012, 9, 30),
-                last_beginning_image_number=123,
```
arowla: Are all these removals just obsolete/unused fields?

webservices/args.py
```diff
@@ -412,11 +412,18 @@ def make_seek_args(field=fields.Int, description=None):
         validate=validate.OneOf(['house', 'senate', 'president']),
         description='Office sought, either President, House or Senate.',
     ),
+    'period': fields.Bool(missing=False),
 }
 
 schedule_a_candidate_aggregate = {
     'candidate_id': fields.List(IStr, required=True, description=docs.CANDIDATE_ID),
     'cycle': fields.List(fields.Int, required=True, description=docs.RECORD_CYCLE),
+    'period': fields.Bool(missing=False),
```
arowla: I'm not quite sure I fully understand the meaning of this argument... can you explain?

jmcarp: If the `period` flag is `True`, then records are aggregated over the relevant election period. For example, if the user requests data for a presidential candidate with `cycle=2012&period=true`, then the response will sum aggregates from 2010 and 2012. This flag has analogous behavior for all the endpoints touched by this patch.

This is probably a signal that the flag isn't named well--suggestions welcome.

webservices/schemas.py
```diff
@@ -252,8 +252,10 @@ class CommitteeSearchListSchema(ApiSchema):
 ScheduleEByCandidateSchema = make_schema(
     models.ScheduleEByCandidate,
     fields={
-        'committee': ma.fields.Nested(schemas['CommitteeHistorySchema']),
-        'candidate': ma.fields.Nested(schemas['CandidateHistorySchema']),
+        'committee_id': ma.fields.Str(),
```
jmcarp: Sorry for the collateral refactor! I noticed that this query run significantly faster when we didn't nest the entire candidate and committee, and now that I've thought about it more, I don't see the benefit of nesting anyhow.

tests/test_elections.py
jmcarp: Yes, exactly. These were some fields that we shoehorned into this endpoint when the frontend required them. It doesn't anymore, and the endpoint is clearer and faster without them.

jmcarp:
Tests updated--ready for more review @arowla.

About the name of the `period` flag: I filed an issue for discussion in #1371. If we can settle on a better name, I'll update the patch.

tests/test_aggregates.py
```diff
@@ -17,68 +19,108 @@
...
+                    candidate_id=self.candidate.candidate_id,
                     committee_id=self.committee.committee_id,
                     cycle=2012,
+                    election_full='true',
                 )
             )
-            self.assertEqual(len(results), 1)
-            self.assertEqual(results[0], schema().dump(aggregate).data)
+            assert len(results) == 1
+            serialized = schema().dump(aggregates[0]).data
```
arowla: These tests might benefit from some comments. I _think_ I've got what's going on here, but a little description would help minimize the head-scratching.

tests/test_elections.py
```diff
@@ -145,27 +163,36 @@ def test_empty_query(self):
...
-                'S',
-            ),
+            'total_receipts': sum(each.receipts for each in totals),
+            'total_disbursements': sum(each.disbursements for each in totals),
+            'cash_on_hand_end_period': sum(each.last_cash_on_hand_end_period for each in totals),
+            'won': False,
+        }
+        self.assertEqual(results[0], expected)
+
+    def test_elections_full(self):
```
arowla: I'm not seeing a test here for when the year parameter and election_full are mismatched. Did I just miss it?

arowla:
LGTM aside from the couple of comments.

tests/test_elections.py
jmcarp: The behavior in this case is to return no results--the redirect we talked about here happens in the webapp but not in the API. I added a test for this behavior in 52c8b9ae4b62c17a6a315c8db9bb3723e485c0c4.

tests/test_aggregates.py
jmcarp: Agreed, I'm not sure I'll understand what's going on here in a few months without some documentation. Added explanations of a few complicated tests in 52c8b9ae4b62c17a6a315c8db9bb3723e485c0c4.

jmcarp:
Thanks for reviewing @arowla! I added documentation and an extra test case based on your comments. Think this is good to merge? :pray:

arowla:
Added one comment, otherwise I'd say it's good to go!

PR #1359, opened 2015-11-15
Feature/nplusone

jmcarp:
Use `nplusone` to detect inefficient lazy loads and unnecessary eager loads. This won't have any effect on production behavior, but will raise errors in tests when potentially inefficient queries are made.


jmcarp:
@arowla this might interest you--this patch adds query profiling with `nplusone`, which I wrote over the summer to find potentially slow queries due to lazy-loading in sqlalchemy and django (and peewee, for some reason).

arowla:
Very cool! I couldn't quite tell from the docs-- does it also profile how long queries take and implement some reporting threshold on the basis of that, or just log at every instance of a lazy-load?

jmcarp:
It only handles lazy loads and unused eager loads (e.g., using `joinedload` in SQLAlchemy or `prefetch_related` in Django, but never accessing the related items). It doesn't replace a standard query profiler based on query times, but it can give specific feedback about how to make queries faster (use a prefetch on a specific relationship, accessed on a specific line). It's basically a port of this thing: https://github.com/flyerhzm/bullet.

tests/test_itemized.py
```diff
@@ -249,7 +249,7 @@ def test_amount_sched_e(self):
         ]
         results = self._results(api.url_for(ScheduleEView, min_amount=100))
         self.assertTrue(all(each['expenditure_amount'] >= 100 for each in results))
-        results = self._results(api.url_for(ScheduleAView, max_amount=150))
+        results = self._results(api.url_for(ScheduleEView, max_amount=150))
```
LindsayYoung: Don't think it is wrong, but  wondered why you changed the test from schedule A to schedule E.

jmcarp: This should've been Schedule E in the first place--I was just fixing a typo.

LindsayYoung: That makes sense then :)

jmcarp:
@LindsayYoung: to test, comment out `query_options` in `webservices/resources/sched_a.py`, then run `py.test -v tests/test_itemized.py`. Tests should fail with messages like "nplusone.core.exceptions.NPlusOneError: Potential n+1 query detected on `ScheduleA.contributor`".

LindsayYoung:
This seems like it is really straight forward, but it is not giving me any errors. I will look at some other PRs to review and come back to this later.

PR #1363, opened 2015-11-19
[WIP] Feature/download utils

jmcarp:
Add miscellaneous helpers for use with data downloads.

* [x] Invoke resource classes outside Flask
* [x] Chunked iteration over database queries
* [x] Download endpoint
* [x] Convert records to CSV
* [x] Name CSV files by query
* [x] Upload CSV files to S3
* [x] Prevent running duplicate tasks
    * https://github.com/TrackMaven/celery-once
    * http://docs.celeryproject.org/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time
* [x] Unit tests for all of the above

For collaboration with @LindsayYoung


data/sql_incremental_aggregates/prepare_schedule_a_aggregate_contributor.sql
```diff
@@ -17,12 +17,14 @@ and (memo_cd != 'X' or memo_cd is null)
...
 
+alter table ofec_sched_a_aggregate_contributor add column idx serial primary key;
+
 -- Create indices on aggregate
-create index on ofec_sched_a_aggregate_contributor (cmte_id);
-create index on ofec_sched_a_aggregate_contributor (cycle);
-create index on ofec_sched_a_aggregate_contributor (contbr_id);
-create index on ofec_sched_a_aggregate_contributor (total);
-create index on ofec_sched_a_aggregate_contributor (count);
+create index on ofec_sched_a_aggregate_contributor (cmte_id, idx);
```
arowla: Clearly, creating these multi-column indexes seems central to the strategy here. I am curious, though, whether you also tried letting Postgres combine single-column indexes for each key + `idx`. My only concern is that these indexes could be getting quite large... have you looked at how much memory each of these take (vs. working memory on the RDS instance)? If the indexes get large enough, it could limit Postgres' ability to keep frequently-used indexes in the cache. But if Postgres can effectively use single-column indexes in combination, it might be slightly slower per-query, but your average time could be faster, since you're less likely to get hit with the swap time necessary for Postgres to load up the index(es) it needs.

tests/test_candidates.py
```diff
@@ -43,33 +40,32 @@ def test_candidate(self):
...
-        self.assertResultsEqual(result['name'], candidate.name)
+        assert result['candidate_id'] == candidate.candidate_id
+        # # most recent record should be first
+        assert result['name'] == candidate.name
         # #address
-        self.assertEqual(result['address_city'], candidate.address_city)
-        self.assertEqual(result['address_state'], candidate.address_state)
-        self.assertEqual(result['address_street_1'], candidate.address_street_1)
-        self.assertEqual(result['address_zip'], candidate.address_zip)
+        assert result['address_city'] == candidate.address_city
```
arowla: It seems like there's a little inconsistency here in whether dict values are asserted one per line, or against an expected dict. As long as these asserts are being changed, is there a reason to keep them in the former style, rather than convert them to the latter? Personally, I prefer the latter (as in where you've used `assert_dicts_subset`). Do they both result in comparably helpful errors when they fail?

PR #1366, opened 2015-11-20
Feature/celery app

jmcarp:
Move celery beat to separate app.

* Extract celery beat from web app
* Use redis service for celery message broker
* Extract development requirements

This patch extracts celery beat from a background process to a separate application, and uses redis instead of sqlite for the celery message broker. This will make the celery process more reliable, since Cloud Foundry only keeps the main process alive. It also lays foundations for the future data download future, which requires a shared message broker across application instances, rather than a separate sqlite broker for each application.

To verify, deploy to the development space using `cf push beat-worker -f manifest_dev.yml`, and check that the refresh task is run periodically.


jmcarp:
Thanks for resolving conflicts @LindsayYoung--this is ready for review and merge. I created redis services for staging and production.

PR #1374, opened 2015-11-27
Feature/candidate nicknames

jmcarp:
Add table of hard-coded candidate nicknames, analogous to committee
"pacronyms". Nicknames are grouped by candidate and added to the
candidate fulltext table on building materialized views.

[Resolves https://github.com/18F/FEC/issues/165]

@leahbannon @LindsayYoung: so far, this only includes Bernie Sanders. We should still go over the major candidate and add them to the nicknames table included in this patch as necessary. And we still likely want to think about a more general fix for nicknames--this is just a quick fix that hard-codes names for important candidates.


jmcarp:
Added "Hilary Clinton" as suggested by @noahmanger and fixed merge conflicts. Pinging @LindsayYoung for review.

LindsayYoung:
There is a conflict, would you like 'load_nicknames' in 'update_schemas' or 'update_all'?

jmcarp:
Conflict resolved--I moved `load_nicknames` to `update_all`.

jmcarp:
Based on discussion in Slack, I think this is good to review and merge. The additional nicknames that came up in discussion are already either handled (Rick Perry, Ted Cruz) or aren't actually nicknames (it's hard to spell "Luetkemeyer", but we don't actually have common misspellings).

@LindsayYoung

manage.py
```diff
@@ -50,6 +50,17 @@ def execute_sql_folder(path, processes):
...
+    import sqlalchemy as sa
+    try:
+        table = sa.Table('ofec_nicknames', db.metadata, autoload_with=db.engine)
+        db.engine.execute(table.delete())
+    except sa.exc.NoSuchTableError:
+        pass
+    load_table(pd.read_csv('data/nicknames.csv'), 'ofec_nicknames')
+
+@manager.command
 def load_pacronyms():
```
LindsayYoung: I notice that if I run update_all more than once I get an error-

```
sqlalchemy.exc.InternalError: (psycopg2.InternalError) cannot drop table ofec_pacronyms because other objects depend on it
DETAIL:  materialized view ofec_committee_fulltext_mv depends on table ofec_pacronyms
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
 [SQL: '\nDROP TABLE ofec_pacronyms'] 
```

jmcarp: Thanks, I updated the patch and don't see this happening anymore. Can you verify that it's behaving as expected on your side as well?

PR #1379, opened 2015-11-30
Feature/notify deploy

### LINKED ISSUE #1375: Add notifications for CF deployments
We should know which branch is deployed to each Cloud Foundry environment at all times. Code can be deployed both manually and automatically, via Travis CI, so it isn't always obvious which code is running where. I propose adding a post-deploy Slack notification, like this:

"openFEC/develop deployed to fec/dev via build c3ad10".

If this pollutes our channel too much, we can add a separate fec-deploy channel, but that's another question.

jmcarp:
* Ping Slack on successful deploy
* Remove redundant new relic invocation

[Resolves #1375]

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,4 +1,5 @@
-python manage.py cf_startup && \
-    (pkill gunicorn || true) && \
-    gunicorn webservices.rest:app --bind 127.0.0.1:8000 --daemon && \
+python manage.py cf_startup &&
+    (pkill gunicorn || true) &&
+    gunicorn webservices.rest:app --bind 127.0.0.1:8000 --daemon &&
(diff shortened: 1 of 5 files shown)


jmcarp:
Ping @LindsayYoung for review.

jmcarp:
Patch updated @LindsayYoung. I wrote up a better explanation in https://github.com/jmcarp/openFEC/commit/5a151059bdf7859fa2223c3ad96ddcb387763fa6, but I modified this to send Slack notifications from Cloud Foundry directly, and not from the machine requesting the deploy. This means the deployer (Travis, or either of us) doesn't have to have a local Slack token, and then deploying via `cf push` will still notify via Slack.

PR #1384, opened 2015-12-04
Feature/date upgrades

LindsayYoung:
- Adds report names
 - Registers documentation for election dates

Working on tests and later will get rid of lazy loading.


webservices/common/models/dates.py
```diff
@@ -1,18 +1,39 @@
+import re
+
+from sqlalchemy import ForeignKey
+
 from webservices import decoders
 
 from .base import db
 
+class ReportNames(db.Model):
+    __tablename__ = 'dimreporttype'
```
jmcarp: Didn't notice this earlier, but it would be best to avoid using tables from the data warehouse in new functionality. Let's ask FEC if there's an equivalent table in FECP that we can use instead.

LindsayYoung: good point!

webservices/args.py
```diff
@@ -280,11 +280,14 @@ def make_seek_args(field=fields.Int, description=None):
...
 
 reporting_dates = {
-    'due_date': fields.List(fields.Date, description='Date the filing is done.'),
+    'min_due_date': fields.Date(description='Date the report is due.'),
+    'max_due_date': fields.Date(description='Date the report is due.'),
     'report_year': fields.List(fields.Int, description='Year of report.'),
     'report_type': fields.List(fields.Str, description='Type of report.'),
-    'create_date': fields.List(fields.Date, description='Date this record was added to the system.'),
-    'update_date': fields.List(fields.Date, description='Date this record was last updated.'),
+    'min_create_date': fields.Date(description='Date this record was added to the system.'),
```
jmcarp: If these fields are repeated exactly in other groups of args, we can break them into their own dictionary--then both views that use them can decorate their `get` methods with the added dictionary as well.

LindsayYoung:
Should we just exclude deleted elections from the results or do they have some type of value?  @paulclark2 @jwchumley

jwchumley:
I think they should be excluded but that is just based on looking at the data and understanding other systems we've built.

webservices/resources/dates.py
```diff
@@ -57,27 +54,30 @@ def get(self, **kwargs):
...
-        'office_sought',
-        'election_date',
-        'trc_election_type_id',
-        'trc_election_status_id',
-        'update_date',
-        'create_date',
-        'election_year',
-        'pg_date',
-    }
+        return self.model.election_date.filter_by(election_status_id=1)
```
LindsayYoung: this isn't working yet

PR #1387, opened 2015-12-07
Feature/date upgrades

jmcarp:
* Fix n+1 queries issues
* Simplify date models

### DIFF
--- a/webservices/resources/dates.py
+++ b/webservices/resources/dates.py
@@ -27,21 +27,21 @@ class DatesResource(ApiResource):
 @doc(description='FEC reporting dates since 1995.')
 class ReportingDatesView(DatesResource):
 
-    model = models.ReportingDates
+    model = models.ReportDate
 
     @property
     def date_column(self):
         return self.model.due_date
 
     filter_multi_fields = [
-        ('due_date', models.ReportingDates.due_date),
-        ('report_year', models.ReportingDates.report_year),
-        ('report_type', models.ReportingDates.report_type),
-        ('create_date', models.ReportingDates.create_date),
-        ('update_date', models.ReportingDates.update_date),
+        ('due_date', models.ReportDate.due_date),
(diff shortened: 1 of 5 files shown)


webservices/resources/dates.py
```diff
@@ -60,24 +58,25 @@ def get(self, **kwargs):
...
-        'office_sought',
-        'election_date',
-        'trc_election_type_id',
-        'trc_election_status_id',
-        'update_date',
-        'create_date',
-        'election_year',
-        'pg_date',
-    }
+    filter_multi_fields = [
```
LindsayYoung: Would the dates be better as `filter_range_fields`?

jmcarp: Sounds reasonable, although now I'm realizing that I don't know what all the dates and other columns actually refer to here--for example, what's `pg_date`, and why would a user want to filter on it? Do we want to include `trc_election_type_id` and `trc_election_status_id`? At minimum, it seems like we'll need `election_date` and `due_date` above to have range filters, since I would guess we'd need those for user-facing views. Seems reasonable to make all date fields behave the same way.

Anyway, the goal of this patch is to only make sure you're not blocked on this work by n+1 queries, so feel free to add or modify filters as you see fit :smiley:

LindsayYoung: Sounds good.

For reference, `pg_date` would be primary/general date. I am hoping that election type and status ids are going to lead to more data when we look at the new tables. They are probably not useful out of context. 

Thanks again!

jmcarp: In that case, what do you think about giving the filter and model field clearer names? Also, I'm still missing what primary/general date means--is it the date of the primary election or the general election? Also, if `trc_election_type_id` and `trc_election_status_id` aren't going to be meaningful to users, should we lose these filters? Should those columns be included in the serialized output?

PR #1390, opened 2015-12-08
Weight typeahead search by committee receipts.

jmcarp:
Weight committee and candidate typeahead results by total receipts.

* Include total committee receipts in committee typeahead view
* Include total receipts summed over authorized and principals in candidate typeahead view
* Order typeahead results by descending receipts

Note: this patch uses total receipts as a very simple heuristic for search relevance. We can implement more elaborate rules if needed, such as including disbursements or weighting results by time, but I propose that we try this out first.

Pinging @LindsayYoung for review as time permits. @noahmanger: let me know if you want to take this for a test drive or demo to FEC, and we can deploy to dev.

### DIFF
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -131,32 +131,3 @@ create index on ofec_candidate_detail_mv_tmp(incumbent_challenge);
 
 create index on ofec_candidate_detail_mv_tmp using gin (cycles);
 create index on ofec_candidate_detail_mv_tmp using gin (election_years);
-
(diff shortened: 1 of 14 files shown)


LindsayYoung:
This breaks the documentation. It is looking for body instead of q. It doesn't make sense to me since it is importing the params from args.names and that clearly has q in it and the endpoint expects q.

jmcarp:
Updated--just needed to use our modified `use_kwargs` instead of the version from `flask-apispec`, which defaults to pulling parameters from the json body. What do you think about the search behavior?

LindsayYoung:
This is great!

PR #1393, opened 2015-12-09
Ensure itemized queues are unique on corresponding primary key.

### LINKED ISSUE #1392: Some schedule E data on dev doesn't show up on production
Looking into causes still.

for example:
https://api.open.fec.gov/v1/schedules/schedule_e/?committee_id=C00541292&api_key=DEMO_KEY
https://fec-dev-api.18f.gov/v1/schedules/schedule_e/?committee_id=C00541292&api_key=DEMO_KEY

I checked and the data is in the underlying data is in the database.

jmcarp:
Nightly updates on itemized tables fail when the same item has been
added or modified more than once within a single update cycle. To fix,
this patch adds primary key constraints on the queue tables and deletes
any existing records in the queues by primary key on insert.

Note: We'll be able to express this more concisely as an upsert once
postgres 9.5 is available on RDS.

[Resolves #1392]

@arowla: Have time to take a look? This is a very small patch--I just want to verify that the approach makes sense here, and that there isn't a smarter way to solve this problem.

### DIFF
--- a/data/sql_incremental_aggregates/update_schedule_a_fulltext.sql
+++ b/data/sql_incremental_aggregates/update_schedule_a_fulltext.sql
@@ -4,15 +4,18 @@ begin
     where sched_a_sk = any(select sched_a_sk from ofec_sched_a_queue_old)
     ;
     insert into ofec_sched_a(
-        select
-            *,
-            to_tsvector(contbr_nm) as contributor_name_text,
-            to_tsvector(contbr_employer) as contributor_employer_text,
-            to_tsvector(contbr_occupation) as contributor_occupation_text,
-            is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text)
+        select distinct on (sched_a_sk)
+            new.*,
+            to_tsvector(new.contbr_nm) as contributor_name_text,
+            to_tsvector(new.contbr_employer) as contributor_employer_text,
+            to_tsvector(new.contbr_occupation) as contributor_occupation_text,
+            is_individual(new.contb_receipt_amt, new.receipt_tp, new.line_num, new.memo_cd, new.memo_text)
                 as is_individual,
-            clean_repeated(contbr_id, cmte_id) as clean_contbr_id
-        from ofec_sched_a_queue_new
+            clean_repeated(new.contbr_id, new.cmte_id) as clean_contbr_id
+        from ofec_sched_a_queue_new new
+        left join ofec_sched_a_queue_old old on new.sched_a_sk = old.sched_a_sk and old.timestamp > new.timestamp
+        where old.sched_a_sk is null
+        order by new.sched_a_sk, new.timestamp desc
     );
 end
 $$ language plpgsql;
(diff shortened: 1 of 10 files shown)


arowla:
So essentially you're only honoring the latest update to the record?

This looks like it will work, though I wonder if it would be more performant to aggregate the new records before inserting, and not doing a delete for each key, each time. Or... what about an UPSERT?

jmcarp:
Took another crack at this, de-duplicating queued records during the nightly updates. The revised patch also solves the problem of records that are added, then deleted, within the same update cycle: queued records are only added if there isn't a queued delete that happened after the queued update.

This should be ready for another review @arowla

LindsayYoung:
This looks good for getting us back to syncing correctly going forward. Am I right in thinking we will need to do some additional work to add the records that we missed because they had repeated keys, or will this deal with those records too?

data/sql_incremental_aggregates/update_schedule_a_fulltext.sql
```diff
@@ -4,15 +4,18 @@ begin
...
+            to_tsvector(new.contbr_employer) as contributor_employer_text,
+            to_tsvector(new.contbr_occupation) as contributor_occupation_text,
+            is_individual(new.contb_receipt_amt, new.receipt_tp, new.line_num, new.memo_cd, new.memo_text)
                 as is_individual,
-            clean_repeated(contbr_id, cmte_id) as clean_contbr_id
-        from ofec_sched_a_queue_new
+            clean_repeated(new.contbr_id, new.cmte_id) as clean_contbr_id
+        from ofec_sched_a_queue_new new
+        left join ofec_sched_a_queue_old old on new.sched_a_sk = old.sched_a_sk and old.timestamp > new.timestamp
+        where old.sched_a_sk is null
```
arowla: I'm missing something about the logic here. Can you explain the `left join` and the `where`?

jmcarp: Yeah! The goal here is to add all records in the added queue that aren't also in the deleted queue with a later timestamp. For example, if somebody at FEC adds a record and then deletes it a minute later, it's going to be in both queues, but we don't want to add it to our ofec_* table. We're using a left join because we want to include all records in the added queue, even if there's no corresponding record in the deleted queue. And we're joining on primary key and relative timestamp to get deleted records with the same key that were deleted *after* the record was added to the added queue. Then we're filtering on `old.sched_a_sk is null` so that we only get records in the added queue that weren't subsequently deleted.

Hope that makes sense. If you can think of a simpler way to implement the behavior, I'm happy to revise. Otherwise, I can add comments here to explain the logic to our future selves.

arowla: Sounds good! And yes, comments would be a plus. :+1:

jmcarp:
Docs added. Think this is good to merge @arowla?

arowla:
:+1: 

Small stylistic note, just since I'm not sure if you were aware of this syntax, but in Postgres, instead of `cast(null as timestamp) as timestamp`, you could write `null::timestamp`. The double colon performs the cast, and the column name will be inferred from the type. It's less explicit, I suppose, but it's far fewer characters. No need to make the change, just wanted to mention it.

jmcarp:
Funny, I've been using that syntax in another pending pull request on this repo--I must've regressed on this one.

PR #1394, opened 2015-12-10
[WIP] First pass at omnibus dates view.

jmcarp:
* Include aggregated general elections
* Include individual non-general elections
* Include aggregated general reporting deadlines
* Include individual non-general reporting deadlines

cc @LindsayYoung


jmcarp:
Included in #1412.

PR #1397, opened 2015-12-13
Create temporary itemized tables to avoid downtime.

jmcarp:
Create temporary itemized tables, then drop existing tables and rename
when ready. Avoids downtime on itemized resources during schema updates.

### DIFF
--- a/data/sql_setup/prepare_schedule_a.sql
+++ b/data/sql_setup/prepare_schedule_a.sql
@@ -1,6 +1,6 @@
 -- Create Schedule A table
-drop table if exists ofec_sched_a;
-create table ofec_sched_a as
+drop table if exists ofec_sched_a_tmp;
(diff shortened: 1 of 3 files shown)


evankroske:
I think that you could further reduce the downtime and risk of schema
updates by applying updates to a separate database and then configuring the
application to talk to the up-to-date database. This technique is called
blue-green deployment:

http://martinfowler.com/bliki/BlueGreenDeployment.html

However, this approach might increase your infrastructure costs too much.
Maybe you could use the staging database for schema updates.

LindsayYoung:
LGTM

Also, thanks for chiming in @evankroske. Like you mentioned, adding blue/green deploys for this would increase costs. We try to save as much taxpayer money as we can. We will continue to improve our processes so we will keep thees suggestions in mind as the project matures.

PR #1386, opened 2015-12-06
Feature/update candidate data

jmcarp:
* Use `cand_valid_fec_yr` instead of `dimcandproperties` for candidate date
* Rebuild test subset with updated constraints between tables

Part of #1337.

Ping @LindsayYoung for review

PR #1399, opened 2015-12-15
Ensure unique candidate election years.

jmcarp:
Fix a regression from #1386 that allowed duplicate election years in
candidate views. Thanks @noahmanger for spotting.

IMO this fix has to go in before the next release, so I'm hoping @LindsayYoung can take a look in the morning. I'm going to deploy this change to the dev database for testing now.


LindsayYoung:
I spot checked some examples and they looked good.

PR #1411, opened 2015-12-18
Add full election flag to committee history view.

jmcarp:
Add an `election_full` flag to the committee history view. When passed,
find all committees associated with the given candidate for the given
election.

[Resolves https://github.com/18F/openFEC-web-app/issues/998]


noahmanger:
So I'm reading https://github.com/18F/openFEC-web-app/issues/998 as asking for both committees to be visible on the "full cycle" option. I see the correct committees showing for each sub-period, but only the most recent one is visible on full cycle.

noahmanger:
Ah I wasn't running the right app branch. This is working well. I have one UI note though but will add it to that repo.

LindsayYoung:
This looks good to me

jmcarp:
Conflicts resolved @LindsayYoung

PR #1419, opened 2015-12-21
Condense reporting deadline categories.

jmcarp:
h/t @noahmanger @jenniferthibault

### DIFF
--- a/data/sql_updates/create_omnibus_dates.sql
+++ b/data/sql_updates/create_omnibus_dates.sql
@@ -21,9 +21,6 @@ returns text as $$
     end
 $$ language plpgsql;
 
-
-
-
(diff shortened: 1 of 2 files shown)


data/sql_updates/create_omnibus_dates.sql
```diff
@@ -21,8 +21,17 @@ returns text as $$
     end
 $$ language plpgsql;
 
-
-
+create or replace function get_report_category(report_type text)
+returns text as $$
+    begin
+        return case
+            when report_type in ('Q1', 'Q2', 'Q3') then 'Q'
```
LindsayYoung: If we want to group things in the API, wouldn't we want to add the year end for quarterly filings and monthly filings too?  Or would the webapp ask for monthly and year end deadlines? That seems reasonable too.

PR #1421, opened 2015-12-22
Add icalendar exports.

jmcarp:
Export calendar dates in ical format for use with Google Calendar, iCalendar, etc.

Note: All exports are automatically filtered to +/- 1 year around the time of the request. Pagination isn't an option here--calendar applications generally expect to get *all* events in a single request. Because there aren't many rows in the calendar table, performance is adequate without pagination--about 0.6s to get all events, with no filters.

@LindsayYoung @noahmanger

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,16 +5,17 @@ Flask-Cors==2.1.0
 Flask-Script==2.0.5
 Flask-RESTful==0.3.3
 Flask-SQLAlchemy==2.0
+python-dateutil==2.4.2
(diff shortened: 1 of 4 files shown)


LindsayYoung:
This is cool, is this ready for review?

jmcarp:
Added csv exports. I haven't written tests for any of this, so that needs to happen before this gets merged into develop, but it's ready to review for merging into feature/dates-2.

LindsayYoung:
lgtm

PR #1423, opened 2015-12-23
individual type code fix [Don't merge]

LindsayYoung:
The code requested was 11A, I don't see an 11AI as a code we want here. 

This is working on https://github.com/18F/openFEC/issues/1308 I don't think this is a complete fix.

### DIFF
--- a/data/functions/individual.sql
+++ b/data/functions/individual.sql
@@ -19,7 +19,7 @@ create or replace function is_inferred_individual(amount numeric, line_number te
 begin
     return (
         amount < 200 and
-        coalesce(line_number, '') in ('11AI', '12', '17', '17A', '18') and
+        coalesce(line_number, '') in ('11A', '12', '17', '17A', '18') and
         not is_earmark(memo_code, memo_text)
     );
 end


jmcarp:
I don't see any itemized receipts with a line number of "11A". I do see a bunch of records marked "11AI" and "11A1", the latter of which is probably a data entry error. It looks like this change would treat *all* 11AI records, of which we're exposing about 42m, as earmarked, and it looks for a line number that doesn't seem to exist.

Counts of our receipts data by line number:
```
 line_num |  count
----------+----------
 11A1     |   184517
 11AI     | 41697261
 11B      |     9111
 11C      |   732086
 11D      |    18536
 12       |   296539
 13       |      629
 13A      |     7865
 13B      |      423
 14       |    10189
 14A      |     2256
 15       |   113363
 16       |    95394
 16A      |        7
 17       |    69241
 17A      |  4547930
 17B      |       41
 17C      |     1235
 17D      |     2256
 18       |  1180325
 18L      |        5
 19A      |      359
```

LindsayYoung:
Glad you checked!

I was going off of what Jeff literally said here https://github.com/18F/openFEC/issues/1141#issuecomment-133054126
```
Individual Contributions:
Rule #1: If it has a receipt_tp in (15,15E, 15J, 18J) then include it
Rule  #2 If it is on line 11a or 12 (F3 or F3x) or line 17 or 17A or 18
(F3P) and the amount is less than $200  and it is not coded include it
        UNLESS  it is a memo entry and includes  "EARMARK','EARMK'
or'ERMK'  in the memo_text field, don't include it.
```

11AI makes sense because that is where the itemized contributions are reported on the [form 3](http://www.fec.gov/pdf/forms/fecfrm3.pdf). I can see that 11A1 have been renamed to 11AI's [before](http://www.fec.gov/elecfil/updatelist.shtml)

PR #1429, opened 2015-12-28
Partition large itemized tables.

jmcarp:
* [x] Partition large tables
* [x] Batch update partitions
* [x] Remove old update scripts


jmcarp:
@arowla have time to review this next week? We may want to change the partition size from two years to one pending benchmarking, but I think that shouldn't get in the way of code review.

jmcarp:
Things to do:

* Reviewers--get familiar with postgres partitioning: http://www.depesz.com/2015/06/07/partitioning-what-why-how/
* Can we use [pg_partman](http://pgxn.org/dist/pg_partman/doc/pg_partman.html) to replace some or all of this code? This might be a good option--the only thing that comes to mind as a potential sticking point would be updating child tables from the itemized queues.
* Is query and download performance on the larger tables adequate now that we've made other performance improvements? I would still guess that partitioning is worthwhile to cut down on potentially slow queries, but there may be other approaches we can take on that front.
* Is the current incremental update strategy viable? It's been slow lately. I'll investigate.

ccostino:
@jmcarp:  Thank you for that blog post, very helpful!

In addition, here's the link to the official docs as well:  http://www.postgresql.org/docs/9.5/static/ddl-partitioning.html

ccostino:
As you read the blog post and get toward the end, there's some code around creating the actual trigger for inserts into a partitioned table (which in turn is describing what happens when using pg_partman), but the code isn't visible due to some styling issue.  When you reach that part, this is what should be displayed:

```sql
create trigger partition_users before insert on users for each row execute procedure partition_for_users();
```

ccostino:
@jmcarp:  Given that `pg_partman` is not an option since it's not available in Amazon RDS this brings us back to managing the partitions ourselves.  With the recent issues that we and the folks at FEC have had with connectivity and performance in the project our hand has been forced in increasing vCPU and RAM availability to address them.  This has had the pleasant side affect of making downloads for the itemized tables somewhat performant as things exist now (e.g., in about a minute or two!), but I think for our long term needs partitioning these tables is still valuable, especially if we need to consider other tables in the future as more data is loaded in.

This brought me back to your PR and after reviewing it in close detail last night I realize that you have it handled quite nicely, and quite completely for that matter.  I pulled it down locally, did a thorough deep dive of how you translated the existing SQL into being supported in the code and how the child tables are built and managed along with the queues, and checked to see how major the conflicts with the current state of the `develop` branch.  It turns out they're very minor:

* Minor conflicts git couldn't resolve with some of the SQL script changes due to the schema changes.
* Adding support for the `pdf_url` column in the partition code itself.
* Accounting for the merge of the workflow PR @LindsayYoung merged earlier today (git couldn't resolve the placement of two `import` statements in `manage.py`).

Once I had everything working the tests pass just fine and everything works as expected.  I'd like to test this in `dev` with the full data set and plan to do so tonight to see what the effect will be on the full data set, especially since you have this setup in a way that does not affect the existing tables (in fact, I can see that you had tested this previously in `dev`).

Would you be amenable to resolving the conflicts in your branch to make it review and merge ready?  I can also resolve the conflicts and make a new PR to carry the torch if need be.  I'd be happy to do a Screen Hero session with you to review, and after having run through this my questions primarily revolve just around your approach and learning more about how your thinking led you there (mainly why the use of class methods).

/cc @LindsayYoung and @noahmanger

ccostino:
Two small additional caveats in that I also had to enable the downloads in the web app (just two minor flag changes in the JS), and enable them in the API (adding the `ScheduleAView` and `ScheduleBView` to the whitelist).

LindsayYoung:
Some follow up tasks to be released with this:
 - [ ] Add all the data. (get rid of the filtering on itemized start date on the schedule tables)
 - [ ] Make 2-year cycle select mandatory
 - [ ] Use unique persistent identifiers instead of idx.
 - [ ] Make identifiers filterable.

jmcarp:
@ccostino: sure, I can resolve the merge conflicts. I'll try to get to this over the next few days.

jmcarp:
Conflicts resolved. Just to be clear, I wouldn't recommend merging this as-is. Like we talked about elsewhere, I recommend taking some time to identify performance goals, make sure FEC shares those goals, and then benchmark the relevant features with and without this kind of partitioning.

ccostino:
Great, thank you @jmcarp!  Yes, we will continue to benchmark some of these metrics and discuss the implications with the folks at FEC.  There's definitely a hardware aspect to this as well that we need to get a clearer picture of with FEC, too (mostly in terms of CPU and RAM we need available to us).

ccostino:
@jmcarp:  We've been reviewing, benchmarking, and testing the partitioned tables extensively over the last week or so and have come to the conclusion that they will suite both our needs and FECs needs based on our findings.  We spoke with the FEC folks and explained our desire and strategy to do this, and we discussed the pros and cons of moving forward with what we had.

In short, there is a desire to be able to continue to search/filter/download based on multiple election cycles (e.g., 2, 4, and 6 years), not just for two year cycles.  However, this wouldn't be performant with the partitioning as we wouldn't be taking advantage of the benefits (more on that in a moment).  Despite this, everyone is in agreement that this is an acceptable trade-off and that partitioning is the appropriate solution to move forward.  It addresses the table size and performance concerns we have and would enable us to load all of the remaining data for the Schedule A and B items.  These benefits would allow us to have downloads working for these items and provide much more value to users a lot sooner, and the FEC folks are on the same page as us in wanting to deliver this highly desired feature.

Going back to performance and the benefits of partitioning:  with the recent hardware resource increases we made we were able to activate downloads for Schedule A and B items even without partitioning in the dev environment, which gave us some great metrics in comparing before and after scenarios.  Even without partitioning downloads are for the most part performant with the additional CPUs and RAM, taking place in roughly under a minute save for one query I came across (and that topped out at 226 seconds).  With partitioning, there are a couple of interesting things to note:

* Performance is comparable, but if you end up in a situation where you are pulling data from multiple child tables it can decrease, and in a couple of cases I found, quite drastically.
* When you query for data within a single election cycle (so only querying a single child table), performance gains range from comparable to noticeably faster, so a net positive gain there.
* Also, just searching and filtering the data on the web site itself remains performant.

@noahmanger, I know you were able to test the non-partitioned downloads and came across a couple of long-running queries as well, were you able to follow up with the partitioned version tests?  Just now I tested one out myself that you reported as running for over 3 minutes; with partitioning, it finished in just under a minute and a half (roughly 107 seconds) instead.

Based on our testing testing and discussions, and verifying our path forward with FEC, we're confident that we want to move forward with this now.  @LindsayYoung has already noted the work we'll be doing as next steps, the main piece of which will be making the two-year election cycle filtering mandatory as that is the major caveat in going with this approach.  However, the stability, performance gains, and scalability of this solution far outweigh leaving things as they are now.  This is especially true when we factor in all of the remaining data to be loaded and all of the data that will be added in the future.  As an aside, the database size is projected to increase by about 40% in just the next couple of months alone, just to give a sense of the exponential increase.

All of this said, I wanted to see if you had any other follow up questions or concerns that you would like to address first before I merge your PR.  As always, thank you immensely for your time, input, and help, @jmcarp!  We greatly appreciate it.

ccostino:
This is ready to be merged in now, which paves the way for the API updates and web app changes required to support it.  The web app changes are already in progress in a PR (https://github.com/18F/openFEC-web-app/pull/1242) and I'll have another PR for the API changes soon.

Huge h/t and thanks to @jmcarp for all of the work and support on this!

PR #1432, opened 2015-12-31
Redirect /docs to documentaion

LindsayYoung:
There were 7,000 hits looking for `/docs` in the last month, right now that just gives a 404. This forwards them to documentation instead. 

(this was a super small thing I noticed after usability testing)


jmcarp:
Merged into release branch.

PR #1433, opened 2015-12-31
After user feedback, it seems that people are confused by what cycle means

LindsayYoung:
The person liked the answer I gave on the phone, so I beefed
up our docs. Happy to improve the descriptions of @emileighoutlaw or
@jmcarp have suggestions.

It seemed that it was a super minor detail where it would be easier to propose a solution and discuss it as needed.


jmcarp:
Merged into release.

PR #1439, opened 2016-01-06
Updates language

emileighoutlaw:
Includes language edits and a few other changes per FEC style:

- from U.S. to US
- removal of third comma in list
- lowercase of presidential
- capitalization Schedule E
- noticed some inconsistent use of periods in descriptions. updated so periods are only in descriptions if those descriptions are complete sentences. we could change this to add periods to all of them later, if that's what we'd like to do.

cc @LindsayYoung @noahmanger

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -149,7 +149,7 @@ candidate_detail = {
     'district': fields.List(District),
     'candidate_status': fields.List(
         IStr(validate=validate.OneOf(['', 'C', 'F', 'N', 'P'])),
-        description='One letter code explaining if the candidate is:\n\
+        description='One-letter code explaining if the candidate is:\n\
         - C present candidate\n\
         - F future candidate\n\
         - N not yet a candidate\n\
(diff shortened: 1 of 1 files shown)


webservices/args.py
```diff
@@ -380,13 +380,13 @@ def make_seek_args(field=fields.Int, description=None):
...
     'committee_id': fields.List(IStr, description=docs.COMMITTEE_ID),
-    'recipient_committee_id': fields.List(IStr, description='The FEC identifier should be represented here the contributor is registered with the FEC'),
+    'recipient_committee_id': fields.List(IStr, description='The FEC identifier should be represented here if the contributor is registered with the FEC'),
     'recipient_name': fields.Str(description='Name of recipient'),
     'disbursement_description': fields.Str(description='Description of disbursement'),
     'recipient_city': fields.List(IStr, description='City of recipient'),
     'recipient_state': fields.List(IStr, description='State of recipient'),
-    'last_disbursement_date': fields.Date(missing=None, description='Filter for records before this date'),
-    'last_disbursement_amount': fields.Float(missing=None, description='Filter for records'),
+    'last_disbursement_date': fields.Date(missing=None, description='Filters for records before this date'),
```
LindsayYoung: Looking at these, it looks like the oned that start with "last" had the wrong explanation, this is used for pagination when you sort by something

jmcarp:
Merged into release.

PR #1412, opened 2015-12-18
Feature/dates 2 [WIP]

LindsayYoung:
This builds off of the omnibus dates work. Has filters and basic descriptions.

Needs better descriptions and tests.

PR #1444, opened 2016-01-09
[WIP] Feature/electioneering new table

LindsayYoung:
Need to follow up on some missing fields and fix per-candidate totals.


jmcarp:
What's next here @LindsayYoung?

LindsayYoung:
I haven't checked to see if we have the additional columns, if we do, we can add them.

jmcarp:
If we're just waiting on columns, how about reviewing and merging as-is, and updating when we have more data? Especially given that we're currently showing incorrect aggregates on other spending in production.

LindsayYoung:
Yea, it looks like Paul is still working on that. I haven't gotten a chance to look at the communication cost stuff yet, and I want to make sure we are using the candidate share column for aggregates.

LindsayYoung:
Ok, updated electioneering totals and communication cost looks pretty straight forward. 

We can merge this and update later with the added columns, if you want.

LindsayYoung:
Ok, updated electioneering totals and communication cost looks pretty straight forward. 

We can merge this and update later with the added columns, if you want, @jmcarp.

data/sql_updates/create_electioneering_aggregate_candidate_view.sql
```diff
@@ -7,17 +7,15 @@ with disbursements as (
...
-    disb.rpt_yr + disb.rpt_yr % 2 as cycle
-from form_94 f9
-join disbursements disb using (link_id)
-where disb.rpt_yr >= :START_YEAR
+    cmte_id,
+    cand_id,
+    sum(calculated_cand_share) as total,
+    count(calculated_cand_share) as count,
+    rpt_yr + rpt_yr % 2 as cycle
+from electioneering_com_vw
```
jmcarp: Now that we're not joining on `sched_b`, we shouldn't need the `with disbursements as ...` above.

jmcarp: Looks like we're dropping the `where rpt_yr >= :START_YEAR`. Should that be included?

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+    # load_date = db.Column(db.DateTime)
+    # update_date = db.Column(db.DateTime)
+    number_of_candidates = db.Column(db.String)
+    calculated_candidate_share = db.Column('calculated_cand_share', db.String)
+    # difference between communication and public distribution dates?
+    communicaion_date = db.Column('comm_dt', db.DateTime)
+    public_distribion_date = db.Column('pub_distrib_dt', db.DateTime)
+    disbursement_date = db.Column('disb_dt', db.DateTime)
+    disbursement_amount = db.Column('reported_disb_amt', db.Numeric(30, 2), index=True)
+    #TODO: add tsvector field
```
jmcarp: Are we waiting on FEC for this? Can we add it now?

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+    # is this similar to transaction_id?
+    sb_link_id = db.Column(db.String, index=True)
+    # transaction_id = db.Column(db.Integer)
+    # filing_type = db.Column(db.String, index=True)
+    # load_date = db.Column(db.DateTime)
+    # update_date = db.Column(db.DateTime)
+    number_of_candidates = db.Column(db.String)
+    calculated_candidate_share = db.Column('calculated_cand_share', db.String)
+    # difference between communication and public distribution dates?
+    communicaion_date = db.Column('comm_dt', db.DateTime)
```
jmcarp: Should be `communication_date`.

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+    public_distribion_date = db.Column('pub_distrib_dt', db.DateTime)
```
jmcarp: Should be `public_distribution_date`.

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+    # file_number = db.Column('file_num', db.Integer)
+    sub_id = db.Column(db.Integer)
+    link_id = db.Column(db.Integer)
+    # is this similar to transaction_id?
+    sb_link_id = db.Column(db.String, index=True)
+    # transaction_id = db.Column(db.Integer)
+    # filing_type = db.Column(db.String, index=True)
+    # load_date = db.Column(db.DateTime)
+    # update_date = db.Column(db.DateTime)
+    number_of_candidates = db.Column(db.String)
```
jmcarp: The original type in `electioneering_com_vw` is `numeric`; we should probably use the same here.

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+    calculated_candidate_share = db.Column('calculated_cand_share', db.String)
```
jmcarp: See above--this column shouldn't be a string.

LindsayYoung: good catch, that was silly of me

webservices/common/models/costs.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class Electioneering(db.Model):
+    __tablename__ = 'ofec_electioneering_mv'
+
+    idx = db.Column(db.Integer, primary_key=True)
+    form_type = db.Column('form_tp', db.String, index=True)
+    committee_id = db.Column('cmte_id', db.String, index=True)
+    candidate_id = db.Column('cand_id', db.String, index=True)
+    candidate_name = db.Column('cand_name', db.String, index=True)
```
jmcarp: This column doesn't seem to be indexed in the materialized view, so we shouldn't mark it as indexed here. I don't think this would be especially useful to index, anyway, unless we want to expose a search on exact match on candidate names.

webservices/common/models/costs.py
LindsayYoung: There isn't a blocker on this. I just wanted to add the basic version to start, but I did not want to forget to add the text search on this.

webservices/common/models/costs.py
LindsayYoung: :+1:

webservices/common/models/costs.py
jmcarp: Cool--if nothing's blocking this, can we do it now?

jmcarp:
Same as my comment on #1412, I'm wondering if we can extract miscellaneous commented-out lines, questions, and TODOs in the code into issues that we can assign and track. Otherwise, we're liable to lose track of tasks that should be done, and forget why certain code is commented out.

LindsayYoung:
A lot of the comments were things we are adding back. I am going to see if I can see the extra fields now.

data/sql_updates/create_electioneering_aggregate_candidate_view.sql
LindsayYoung: yes to both of these

webservices/common/models/costs.py
LindsayYoung: I'll add that now

PR #1446, opened 2016-01-11
Abort deploy on invalid configuration.

jmcarp:
Abort deploy if:
* Environment variables or credentials missing
* Cloud Foundry services missing
* Required tables missing

Part of https://github.com/18F/openFEC-web-app/issues/1020.

cc @LindsayYoung


LindsayYoung:
I got it to throw an error if there was a missing table, it does not build and throws the configuration error. Woo!

Issue #1416, opened 2015-12-18
Add data dictionary to download manifest


noahmanger:
So that users understand the data they're getting in their downloads, include an automatically generated data dictionary in the manifest file.

noahmanger:
Tasks for this sprint:
- [ ] FEC build out the dictionary in a doc
- [ ] 18F add to the API
- [ ] Add to the download manifest

LindsayYoung:
I don't think we should do a document, I would like to get Paul or Jeff to propose edits directly.

noahmanger:
:+1:

noahmanger:
Status: data is here but we need to add it to the manifest.

noahmanger:
@LindsayYoung this issue doesn't strictly fall into the priorities we identified. Should we remove it from "in progress"?

vrajmohan:
Given the move away from a zipped file to a CSV file, this issue is no longer relevant.

PR #1449, opened 2016-01-19
Feature/document fields

LindsayYoung:
Adds documentation to most of the fields. 

I am trying to make a good base that puts Paul and Jeff in a better place to improve it with more specifics. 

It mostly uses existing work for documenting arguments and I added a few straight forward things. 

There is also a function that makes some generic definitions from the variable name. Not sure how helpful this is, but it makes the documentation look a lot more filled out. Happy to get feedback on that approach. 

I am getting a swagger integration error from py.test, but it all looks like it is all working to me.


LindsayYoung:
First step of #1416 
Need to add the definitions to the manifest too.

jmcarp:
The failing swagger tests means that we're generating invalid swagger. Here's the actual error message:

screen shot 2016-01-25 at 10 12 24 am

webservices/docs.py
```diff
@@ -469,3 +470,187 @@
...
+
+def add_period(var):
+    return var + ' total for the reporting period'
+
+
+def add_ytd(var):
+    return var + ' total for the year to date'
+
+
+def make_generic_def(var):
```
jmcarp: I'm not sure I see the point of this helper. It looks like we're more or less copying the column name into the documentation field, modulo some spacing cleanup. But we aren't actually adding any information to the documentation--we're showing the same information twice. Imagine this in the download manifest:

aggregate_amount_personal_contributions_general: Aggregate amount personal contributions general

LindsayYoung: My thinking was that it was better than nothing, but I can be persuaded to remove it.

LindsayYoung:
I need to fix the swagger error and resolve conflicts, but in the meantime, I wanted to flag @emileighoutlaw on the additional language for the documentation

emileighoutlaw:
Thanks thanks! Taking a look now.

emileighoutlaw:
This is rad, @LindsayYoung! I just suggested a handful of tiny copyedits.

webservices/docs.py
```diff
@@ -469,6 +476,14 @@
 - The communication is distributed within 60 days prior to a general election or 30 days prior to a primary election to federal office.
 '''
 
+# merge question????????????????
```
LindsayYoung: @emileighoutlaw in merging the docs, I noticed we have 2 versions of electioneering, I like the first one because it is more specific, with days and says federal candidate instead of House and Senate, (because I see presidential electioneering so only house or senate doesn't make sense to me). Wondered if this came up for you before.

Data source:
http://www.fec.gov/finance/disclosure/electioneering.shtml

emileighoutlaw: Oh! I hadn't seen that before, but I think you're definitely right!

webservices/docs.py
```diff
@@ -481,8 +482,14 @@
     - U Unregistered Committee Notice
 '''
 DATES_TAG = '''
-Regular reporting deadlines and election dates.
+Reporting deadlines, election dates FEC meetings, events etc.
 '''
+
+CALANDAR_DATES = '''
```
jmcarp: CALENDAR

webservices/tasks/download.py
```diff
@@ -168,7 +168,7 @@ def make_manifest(resource, row_count, path):
         fp.write('Time: {}\n'.format(resource['timestamp']))
         fp.write('Resource: {}\n'.format(resource['path']))
         fp.write('*Count: {}\n'.format(row_count))
-        fp.write('Filters:\n\n')
+        fp.write('Fields:\n\n')
```
jmcarp: Why rename this? Did we add fields?

LindsayYoung: I was in the process of adding them, I wanted that to be a separate pr, I thought I rolled that back.

webservices/docs.py
LindsayYoung: Thanks! I have a serious issue with spelling that right, and It apparently doesn't matter how many times I have had to type that in the last few months.

PR #1450, opened 2016-01-19
Use election full flag in candidate history view.

jmcarp:
The `CandidateHistoryView` resource is meant to take an `election_full`
flag, but previously hadn't included the flag in its `args`. This patch
adds the missing request parser and a regression test.

[Resolves #1031]


jmcarp:
cc @LindsayYoung

LindsayYoung:
:+1:

PR #1457, opened 2016-01-25
Feature/indivual

### LINKED ISSUE #1435: Update individual function
After talking to Paul there are 2 updates to the individual function we will want to make:

 - Add new individual party codes as individual contributions: 30, 30T, 31, 31T, 32
 -  [here](https://github.com/18F/openFEC/blob/develop/data/functions/individual.sql#L22) on line 22, it should be counted as an earmark and excluded from our individual filtering if the line has a memo code of X _OR_ it has  “earmark”, “earmk”, or “ermk” in the memo_text.

this comes from:
https://docs.google.com/document/d/1dZpMv8K6hikRCHnXUSiEt0hQ4twRQNBeh3bZD786yJs/edit#

LindsayYoung:
Adds the new political party individual codes, looks for memo_code or earmark. Adds documentation about our definitions of individual. 

closes #1435

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -5,7 +5,6 @@ Unique identifier for the electronic or paper report. This number is used to con
 PDF URLs to the original document.
 '''
 
-
 CANDIDATE_ID = '''
 A unique identifier assigned to each candidate registered with the FEC.
 If a person runs for several offices, that person will have separate candidate IDs for each office.
@@ -268,11 +267,10 @@ or `/totals` endpoints.
 
 When comparing the totals from reports to line items. the totals will not match unless you
 only look at items where `"is_individual":true` since the same transaction is in the data
-multipliable ways to explain the way it may move though different committees as an earmark.
+multiple ways to explain the way it may move though different committees as an earmark.
 
-For the Schedule A aggregates, such as by_occupation and by_state, include only individual
-contributions, i.e. receipts from line numbers 11AI and 17A. Again, "memoed" items are not
-included to avoid double counting.
+For the Schedule A aggregates, such as by_occupation and by_state, include only unique individual
+contributions. See below for full methodology.
 
 ### Methodology for determining individual contributions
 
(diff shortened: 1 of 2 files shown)


noahmanger:
Follow up: discussed over video with @jwchumley and @PaulClark2 and they agreed we should do it this way. @jmcarp feel free to merge if you're happy with it.

webservices/docs.py
```diff
@@ -267,13 +267,42 @@
 or `/totals` endpoints.
 
 When comparing the totals from reports to line items. the totals will not match unless you
```
jmcarp: "." -> ","

webservices/docs.py
```diff
@@ -267,13 +267,42 @@
...
-take out items where `"memoed_subtotal":true`. Memoed items are subtotals of receipts
-that are already accounted for in another schedule a line item.
+only look at items where `"is_individual":true` since the same transaction is in the data
+multipliable ways to explain the way it may move though different committees as an earmark.
```
jmcarp: "multipliable" -> "multiple". Also, this sentence could be confusing--some wordsmithing and/or an example might help. Thoughts @emileighoutlaw?

webservices/docs.py
```diff
@@ -267,13 +267,42 @@
...
+|---|---
+|10 |Contribution to Independent Expenditure-Only Committees (Super PACs), Political Committees with non-contribution accounts (Hybrid PACs) and nonfederal party "soft money" accounts (1991-2002) from a person (individual, partnership, limited liability company, corporation, labor organization, or any other organization or group of persons)
+|15 | Contribution to political committees (other than Super PACs and Hybrid PACs) from an individual, partnership or limited liability company
+|15E | Earmarked contributions to political committees (other than Super PACs and Hybrid PACs) from an individual, partnership or limited liability company
+|15J | Memo - Recipient committee's percentage of contribution from an individual, partnership or limited liability company given to joint fundraising committee
+|18J | Memo - Recipient committee's percentage of contribution from a registered committee given to joint fundraising committee
+|30, 30T, 31, 31T, 32 | Individual party codes
+
+For receipts under $200:
+We check the following codes and see if there is "earmark" (or a variation) in the
```
jmcarp: As of this patch, we don't just check the description (which we might want to call "memo text" to be consistent with the API output)--we also check the memo code. I think this would be helpful to clarify here.

webservices/docs.py
```diff
@@ -267,13 +267,42 @@
...
+description of the contribution.
+|Line number | description
+|---|---
+|11Ai | The itemized individual contributions from F3 schedule A
+|12 | Nonfederal other receipt - Levin Account (Line 2)
+|17 | Itemized individual contributions from Form 3P
+|17A | Itemized individual contributions from Form 3P
+|18 | Itemized individual contributions from Form 3P
+
+Of those transactions,[under $200 OR having the codes 11A, 12, 17, 17A, or 18], we then want to exclude earmarks.
```
jmcarp: The way the `is_inferred_individual` check is implemented, a receipt is individual if it's < $200, has a line number in [11AI, 12, 17, ...], and is neither memoed nor includes "earmark" in the memo text. Since we're using `and` in the sql, is the `or` here inconsistent?

webservices/docs.py
```diff
@@ -267,13 +267,42 @@
...
+|10 |Contribution to Independent Expenditure-Only Committees (Super PACs), Political Committees with non-contribution accounts (Hybrid PACs) and nonfederal party "soft money" accounts (1991-2002) from a person (individual, partnership, limited liability company, corporation, labor organization, or any other organization or group of persons)
+|15 | Contribution to political committees (other than Super PACs and Hybrid PACs) from an individual, partnership or limited liability company
+|15E | Earmarked contributions to political committees (other than Super PACs and Hybrid PACs) from an individual, partnership or limited liability company
+|15J | Memo - Recipient committee's percentage of contribution from an individual, partnership or limited liability company given to joint fundraising committee
+|18J | Memo - Recipient committee's percentage of contribution from a registered committee given to joint fundraising committee
+|30, 30T, 31, 31T, 32 | Individual party codes
+
+For receipts under $200:
+We check the following codes and see if there is "earmark" (or a variation) in the
+description of the contribution.
```
jmcarp: This doesn't seem to be displaying as a table:

screen shot 2016-02-03 at 3 55 37 pm

PR #1458, opened 2016-01-25
Looking to update data 4 am eastern, 9 GMT to get fresher data

### LINKED ISSUE #1448: Move back 18F updates to reduce lag.
It looks like we are running them at midnight GMT (7 pm eastern). @paulclark2 is going to confirm with Rohan and the DBAs that we can run these updates later (maybe 3  am eastern), so that we are up to date with the website. 

Code for scheduling task is [here].(https://github.com/18F/openFEC/blob/209160bacb7cedc345c0c1e1a1a24871d4e8845e/webservices/tasks/__init__.py#L32)

Paul, can you post the time you think the updates will be done by on this issue? Thanks!

LindsayYoung:
Starting our refresh at 4 am will get us all the main tables except for calendar data. That isn't done until 7am and I think if we waited that long to do our update, that would impact performance. 

@jmcarp do you agree that starting our data refresh at 4 will give us enough time so we don't run into morning performance issues?

Closes #1448

### DIFF
--- a/webservices/tasks/__init__.py
+++ b/webservices/tasks/__init__.py
@@ -29,7 +29,7 @@ app.conf.update(
     CELERYBEAT_SCHEDULE={
         'refresh': {
             'task': 'webservices.tasks',
-            'schedule': crontab(minute=0, hour=0),
+            'schedule': crontab(minute=0, hour=9),
         },
     }
 )


jmcarp:
According to #1448, this timing means we should catch everything but the calendar data, which may not finish transferring until 6:00 EST. Is that cool?

LindsayYoung:
Do you want to make the calendar import a separate process? It will eventually need to be updated more frequently anyway.

PR #1459, opened 2016-01-25
Extending about.yml services

DavidEBest:
Updating the about.yml services entry to match the README.


jmcarp:
Is the `services` field being used now?

DavidEBest:
It will be used for the Compliance Toolkit stuff we are working on.

PR #1462, opened 2016-01-26
Allow multiple queries on full-text columns.

jmcarp:
Permit queries like `contributor_name=smith&contributor_name=jones`.
Required for https://github.com/18F/openFEC-web-app/issues/1022.

cc @LindsayYoung


jmcarp:
Note: this patch modifies full-text search on contributor name (receipts) and recipient name (disbursements) to allow for searching on both name and committee ID. If we decide that *all* filter typeaheads should support search on both ID and text, we'll need to make this change to a few more columns.

cc @noahmanger @LindsayYoung @jenniferthibault

jmcarp:
This is part of the consolidated typeahead feature. Assigning to @ccostino for review.

webservices/resources/candidates.py
```diff
@@ -32,6 +32,7 @@ class CandidateList(ApiResource):
     schema = schemas.CandidateSchema
     page_schema = schemas.CandidatePageSchema
     filter_multi_fields = filter_multi_fields(models.Candidate)
+    filter_fulltext_fields = [('q', models.CandidateSearch.fulltxt)]
```
ccostino: What specifically is going on here?  It looks like this is just setting a property on the class that is a list value comprising of tuples that represent each field, but I just want to be sure.  What does the `'q'` in the tuple represent in this case?

webservices/args.py
```diff
@@ -126,7 +126,7 @@ def make_seek_args(field=fields.Int, description=None):
     }
 
 names = {
-    'q': fields.Str(required=True, description='Name (candidate or committee) to search for'),
+    'q': fields.List(fields.Str, required=True, description='Name (candidate or committee) to search for'),
```
ccostino: Just checking to make sure my assumption is correct on these changes - this is being changed from accepting a single value for a key to being able to support multiple values for the same key?

ccostino:
@jmcarp:  I only had a couple of questions more for just verifying my own understanding, otherwise this looks good to merge.  Thanks!

webservices/args.py
jmcarp: :+1:

webservices/resources/candidates.py
jmcarp: For implementation, check out `ApiResource` in `webservices/common/views.py`. This particular class variable is a list of tuples relating query parameters (in this case, "q") to tsvector columns for use in filtering. Basically, DRF-lite.

ccostino: Okay, I see now, thank you!

Issue #1455, opened 2016-01-22
Caucuses and primary conventions should be filterable as "primaries"


noahmanger:
Currently applying the "Primary election" filter hides the Iowa caucuses and other primary conventions. I know we had this conversation before, but I think it's important that all of these events be filterable as they are commonly understood, which is as primary elections. 

Is there a reason we can't or shouldn't do that?

LindsayYoung:
The categories in the API don't line up with the 3 election types we are showing. 

We are not creating or tagging any of the data. We are creating categories based on their election codes. The Iowa caucus is tagged as a convention, "election-C" 

The options for all the categories are:
 AOs and Rules
 Commission Meetings
 Conferences
 Conferences and Outreach
 EC Periods
 election-C
 election-G
 election-GR
 election-P
 election-PR
 election-R
 election-SC
 election-SG
 election-SGR
 election-SP
 election-SPR
 Executive Sessions
 FEA Periods
 Federal Holidays
 IE Periods
 Litigation
 Open Meetings
 Other
 Public Hearings
 report-12C
 report-12G
 report-12GR
 report-12P
 report-12PR
 report-12R
 report-12S
 report-12SC
 report-12SG
 report-12SGR
 report-12SP
 report-12SPR
 report-30G
 report-30GR
 report-30R
 report-30S
 report-30SC
 report-30SG
 report-30SGR
 report-M10
 report-M11
 report-M12
 report-M2
 report-M3
 report-M4
 report-M5
 report-M6
 report-M7
 report-M8
 report-M9
 report-MY
 report-Q1
 report-Q2
 report-Q3
 report-YE
 Roundtables
 State Outreach Workshops


Here are what the letters mean in the election codes:

G = general
P = primary
R = runoff
C = convention
S = special

noahmanger:
Gotcha. So what I'm suggesting is that either a) they should be categorized that way (for the sake of mapping onto a typical understanding of what kind of thing they are), or b) that we should have our primary filter also filter type C events, or c) if neither of those options work, we at least need to expose filters for those other types.

My preference is A or B, but I imagine there's side-effects to those that we may want to avoid. What do you think?

LindsayYoung:
A) If we want new data, that would be something that is a task that is better-aligned for FEC experts and not the dev team. 

B) We would have to confirm that all convention types can be classified as primary. I am not sure there aren't outliers that would be wrong. 

[...]

PR #1464, opened 2016-01-27
Feature/dates 2 tweaks

LindsayYoung:
Adds house districts and re-orders some fields as requested by @noahmanger


data/sql_updates/create_omnibus_dates.sql
```diff
@@ -1,14 +1,36 @@
+drop materialized view if exists ofec_omnibus_dates_mv_tmp;
+
+-- pre-processing to get the right district-state combos for House members and
+-- the election doesn't exist if the status ID is not 1
+drop view if exists tmp_election_vw;
+create view tmp_election_vw as
+    select
+        *,
+        case
+            when office_sought = 'H' and election_district != ' ' then array_to_string(
```
jmcarp: What happens when `election_district` is missing but not `' '`? It looks like it can also be `null`.

data/sql_updates/create_omnibus_dates.sql
```diff
@@ -1,14 +1,36 @@
...
+                array[
```
jmcarp: With just two values, I think `election_state || '-' || election_district` would be easier to read.

data/sql_updates/create_omnibus_dates.sql
```diff
@@ -1,14 +1,36 @@
+drop materialized view if exists ofec_omnibus_dates_mv_tmp;
+
+-- pre-processing to get the right district-state combos for House members and
+-- the election doesn't exist if the status ID is not 1
+drop view if exists tmp_election_vw;
```
jmcarp: Since we're already using common table expressions (`with subquery as (select ...)`), how about continuing that pattern instead of adding a temporary view?

jmcarp:
I'm not sure exactly which changes @noahmanger was looking for here. How about running the proposed view updates on dev and verifying with @noahmanger that things are behaving as expected?

LindsayYoung:
data part to support #1455 combines different election types into one elections filter.

data/sql_updates/create_omnibus_dates.sql
LindsayYoung: The aggregate array function work well on all cases except when there is a space string. 
 We want "CA-4" or "CA" but not "CA- "

data/sql_updates/create_omnibus_dates.sql
LindsayYoung: Then we would have to filter out the nones, right now they are handled correctly.

data/sql_updates/create_omnibus_dates.sql
LindsayYoung: That makes sense

noahmanger:
Cool cool. So the combined elections filter looks like it works as expected. Curious what the purpose of the districts is though? Are you wanting to expose these in the calendar?

noahmanger:
Thanks for moving the state abbreviation to the beginning of the title. One more small request, can we not include it at the end of the summary? I added a separate field in the tooltip for the states:

screen shot 2016-01-28 at 3 39 48 pm

LindsayYoung:
@noahmanger So the summary part is states when it is a senatorial or presidential race but it should be state and district for house races, FL-2 etc. 

The problem that I am trying to solve for is when there is something going on for one house race, you would not be able to tell which race it is. For example the OH-8 special primary on March 15 http://www.fec.gov/calendar/calendar.shtml?month=3&year=2016&day=15&categoryId=-1&eventId=-1

LindsayYoung:
Also, if we don't include the states in the description field, they will not show up as part of people's downloads and subscriptions.

noahmanger:
Ooooh. Thanks for clarifying. That all makes sense—disregard my request :) 

I'd love @emileighoutlaw 's thoughts on how we should best format the event title (`description`) and `summary`. Right now they just all kind of bleed in to each other and are repetitive.

LindsayYoung:
closes #18F/openFEC/issues/1474
Adds "Due" for reports 18F/openFEC/issues/1475

emileighoutlaw:
Just popping in to add a note per my amazing discussion with Lindsay— Our plan is to hide the event title when clicked. That should address the issue of repetition between `description` and `summary`.

LindsayYoung:
reports should be unified as described in #1474

PR #1473, opened 2016-01-29
[WIP] Add candidate fundraising history endpoint.

### LINKED ISSUE #1466: API Endpoint for totals over time
#For some of the new views, we will want to create a browse view for exploring financials. To do that, we will want to default to the most current data or have the user select a 2 year slice, or full election year in certain candidate views. Then they can filter by total receipts, total disbursements, cash on hand and debt. (The totals will be the totals of the time period selected and not just the most recent.)

jmcarp:
* [x] Add candidate fundraising history
* [x] Add committee fundraising history
* [x] Refactor existing code to use pre-aggregated totals
* [x] Update unit tests

[Resolves #1466]


jmcarp:
Assigning to @LindsayYoung for review. The specialized committee endpoints will need more work that can happen in a subsequent patch, but the candidate views should be in good shape. The complementary changes to the frontend live in https://github.com/18F/openFEC-web-app/pull/1073, which will also pull in https://github.com/18F/fec-style/pull/248.

webservices/rest.py
```diff
@@ -192,7 +192,8 @@ def add_aggregate_resource(api, view, schedule, label):
 api.add_resource(candidate_aggregates.ScheduleABySizeCandidateView, '/schedules/schedule_a/by_size/by_candidate/')
 api.add_resource(candidate_aggregates.ScheduleAByStateCandidateView, '/schedules/schedule_a/by_state/by_candidate/')
 
-api.add_resource(candidate_aggregates.TotalsCandidateView, '/candidate/<candidate_id>/totals/')
```
LindsayYoung: Can we keep this url and direct to the TotalsCandidateView? I feel like this would be a breaking change that we don't need to make.

webservices/rest.py
```diff
@@ -192,7 +192,8 @@ def add_aggregate_resource(api, view, schedule, label):
...
+api.add_resource(candidate_aggregates.TotalsCandidateView, '/candidates/totals/')
```
LindsayYoung: Any reason we don't want to add these endpoints to the docs yet?

PR #1481, opened 2016-02-02
Use ProxyFix middleware.

jmcarp:
Since we're running the API behind a proxy, we need to use the `ProxyFix`
middleware to pull the forwarded URL, scheme, etc. off the request. This
fixes redirects from https to http. I'm guessing this issue emerged when
we dropped the unnecessary nginx proxy.


jmcarp:
Flagging @LindsayYoung

PR #1484, opened 2016-02-04
Fix path to refresh task in scheduler.

jmcarp:
Restore nightly refresh to celery-beat.


LindsayYoung:
LGTM, I can see it is running, it is very straight forward anything else I should do to review this?

PR #1487, opened 2016-02-05
Handle related fields in data export.

jmcarp:
Fixes an error on downloading itemized independent expenditures such
that table headers were inconsistent with field names.


jmcarp:
Merging for now to get IE downloads working, but let's do a post-mortem review when time permits.

PR #1497, opened 2016-02-06
Feature/costs sort pdf

jmcarp:
Note: the date column is called `transaction_date` for communication costs and `disbursement_date` for electioneering.


data/sql_updates/create_electioneering_view.sql
```diff
@@ -18,9 +18,11 @@ create index on ofec_electioneering_mv_tmp (cand_office_st);
 create index on ofec_electioneering_mv_tmp (cand_office_district);
 -- These will be added back in
 -- create index on ofec_electioneering_mv_tmp (election_tp);
--- create index on ofec_electioneering_mv_tmp (receipt_dt);
```
LindsayYoung: I think it would still be nice to have receipt date, it was one of the things we asked for

PR #1500, opened 2016-02-08
Filter itemized IEs on filing form.

jmcarp:
Prerequisite for hiding non-F3X items on the webapp.


LindsayYoung:
LGTM

I did notice that sometimes we call it form_type and sometimes we say filling_form, depending on what the underlying table says. I don't think we should address this in this PR, just wanted to note it.

PR #1504, opened 2016-02-09
Drop contribution aggregate by contributor.

### LINKED ISSUE #1303: To have confidence in our data, verify logic for contributor aggregates
All receipts aggregates from individuals use a custom filter described by @jwchumley and @PaulClark2 to remove duplicate contributions (receipt type is one of "15", "15E", "15J", "18J" or amount < 200 and "earmark" not in memo, etc.). We handle the problem of duplicate earmarked contributions differently for aggregates of committee contributions, since this filter only applies to contributions from individuals. I don't think we've discussed this logic specifically, so I'd like to run our query by you and verify that it's correct.

Here's how we tally up contributions aggregated by donor committee. We're filtering out receipts with memo_code "X" (to filter out memoed subtotals), and with entity_type != "IND" (to filter out contributions routed through Club for Growth, ActBlue, etc.). Does this make sense @jwchumley @PaulClark2?

cc @LindsayYoung 

```
create table ofec_sched_a_aggregate_contributor as
select
    cmte_id,
    rpt_yr + rpt_yr % 2 as cycle,
    contbr_id,
    max(contbr_nm) as contbr_nm,
    sum(contb_receipt_amt) as total,
    count(contb_receipt_amt) as count
from sched_a
where rpt_yr >= :START_YEAR_AGGREGATE
and contb_receipt_amt is not null
and contbr_id is not null
and coalesce(entity_tp, '') != 'IND'
and (memo_cd != 'X' or memo_cd is null)
group by cmte_id, cycle, contbr_id
;
```

jmcarp:
Instead, use disbursement aggregate by recipient. This patch also joins
disbursement aggregates on related committees to expose committee names
in the API, as needed for the "between committees" view on beta.fec.gov.
For consistency, we might move the join logic to the `BaseAggregate`
class.

[Resolves #1303]


LindsayYoung:
It looks good to me. 

Since we are removing an endpoint, I am wondering if we should do a mailing to our google group about the upcoming change. We can add an explanation to why it is gone on the the schedule A section of our docs. 

Also wanted to confirm we switched endpoints and links on the front end.

jmcarp:
Frontend changes are here: https://github.com/18F/openFEC-web-app/pull/1086. It would probably be best to review and merge the two pull requests together to verify that everything works correctly together.

LindsayYoung:
sounds good, I can take a look now

PR #1505, opened 2016-02-10
Hotfix/date sort bug

LindsayYoung:
The validation on the reporting and election endpoints is not working, I think we should take that validation down as a quick fix, but I would still like to understand why it suddenly had trouble validating these endpoints.


LindsayYoung:
OK, I think it is not validating because there is not an index on due date. I will look at that now.

LindsayYoung:
We don't create indexes on those, and I don't think we need to there is not that much data there.

jmcarp:
@LindsayYoung, could you resolve the merge conflict? Hotfix PRs should normally target master, but since this change is so small, it's not worth retargeting.

noahmanger:
Is this still relevant and needed?

noahmanger:
Checking on this again. Still needed or should we close?

LindsayYoung:
Let's kill it.

PR #1512, opened 2016-02-12
Revert null position option.

jmcarp:
Allowing users to control where nulls appear in sort order has the
undesirable consequence of permitting queries that sort without relying
on an index. This patch removes the `sort_nulls_large` option to prevent
users from sending potentially slow queries of this type.


LindsayYoung:
Could we all schedule a time to talk this over?

PR #1514, opened 2016-02-12
reformats report code

### LINKED ISSUE #1508: Format report due date decriptions
In the API we should reformat the report descriptions to be more readable. 

**Current:**
screen shot 2016-02-11 at 3 55 44 pm

**Proposed:**
March monthly report due. Report code: M3

@emileighoutlaw what do you think of that change?

LindsayYoung:
Resolves #1508

### DIFF
--- a/data/sql_updates/create_omnibus_dates.sql
+++ b/data/sql_updates/create_omnibus_dates.sql
@@ -139,9 +139,8 @@ with elections_raw as(
         array_to_string(array[
             expand_office_description(office_sought::text),
             clean_report(rpt_tp_desc::text),
-            'Due, Report (',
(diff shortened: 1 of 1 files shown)


data/sql_updates/create_omnibus_dates.sql
```diff
@@ -139,9 +139,8 @@ with elections_raw as(
         array_to_string(array[
             expand_office_description(office_sought::text),
             clean_report(rpt_tp_desc::text),
-            'Due, Report (',
+            'Due, Report code:',
```
noahmanger: Minor thing: can this be a period after "Due"?

PR #1518, opened 2016-02-14
Feature/update electioneering schema

jmcarp:
Add columns to EC and CC models; remove nested committees from EC output.

h/t @PaulClark2

Review with https://github.com/18F/openFEC-web-app/pull/1097.


webservices/common/models/costs.py
```diff
@@ -69,9 +74,5 @@ def pdf_url(self):
         return utils.make_report_pdf_url(self.beginning_image_number)
```
LindsayYoung: If it is beginning image number, I think that will take the user to the first page of the filing and not the page with the transaction. I think that is different than what we do with other transactions.

PaulClark2: The EC view has both the Form 9 beginning image number (f9_begin_image_num) and the SB (transaction) image number (sb_image_num)

webservices/common/models/costs.py
```diff
@@ -58,9 +63,9 @@ class Electioneering(db.Model):
     disbursement_amount = db.Column('reported_disb_amt', db.Numeric(30, 2), index=True)
     purpose_description = db.Column('disb_desc', db.String)
     report_year = db.Column('rpt_yr', db.Integer, index=True)
-
-    committee = utils.related_committee('committee_id')
-    candidate = utils.related_candidate('candidate_id')
+    election_type = db.Column('election_tp', db.String)
```
LindsayYoung: I was expecting P, G, S and R or something like that for election_type, instead there is:
```
 election_tp 
─────────────
 P2008
 P2016
 S
 O2011
 C
 G2012
 P2010
 P2014
 R
 G2008
 P2012
 R2008
 O2008
 G2014
 O2010
 G
 P
 G2010
```
I am not sure why sometimes it includes the year and sometimes it doesn't

webservices/common/models/costs.py
LindsayYoung: Thanks paul!

webservices/common/models/costs.py
PaulClark2: EYYYY (election plus election year) is correct for electronic filings.   Single character election types are from paper filings.

LindsayYoung: What do you think about displaying them as just the election in the API? We already have several dates on this endpoint, so year is covered

Thanks again!

PaulClark2: That makes sense to me.

Issue #1517, opened 2016-02-14
Discussion: revisit branching and continuous delivery workflow


jmcarp:
Our repos currently use the git-flow branching workflow, which we integrate with continuous delivery. Successful builds on the develop branch are automatically deployed to the development server, on the master branch to staging, and tagged commits to production. We initially settled on this workflow because it's familiar to many developers and includes good tooling, but it's becoming increasingly clear that the model is confusing for people on both the 18F and FEC teams. I'd like to discuss alternative branching and deployment workflows, with the goal of finding a pattern that makes sense to everyone on the team. Specifically, I'm not convinced that the complexity of git-flow is justified for our applications.

As specific alternatives, I'd like to consider GitHub Flow and GitLab Flow, both of which are simpler than git-flow but might suit our needs at least as well. Both of these workflows are described in this article, which I recommend at least skimming if you want to weigh in: http://doc.gitlab.com/ee/workflow/gitlab_flow.html.

cc @noahmanger @LindsayYoung @ccostino. And, now that we have the author of Continuous Delivery on the 18F team, flagging @jezhumble in case you're interested and want to advise.

jezhumble:
Hi! Since you asked, here's what I think, although with the caveat that it's based on git-flow as canonically defined, and I don't know how your team has implemented it; practice > theory.

I don't consider git-flow to be continuous integration, let alone continuous delivery, for the simple reason that continuous integration is defined as all developers merging _into_ master on at least a daily basis. Indeed, allowing work to remain on feature branches unmerged _into_ master for any length of time is pretty much the opposite of "flow". The amount of time you are going to spend engaged in re-work to deal with merge issues, and generally futzing with version control, will increase non-linearly with both the number of branches and the length of time they remain unmerged. Git-flow is basically the old-school ClearCase workflow but with much better tooling (git) which take it from utterly unmanageable to bearable (but far from ideal).

If you're working on a project with community contributions, there are good reasons to deviate from this, in which case I think the GitHub model is OK, although I still think feature branches should be kept short (less than a day's work), because when there is a long diff between a feature branch and master it becomes hard to reason about its impact, and this slows down code review (IMHO code reviews should happen in-process, not asynchronously). I have not checked out GitLab flow.

[...]

PR #1502, opened 2016-02-09
Copy to csv proof of concept.

jmcarp:
Use the `copy` command of `psycopg2` to export queries to csv, rather
than loading and marshalling data in python. Based on some simple
benchmarks, this is much faster than the current implementation. Some
features are missing, though: this proof of concept doesn't include
fields implemented as properties, or nested fields. Fields excluded
using marshmallow are currently being included, and array types aren't
being rendered nicely.

Next, I'd like to do some more thorough benchmarking and discuss which
formatting niceties should be restored, and how. Paging @LindsayYoung
and @ccostino.

Update: nesting doesn't seem like an essential feature for download, or possibly even a useful feature, so I'm going to assume we don't need to restore it in this patch.

Tasks:
* [ ] Verify behavior for all views of interest
* [ ] Add dynamically generated columns
* [ ] Choose appropriate size caps
* [ ] Review updates with team

[Resolves #1489]

PR #1521, opened 2016-02-16
[WIP] Manage schema updates with airflow.

### LINKED ISSUE #1515: Explicitly track dependencies between data views
Some of our data views are built directly from the raw FEC tables, but others depend on a chain of our own views. For example, the candidate full-text search view (ofec_candidate_fulltext_mv) depends on the candidate detail view (ofec_candidate_detail_mv). If the views are created in the wrong order, the process will crash. If the views are updated in the wrong order, some views may be stale.

At the moment, we're using the very brittle approach of running sql files in alphabetical order by file name. Instead, I propose that we explicitly declare dependencies between data import and refresh steps and ensure that all steps run in an appropriate order. Airbnb's airflow library is designed to solve exactly this kind of problem and might be a good candidate--or if that's overkill, we can build a directed graph of view dependencies and run steps in topologically sorted order.

jmcarp:
Instead of running schema updates in order by filename, explicitly declare dependencies between tasks and manage updates using airflow. This patch is a work in progress but includes enough work to give a sense of the approach--I'd like to discuss with @LindsayYoung and @ccostino, and anybody else who's interested, and decide how to proceed.

[Resolves #1515]

I thought I'd include a few screenshots of the airflow interface. I find it really useful to be able to see the dependencies between data loading tasks visually, and it's also helpful to be able to track the progress of the various update tasks in real time:

screen shot 2016-02-15 at 11 36 02 pm
screen shot 2016-02-15 at 11 36 18 pm


LindsayYoung:
Can we schedule a time to talk about this one?

jmcarp:
Sounds good! Want to pick a time when we're all free? While we're at it, we also need to talk over #1502 and #1517, the latter of which should include @noahmanger.

noahmanger:
Great. Yeah I'd love to be in these convos.

LindsayYoung:
@noahmanger I will send you an invite

noahmanger:
Do we still want to review this patch this sprint? Or should we move it to the backlog?

ccostino:
@noahmanger:  Actually if it'd be possible to do so, I think it'd be a good thing.  Not only does this patch provide a means of executing scripts in the correct order and managing them, it also drops the materialized views so we're dealing with just the tables directly.  This should theoretically simplify things from a data management standpoint.

@jmcarp, do you agree, or do you feel this should come at a later point when we can devote more time to it directly?

PR #1523, opened 2016-02-16
Use S3 service in Cloud Foundry.

### LINKED ISSUE #1522: Manage S3 buckets using Cloud Foundry service
* Reduce overhead around manually provisioning buckets and setting credentials
* Use separate buckets for dev / stage / prod environments

jmcarp:
[Resolves #1522]

### DIFF
--- a/manifest_dev.yml
+++ b/manifest_dev.yml
@@ -4,6 +4,7 @@ host: fec-dev-api
 services:
   - fec-creds-dev
   - fec-redis-dev
+  - fec-s3-dev
 env:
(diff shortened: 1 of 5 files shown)


jmcarp:
Assigning to @LindsayYoung for review. This isn't urgent and doesn't need to go into the upcoming release, but it's also pretty simple.

LindsayYoung:
This looks good, I noticed that the s3 service is in dev but not added to stage or prod yet, should we wait to merge after we add the service to staging and production so we don't forget?

jmcarp:
I added the service to all our spaces earlier this week.

Issue #1526, opened 2016-02-16
Properly decode Caucus elections from Convention elections


noahmanger:
As discussed in Slack, caucuses and conventions are both coded `C` in the data, making it difficult to properly label the events. I'd like to consolidate our discussion around this topic here.

I see there's three options:

### Option 1: FEC fixes the underlying data
Though this sounds like there's some significant challenges because of the way the data is entered.

### Option 2: Create some logic for deriving the correct meaning of "C"
I'm wondering if it would be possible to decode these events based on the state they occur in? Do we have a list of which states hold conventions and which hold caucuses? And then for the four states that have both, we could just manually set them? Could this just be a one time manual process for 2016 elections?

### Option 3: Decode them all as "Caucus or convention"
Simplest quick fix for the time being. Not the prettiest solution, but would this work?

noahmanger:
cc @PaulClark2 @jwchumley @amykort

AmyKort:
I am out of the office until 02/18/2016.




Note: This is an automated response to your message  "Re: [openFEC]
Properly decode Caucus elections from Convention elections (#1526)" sent on
2/16/2016 5:32:46 PM.

This is the only notification you will receive while this person is away.

jwchumley:
I'm testing the entry system this morning. In the test system I've added CON and CAU for convention and caucus. I'm going to create some elections and add some candidates and run some reports to make sure nothing breaks.

AmyKort:
I am out of the office until 02/18/2016.




Note: This is an automated response to your message  "Re: [openFEC]
Properly decode Caucus elections from Convention elections (#1526)" sent on
2/17/2016 7:01:16 AM.

This is the only notification you will receive while this person is away.

jwchumley:
Here are the 2016 elections that should be either CON or CAU:

TRC_Election_type_id	TRC_ELECTION_ID
CAU	                                1978
CAU	                                 2313
CAU	                                 2321
CAU	                                 2338
CAU	                                 1987
CON	                                 2322
CON	                                 2340
CAU	                                 2323
CAU	                                 2310
CAU	                                 2020
CAU	                                 2324
CAU	                                 2316
CAU	                                 2023
CAU	                                 2157
CAU	                                 2032
CAU	                                 2314
CAU	                                 2041
CAU	                                 2326
CAU	                                 2341
CAU	                                 2065
CAU	                                 2325
CAU	                                 2052
CON	                                 2335
CON	                                 2334
CON	                                 2333
[...]

PR #1528, opened 2016-02-17
Adding additional decoding for caucuses and conventions

LindsayYoung:
as outlined in #1526


LindsayYoung:
Tests not passing, going to look at that now.

LindsayYoung:
Everything works when I run it locally but the tests don't pass because they can't seem to find the helper functions.

LindsayYoung:
think it is a type error, will fix now

LindsayYoung:
@jmcarp ready for review

PR #1530, opened 2016-02-18
Feature/hide prez fix caucus

LindsayYoung:
Hopefully the last round of calendar for a while
 - fixes our list of manual caucus convention coding 
 - hides pre-primary presidential reports on even years, look at the domination update to see why


LindsayYoung:
h/t @PaulClark2 for finding the over-zealous filtering

PR #1531, opened 2016-02-18
Update exclusion rule.

jmcarp:
As written, this patch excluded all odd-year reports and all reports
with the 12C or 12P types. This update only excludes 12C and 12P reports
in even years.


LindsayYoung:
Thanks for catching that!

LindsayYoung:
OK, Paul also caught that this is excluding all pre-primary and not just the presidential ones, I will update that now.

PR #1534, opened 2016-02-22
Feature/statitutory candidate flag

LindsayYoung:
Filter by candidates that have ever raised $5,000 

This wraps up the dependency for 18F/openFEC-web-app/issues/859

### DIFF
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -32,6 +32,18 @@ with
             max(fec_election_yr) as max_cycle
         from fec_yr
         group by cand_id
+    ),
+    five_thousand as (
+        select
+            distinct on (cand_id)
+            cand_id,
+            case when array_agg(fec_yr.cand_status)::text[] && array['C', 'F', 'P'] then
+                true::boolean
+            else
+                false::boolean
+            end as five_thousand_flag
+        from fec_yr
+        group by cand_id, cand_status
     )
 select distinct on (fec_yr.cand_id, fec_yr.fec_election_yr)
     row_number() over () as idx,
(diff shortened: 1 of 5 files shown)


jmcarp:
I sent some suggested revisions in #1535. Also, I'm wondering if the flag should have "statutory" in the name, since that's what it's checking most directly--maybe `is_statutory` or `statutory_candidate`. What do you think @emileighoutlaw?

LindsayYoung:
I like `five_thousand_flag` because I did not want people to misconstrue us as weighing in on “testing the waters" political controversy-- that might be over cautious. It also mirrors the proposed language on the website. 

Always happy to her Emileigh's opinion too.

jmcarp:
So I took a look at the data, and I don't think `five_thousand_flag` is an accurate description of what's going on here. The data include both statutory candidates who haven't raised $5000 and candidates who've raised $5000 who aren't classified as statutory candidates.

```
fec=> select id as candidate_id from ofec_candidate_fulltext_mv where receipts >= 5000 except all select distinct(candidate_id) from ofec_candidate_history_mv where five_thousand_flag order by candidate_id;
candidate_id
--------------
 H0AL03184
 H0AL06088
 H0AL07110
 H0CA10081
 H0CA10099
 H0CA10107
 H0CA10156
 H0CA10172
 H0CA10206
 H0CA11311
 ...
(455 rows)
```

```
fec=> select distinct(candidate_id) from ofec_candidate_history_mv where five_thousand_flag except all select id as candidate_id from ofec_candidate_fulltext_mv where receipts >= 5000;
candidate_id
--------------
 H8CA33011
 P00002162
 H0CA07046
 S2NJ00130
 H8MI09019
 H4TN05053
 H2MT01037
 S2CA00153
 H8MI18010
 P60003555
...
(1704 rows)
```

If we're really trying to filter by fundraising, let's forget about `cand_status` and actually filter by fundraising. And if we want to filter by candidate status, let's not refer to that filter in terms of financials. Unless I'm interpreting these queries incorrectly--correct me if I'm wrong.

LindsayYoung:
We should make this using the financial information. Thank you for investigating this, Josh!

PR #1535, opened 2016-02-22
Fix and clean up statutory candidate flag.

jmcarp:
As implemented, the statutory candidate flag includes a number of false
negatives due to a `group by` on `cand_status`, combined with a
`distinct` on `cand_id`. The `group by` yields potentially many rows per
candididate, with some `true` and some `false`; the `distinct` selects
just one of these rows, which could be a `false` when other rows are
`true`. This patch drops these two expressions. And now that the
grouping logic is the same as in a previous select, it also combines
selects to make the view slightly faster to build.

* Drop group by on `cand_status`
* Drop unnecesary distinct on `cand_id`
* Move status aggregate into existing select
* Simplify case to plain &&

BTW: any `case` statement that only yields `t` or `f` should be possible to simplify to a boolean operation.

### DIFF
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -25,25 +25,15 @@ with
         ) dedup
         group by cand_id
     ),
-    cycles as (
+    aggregates as (
         select
             cand_id,
             array_agg(fec_election_yr)::int[] as cycles,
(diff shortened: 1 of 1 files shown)


LindsayYoung:
That is much better! Also, thanks for the great explanation, @jmcarp

PR #1542, opened 2016-02-25
Use pre-computed election ranges to round up to next election.

jmcarp:
* Use pre-computed election ranges for `election_full` flag
* Limit election ranges to maximum office duration

[Resolves https://github.com/18F/openFEC-web-app/issues/1107]

### DIFF
--- a/data/functions/utils.sql
+++ b/data/functions/utils.sql
@@ -4,3 +4,14 @@ begin
     return year + year % 2;
 end
 $$ language plpgsql immutable;
+
+create or replace function election_duration(office text)
+returns integer as $$
+begin
+    return case office
+        when 'S' then 6
+        when 'P' then 4
+        else 2
+    end;
+end
+$$ language plpgsql;
--- a/data/sql_updates/create_candidate_history_view.sql
+++ b/data/sql_updates/create_candidate_history_view.sql
@@ -115,7 +115,10 @@ with years as (
 select distinct on (years.candidate_id, years.cand_election_year)
     years.candidate_id,
     years.cand_election_year,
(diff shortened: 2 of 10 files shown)


LindsayYoung:
This fixes  18F/openFEC-web-app#1107 but introduces an error on the other end of the spectrum. 

If you want to see the full cycle for an election in the future, it doesn't show summary results:
http://localhost:3000/candidate/S2KY00012/?cycle=2020&election_full=true

LindsayYoung:
OK, it is not isolated, other candidates in 2020 also not going on full cycle:
http://localhost:3000/candidate/S2TN00058/?cycle=2020&election_full=true

Candidates running on 2018 have the full cycle same problem:
http://localhost:3000/candidate/S2MO00429/?cycle=2018&election_full=true

jmcarp:
Good catch. Those issues don't appear related to this patch, though--looks like they originate from #1473. I'll push a fix here in a few minutes.

LindsayYoung:
Looking back at the test cases but I am going to try again because I think my last migration might have been clobbered. 

Bugs from #1473:
http://localhost:3000/candidate/S2KY00012/?cycle=2020&election_full=true
http://localhost:3000/candidate/S2KY00012/?cycle=1982&election_full=true  -error
http://localhost:3000/candidate/S2MO00429/?cycle=2018&election_full=true
Reported bug #1107:
http://localhost:3000/candidate/S2KY00012/?cycle=1982&election_full=true  -error

LindsayYoung:
Woo Hoo! they are all working. I will poke around a but and merge if it all looks okay!

PR #1543, opened 2016-02-25
Adding some more inline docs closes #1533

### LINKED ISSUE #1533: add some calendar docs
Since it is complicated, add some one liners explain the calendar data.

LindsayYoung:
Just a little inline documentation. @ccostino since you haven't looked at it maybe you would be a good to just eyeball this file and judge if this is helpful or if there are additional things to document. We don't need to spend a lot of time on it, hopefully this script will be replaced with a better input method.

### DIFF
--- a/data/sql_updates/create_omnibus_dates.sql
+++ b/data/sql_updates/create_omnibus_dates.sql
@@ -58,8 +58,7 @@ $$ language plpgsql;
 
 
 -- Not all report types are on dimreporttype, so for the reports to all have
--- titles, I am adding a case. Ideally, we would want the right mapping,
--- that is one of the things we have asked for.
+-- titles, I am adding a case. Ideally, we would want the right mapping.
 create or replace function name_reports(office_sought text, report_type text, rpt_tp_desc text, election_state text[])
 returns text as $$
     begin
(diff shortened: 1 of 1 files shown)


ccostino:
Thank you, @LindsayYoung, I'll give this a quick pass!

data/sql_updates/create_omnibus_dates.sql
```diff
@@ -58,8 +58,7 @@ $$ language plpgsql;
 
 
 -- Not all report types are on dimreporttype, so for the reports to all have
--- titles, I am adding a case. Ideally, we would want the right mapping,
--- that is one of the things we have asked for.
+-- titles, I am adding a case. Ideally, we would want the right mapping.
 create or replace function name_reports(office_sought text, report_type text, rpt_tp_desc text, election_state text[])
```
ccostino: So when the description (`rpt_tp_desc`) is `null`, just the report type itself (`report_type`) is displayed in its place as the title?

ccostino:
@LindsayYoung, this looks fine to me and covers some of the things we had reviewed last week, I only had one follow up question for clarification.  I am assuming some of these things will disappear with the move off of the data warehouse?

Either way, thank you!  Happy to merge now that I've had a chance to review.

data/sql_updates/create_omnibus_dates.sql
LindsayYoung: Yes, sometimes I found that was not populated so I show the code if we can't identify it.

LindsayYoung:
When we talk about transitioning these scripts, we are thinking about creating a data entry system to create the dates in a consistent way.  Only dimreporttype is data warehouse, so that would be a small change. And I think we just need to hard-code it.

PR #1544, opened 2016-02-26
Feature/filter candidate election year

jmcarp:
Filter candidates by election year.

Part of https://github.com/18F/openFEC-web-app/issues/1002.


LindsayYoung:
To recap, we are getting rid of the worse name search field, in favor of q, and using election year for sorting results on the candidate table. I spot checked it locally and it looks good to me.

PR #1548, opened 2016-02-29
Feature/money flags

LindsayYoung:
Adding two flags to the candidate history view
- five_thousand_flag
- federal_fund_flag

Doing this by creating a table after table creation that joins candidate history and candidate totals. Open to better ways of ding this if you have ideas. 

Resolves https://github.com/18F/openFEC/issues/1469
Resolves https://github.com/18F/openFEC/issues/1476


noahmanger:
Is this ready for review? Would love to get this and the related changes merged in to finish out this sprint.

LindsayYoung:
ahhh look at failing tests later

data/sql_updates/post_totals_create_candidate_aggregates.sql
```diff
@@ -8,7 +8,8 @@ with totals as (
         receipts,
         disbursements,
         last_cash_on_hand_end_period,
-        last_debts_owed_by_committee
+        last_debts_owed_by_committee,
+        cast('false' as boolean) as federal_funds_flag
```
ccostino: If I'm understanding this correctly, this is just defaulting to `false` in this case because the flag is only meant for presidential candidates, right?

LindsayYoung: Yes

jmcarp: By the way, why the `cast` here? Can we not get the same value with `false`?

LindsayYoung: At one point I was getting a casting error and I think I was over-zealous about getting rid of it. I can simplify that.

jmcarp:
Our sqlalchemy models are configured in a pretty unusual way, with multiple models mapped to the same table in a few cases. I set this up a long time ago to reduce duplicated code, but now I'm wondering if we can simplify the code further. Right now, we have a `Candidate` model and a `CandidateDetail` model. Both models read from the candidate detail table, but the `Candidate` model hides some of the columns. This structure predates me on the project, and I never really understood why it was useful. Is there a good reason to expose two views onto the same table like this?

Either way, we can ignore failures related to potentially unused eager loads by disabling the eager load checks in `nplusone`. To do that, add the following to `tests/common.py` after line 17:

```python
rest.app.config['NPLUSONE_WHITELIST'] = [
    {'label': 'unused_eager_load'},
]
```

With that configuration, there are still some tests failing, at least on my local install. I also didn't see any tests for the new behavior--can we add tests to verify that the new flags are appearing where they're supposed to, with the right values?

LindsayYoung:
I disabled the nplusone tests, we should re-enable them later. Tell me if we should delete them for now.

jmcarp:
@LindsayYoung: I just released nplusone 0.7.0, which handles checks on eager loads more intelligently. When I run this branch with the latest nplusone, the warnings about unused eager loads go away, but I still see several warnings about lazy loads. That actually makes sense, because many of the candidate models and views are now associated with `CandidateFlags`, but only the `CandidateList` view uses the necessary eager load. That eager load doesn't seem to be happening in `CandidateSearch` or `CandidateView`. It *appears* to be happening in `CandidateHistoryView`, but it actually isn't, since we override the `build_query` method without calling `super` or explicitly applying `query_options`. So I would recommend updating nplusone, re-enabling the nplusone checks, and adding those eager loads.

Does that make sense? If you don't have time, I can send you a sub-PR.

data/sql_updates/post_totals_create_candidate_aggregates.sql
```diff
@@ -8,7 +8,8 @@ with totals as (
         receipts,
         disbursements,
         last_cash_on_hand_end_period,
-        last_debts_owed_by_committee
+        last_debts_owed_by_committee,
+        'false' as federal_funds_flag
```
jmcarp: Does this need to be a string? Can we use `false as federal_funds_flag` instead? If we can, then we also shouldn't need the `cast` below.

data/sql_updates/post_totals_create_candidate_aggregates.sql
```diff
@@ -51,7 +55,9 @@ election_aggregates as (
         candidate_id,
         election_year,
         sum(receipts) as receipts,
-        sum(disbursements) as disbursements
+        sum(disbursements) as disbursements,
+        sum(receipts) >= 5000 as five_thousand_flag,
+        array_agg(federal_funds_flag) @> array[cast('true' as boolean)] as federal_funds_flag
```
jmcarp: Same question as above--do we need the `cast` here? How about something like `true != any array_agg(federal_funds_flag)`?

LindsayYoung:
@jmcarp thank you for the great comments on this post! 

I think the candidate models diverged right before you came on to get them faster. The calls for affiliated committees were slowing them down. I am not opposed to simplifying anything if we can.

I can take a first pass at making these changes, then I would like to follow up with you, Josh. 

Thanks again!

ccostino:
@jmcarp: Thank you for the details, this is a huge help.  Is there a reason that you're aware of as to why the `build_query` method is overridden in the `CandidateHistoryView` resource without calling `super()` like the other ones do?  Is there any harm that you see in changing that to work like the others?

jmcarp:
@ccostino: the best bet is to read the source, but I don't see a reason against calling `super` in `CandidateHistoryView`.

ccostino:
@jmcarp: Yes, and from what I could tell nothing seemed out of the ordinary in looking it over.  Modifying the call to use the super method seemed to continue to work fine and all of the tests continued to pass (we'll check this out further to verify).  I was asking more for if you had historical context around the implementation just to make sure I wasn't missing something else (and no worries if not!).

tests/common.py
```diff
@@ -5,7 +5,7 @@
 import subprocess
 
 from webtest import TestApp
-from nplusone.ext.flask_sqlalchemy import NPlusOne
+# from nplusone.ext.flask_sqlalchemy import NPlusOne
```
LindsayYoung: I'll add this back

webservices/resources/candidates.py
```diff
@@ -84,7 +98,9 @@ class CandidateSearch(CandidateList):
 
     schema = schemas.CandidateSearchSchema
     page_schema = schemas.CandidateSearchPageSchema
-    query_options = [sa.orm.subqueryload(models.Candidate.principal_committees)]
+    query_options = [
+        sa.orm.subqueryload(models.Candidate.principal_committees),
```
ccostino: We've made a note of this as being something that is still problematic with `nplusone` when changing it to `joinedload` and will address in the future.

ccostino:
I am going to merge this now so that we don't lose momentum of getting this done, @LindsayYoung and I   walked through this again earlier today and reviewed things as they are.  We've noted the remaining `subqueryload` call and have an issue made to re-enable `nplusone` again as well.

webservices/resources/candidates.py
jmcarp: Curious--how can you tell this is problematic with `nplusone`? It looks like tests failed with `joinedload` even with `nplusone` disabled.

ccostino: @jmcarp:  You're absolutely right, I'm sorry.  I mixed up two separate issues we ran into and confused the two, and in my haste attributed the problem incorrectly.  Please disregard the note above, the issues is with the way the joins themselves are constructed and the SQL that's ultimately generated, **not** `nplusone`!

PR #1552, opened 2016-03-03
add tmp to candidate views

LindsayYoung:
Not sure if this is the whole problem but this might take the view down for a minute.


jmcarp:
I'm not sure this name change will have any effect on anything. When we rename from `my_view_tmp` to `my_view`, we only check materialized views, and `ofec_committee_totals_tmp` is a table. If we want the rename to take effect, we can change the `create table` to a `create materialized view`.

noahmanger:
Is this still something we want to review and merge? If so, can we assign a reviewer?

LindsayYoung:
Closing this in favor of some airflow-like solution for better restructuring data flows later

PR #1555, opened 2016-03-07
Feature/all day events

jmcarp:
Handle multi-day events.

* Use `use_time` flag from `cal_event` table
* Localize and optionally format all dates
* Fix stray linter errors

[Resolves https://github.com/18F/FEC/issues/263]

### DIFF
--- a/data/functions/utils.sql
+++ b/data/functions/utils.sql
@@ -4,3 +4,14 @@ begin
     return year + year % 2;
 end
 $$ language plpgsql immutable;
+
+create or replace function election_duration(office text)
+returns integer as $$
+begin
+    return case office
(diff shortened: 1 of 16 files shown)


jmcarp:
Recommend verifying this with https://github.com/18F/fec-cms/pull/225 checked out. Specific points to verify:

* Multi-day events, like conferences, are displayed correctly (spanning multiple days, and without hours or minutes)
* Time-bounded events, like webinars, are displayed in the correct local timezone
* All of the above also works on Google Calendar

I'm open to suggestions about improving tests here, but since some bugs we've run into happen on ical / google calendar, I'm not sure how to test some behaviors effectively.

PR #1557, opened 2016-03-07
Fix two typos in 'Quick(ish) start' section

fureigh:
This quick pull request fixes two typos:

 * `create up your environment` --> `create your environment`
 * `make sure you environment` --> `make sure your environment`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ Prior to running, ensure you have the following requirements installed:
 
 Clone this repository.
 
-Create a Python environment for your project. We use virtualenv and virtualenv wrapper, but feel free to create up your environment with your preferred set up.
+Create a Python environment for your project. We use virtualenv and virtualenv wrapper, but feel free to create your environment with your preferred set up.
 
-You will need your Python 3 path to make sure you environment is pointing to the right Python version. You can find that out by running:
+You will need your Python 3 path to make sure your environment is pointing to the right Python version. You can find that out by running:
 ```
 which python3
 ```


LindsayYoung:
Thanks for catching that!

jmcarp:
Thanks @fureigh. Now that you mention it, I would actually like to drop this text altogether. Users don't necessarily need to use a virtualenv to run the code (they might be using docker or vagrant), and if they do use a virtualenv, they don't need to use virtualenvwrapper.

LindsayYoung:
Or vagrant set up is not documented and in the line before, it says "feel free to create up your environment with your preferred set up." 

I added this after setting some FEC people up, this is how they are configured and I wanted them to be able to refer to the docs for that.

jmcarp:
Right, I'm not talking about our vagrant setup in particular--I'm just suggesting that some people may want to develop FEC using vagrant, and they may not need a virtualenv. This section of the docs is less about running FEC in particular and more about running python projects in general. For folks who aren't familiar with python development, I'd prefer just linking to a more complete existing guide, like https://pages.18f.gov/dev-environment-standardization/languages/python/.

ccostino:
@fureigh, thanks for submitting these quick fixes!

@LindsayYoung and @jmcarp, these instructions are among the things I have listed that I'd like to take a closer look at and improve in general.  I understand the distinction between the steps necessary in getting things setup for this project in particular vs. setting up a Python project in general; that would simplify things here by linking out to the Python guide to help folks setup their Python environment as they prefer before going through the specific steps needed for this project.  However, maintaining a quick start section would be beneficial for anyone looking to get up and running quickly that may not be familiar with the Python ecosystem.  We've been talking about this quite a bit in the documentation working group and discussing what makes a valuable README as well as valuable installation/setup instructions vs. repeating ourselves all over the place.

All of that said, those are much larger edits to the overall README, and these are two quick fixes we can put in place now while these instructions remain prominent.  Let's merge this PR, and I'll also create a new issue to capture the thoughts around overall README improvement more formally so that we can track the work better.

Sound good to everyone?

fureigh:
@ccostino That all sounds good to me. Thanks!

ccostino:
Excellent, I created #1559 to track this larger effort. :-)

PR #1564, opened 2016-03-14
Use external library for COPY TO.

jmcarp:
The PostgreSQL `COPY TO` helpers have been externalized and generalized
at https://pypi.python.org/pypi/sqlalchemy-postgres-copy. This patch
switches to the external library for `COPY TO` logic.


ccostino:
Looks good, thank you @jmcarp!  I took a peak into the library you created and tested out the functionality again, all is still working very well with this.

PR #1569, opened 2016-03-21
Add manifest files for cf-ssh access

ccostino:
This changeset adds additional manifest files so that we are able to use the `cf-ssh` cli tool in order to connect directly to our servers.

**NOTE:**  This is just a stop-gap measure fow now, we will want to improve this in the near future, but we need to be able to access these machines now.  The current problem appears to be that you are unable to use a manifest that inherits from another one when using the `cf-ssh` cli-tool, so that will be a place to start investigating further.

**USAGE:**  Just to be sure things work as expected I make sure that CF is configured to access the same target space I want to use `cf-ssh` with:

`cf target -o fec -s <space>`

And then to use these with `cf-ssh`:

`cf-ssh -f <ssh manifest file for desired space>`

After a few moments of deploying an `api-ssh` app, you should be presented with a `tmate` window in your terminal for the remote app.

/cc @LindsayYoung

### DIFF
--- a/ssh_manifest_dev.yml
+++ b/ssh_manifest_dev.yml
@@ -1,0 +1,31 @@
+---
+path: .
+memory: 512M
+stack: cflinuxfs2
+domain: 18f.gov
+buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
+env:
+  NEW_RELIC_CONFIG_FILE: newrelic.ini
+  NEW_RELIC_LOG: stdout
+  WEB_CONCURRENCY: 4
+  NEW_RELIC_APP_NAME: OpenFEC API (development)
+  NEW_RELIC_ENV: development
+applications:
+- name: api
+- name: celery-beat
+  instances: 1
+  memory: 256M
+  no-route: true
+  command: celery beat --app webservices.tasks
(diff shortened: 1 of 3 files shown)


jmcarp:
Not sure this is the right direction. If you want to `cf-ssh` into an application, IMO the simplest thing to do is to use `cf create-app-manifest`:

```
cf create-app-manifest my-app ssh-manifest.yml
cf-ssh -f ssh-manifest.yml
```

See https://github.com/18F/fec-cms/pull/187 for a documented example.

ccostino:
@jmcarp:  By doing that will it account for the current target space you are in?  And is the idea that the created manifest file is just a temporary artifact for whatever task you're currently trying to accomplish?

ccostino:
I just gave those instructions a shot with the API too, I think you're right.  @LindsayYoung, can you see if the `create-app-manifest` instructions also work for you (just be sure to set your target space first)?  If so then that is the better way to go and I'll withdraw this PR and add this to my list of things to update in the README in https://github.com/18F/openFEC/issues/1559

LindsayYoung:
Thanks, I will give this a try later!

ccostino:
We just confirmed that this works for @LindsayYoung, so I'm closing this PR. :-)

PR #1570, opened 2016-03-22
Fixes docs.py typos

nbedi:
Quick typos I found while working with the API and reading docs.


LindsayYoung:
Thank you for taking the time to submit this PR!

PR #1592, opened 2016-04-01
adding simple search for legal documents

anthonygarvan:


webservices/rest.py
```diff
@@ -215,6 +216,8 @@ def add_aggregate_resource(api, view, schedule, label):
 
 api.add_resource(download.DownloadView, '/download/<path:path>/')
 
+api.add_resource(legal.UniversalSearch, '/legal/search/')
```
adborden: Curious about the name `UniversalSearch`, are you thinking this will encompass more things than just legal resources?

Only ask because I'm seeing similar themes come up in the UI/web-app side so it would be good to be on the same page as the API. I don't have an opinion myself, yet.

webservices/resources/legal.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@doc(
+    tags=['search'],
+    description=docs.LEGAL_SEARCH
+)
+
+class UniversalSearch(utils.Resource):
+    def __init__(self):
+        es_conn = env.get_credential('ES_CONN')
```
jmcarp: * The `get_credential` method takes an optional default argument, so you can write `env.get_credential('ES_CONN', 'http://localhost:9200')`
* I'm guessing this will use the cloud.gov elastic service in real life, so it might be worth writing the configuration for that in this patch. Example at https://github.com/18F/atf-eregs/blob/master/atf_eregs/settings/prod.py#L33-L40.
* AFAICT, this is going to make a new `ElasticSearch` client every time a request gets dispatched to this view. I'm guessing that's not what we want to do.

jmcarp:
Tests are failing because the swagger markup from the new endpoint isn't valid. Swagger validation errors are awful, and I didn't dig into the error that's showing up here, but I'm guessing the validator is complaining because we aren't defining responses for the new endpoint. We can add a response by using the `@marshal_with` decorator and passing a `Schema`, or by manually adding swagger json in `@doc`. Or, even simpler, we could skip adding the endpoint to the swagger docs for the moment.

webservices/rest.py
anthonygarvan: Hmm maybe could think of a better name. I mean universal in the context of legal documents, as in not AO-specific or reg-specific. Probably just `search` would be better for now.

webservices/resources/legal.py
anthonygarvan: - thanks!
- I added the elastic search to the credential service and it seemed to work fine with get_credential on dev. Is that not the right way to do it for openFEC?
- Yep, good catch.

ccostino:
Thanks, @anthonygarvan!

PR #1603, opened 2016-04-05
Hotfix/name back

LindsayYoung:
adding name=x back to be backwards compatible
improved docs slightly
closes 18F/openFEC/issues/1593


ccostino:
Thanks, @LindsayYoung, I'm glad this eventually worked out!

PR #1605, opened 2016-04-07
Update deployer creds

adborden:


.travis.yml
```diff
@@ -12,8 +12,8 @@ addons:
 
 env:
   global:
-    - secure: "cpOgenTtFAeKEdUBcynR1sZFsx7tSqKUytWAQ7oxDQgSbuHNhlazOfyQd9qytyKsc2jFZgfctGajAoZ/LvN71R21MxyQyFQDUg3RTsY9mNjg6GtcPyH+iUdIUxJkhEu/tdJqz5pzuz+hH8FGtTLP4A83N7En/apyAxQaJRL5A1Y="
-    - secure: "WXquMki0USRcQ8vVvwKOk3mcUmoMufMLdDfbiIR/IZ9D1JL2Qu7/431BG9UBLDqs8OZXQRQNFrGhZpz6roczbcePZz7Wl5cdj8yd48WunkRpZDG1fIepHqppb3ydi/UfGXBPA6CaMAmjR1ErBchDl88ykQZE0L6mAdx4iyflOAg="
+    - secure: "L0e1dTMLeUQGTrDfmWkTOUvg/zpQ1b8yidnZvI/nGVKW9OY1r4VqKc+tPLWtb9/e8D+jBUKmomjYR1TYL8xmm1DxWThx5AESr4FG6kndsUrDAxlG7PRZECumoG4FehsbPVKr5OP+ZC+GZY6V/6NdvijNKxvSae9zda3yCaGu7EE=" # FEC_CF_USERNAME
+    - secure: "qpKjTwlaNDdbFFO5oqDJ9PsVFsxowYbh0z4whQSjip7F2JWpUOT4ympPoeNikmVSevuxQkCCYsT8AnrSWyhR5I9Lg4xEoOezO+gRfY/sVawX3Obnwv46dUsvxaoStCTIwpmTxqyLZaRn7VkgPrnW+rEn9sIardgTm9RLuTa3JmA=" # FEC_CF_PASSWORD
```
ccostino: I hesitate including the comments a bit due to the nature of encrypting and setting these as secure environment variables in the first place (and the fact that order doesn't matter).  Do you mind taking them out?

adborden: I consider the variable names as [part of the code](https://github.com/18F/openFEC/blob/develop/tasks.py#L170). They're already in the deploy task and they show up in the travis logs. In theory, this should have been a one-line change but because I wasn't sure which variable is which, I had to replace both of them which is a bit tedious. In repos where there are many encrypted variables, this becomes untenable.

ccostino: Fair enough, I think I'm being overly cautious here and didn't consider that aspect of not knowing which is which.  Thanks!

ccostino:
Thanks for getting this squared away and updated, @adborden!

PR #1606, opened 2016-04-08
Parse ES url from VCAP_SERVICES

adborden:


ccostino:
Thanks for the catch, @adborden!

PR #1610, opened 2016-04-13
Run database tasks in dependency order.

jmcarp:
This is an implementation of #1515 that's simpler than the airflow version, and hopefully easier to review. This doesn't switch from views to tables, but that would be a simple, and independent, change to make later on. Switching from this to airflow later on would also be relatively straightforward.

Most of the diffs are just renaming and splitting up sql scripts--the actual code changes (and there's not much changing here) mostly live in `webservices/flow.py`.

ping @LindsayYoung @ccostino


ccostino:
Awesome, thank you @jmcarp!

LindsayYoung:
Thanks @jmcarp :+1:

webservices/flow.py
```diff
@@ -0,0 +1,58 @@
...
+
+here, _ = os.path.split(__file__)
+home = os.path.join(here, os.pardir)
+script_path = os.path.join(home, 'data', 'sql_updates')
+
+def get_graph():
+    """Build a `DiGraph` that captures dependencies between database migration
+    tasks. Each node represents a migration script, and each edge represents
+    a dependency between tasks. Tasks can be ordered using topological sort.
+    """
```
ccostino: @jmcarp: Just a quick point of order, if we add or modify the SQL scripts that need to be managed by this, we'll have to update the graph here as well, correct?  Just considering where we might want to make a note of that for other developers who come into the project in the future.

jmcarp: Only if the script has dependencies or if other scripts depend on it. Adding a script like the one that generates calendar dates, which doesn't depend on any other migrations, wouldn't require a change to this file.

ccostino: Ah right, good point for clarification.  Thanks!

LindsayYoung:
Sorry if this is a dumb question- How do you see the graph?

jmcarp:
The word graph gets a little confusing here--it could be the data structure (a directed acyclic graph) that represents task dependencies, or it could be a picture of that data. To see the data as a networkx object, you can run the `get_graph` function from `flow.py`. Pictures of graph data were generated by airflow in the larger PR that I sent a while back, but I haven't implemented them here. If that would be useful to see, I'll add a quick helper to output an image.

webservices/flow.py
LindsayYoung: I am confused, if we add a new script, even with no dependencies, it won't be included in the `update_schemas` command if it is not in this function, or am I missing something?

LindsayYoung:
I asked because I saw the graph at one point. I am thinking it is ok not to view the image, do you agree @ccostino? I think we are wanting to move to more formal migrations anyway. Thanks Josh!

webservices/flow.py
jmcarp: Lines 15-18 include all `*.sql` files in the `sql_updates` directory.

LindsayYoung: Ah, that makes sense, thanks!

ccostino:
Yes, the image of the graph was in the full Airflow PR - no need to recreate here!

PR #1611, opened 2016-04-14
adding feature space manifest

anthonygarvan:
This adds the manifest for the new `feature` space. I have already deployed this on the space, so I know it works :).

### DIFF
--- a/manifest_feature.yml
+++ b/manifest_feature.yml
@@ -1,0 +1,11 @@
+---
+inherit: manifest_base.yml
+host: fec-feature-api
+services:
+  - fec-creds-feature
+  - fec-redis-feature
+  - fec-s3-feature
+  - fec-search-feature
+env:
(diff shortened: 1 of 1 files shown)


ccostino:
Thanks, @anthonygarvan!  We might also want to consider adding support in the `./tasks.py` file used by `invoke`.  Perhaps we could tie it to branches that start with `feature` just like branches that start with `release` are tied to staging.  Any thoughts on this?

/cc @LindsayYoung and @adborden

LindsayYoung:
I like the idea of the feature space being purposefully deployed so it doesn't get clobbered if we are using it for demos or testing.

anthonygarvan:
Yeah I agree with @LindsayYoung, feature should not be automatically deployed.

ccostino:
Ah good point, and this gives us the ability to retain auto-deploys on dev so that it stays updated with everything merged to-date.  Merging!

PR #1615, opened 2016-04-15
Feature/load legal docs

anthonygarvan:
This commit adds the capability to load documents through the API. It does this by matching the api key against a set of comma-separated whitelisted tokens provided through a new `WRITE_AUTHORIZED_TOKENS` credential. I tested it in the feature environment.


jmcarp:
Just curious--do these documents come from the fec oracle db or someplace else? Who's going to be pushing these to the api?

webservices/resources/legal.py
```diff
@@ -9,8 +9,15 @@
 from webargs import fields
 from pyelasticsearch import ElasticSearch
 from webservices.env import env
+from flask import request
 
 
+if env.get_credential('WRITE_AUTHORIZED_TOKENS'):
```
jmcarp: How about `env.get_credential('WRITE_AUTHORIZED_TOKENS', '12345, ...')`? Also, do the default tokens have any meaning? Do we need them? Seems like it would be easy to forget to add this credential to a user-provided service and open up the write api to everybody who takes the time to look at the source.

webservices/resources/legal.py
```diff
@@ -23,3 +30,18 @@ def get(self, **kwargs):
...
+        data = request.get_json()
+        if data['api_key'] in write_authorized_tokens:
+            es.bulk((es.index_op(doc, id=doc['doc_id']) for doc in data['docs']),
+              index='docs',
+              doc_type=data['doc_type'])
+
+            es.refresh(index='docs')
+            return {'success': True}
+        else:
+            return {'success': False, 'message':
```
jmcarp: This looks to return a 200 even though it's an error.

webservices/resources/legal.py
```diff
@@ -9,8 +9,15 @@
 from webargs import fields
 from pyelasticsearch import ElasticSearch
 from webservices.env import env
+from flask import request
 
 
+if env.get_credential('WRITE_AUTHORIZED_TOKENS'):
+    write_cred = env.get_credential('WRITE_AUTHORIZED_TOKENS') 
+else:
+    write_cred = '12345, abc, def'
```
ccostino: Per our code review session:  Let's change this to an environment variable only so we don't run the risk of introducing an issue with bad credentials letting folks in.

webservices/rest.py
```diff
@@ -217,6 +217,7 @@ def add_aggregate_resource(api, view, schedule, label):
 api.add_resource(download.DownloadView, '/download/<path:path>/')
 
 api.add_resource(legal.Search, '/legal/search/')
+api.add_resource(legal.Load, '/legal/load/')
```
ccostino: Per our code review session:  Let's change this URL to be something different since this is a write.

anthonygarvan:
@jmcarp - they are in Oracle. I currently have a script that goes through CSVs dumped out by Oracle, formats them, and ships them to this api. We will probably be doing this ourselves for a while but long term FEC should be able to do it.

jmcarp:
@anthonygarvan: is this something that you can push to the celery cron worker instead, for consistency with the other periodic data updates? Then you wouldn't need to worry about a new endpoint or auth tokens.

webservices/resources/legal.py
```diff
@@ -1,25 +1,17 @@
...
 
 from webservices import args
 from webservices import docs
 from webservices import utils
 from webservices.utils import use_kwargs
 from webargs import fields
-from pyelasticsearch import ElasticSearch
-from webservices.env import env
-
+from flask import request
```
jmcarp: Some of these imports are unused. Do you have flake8 running in your editor?

webservices/resources/load.py
```diff
@@ -0,0 +1,30 @@
+from webservices.env import env
+from flask import request
+from webservices import utils
+import os
+
+es = utils.get_elasticsearch_connection()
+
+if env.get_credential('WRITE_AUTHORIZED_TOKENS'):
```
jmcarp: The `get_credential` method checks the environment, so this block can be replaced with `write_cred = env.get_credential('WRITE_AUTHORIZED_TOKENS', '')`.

https://github.com/jmcarp/py-cfenv/blob/master/cfenv/__init__.py#L62-L70

webservices/resources/load.py
```diff
@@ -0,0 +1,30 @@
...
+        data = request.get_json()
+        if data['api_key'] in write_authorized_tokens:
+            es.bulk((es.index_op(doc, id=doc['doc_id']) for doc in data['docs']),
+              index='docs',
+              doc_type=data['doc_type'])
+
+            es.refresh(index='docs')
+            return {'success': True}
+        else:
+            return {'success': False, 'message':
```
jmcarp: Sorry for repetition, but my comment got lost earlier. This appears to return a 200 for an auth error--I'm guessing we want a different status code here.

anthonygarvan: good catch, thanks!

webservices/resources/legal.py
```diff
@@ -1,21 +1,8 @@
...
-from pyelasticsearch import ElasticSearch
-from webservices.env import env
-
 
-es_conn = env.get_service(label='elasticsearch-swarm-1.7.1')
-if es_conn:
-    es = ElasticSearch(es_conn.get_url(url='uri'))
-else:
-    es = ElasticSearch('http://localhost:9200')
+es = utils.get_elasticsearch_connection()
```
adborden: Nice

PR #1623, opened 2016-04-16
Feature/search enhancements

anthonygarvan:
Various improvements to AO search (detailed in 18F/fec-eregs#48).

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -126,7 +126,9 @@ names = {
 }
 
 query = {
-    'q': fields.Str(required=True, description='Text to search legal documents for')
+    'q': fields.Str(required=True, description='Text to search legal documents for'),
+    'from_hit': fields.Int(required=False, description='Get results starting from this index.'),
+    'hits_returned': fields.Int(required=False, description='Number of results to return (max 10).')
 }
 
 candidate_detail = {
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -126,7 +126,7 @@ names = {
 }
 
 query = {
-    'q': fields.Str(required=True, description='Text to search legal documents for'),
+    'q': fields.Str(required=True, description='Text to search legal documents for.'),
     'from_hit': fields.Int(required=False, description='Get results starting from this index.'),
     'hits_returned': fields.Int(required=False, description='Number of results to return (max 10).')
 }
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -41,7 +41,6 @@ class Search(utils.Resource):
                          es_from=from_hit)['hits']['hits']
 
         for hit in hits:
-            print(hit)
             hit["pdf_url"] = 'http://saos.fec.gov/aodocs/%s.pdf' % hit['_source']['AO_No']
 
         return {'results': hits, 'count': count['count']}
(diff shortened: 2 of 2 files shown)


webservices/args.py
```diff
@@ -126,7 +126,9 @@ def make_seek_args(field=fields.Int, description=None):
 }
 
 query = {
-    'q': fields.Str(required=True, description='Text to search legal documents for')
+    'q': fields.Str(required=True, description='Text to search legal documents for'),
+    'from_hit': fields.Int(required=False, description='Get results starting from this index.'),
+    'hits_returned': fields.Int(required=False, description='Number of results to return (max 10).')
```
adborden: Can we up the max, 10 seems pretty low for search. I would prefer a max of 200 with a default of 20, and include those in the docs.

webservices/args.py
```diff
@@ -126,7 +126,9 @@ def make_seek_args(field=fields.Int, description=None):
 }
 
 query = {
-    'q': fields.Str(required=True, description='Text to search legal documents for')
+    'q': fields.Str(required=True, description='Text to search legal documents for'),
```
adborden: Missing the period for consistency, although there is a lot of inconsistencies in the api docs https://api.open.fec.gov/developers/

+ @emileighoutlaw in case you want to review any of this copy.

adborden:
@jmcarp had suggested we move the transform logic from the web app into the API, and I totally agree. Would now be a good time to do that?

https://github.com/18F/openFEC-web-app/pull/1175#issuecomment-208591719

webservices/args.py
emileighoutlaw: Thanks for pinging me! Taking a look now. :)

emileighoutlaw: Your changes look 👍, @adborden. Thanks for adding me to this review

webservices/resources/legal.py
```diff
@@ -19,7 +19,30 @@
...
+        count = es.count(query)
+
+        query["highlight"] = {"fields": {"text": {}}}
+        query["_source"] = {"exclude": "text"}
+        hits_returned = min([200, hits_returned])
+        hits = es.search(query, index='docs', size=hits_returned,
+                         es_from=from_hit)['hits']['hits']
+
+        for hit in hits:
+            print(hit)
```
adborden: Stray `print`?

anthonygarvan: doh!

edmullen:
- I’m seeing a number of scenarios where the AO name search doesn’t return the exact match at the top. This one doesn't include 2014-02 at all: 
   - https://fec-proxy.18f.gov/data/legal/search/?search_type=all&search=2014-02
   - https://fec-web-adborden.18f.gov/legal/search/?search_type=all&search=2014-02
- I'm also wondering about the labelling. The names display as "2014-02", not "AO 2014-02". I haven't seem FEC refer to specific AOs without the "AO" part elsewhere. At least on the results display, I think we should include the AO.

adborden:
> - I'm also wondering about the labelling. The names display as "2014-02", not "AO 2014-02". I haven't seem FEC refer to specific AOs without the "AO" part elsewhere. At least on the results display, I think we should include the AO.

This one's one me, so I'll add that.

--
Aaron D Borden
Human and Hacker

PR #1628, opened 2016-04-27
Feature/house, senate, presidental, pac and party table switch

LindsayYoung:
That was easier than I thought it would be so far!

There are some minor changes we will want to communicate to users:
 - This adds `receipt_date`
 - Keeps the `is_amended` filter and adds that as a boolean to the output 
 - this removes `load_date` - the date it was loaded in the DB, not a great date but did not have the receipt date before  
 - this removes `expire date` we don't need that for finding out if a record is amended anymore


LindsayYoung:
(need to add new test data for tests to pass)

LindsayYoung:
missing election type on IE
need new tests subset
want to confirm the variables with slightly different names

LindsayYoung:
FYI, this PR is almost done I to update the integration tests, but this is put on hold because calendar work is taking precedence.

LindsayYoung:
Really close on this, still working on updating the tests.

jontours:
I made some changes and currently the only test that is failing is 'test_update_schemas' in the integration tests.  That test basically tests that all the data-backed models actually have one or more rows in them.  After a quick perusal of the subset test database, some of the tables that are serving as roots for a lot of the models are empty, e.g. fec_vsum_f3p.  As soon as we get some data for these tables (they do have data in the dev database) then they will pass.

jontours:
Looks good!  I can merge it if you want.  I'm curious did you use rdmbs-subsetter to generate the new subset?  And do you know why those tables were empty in the old subset?

LindsayYoung:
I copied data from dev. I just had to make sure there was enough overlapping data to make the totals.  

I think they were not populated in the test dataset because they had not been loaded to our databases yet.

noahmanger:
Is this ready to be merged?

LindsayYoung:
Yes, but we don't want it to be a part of this release

LindsayYoung:
test dataset is the conflict :/

jontours:
eek, I was hoping it wouldn't be something like that.

LindsayYoung:
I might still have some sample data in a tmp folder, I am hoping it shouldn't be too bad.

Issue #1629, opened 2016-04-28
fix nplusone issues


LindsayYoung:
We are doing eager loading, but getting nplus one errors. It is going to be commented out on the money flags pr #1548 , but we would like to confirm there aren't n+1 queries happening and reenable the tool if we can.

LindsayYoung:
Josh found them and fixed them.

ccostino:
Closing this as it was finished!

PR #1631, opened 2016-04-29
Eager-load flags on candidate views and restore nplusone.

jmcarp:
We introduced a handful of lazy-loaded relationships in #1548. For
example, loading N records via `CandidateView` currently runs one query
(to fetch candidate detail rows) followed by N additional queries (one
to fetch the flags row for each candidate row). This patch uses eager
loading for all candidate views that expose related attributes via
flags, and restores `nplusone` to prevent future performance
regressions.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -5,7 +5,7 @@ import unittest
 import subprocess
 
 from webtest import TestApp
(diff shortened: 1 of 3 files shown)


LindsayYoung:
Thank you Josh!

LindsayYoung:
closes #1629

PR #1644, opened 2016-05-07
Updated subset.dump file

ccostino:
This changeset updates our test subset of data with a sampling of data taken around midnight of May 7, 2016. This subset was pulled from the `dev` database.

There have been numerous updates to the data set since the last subset was pulled so we felt it was time to refresh it.

/cc @LindsayYoung


ccostino:
Uh oh, it looks like this data dump requires some reconciliation with our tests.

LindsayYoung:
I am a little surprised that adding tables would break the tests. I will take a look. I needed to update the integration tests on #1628 anyway.

ccostino:
Okay great, thank you @LindsayYoung!

ccostino:
After chatting with @LindsayYoung, I'm going to update this subset again with a new data dump since we just got a new table.

LindsayYoung:
I am going to close this because the test dataset in #1628 is now the most complete

PR #1645, opened 2016-05-07
Build assets on CI.

jmcarp:
* Switch to vetted python buildpack
* Install swagger-ui on travis and include in deploy
* Remove `node_modules` from `.cfignore`

Part of https://github.com/18F/openFEC/issues/1643.

Note--we could simplify this further and `git clone` swagger-ui instead of `npm install`-ing it, since I'm guessing this repo isn't going to accumulate more npm requirements. I don't have a preference about this--just thought I'd mention the possibility.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,2 +1,0 @@
-https://github.com/cloudfoundry/python-buildpack#v1.5.1
-https://github.com/cloudfoundry/nodejs-buildpack.git#v1.5.0
--- a/.cfignore
+++ b/.cfignore
@@ -1,7 +1,5 @@
 cache
-node_modules
 
-data/bin/set_env_vars.real.sh
 .DS_Store
 
 # Byte-compiled / optimized / DLL files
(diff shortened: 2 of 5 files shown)


ccostino:
@jmcarp:  Are you ready for this to be reviewed?  Also, regarding the `git clone` of swagger-ui as opposed to the `npm install` of it, if you feel it would help simplify things more than I think it makes sense to change as well if it's not a heavy lift.

jmcarp:
Yep, these PRs are ready to test out. Once these are in good shape, we should probably make the same change on openfec-web-app--if people on the team are busy, I can do that.

ccostino:
Okay great, thank you!  We're just about finished getting the next release prepped for deployment, which should be happening tonight.  I imagine one of us will have a chance to take a look if you wouldn't mind helping review after?

ccostino:
@jmcarp:  Sorry for the delay, but I just got around to testing this - all seems fine with a manual deploy (aside from accounting for other recent changes we had to make that aren't accounted for in this branch yet).  I don't have any questions around this and the changes are straightforward, if you're set then we're ready to merge!

jmcarp:
If it's working, then merge at will.

PR #1647, opened 2016-05-09
Updated test data subset instructions

ccostino:
This changeset adds more details to the test data subset instructions found in the README.

The existing instructions were not completely clear in what information you needed to provide in order for the commands to work, or that you needed to create a new database for yourself that was completely empty.

/cc @emileighoutlaw and @LindsayYoung

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,40 +191,24 @@ py.test
 #### The test data subset
 If you add new tables to the data, you'll need to generate a new subset for testing. We use this nifty subsetting tool: [rdbms-subsetter](https://github.com/18F/rdbms-subsetter).
 
-To build a new test subset, first create a new database for yourself:
+To build a new test subset, first delete and recreate the test database:
 
 ```
-createdb cfdm_new_subset
+dropdb cfdm_test
+createdb cfdm_test
 ```
 
 Now use the `build_test` invoke task to populate the new database with the new subset:
 
 ```
-invoke build_test <source> <dest>
+invoke build_test <source> postgresql://:@/cfdm_test
 ```
 
-where both `source` and `dest` are valid PostgreSQL connection strings. If you're following these instructions, the command will look like the following:
-
-```
-invoke build_test <source> postgresql://:@/cfdm_new_subset
-```
-
+where `source` is a valid PostgreSQL connection string.
 To update the version-controlled test subset after rebuilding, run:
 
 ```
-invoke dump <source> data/subset.dump
-```
-
-where `source` is the database containing the newly created test subset. Again, if you're following these instructions then the command will look like this:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,22 +191,41 @@ py.test
 #### The test data subset
 If you add new tables to the data, you'll need to generate a new subset for testing. We use this nifty subsetting tool: [rdbms-subsetter](https://github.com/18F/rdbms-subsetter).
 
-To build a new test subset, use the `build_test` invoke task:
+To build a new test subset, first create a new database for yourself:
+
+```
+createdb cfdm_new_subset
```
LindsayYoung: I made them before by overwriting the current dump. Any reason we would not want to do that?

ccostino: Did you overwrite the current dump directly when doing this or your `cfdm_test` database?  We certainly could, I happened to run into an issue for some reason when doing so because some data/structures already existed (I can't remember the specific error), and overwriting the previous dump itself didn't work for me due to the need to supply a Postgres connection string as opposed to a file path.

If there's a way to do this without having to go through some of these extra steps, I'd much rather do that! :-)

LindsayYoung: What about deleting the current test data and then running the invoke command?

ccostino: Ah okay, which `invoke` command are you referring to?  The `build_test` one or the `dump` one?  If you're referring to the `build_test` one, it appears that needs a connection string to a database for both the source and destination and not a file path.

Could you please paste in the last one you ran (or tried to run) when you have a moment so I can see?  I think I'm either missing something or just generally confused and I want to make sure I'm on the same page as you.

emileighoutlaw: Are we still waiting on something here @ccostino and @LindsayYoung ? I'm ready to merge whenever y'all are!

ccostino: Still waiting for the moment, @emileighoutlaw, but thanks for being at the ready! :-)

LindsayYoung: I was talking about the way it is currently described:
```
invoke build_test <source> <dest>
invoke dump <source> data/subset.dump
```
It works fine the first time and if you running the build test again, you can delete and recreate the db first.

ccostino: Ah okay, you're referring to just using the existing `cfdm_test` database that we start out with when setting things up locally, right?  In that case you still need to drop the database and start again, but I can update the instructions to reflect that instead of just creating an entirely new database.

LindsayYoung: Yea, I think that is simpler than renaming.

LindsayYoung:
Thanks Carlo and Emileigh!

ccostino:
Thank you for the feedback and clarification! :-)

PR #1657, opened 2016-05-12
Feature/add regulations

anthonygarvan:
This changes the search format to delivery suitable JSON to the web app, and adds a filter by type. Should be merged simultaneously with 18F/openFEC-web-app#1212


ccostino:
Thanks again @anthonygarvan, great to run through a live pair review!

PR #1664, opened 2016-05-19
Feature/reporting periods states

### LINKED ISSUE #1661: state filtering for reporting periods
Add state filtering for reporting periods

Part of #1619

LindsayYoung:
Creating the IE and CC periods from the new `trc_election_dates` table, this has the advantage of making them filterable by state. 

*Needs new test data with the new table for integration tests to pass.

The microcopy changes are already merged here. 

closes  #1661


LindsayYoung:
So I don't forget what I am doing on Monday here are the calendar issues:

Still working on:
- Why does the api IEs look right and the web look wrong?
- May 2 looks like there are duplicates, VA caucus or convention
- May 2 VA districts not in description, I think it is a data issue but need to confirm
    - add more districts to elections?
- May 18 missing election description  (not on FEC's calendar)
- make the IE and EC entries start with state if 3 or less
- Needs new test data set!

LindsayYoung:
OK, this looks like it is working. This will need a lot of FEC review and a new test data set.

LindsayYoung:
noticed a small thing with repeated states, will look into that later

LindsayYoung:
I think this is as good as I can make this for now. The 24 hour reports have summary and description switched and I have no idea why. 

We will need a lot of FEC review on this by info division.

LindsayYoung:
Tests are passing locally.

Remaining issue: I think this is as good as I can make this for now. The 24 hour reports have summary and description switched and I have no idea why.

Also needs tests ;)

LindsayYoung:
OK, I got the same errors when I updated my requirements

ccostino:
@jontours and I reviewed this with @LindsayYoung this morning, all looks fine, thank you for the walkthrough!

PR #1671, opened 2016-05-20
Use better party discription on calendar

LindsayYoung:
Cleaner party names, but uppercased. Also switches one instance off of a warehouse table.

@emileighoutlaw can you confirm uppercased parties are better that title cased parties that have junk in them sometimes.


emileighoutlaw:
Yeah — in general, I'm not stoked about uppercase, but in this case I totally agree it's the right call.

ccostino:
This looks fine, thanks @LindsayYoung!

PR #1683, opened 2016-05-23
Fixed database string, changes dashes to underscores

jontours:
Although this is a very minor fix.  I'm wondering how people are getting tests to pass locally? (guessing that the sql string is set via environment variable)  And finally, are the tests being actively used and maintained?  Sorry for all the questions.  This pull request was more for me to get comfortable with the git flow workflow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
     - secure: "qpKjTwlaNDdbFFO5oqDJ9PsVFsxowYbh0z4whQSjip7F2JWpUOT4ympPoeNikmVSevuxQkCCYsT8AnrSWyhR5I9Lg4xEoOezO+gRfY/sVawX3Obnwv46dUsvxaoStCTIwpmTxqyLZaRn7VkgPrnW+rEn9sIardgTm9RLuTa3JmA=" # FEC_CF_PASSWORD
 
 before_script:
-  - psql -c 'create database "cfdm-unit-test";' -U postgres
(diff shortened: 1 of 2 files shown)


jontours:
Ah, well I see the travis build failed with that string.  So maybe setting the postgres string via command line is the way to go.

ccostino:
Huh, that is interesting!  Did the tests not work for you locally without setting the `SQLA_CONN` env variable?   If not then this seems an appropriate change, and I suspect that the issue may also be tied to a Travis CI config setting.  In looking at `.travis.yml`, (I see the database is named differently)[https://github.com/18F/openFEC/blob/develop/.travis.yml#L19].

If you were to change `.travis.yml` to match this change as a part of this PR things should work again, and then we'd have consistent names within the project configuration and the documentation.  Thanks, @jontours!

jontours:
Hey Carlo,

Thanks for reviewing the merge request and providing help.  Pending approval, what is the next step?  Does the reviewer then issue a git flow finish?  

To answer your question though, yes after changing the database name in code (in common.py) the tests then started passing.  So with this change the configuration should match the readme.

LindsayYoung:
You can merge PRs to develop via GitHub. 

Create your branches with git flow so they are properly named to the git flow convention. The git flow process of finishing and releasing is mostly for release branches, and deployment to staging and production.

ccostino:
Thanks, @jontours and @LindsayYoung!

(we had also caught up in Slack some more about team norms and process regarding who merges what and when).

PR #1686, opened 2016-05-24
Add title case to party

LindsayYoung:
This was part of the next PR but there isn't a reason for it to wait. 
I ran this on staging if you want to take a look.
The other more complicates state filtering  PR is on dev (as of 05/24/2016).


emileighoutlaw:
Looks good to me. Merging now.

PR #1689, opened 2016-05-25
Changing the 5k flag to has_raised_funds

LindsayYoung:
We are going to have to run more sql updates for this.

### DIFF
--- a/data/sql_updates/candidate_aggregates.sql
+++ b/data/sql_updates/candidate_aggregates.sql
@@ -31,7 +31,7 @@ cycle_totals as (
         false as is_election,
         sum(totals.receipts) as receipts,
         sum(totals.disbursements) as disbursements,
-        sum(receipts) >= 5000 as five_thousand_flag,
+        sum(receipts) > 0 as has_raised_funds,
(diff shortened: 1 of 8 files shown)


LindsayYoung:
Running the schema updates as the tests run.

data/sql_updates/candidate_aggregates.sql
```diff
@@ -107,3 +107,5 @@ create index on ofec_candidate_totals_mv_tmp (cycle);
 create index on ofec_candidate_totals_mv_tmp (is_election);
 create index on ofec_candidate_totals_mv_tmp (receipts);
 create index on ofec_candidate_totals_mv_tmp (disbursements);
+create index on ofec_candidate_totals_mv_tmp (has_raised_funds);
+create index on ofec_candidate_totals_mv_tmp (federal_funds_flag);
```
ccostino: Ah good catch!

LindsayYoung:
The flags work for me except `/committee/{committee_id}/candidates/` I don't think they should be there anyway.

ccostino:
Okay you're good to go with this, thanks @LindsayYoung!

PR #1692, opened 2016-05-26
Feature/totals flag extension

jontours:
Added the has_raised_funds and federal_funds_flag to the "/candidate/totals" endpoint.  At first, I thought my code was wrong because there were some candidates being shown in the results with negative receipts for certain cycles.  However after taking a look at the candidate_flags table, they indeed were listed as 'true' for has_raised, but, the summation over all cycles put them over 0 dollars.

Which is how the filter works for the currently implemented "/candidates" endpoint as well.  I don't see this as an issue other than if a user takes a look at certain cycles in isolation and sees a negative or zero amount.  From an outside user I don't know if the language conveys that the candidate has raised more than zero dollars for all time.  

With that being said I did play around with the api and verify that the users are being properly filtered, but not an official test or anything, although that may be something worth writing.  Also, for the candidates/{candidateType} endpoints on the web app there are no selections for these filters, so if it's to be presented those will have to be added.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -468,6 +468,8 @@ candidate_totals = {
     'max_cash_on_hand_end_period': Currency(description='Maximum cash on hand'),
     'min_debts_owed_by_committee': Currency(description='Minimum debt'),
     'max_debts_owed_by_committee': Currency(description='Maximum debt'),
+    'federal_funds_flag': fields.Bool(description=docs.FEDERAL_FUNDS_FLAG),
+    'has_raised_funds': fields.Bool(description=docs.HAS_RAISED_FUNDS),
 }
 
 totals_committee_aggregate = {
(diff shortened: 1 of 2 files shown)


LindsayYoung:
I can see that "has_raised_funds" might not imply ever, but the documentation has it as, "A boolean that describes if a candidate's committee has ever received any receipts for their campaign for this particular office. (Candidates have separate candidate IDs for each office.)"  

I was depending on the documentation to explain the finer points. I am not a fan of misleading variable names, so chime in again if you don't think it is enough. 

I think you bring up a good point and we will want to be explicit in the web app as well.

LindsayYoung:
Looks good and it works. I can merge if you are happy with it.

Also, as you noted in your comment we have two ways of doing filtering. The older models did not use the helper methods. It is a little bit of technical debt that is hard to prioritize.

jontours:
Yeah you can go ahead and merge it.  If I feel it needs to be refactored to adhere to that filtering pattern, I can open another pull request.  Thanks @LindsayYoung!

LindsayYoung:
Thank you!

PR #1693, opened 2016-05-27
fix report year definition

LindsayYoung:
Incorrect definition was overwriting correct definition.


ccostino:
Thanks for the catch, @LindsayYoung!

PR #1694, opened 2016-06-02
Feature/move from warehouse tables

jontours:
This is a pretty beefy pull request.  But I'll attempt at explaining the changes.

a) Changed the aggregates (prepare and update) sql to reference the new fec_vsum_sched_a table
b) Migrated test data from sched_a to fec_vsum_sched_a table in postgres using a sql insert statement (local test DB)
c) Regenerated the test data dump after this migration
d) Used this new subset to expose all breaking changes in the code using the integration tests
e) Modified the code to get all of the tests to pass

This pull request may be problematic if some fields are no longer being exposed with the new table, but from my inspection I couldn't find any.  And if anything the new table adds new columns that we aren't currently exposing (e.g. all the "desc" columns).  

Also the old sched_a table had a column named "sched_a_sk".  This was also used by us for the primary key on the "ofec_sched_a" table.  The new table uses sub_id as a primary key, therefore, any point in our code where "sched_a_sk" is referenced has been changed to "sub_id".  I think we'll need to discuss further if this key is indeed a proper unique constraint.

And finally, it's probably best to wait to merge these commits, and the upcoming sched_b table, after @ccostino finishes the partitioning task.  As this migration may affect that, we can discuss further at the sprint review.  However in the meantime if any one wants to pull this branch and check out the changes I made, and suggest feedback, we can commit changes.

### DIFF
--- a/webservices/common/models/itemized.py
+++ b/webservices/common/models/itemized.py
@@ -9,19 +9,27 @@ from .base import db
 class BaseItemized(db.Model):
     __abstract__ = True
 
+    """These weren't exposed in the API previously, although they were in the old tables
+    gonna leave uncommented until this is clarified
+
+    candidate_office = db.Column('cand_office', db.String)
+    candidate_office_description = db.Column('cand_office_desc', db.String)
+    candidate_office_district = db.Column('cand_office_district', db.String)
+    """
     committee_id = db.Column('cmte_id', db.String, doc=docs.COMMITTEE_ID)
     committee = utils.related_committee_history('committee_id', cycle_label='report_year')
     report_year = db.Column('rpt_yr', db.Integer, doc=docs.REPORT_YEAR)
     report_type = db.Column('rpt_tp', db.String, doc=docs.REPORT_TYPE)
     entity_type = db.Column('entity_tp', db.String)
-    image_number = db.Column('image_num', db.String, doc=docs.IMAGE_NUMBER)
-    memo_code = db.Column('memo_cd', db.String)
-    memo_text = db.Column(db.String)
+    entity_type_desc = db.column('entity_tp_desc', db.String)
+    election_type = db.Column('election_tp', db.String, doc=docs.ELECTION_TYPE)
+    election_type_full = db.Column('election_tp_desc', db.String, doc=docs.ELECTION_TYPE)
 
+    fec_election_type_desc = db.column('fec_election_tp_desc', db.String)
+    fec_election_year = db.column('fec_election_yr', db.String)
+    image_number = db.Column('image_num', db.String, doc=docs.IMAGE_NUMBER)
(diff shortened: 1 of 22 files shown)


ccostino:
Just to recap some of the changes we reviewed this morning:

* Merge in latest `develop` branch updates.
* Remove `swagger-tools` from `packages.json` and install it globally (`npm install -g swagger-tools` - we'll update the README to account for this step).
* Double check on some of the fields that changed in the models (follow up with Paul and Jeff).
* Remove commented out lines referencing things we no longer need.
* Work in updates for Schedule B since they are roughly the same changes.

Thanks a ton for this work, @jontours!

jontours:
Generated two new feature branches from this one, that way a and b aren't held up by e update.
The features are sched_a_b_transition and sched_e_transition.

webservices/common/models/itemized.py
```diff
@@ -9,25 +9,29 @@
...
+    fec_election_year = db.column('fec_election_yr', db.String)
     image_number = db.Column('image_num', db.String, doc=docs.IMAGE_NUMBER)
-    memo_code = db.Column('memo_cd', db.String)
-    memo_text = db.Column(db.String)
-    filing_type = db.Column(db.String)
     filing_form = db.Column(db.String)
     link_id = db.Column(db.Integer)
-    sub_id = db.Column(db.Integer)
-    original_sub_id = db.Column('orig_sub_id', db.Integer)
-    amendment_indicator = db.Column('amndt_ind', db.String)
```
LindsayYoung: I think amendment indicator might be acton_code in the new data.

webservices/common/models/itemized.py
```diff
@@ -9,25 +9,29 @@
...
+    fec_election_type_desc = db.column('fec_election_tp_desc', db.String)
+    fec_election_year = db.column('fec_election_yr', db.String)
     image_number = db.Column('image_num', db.String, doc=docs.IMAGE_NUMBER)
-    memo_code = db.Column('memo_cd', db.String)
-    memo_text = db.Column(db.String)
-    filing_type = db.Column(db.String)
     filing_form = db.Column(db.String)
     link_id = db.Column(db.Integer)
-    sub_id = db.Column(db.Integer)
-    original_sub_id = db.Column('orig_sub_id', db.Integer)
```
LindsayYoung: This might be back reference id

LindsayYoung: Ignore that

webservices/common/models/itemized.py
jontours: Yes think I meant to change that. I'll assign 'amendment_indicator' to 'action_cd'

LindsayYoung:
This was incorporated in another PR

PR #1713, opened 2016-06-09
[WIP] Enable support to retrieve itemized data limited by cycle

### LINKED ISSUE #1636: Make partitioned schedule A and B tables work for downloads
This is follow up work to be done after https://github.com/18F/openFEC/issues/1429 is merged.  The following things need to happen:

* Enable (whitelist) the Schedule A and Schedule B models in the download tasks.
* Coordinate with @noahmanger on https://github.com/18F/openFEC-web-app/pull/1242 for restricting downloads to 2 year cycles

ccostino:
Closes #1636 

This changeset enables support to correctly retrieve schedule A and schedule B data from the API for searching, filtering, and downloading based on the partitioning work we just completed.  It exposes the `transaction_year` column we've added, makes it required and supplies a default value of the end year in the current latest cycle if it's not present.

These changes will need to be communicated out to API users as a part of merging this or soon after.

https://github.com/18F/openFEC-web-app/pull/1242 is also dependent on this PR in order to properly work.

This is ready for review, but we are going to spend a bit of time testing and benchmarking these changes during this sprint to ensure we're seeing the performance gains we are expecting to see.  Any follow up changes will come in subsequent PRs.

/cc @LindsayYoung, @jontours, @noahmanger


tests/common.py
```diff
@@ -14,7 +14,7 @@
 
 TEST_CONN = os.getenv('SQLA_TEST_CONN', 'postgresql:///cfdm_unit_test')
 
-rest.app.config['NPLUSONE_RAISE'] = True
+rest.app.config['NPLUSONE_RAISE'] = False
```
jmcarp: Did you want to check this in?

ccostino: Ugh, no, I didn't.  Reverting!

ccostino: Actually before I revert, the reason why I had shut this off in the first place is because I seeing a bunch of exceptions raised while testing.  It seems all of the errors are being raised around `BaseConcreteCandidate.flags`.  I thought we addressed these already?

ccostino:
@jmcarp: Looks like it's just an issue I'm having locally with those errors being raised, which is definitely curious.  I'll see if I can uncover what's going on.

tests/common.py
LindsayYoung: Yea, Josh helped me with that before. 

Also we are not getting nplusone errors on develop, where we are checking for it. 
https://github.com/18F/openFEC/blob/develop/tests/common.py#L17

ccostino: That's what I thought - I'll have to investigate and see what's going on locally for me.

PR #1723, opened 2016-06-10
small definition tweak

LindsayYoung:
I wanted to emphasize that it was a two-year period, so we don't assume they already know our two year cycle construct. Also I think it is more helpful to mention the field names by what they see in the API rather than the underlying SQL tables that users don't see. 

I ran it by @emileighoutlaw too get her opinion too. 

Thanks!

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -248,13 +248,13 @@ years so this can differ from underlying form's receipt date.
 '''
 
 TRANSACTION_YEAR = '''
-This is a derived field for storing the year a transaction took place in the
-Itemized Schedule A and Schedule B tables. In cases where we have the date of
-the transaction (contb_receipt_dt in Schedule A, disb_dt in Schedule B), we
-make use of the year in that date and adjust it to fit within the appropriate
-cycle that it belongs to. If we do not have the date, we fall back to using the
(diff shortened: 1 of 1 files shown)


webservices/docs.py
```diff
@@ -319,6 +319,10 @@
 
 SCHEDULE_A = SCHEDULE_A_TAG + '''
 
+The data is divided in two-year periods, called `transaction_year` this is derrived from
```
ccostino: Accounting for a couple of grammatical things here, should this instead say:

"The data is divided in two-year periods, called `transaction_year`, which is derived from"...

webservices/docs.py
```diff
@@ -352,6 +356,11 @@
 
 SCHEDULE_B = SCHEDULE_B_TAG + '''
 
+The data is divided in two-year periods, called `transaction_year` this is derrived from
```
ccostino: Same question as above regarding grammar/spelling.

ccostino:
This looks good to me except for the couple of lines I noted.  @emileighoutlaw, when all is to your liking, please feel free to merge! :-)

ccostino:
Thank you @LindsayYoung and @emileighoutlaw!

PR #1724, opened 2016-06-10
Fix for flask import warning statements

jontours:
Fixed this problem using the "import x as z" statement, so as not to break all the references to restful and cors.  Warnings can be verified as gone by starting up the api server.


LindsayYoung:
👍

jontours:
thanks! @LindsayYoung

PR #1728, opened 2016-06-14
Update CloudFoundry login setup

ccostino:
This changeset accounts for recent changes with our CloudFoundry setup and only logs you in if necessary now.

h/t @adborden for the same work in the web app.

/cc @LindsayYoung, @anthonygarvan, @adborden, @jontours, @noahmanger

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ export SQLA_CONN=<psql:address-to-box>
    
    *Note: If you modify your local environment to run these applications at a different address, be sure to update these environment variables to match.* 
 
-3. Run: 
+3. If you do not login to CloudFoundry with SSO (single sign-on), run:
 
    ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -151,7 +151,7 @@ export SQLA_CONN=<psql:address-to-box>
 
    Create these account credentials to gain full access to the application. You can set them to any username and password of your choosing.  
    
-   *Note: 18F and FEC team members will have additional environment variables to set up. Please reach out to a team member for detailed information.*
+   *Note: 18F team members should not set these environment variables. 18F and FEC team members will also have additional environment variables to set up. Please reach out to a team member for detailed information.*
```
LindsayYoung: Maybe, we want to be a bit more explicit "18F team members should not set these environment variables locally. These are stored in a cf service."

PR #1731, opened 2016-06-15
2nd try at new travis and local log in

LindsayYoung:
The login stays the same for Travis, we use the new format. Correct me if I am wrong, but the new way doesn't work on Travis and the old way does.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -181,14 +181,15 @@ def deploy(space=None, branch=None, yes=False):
         return
 
     # Log in if necessary
-    args = (
-        ('--a', 'https://api.cloud.gov'),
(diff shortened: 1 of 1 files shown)


ccostino:
I think you're right, @LindsayYoung, this'll let Travis still login correctly to cloud.gov.  Locally we just need to make sure we login for ourselves first before doing anything and not set those environment variables.

adborden:
Hmm, we want to be careful here, if you don't have the env variables set (when deploying locally), you're never targeting an org/space so you might actually deploy to the wrong place.

We should add `cf target -o fec -s %(space)` after the login, but before doing anything else.

PR #1751, opened 2016-06-21
Feature/rad

### LINKED ISSUE #1740: RAD analyst endpoint
So that API users can find the RAD analyst associated with a particular committee, create an endpoint that returns RAD analysts by committee ID.

**Criteria for completion:**
- [ ] API client can send a `committee_id` to the endpoint and receive an analyst name in response

Requested week of the 27th or the week of the 4th

LindsayYoung:
Adds an endpoint for RAD analyst lookup
Includes tests and updated subset.dump

closes #1740

@emileighoutlaw there is some copy in the docs file that describes RAD, I got it from the current site. Maybe you know something better.

Also noticed that we were not populating committee type options for the totals endpoint not super related but, I fixed it in webservices/resources/totals.py


tests/test_integration.py
```diff
@@ -77,6 +77,7 @@ def setUpClass(cls):
 
     def test_update_schemas(self):
         for model in db.Model._decl_class_registry.values():
+            print(model)
```
ccostino: Is this an extraneous `print` statement from debugging?

ccostino:
Thanks, @LindsayYoung!  This looks good to me and appears to be working just fine - I only noticed one thing and it's minor. :-)

emileighoutlaw:
I pushed a small content edit to this, but I think it looks great, Lindsay! 👍 from me

tests/test_integration.py
LindsayYoung: Happy to take it out if it is too much of a style violation. 

We might want to keep it, because then the test tells you what model it fails at so you know what you need to add data to. I have added that print statement in at least 4 times now and it only prints if the test fails.

jontours: Oh yeah, I forgot it doesn't print to console unless it fails, which kind of had me scratching my head.  I'd advocate for keeping it then.

ccostino:
Thanks, @LindsayYoung!

PR #1752, opened 2016-06-21
Address itemized downloads feedback

ccostino:
This changeset addresses a few things that were flagged as needing to be changed based on the recent feedback we received.  These changes include the following:

* Enabling the display of the Memo Code field (memo_cd in the database) in the CSV output (this has the side effect of displaying it in the API as well)
* Displaying the time zone label with the timestamp given in the manifest (the timestamp is processed as UTC).

These modifications are a part of the full list in https://github.com/18F/openFEC/issues/1748

/cc @LindsayYoung, @jontours, @noahmanger

### DIFF
--- a/webservices/schemas.py
+++ b/webservices/schemas.py
@@ -250,7 +250,6 @@ ScheduleASchema = make_schema(
     },
     options={
         'exclude': (
-            'memo_code',
             'contributor_name_text',
             'contributor_employer_text',
             'contributor_occupation_text',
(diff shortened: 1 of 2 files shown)


webservices/schemas.py
```diff
@@ -250,7 +250,6 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
     },
     options={
         'exclude': (
-            'memo_code',
```
LindsayYoung: It was included there because it was in there under a different name, I think we will want tok phase out that name too. I can add that to the email.

ccostino: Thanks, @LindsayYoung!

Issue #1760, opened 2016-06-22
Schedule A and B updates are failing


ccostino:
With the table partitioning now in place for the Schedule A and B data and cloud.gov stability issues fixed, we're now started to see some issues with the updates to these tables with the queue tables we have to support them.

In short, the queue tables are not accounting for the new `two_year_transaction_period` value that we calculate and store, and the logic behind updating the child tables still references the `rpt_yr` directly.  We need to add support for storing the `two_year_transaction_period` value to the queue tables and base the checks on those, and then update the logic in the child update process to check against that date instead of `rpt_yr` directly.

PR #1763, opened 2016-06-23
Fix Schedule A and B updates

ccostino:
Addresses #1760 

This changeset addresses an issue with failing updates for the Schedule A and B tables.  The update process did not take into account the table partitioning and was still basing the checks off of the report year (`rpt_yr`) instead of the new calculated value (`two_year_transaction_period`).  In order to fix this, I had to do the following:

* Add the `two_year_transaction_period` column to both sets of queue tables for Schedule A and B.
* Remove the `two_year_transaction_period` column from the column factory in the select statement generation for processing updates.
* Add full support for the `timestamp` column in the Schedule B queue tables.

I do not believe this is entirely straightforward so there is a comment in the code that refreshes the child tables during an update process to help explain the reasoning behind what is happening.

h/t to @LindsayYoung for working through the approach with me!

/cc @LindsayYoung, @jontours


LindsayYoung:
This looks good :+1:

We run this on dev and staging and then I will merge when the data is ready for this code.

PR #1769, opened 2016-06-24
Reverting #1645 this puts us back on the old multi buildpack 

LindsayYoung:
that expires in January, but the documentaion works now


jontours:
One thing to note with this, this weekend when @ccostino and I kicked off partitioning on production, the buildpack of "python_buildpack" was needed for the pip installs to run correctly.  So just keep in mind when starting a one-off task.  Other than that I'll go ahead an merge.

LindsayYoung:
Yes, this is a temporary fix we should upgrade build packs soon: https://github.com/18F/openFEC/issues/1770

PR #1771, opened 2016-06-27
Upgrade to latest invoke.

jmcarp:
Low priority, but I noticed that invoke deprecated its old task definition api and now requires tasks to take a context argument. This patch pins to the latest invoke and uses the new task api. It's also fine to stick with an older invoke, but IMO it's worth upgrading, if only so that our tasks look like the tasks in the invoke docs. Plus miscellaneous bug fixes invoke has made between 0.10 and 0.13.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 cfenv==0.4.0
-invoke==0.10.1
+invoke==0.13.0
 psycopg2==2.6
 Flask==0.11
 Flask-Cors==2.1.0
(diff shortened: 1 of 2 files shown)


LindsayYoung:
Thanks!

ccostino:
Thank you, @jmcarp!

ccostino:
Thanks again, @jmcarp!  I just had a chance to look closely and this is fine.  I'll merge and we'll work in some PRs to update the web app and CMS too. :-)

PR #1772, opened 2016-06-27
Modify pac/party committee for filtering, sorting [WIP]

jontours:
Open to different ways of doing this, i.e. let me know if there is something glaringly wrong with this strategy.  

That strategy consists of adding materialized views to support pac/party information, and new models and filters to support them.  At first adding MV(s) for these seemed heavy handed, but maybe not if more and more super pacs come into existence.

Work left to be done is perhaps adding a new view class for this functionality, although it may not be needed.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -231,7 +231,7 @@ reports = {
 
 totals = {
     'cycle': fields.List(fields.Int, description=docs.RECORD_CYCLE),
-    'type': fields.List(fields.Str, description=docs.COMMITTEE_TYPE)
+    'pac_party_type': fields.List(fields.Str, description=docs.PAC_PARTY_TYPE)
 }
 
 
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -231,6 +231,7 @@ reports = {
 
 totals = {
     'cycle': fields.List(fields.Int, description=docs.RECORD_CYCLE),
+    'type': fields.List(fields.Str, description=docs.COMMITTEE_TYPE)
 }
 
 
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -231,7 +231,6 @@ reports = {
 
 totals = {
     'cycle': fields.List(fields.Int, description=docs.RECORD_CYCLE),
-    'pac_party_type': fields.List(fields.Str, description=docs.PAC_PARTY_TYPE)
 }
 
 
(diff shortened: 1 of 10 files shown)


noahmanger:
I was just testing this out locally and I'm seeing `sqlalchemy.exc.NoSuchTableError: ofec_totals_parties_mv` at `/totals/pac` and `/totals/party` . If it's too in progress for me to try it out, that's cool. But if this is just something simple I need to do to get it running, then I could get the frontend in shape.

jontours:
Hey @noahmanger, if you run "python manage.py update_all" that should create the tables.  Make sure your first do a pg_restore though, and pulled the last commit, as it has a modified pg_dump subset of data.

noahmanger:
What more is needed here? Seems to be working for me with the test data.

jontours:
I would say it's done at this point, pending review from other developers.  I would like to get the test passing before merging, but that shouldn't be a problem.

webservices/args.py
```diff
@@ -231,6 +231,7 @@ def make_seek_args(field=fields.Int, description=None):
 
 totals = {
     'cycle': fields.List(fields.Int, description=docs.RECORD_CYCLE),
+    'type': fields.List(fields.Str, description=docs.COMMITTEE_TYPE)
```
LindsayYoung: I think we don't need this because committee type is passed in as part of the url here https://github.com/18F/openFEC/pull/1772/files#diff-36f6a3b25691ff374f1941da4493c150L27

jontours:
I agree with that.  But this was to add filtering for whatever model is being hit for that request.  For example: at the pac endpoint if ?type='O' is passed in as a query parameter, only committees of that type are returned.  It's nothing that was required but more a nice to have, I can take it out if we feel it's not necessary.

webservices/args.py
jontours: I agree with that. But this was to add filtering for whatever model is being hit for that request. For example: at the pac endpoint if ?type='O' is passed in as a query parameter, only committees of that type are returned. It's nothing that was required but more a nice to have, I can take it out if we feel it's not necessary.

LindsayYoung: I did not realize this was more granular. Let's keep it

noahmanger:
Woo! Have the necessary updates run on dev yet to test the frontend against?

jontours:
Yup!  The tables are on dev so if you point to the dev api you should be able to test the endpoint.

Issue #1716, opened 2016-06-09
Review and add redundancy to data feed


LindsayYoung:
It seems like there are some transactions that are slipping though the cracks on the large, schedule tables.

The current issues will be solved with partitioning since we are rebuilding those tables but the data integrity is extremely important so: 

    1) we should identify the source of the missing data
    2) we should add some extra processes to check and fix missing data

We might want to do 2 first to help us identify when the problems are occurring and make sure the data is always accounted for.

noahmanger:
Another instance of missing data: https://github.com/18F/openFEC/issues/1720

LindsayYoung:
I think that person downloaded filings and then did not find filings again. This is data in a FEC table that is not in a ofec_ table.

noahmanger:
It's unclear what they did or didn't try to do. I think they're point about downloading receipts was potentially mistaken. All I know is that that committee has reported itemized receipts but they're not showing up.

LindsayYoung:
OK, thank for clarifying I just read their part before

PaulClark2:
I agree with Noah's comment. However, I can confirm there is a difference however between beta and the presidential map for contributions from New Mexico to Mr. Trump. The map and the DB (PG sched_a and FECP & DW sched_a) have 53 Trump contributions from New Mexico `select * from sched_a s where s.cmte_id = 'C00580100' and s.contbr_st = 'NM';`  Beta only has 27 contributions from New Mexico.

noahmanger:
Criteria for completion: 
- [ ] We have a backup system for transfer of records
- [ ] We have a system for alerting us when our counts get out of line

noahmanger:
So is the only step on this to add the alerting system? (Re criteria for completion above).

LindsayYoung:
Correct. We still want to use counts to flag any discrepancies going forward. I have not started that yet

ccostino:
Given that we just switched tables off of the data warehouse versions, these alerts should be based on count discrepancies between the `fec-vsum_sched_*` tables and our `ofec_sched_*` versions, correct?  Or is there something else we should be basing the counts off of instead?

Also, should these alerts take place along with our refresh task, be an entirely separate check, or hang off of another event like the nightly updates?

LindsayYoung:
[...]

PR #1773, opened 2016-06-27
Feature/data load fallback

LindsayYoung:
This will rebuild the data on Saturdays to ensure there is no missing data. 

This is the first part of #1716

### DIFF
--- a/webservices/tasks/__init__.py
+++ b/webservices/tasks/__init__.py
@@ -29,8 +29,13 @@ app.conf.update(
     CELERYBEAT_SCHEDULE={
         'refresh': {
             'task': 'webservices.tasks.refresh.refresh',
-            'schedule': crontab(minute=0, hour=9),
+            'schedule': crontab(minute=0, hour=9, day_of_week='sun-fri'),
         },
+        'refresh_all': {
+            'task': 'webservices.tasks.refresh.refresh_all',
+            'schedule': crontab(minute='*', hour='*', day_of_week='mon'),
+            # 'schedule': crontab(minute=0, hour=9, day_of_week='sat'),
+        }
     }
 )
 
--- a/webservices/tasks/__init__.py
+++ b/webservices/tasks/__init__.py
@@ -31,7 +31,7 @@ app.conf.update(
             'task': 'webservices.tasks.refresh.refresh',
             'schedule': crontab(minute=0, hour=9, day_of_week='sun-fri'),
         },
-        'refresh_all': {
+        'refresh_and_rebuild': {
             'task': 'webservices.tasks.refresh.refresh_and_rebuild',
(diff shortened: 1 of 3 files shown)


LindsayYoung:
In watching the partitioning today, I noticed that the aggregates went down during partitioning. So with this PR aggregates will be unavailable for some time on Saturdays. I think it is still worth it, I just wanted to be upfront.

ccostino:
We just discovered _why_ the aggregates went down during partitioning.  First, in performing this process we simply ran `update_all` to account for all changes that needed to happen recently.  The updates that handle the aggregates simply drop the tables and recreate them as opposed to creating temporary tables first and then dropping and renaming things at the end.

Second, there was a missing dependency in recent releases with the new way we're managing which order to run the update scripts in.  This is fixed in the upcoming release, but something we'll need to double check as well to be sure.

LindsayYoung:
for reference, the temp table issue is here https://github.com/18F/openFEC/issues/1775

webservices/tasks/__init__.py
```diff
@@ -29,8 +29,12 @@ def redis_url():
     CELERYBEAT_SCHEDULE={
         'refresh': {
             'task': 'webservices.tasks.refresh.refresh',
-            'schedule': crontab(minute=0, hour=9),
+            'schedule': crontab(minute=0, hour=9, day_of_week='sun-fri'),
         },
+        'refresh_all': {
+            'task': 'webservices.tasks.refresh.refresh_all',
+            'schedule': crontab(minute=0, hour=9, day_of_week='sat'),
```
ccostino: Just to make sure I have my times right, this will be starting at 9 AM UTC (http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#time-zones), correct?

LindsayYoung:
@ccostino good point about the partitioning taking more than a day. Added the rebuild and refresh commands so this is ready for review again.

webservices/tasks/__init__.py
```diff
@@ -29,8 +29,12 @@ def redis_url():
     CELERYBEAT_SCHEDULE={
         'refresh': {
             'task': 'webservices.tasks.refresh.refresh',
-            'schedule': crontab(minute=0, hour=9),
+            'schedule': crontab(minute=0, hour=9, day_of_week='sun-fri'),
         },
+        'refresh_all': {
```
ccostino: Whoops, should probably change this to `refresh_and_rebuild` too - when I saw it come up in the logs I was like hey wait a minute...!  Oh. :-)

LindsayYoung: Good catch!

ccostino:
Thank you, @LindsayYoung!

PR #1774, opened 2016-06-27
Expose memo code

LindsayYoung:
Adding this early, people might want to start using that now since we are depreciating the other name.


jontours:
Looks good.

PR #1777, opened 2016-06-29
Update schemas before schedules and aggregates 

LindsayYoung:
This reduces the channce that a dependency will not be built in advance

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -197,13 +197,13 @@ def update_all(processes=1):
     load_pacronyms()
     load_nicknames()
     load_election_dates()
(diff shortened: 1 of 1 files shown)


LindsayYoung:
This is making me think, yet again, we are overdue for adding a migration framework. (To be prioritized in a future sprint) @ccostino @jontours 

maybe: https://flask-migrate.readthedocs.io/en/latest/

LindsayYoung:
I am going to close this for now

ccostino:
Yes, I think that makes sense @LindsayYoung!

PR #1780, opened 2016-06-30
Schedule A and B migration

jontours:
This is the actual pull request to be reviewed for schedule A and B migration.  We had some issues with Git and merging in develop, as it had been about 5 weeks of commit history to consider.  All other pull requests for this feature can be ignored.


data/functions/utils.sql
```diff
@@ -5,6 +5,15 @@ begin
 end
 $$ language plpgsql immutable;
 
+create or replace function get_transaction_year(transaction_date timestamp, report_year numeric)
```
ccostino: Interesting, can you overload functions in Postgres like with C++/C# where you provide multiple method declarations that accept different arguments?

webservices/common/models/itemized.py
```diff
@@ -109,20 +114,32 @@ class ScheduleB(BaseItemized):
...
-    record_number = db.Column('record_num', db.Integer)
-    report_primary_general = db.Column('rpt_pgi', db.String)
-    receipt_date = db.Column('receipt_dt', db.Date)
+    original_sub_id = db.Column('orig_sub_id', db.Integer)
+    fec_election_type_desc = db.Column('fec_election_tp_desc', db.String)
+    fec_election_year = db.Column('fec_election_tp_year', db.String)
+    #These are deprecated
+    #record_number = db.Column('record_num', db.Integer)
+    #report_primary_general = db.Column('rpt_pgi', db.String)
+    #receipt_date = db.Column('receipt_dt', db.Date)
```
ccostino: Shouldn't we just remove these if they're deprecated?

webservices/partition/sched_a.py
```diff
@@ -57,7 +57,7 @@ def index_factory(cls, child):
             sa.Index(None, c.rpt_yr),
             sa.Index(None, c.entity_tp),
             sa.Index(None, c.image_num),
-            sa.Index(None, c.sched_a_sk),
+            sa.Index(None, c.sub_id),
```
ccostino: D'oh!  Apparently I missed that one, sorry.  We could also reference it like `c[cls.primary]` like I had elsewhere so we only ever change the primary key label in one spot.

webservices/partition/sched_b.py
```diff
@@ -51,7 +51,7 @@ def index_factory(cls, child):
         return [
             sa.Index(None, c.rpt_yr),
             sa.Index(None, c.image_num),
-            sa.Index(None, c.sched_b_sk),
+            sa.Index(None, c.sub_id),
```
ccostino: Same note as the `sched_a` version:  `c[cls.primary]` could be used.

data/functions/utils.sql
ccostino: Indeed it can:  https://www.postgresql.org/docs/current/static/xfunc-overload.html

ccostino:
Awesome, thank you so much @jontours!  I had just a few notes and I want to make sure @LindsayYoung has a chance to look too, but I think we're in good shape with this.

webservices/partition/sched_a.py
jontours: Yeah that would probably make sense, I can make that change.

webservices/common/models/itemized.py
jontours: I definitely think they should be removed, I left them more as a reminder to @LindsayYoung as two what will be deprecated, in correspondence with the api release documentation.  Which could also be done with a diff on the table schemas.

data/functions/utils.sql
jontours: I didn't look it up before trying.  Sometimes it's better to be lucky than good!

webservices/common/models/itemized.py
LindsayYoung: These were good fields :(

tests/test_integration.py
```diff
@@ -24,18 +24,17 @@ def make_factory():
     class SchedAFactory(SQLAlchemyModelFactory):
         class Meta:
             sqlalchemy_session = db.session
-            model = automap.classes.sched_a
-        load_date = datetime.datetime.utcnow()
```
LindsayYoung: we can use pg_date for this

webservices/common/models/itemized.py
```diff
@@ -65,23 +62,24 @@ class ScheduleA(BaseItemized):
...
-    election_type_full = db.Column('election_tp_desc', db.String, doc=docs.ELECTION_TYPE)
     back_reference_transaction_id = db.Column('back_ref_tran_id', db.String)
     back_reference_schedule_name = db.Column('back_ref_sched_nm', db.String)
-    national_committee_nonfederal_account = db.Column('national_cmte_nonfed_acct', db.String)
-    record_number = db.Column('record_num', db.Integer)
-    report_primary_general = db.Column('rpt_pgi', db.String)
-    form_type_full = db.Column('form_tp_cd', db.String, doc=docs.FORM_TYPE)
-    receipt_date = db.Column('receipt_dt', db.Date, doc=docs.RECEIPT_DATE)
     increased_limit = db.Column(db.String)
-    load_date = db.Column(db.DateTime, doc=docs.LOAD_DATE)
```
LindsayYoung: the pg_date comment should have gone here

webservices/common/models/itemized.py
```diff
@@ -109,20 +112,28 @@ class ScheduleB(BaseItemized):
     recipient_zip = db.Column(db.String)
     disbursement_type = db.Column('disb_tp', db.String)
     disbursement_description = db.Column('disb_desc', db.String)
-    disbursement_date = db.Column('disb_dt', db.Date)
+    disbursement_date = db.Column('disb_dt', db.TIMESTAMP(timezone=False))
```
LindsayYoung: We just want the date, not a timestamp, right?

webservices/partition/sched_b.py
```diff
@@ -51,7 +51,7 @@ def index_factory(cls, child):
         return [
             sa.Index(None, c.rpt_yr),
```
LindsayYoung: We don't need an index on this anymore, correct?

webservices/common/models/itemized.py
jontours: The FEC is storing it as a Timestamp though, I could double check.  I originally tried assigning it to a Date in the api code, and was getting an error.  I can verify though.

webservices/partition/sched_b.py
jontours: I can't speak for the need to index on this column, it did carry over to the new vsum tables, if it's not being sorted on this column or being referenced I don't see any need to index it.

LindsayYoung:
OK, going to review this now

webservices/partition/base.py
```diff
@@ -105,7 +105,7 @@ def create_constraints(cls, child, cycle):
         start, stop = cycle - 1, cycle
         cmds = [
             'alter table {child} alter column {primary} set not null',
-            'alter table {child} alter column load_date set not null',
+            'alter table {child} alter column filing_form set not null',
```
LindsayYoung: Why are we setting filing_form to not null here?

jontours: It's a non-null table constraint configured by the `fec_vsum_` tables, if you take that line out the tests fail as the child tabes aren't created with that matching constraint.

LindsayYoung: OK, so we are just mirroring their structure

jontours: Yup, I'm guessing the FEC must have a good reason to enforce that constraint, interestingly it's the only column with that constraint, besides the PK.

LindsayYoung:
Thank you @jontours

jontours:
No problem, thanks to @LindsayYoung and @ccostino for giving this a good look over.

PR #1799, opened 2016-07-05
Adding general reporting page link for reporting periods

LindsayYoung:
We can add more specific links later.


ccostino:
Thanks, @LindsayYoung!

PR #1801, opened 2016-07-05
Fixing args so that the datatype for image numbers are strings

### LINKED ISSUE #1782: image numbers
They need to stay as numbers in dev or it will break code

LindsayYoung:
This will fix the detail panel on web
fix #1782


noahmanger:
Closing this since it was merged into the release.

PR #1803, opened 2016-07-05
Fix for candidate totals bug

jontours:
This fix is by no means pretty, if anything don't merge it.  If worse comes to worse it can be merged to buy us more time, but it works.  Figuring out the real data issue should be the better fix.

### DIFF
--- a/data/sql_updates/totals_house_senate.sql
+++ b/data/sql_updates/totals_house_senate.sql
@@ -48,7 +48,7 @@ select
     max(last.rpt_yr) as last_report_year
 from
     fec_vsum_f3 hs
-    left join last using (cmte_id, election_cycle)
+    inner join last using (cmte_id, election_cycle)
 where
     hs.most_recent_filing_flag like 'Y'
     and hs.election_cycle >= :START_YEAR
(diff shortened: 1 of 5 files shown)


noahmanger:
Once this is merged, we should to merge this on the web app https://github.com/18F/openFEC-web-app/pull/1332

ccostino:
@noahmanger, @LindsayYoung, what do you think based on @jontours note?  Do you think we still have time to investigate the potential root cause?

noahmanger:
Because the release introduces a very breaking change to production, we need it fixed by the time of the release. I'd rather not push back the release for this, so I think my vote is merge in the patch. But I'll let y'all decide in the AM.

LindsayYoung:
OK, I have looked at this and I think it is a good fix. We can change it later if we have a better idea on how to handle cases like this.

I think this happened when we went form a SQL join to a sqlalchmy join for the candidate financial information. The former function seemed not to be able to find the column and figure out if it was null with the new join. This fixes the problem by explicitly looking for the column in the model. 

Works well locally. Checked a few other endpoints and they did not seem impacted.

Thank you @jontours for jumping on this and fixing it!

PR #1805, opened 2016-07-08
Feature/build swagger on ci

### LINKED ISSUE #1770: switch off multi-buildpack
Make sure docs still work.

adborden:
This get's us off of the multibuildpack which is deprecated. Thanks to @jmcarp for getting most of it done. The last piece is getting swagger-ui out of node_modules so we're not shipping any dev dependencies to cloud.gov. This reverts the last fix which saved us from a broken swagger docs page and adds a quick build step to pull the built swagger-ui assets out of `node_modules`.

I tested this via [travis](https://travis-ci.org/18F/openFEC/builds/143403613) on the feature space, so hopefully there's no more surprises.

Fixes #1770

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "dependencies": {
      "swagger-ui": "git+https://github.com/18F/swagger-ui.git"
   },
+  "scripts": {
+    "build": "mkdir -p static/swagger-ui && cp -R node_modules/swagger-ui/dist static/swagger-ui/"
+  },
   "engines": {
     "node": "0.12.7",
     "npm": "2.11.3"
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,2 +1,0 @@
-https://github.com/cloudfoundry/python-buildpack#v1.5.1
-https://github.com/cloudfoundry/nodejs-buildpack.git#v1.5.0
--- a/.cfignore
+++ b/.cfignore
@@ -1,5 +1,4 @@
 cache
-node_modules
 
 .DS_Store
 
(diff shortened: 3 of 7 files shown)


ccostino:
Thank you, @adborden!  I'll take a closer look for review - does this still require we install `swagger-tools` for ourselves locally per the current README instructions (`npm install -g swagger-tools`), or is it possible we can place that in the `package.json` requirements as well to keep dependencies local to the app?

package.json
```diff
@@ -2,6 +2,9 @@
   "dependencies": {
      "swagger-ui": "git+https://github.com/18F/swagger-ui.git"
   },
+  "scripts": {
+    "build": "mkdir -p static/swagger-ui && cp -R node_modules/swagger-ui/dist static/swagger-ui/"
+  },
   "engines": {
     "node": "0.12.7",
```
ccostino: We should really update this...  Separate PR though to test.

ccostino:
Thanks again, @adborden, we're going to merge this and can address additional improvements in a subsequent PR(s) if need be!

adborden:
> Thank you, @adborden! I'll take a closer look for review - does this still require we install swagger-tools for ourselves locally per the current README instructions (npm install -g swagger-tools), or is it possible we can place that in the package.json requirements as well to keep dependencies local to the app?

From slack conversations `swagger-tools` might still be required for the integration tests. There's no specific reference to it in the code but perhaps a binary installed by the package. It's also possible that the integrations are using [flask-apispec](https://github.com/jmcarp/flask-apispec) for the swagger tests, in which case we can drop the swagger-tools dependency. We'll have to dig in a bit more.

If we move the dependency to the `package.json` we'll have to adjust the `PATH` so the integration tests pick up the right bin path.

PR #1807, opened 2016-07-11
Adjust aggregate table processing to make use of temporary tables

### LINKED ISSUE #1775: make aggragate tables as _tmp
so that we don't have table downtime on refresh

ccostino:
Fixes #1775

This changeset adjusts the processing for the aggregate tables to make use of temporary tables first. Prior to this, any updates to the aggregates in the database would cause the site to not yield the data that we expect to see.  We would like to achieve as little downtime as possible and the rest of our database processing makes use of creating temporary tables and views to accomplish this, so we are making use of them here as well.

For the initial approach I opted to do the renaming inline with the table processing.  Most of our materialized views are renamed separately at the end of processing, but if we want to follow those steps we'll need to move the creation of all of the triggers for each of these tables.

There is one instance of a table that has dependent objects, the `ofec_sched_a_aggregate_size` table and its corresponding `ofec_sched_a_aggregate_size_merged_mv` materialized view.  In order to address the downtime and renaming dance I modified the table reconstruction to just rename the existing table and shift the dropping of it to when the materialized view itself is reconstructed.  It's an added step in the materialized view recreation that deals with a different relation, but I've noted this with comments.

/cc @LindsayYoung, @jontours

### DIFF
--- a/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
+++ b/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
@@ -11,8 +11,8 @@ end
 $$ language plpgsql;
 
 -- Create initial aggregate
-drop table if exists ofec_sched_a_aggregate_size cascade;
-create table ofec_sched_a_aggregate_size as
+drop table if exists ofec_sched_a_aggregate_size_tmp cascade;
+create table ofec_sched_a_aggregate_size_tmp as
 select
     cmte_id,
     rpt_yr + rpt_yr % 2 as cycle,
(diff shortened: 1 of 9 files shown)


data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
```diff
@@ -27,11 +27,15 @@ group by cmte_id, cycle, size
...
-create index on ofec_sched_a_aggregate_size (total);
-create index on ofec_sched_a_aggregate_size (count);
+create index on ofec_sched_a_aggregate_size_tmp (cmte_id);
+create index on ofec_sched_a_aggregate_size_tmp (cycle);
+create index on ofec_sched_a_aggregate_size_tmp (size);
+create index on ofec_sched_a_aggregate_size_tmp (total);
+create index on ofec_sched_a_aggregate_size_tmp (count);
+
+-- Remove previous aggregate and rename new aggregate
+drop table if exists ofec_sched_a_aggregate_size cascade;
```
LindsayYoung: I think this will also drop the schedule A size merged view, until we re-create it. Maybe there is a better way or place to rename and drop the old table.

ccostino: You're right, it will. :-(  I ran into this when I initially made the change because I didn't have the `cascade`.  This dependency does make it a bit difficult to account for but I'm not sure of a good way at the moment to improve it given the order of the other steps.

LindsayYoung: Not sure it is a great idea, but what if we add a table rename step like we have for materialized views for a quick switch at the end?

PR #1808, opened 2016-07-14
Feature/reorder-columns

noahmanger:
I offered to take this task off @LindsayYoung 's plate, so here it is: my first API PR!

This re-orders the properties of the models in order to have the columns of the downloads be ordered in a more logical way. 

Over the course of doing this, I realized it's impossible to really get it feeling 100% right, since there will always be a lot of empty columns for certain rows and there will always be differences of opinion about the best order, but nonetheless, I tried to add some rhyme and reason to the order. 

Here's the order for itemized A, B and E (with questions & explanation):
- **Committee name & ID**: Currently only committee IDs are showing up. Is there a reason committee names aren't? It looks like they're in the model.
- **Report info**: Basic info about the report the transaction was filed on. AFAICT, because this is in the base model, it needs to come first, though there's certain columns that are showing up out of order (like `pdf_url` which shows up near the end). Is there a way to control this? I think ideally it would all be near the end.
- **Contributor (sched A) / Recipient (sched B) / Payee (sched E) info**: All pertinent info related to the other party of the transaction.
- **Primary transaction info:** The date, amount, and any description and memo data for the transaction
- **Transaction meta info:** This is the stuff that seemed to be more secondary info that wasn't crucial to understanding the transaction

In general, I also organized like properties to always be by like (e.g. `sub_id` next to `original_sub_id`). This was the case for most things, but in some places I shuffled things around.

I also tweaked the candidates and committee models to be organized in a more natural way as well. I found filings and the costs models to not really need anything. 

Last, I'm finding there's certain columns throughout that just aren't showing up and others that are still showing up out of order. I can go through and document which ones aren't, but wanted to see if that's expected first.

Resolves https://github.com/18F/openFEC/issues/1581

### DIFF
--- a/webservices/common/models/itemized.py
+++ b/webservices/common/models/itemized.py
@@ -59,6 +59,8 @@ class ScheduleA(BaseItemized):
     contributor_occupation = db.Column('contbr_occupation', db.String, doc=docs.CONTRIBUTOR_OCCUPATION)
     contributor_occupation_text = db.Column(TSVECTOR)
     contributor_id = db.Column('clean_contbr_id', db.String, doc=docs.CONTRIBUTOR_ID)
+    contributor_aggregate_ytd = db.Column('contb_aggregate_ytd', db.Numeric(30, 2))
+    is_individual = db.Column(db.Boolean, index=True)
 
     # Primary transaction info
     receipt_type = db.Column('receipt_tp', db.String)
@@ -82,7 +84,6 @@ class ScheduleA(BaseItemized):
     amendment_indicator_desc = db.Column('action_cd_desc', db.String)
     schedule_type = db.Column('schedule_type', db.String)
     schedule_type_full = db.Column('schedule_type_desc', db.String)
-    is_individual = db.Column(db.Boolean, index=True)
     increased_limit = db.Column(db.String)
     load_date = db.Column('pg_date', db.DateTime)
     transaction_id = db.Column('tran_id', db.Integer)
--- a/webservices/common/models/itemized.py
+++ b/webservices/common/models/itemized.py
@@ -9,8 +9,8 @@ from .base import db
 class BaseItemized(db.Model):
     __abstract__ = True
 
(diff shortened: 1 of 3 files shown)


noahmanger:
Aaand definitely gonna need a hand with that failing test.

LindsayYoung:
Thank you again and congrats on your first API contribution. I am happy to give a hand with the tests.

webservices/common/models/itemized.py
```diff
@@ -47,7 +43,7 @@ class ScheduleA(BaseItemized):
         )'''
     )
     contributor_name = db.Column('contbr_nm', db.String, doc=docs.CONTRIBUTOR_NAME)
-    contributor_prefix = db.Column('contbr_prefix', db.String)
+    contributor_name_text = db.Column(TSVECTOR)
```
LindsayYoung: The items that were at the bottom in the auxiliary fields section before are things that we use internally but is not exposed to the user. The TSVECTOR fields are a kind of index for searching, they break up text into pieces and weight them. These fields return text matches quickly. The output looks like `'barber':1A 'joe':3A 'rick':2A`

webservices/common/models/itemized.py
```diff
@@ -59,43 +55,49 @@ class ScheduleA(BaseItemized):
     contributor_state = db.Column('contbr_st', db.String, doc=docs.CONTRIBUTOR_STATE)
     contributor_zip = db.Column('contbr_zip', db.String, doc=docs.CONTRIBUTOR_ZIP)
     contributor_employer = db.Column('contbr_employer', db.String, doc=docs.CONTRIBUTOR_EMPLOYER)
+    contributor_employer_text = db.Column(TSVECTOR)
     contributor_occupation = db.Column('contbr_occupation', db.String, doc=docs.CONTRIBUTOR_OCCUPATION)
-    contributor_aggregate_ytd = db.Column('contb_aggregate_ytd', db.Numeric(30, 2))
```
LindsayYoung: This was removed, I'll add it back

PR #1810, opened 2016-07-14
prioritizing final opinion document category

anthonygarvan:
This commit prioritizes Final Opinions in the search results for advisory opinions. It does not exclude other document types.


ccostino:
Thanks, @anthonygarvan!

PR #1812, opened 2016-07-14
Feature/election comms table migration

jontours:
This migrates to the new form `fec_vsum_76` for communication costs.  All tests are passing and functionally it should be good to go.  The only column that is non-existent in this table is party affiliation, will that be a problem.  If so I'm sure there is a fairly easy way to join on the committee table to get that information.


noahmanger:
Nice. Is this ready for review to make it into release?

jontours:
Yes, it is ready to be reviewed.

webservices/common/models/costs.py
```diff
@@ -6,7 +6,9 @@
 class CommunicationCost(db.Model):
     __tablename__ = 'ofec_communication_cost_mv'
 
-    idx = db.Column(db.Integer, primary_key=True)
+    idx = db.Column(db.Integer)
```
LindsayYoung: we can get rid of this, now we have sub_id

webservices/common/models/costs.py
```diff
@@ -34,6 +36,47 @@ class CommunicationCost(db.Model):
...
+    #form_type_code_full = db.Column('form_tp_desc', db.String)
+    #This seems to be deprecated
+    #schedule_type_code = db.Column('form_tp', db.String, index=True)
+    tran_id = db.Column(db.String)
+    file_number = db.Column('file_num', db.Integer)
+    image_number = db.Column('image_num', db.String, index=True)
+
+    #deprecated as per usual
+    #report_year = db.Column('rpt_yr', db.Integer, index=True)
+    #pdf_url = db.Column(db.String)
```
LindsayYoung: We still want to make the pdf url. We construct it using the beginning image number. Did we take that away other places? I would want to add it back

data/sql_updates/communication_cost.sql
```diff
@@ -3,26 +3,30 @@ create materialized view ofec_communication_cost_mv_tmp as
...
-create index on ofec_communication_cost_mv_tmp (transaction_dt);
-create index on ofec_communication_cost_mv_tmp (transaction_amt);
+create index on ofec_communication_cost_mv_tmp (s_o_cand_office_st);
+create index on ofec_communication_cost_mv_tmp (s_o_cand_office_district);
+create index on ofec_communication_cost_mv_tmp (s_o_cand_office);
+create index on ofec_communication_cost_mv_tmp (s_o_ind);
+--create index on ofec_communication_cost_mv_tmp (cand_pty_affiliation);
+--create index on ofec_communication_cost_mv_tmp (transaction_dt);
+--create index on ofec_communication_cost_mv_tmp (transaction_amt);
+create index on ofec_communication_cost_mv_tmp (communication_dt);
```
LindsayYoung: maybe we should re-name this `transaction_dt` to keep it the same

data/sql_updates/communication_cost.sql
```diff
@@ -3,26 +3,30 @@ create materialized view ofec_communication_cost_mv_tmp as
...
+create index on ofec_communication_cost_mv_tmp (communication_cost);
```
LindsayYoung: Again, maybe we should call this `transaction_amount` to keep it the same

jontours:
Based on @PaulClark2's comment, the table we are migrating to is fec_vsum_f76?  Because I used nml_form_76.

webservices/common/models/costs.py
jontours: I may have unintentionally commented that out.

data/sql_updates/communication_cost.sql
jontours: I'll have to check but in the table I was looking at (which I'm not sure is even the right table) didn't have a column for transaction_dt or amt.  Hence I left them commented out to bring attention to that potential problem.

data/sql_updates/communication_cost.sql
```diff
@@ -3,26 +3,30 @@ create materialized view ofec_communication_cost_mv_tmp as
 select
     row_number() over () as idx,
     *,
+    s_o_cand_id as cand_id,
+    org_id as cmte_id,
     report_pdf_url(image_num) as pdf_url
-from communication_costs_vw
-where rpt_yr >= :START_YEAR
+from nml_form_76
```
LindsayYoung: We actually want `fec_vsum_f76`. I gave you the wrong instructions- I just found out  `nml_form_76` is also a data warehouse table.

LindsayYoung: Don't worry, it is the same structure.

webservices/common/models/costs.py
LindsayYoung: Actually, we do have it above on line 38. Though tell me if we took it out in other models.

data/sql_updates/communication_cost.sql
LindsayYoung: I am just thinking of renaming it in the sql to keep it consistent with what we had. `communication_cost` in this table is the same as `transaction_amount` in the old table.

data/sql_updates/communication_cost.sql
jontours: ok, that's good then.

data/sql_updates/communication_cost.sql
jontours: Ah, gotcha!  Will do.

webservices/common/models/costs.py
jontours: All the models I transitioned I left it in.  I remember that because it was kind of an oddball compared other columns.

data/sql_updates/communication_cost.sql
jontours: Ok I think for this file the references will have to stay as communication_dt and _cost.  If you look in costs.py  we are aliasing the columns so externally it appears that those columns never went away.  Otherwise I believe the column would have to renamed in the database.

jontours:
This should be ready to go.

webservices/common/models/costs.py
```diff
@@ -6,34 +6,40 @@
...
+    candidate_office_district = db.Column('s_o_cand_office_district', db.String, index=True)
+    candidate_office = db.Column('s_o_cand_office', db.String, index=True)
+    candidate_office_full =db.Column('s_o_cand_office_desc', db.String)
+    transaction_date = db.Column('communication_dt', db.Date, index=True)
+    transaction_amount = db.Column('communication_cost', db.Numeric(30, 2), index=True)
     transaction_type = db.Column('transaction_tp', db.String)
-    receipt_date = db.Column('f7_receipt_dt', db.String)
-    purpose = db.Column(db.String)
     communication_type = db.Column('communication_tp', db.String, index=True)
+    communication_type_full = db.Column('communication_tp_desc', db.String)
```
LindsayYoung: I think this is the same as purpose in the old table, right @PaulClark2 ?

jontours: So did he answer this to you directly or is this question still pending?

LindsayYoung:
We need fec_vsum_76 for this PR

webservices/common/models/costs.py
PaulClark2: I'm not sure how I missed this. `fec_vsum_76` would be the replacement for DW table `nml_form_76`

jontours: Thanks for the verification @PaulClark2, I believe the schemas should be quite similar so it's no big deal.

LindsayYoung:
We have this now, correct?

jontours:
If we do it's not on dev.

LindsayYoung:
Yea, I don't see it on production either

PaulClark2:
untitled

connected to `goldengate.open.fec.gov`

jontours:
Ahh, yeah it is there, thanks @PaulClark2

LindsayYoung:
Thanks @PaulClark2 I was missing the "f" when I checked

webservices/schemas.py
```diff
@@ -359,7 +359,7 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
 
 CommunicationCostSchema = make_schema(
     models.CommunicationCost,
-    options={'exclude': ('idx', )},
+    #options={'exclude': ('idx', )},
```
LindsayYoung: We can get rid of this line.

LindsayYoung:
This all looks good to me. There merge conflict is just the test data.

jontours:
@LindsayYoung This is ready for review.

PR #1813, opened 2016-07-14
Primary Key fix for partitioned tables

jontours:
@LindsayYoung @ccostino This is a pretty simple change.  If anything else needs to be changed with this let me know, but I've verified with my test database that this does indeed add the primary keys after running update_all.


LindsayYoung:
Thanks!

We can apply this today

PR #1814, opened 2016-07-14
Getting rid of daily updates on feature 

### LINKED ISSUE #1626: Turn off nightly updates on feature branch
Feature and dev a are sharing a DB so only one can run updates at a time.

LindsayYoung:
So they don't collide with dev on the same RDS everynight

fix #1626

### DIFF
--- a/webservices/tasks/__init__.py
+++ b/webservices/tasks/__init__.py
@@ -10,6 +10,21 @@ from raven.contrib.celery import register_signal, register_logger_signal
 from webservices.env import env
 from webservices.tasks import utils
 
+
+# Feature and dev are sharing the same RDS box so we only want dev to update
+schedule = {}
+if not os.getenv('FEATURE'):
+    schedule = {
+        'refresh': {
+            'task': 'webservices.tasks.refresh.refresh',
+            'schedule': crontab(minute=0, hour=9, day_of_week='sun-fri'),
+        },
+        'refresh_and_rebuild': {
+            'task': 'webservices.tasks.refresh.refresh_and_rebuild',
+            'schedule': crontab(minute=0, hour=9, day_of_week='sat'),
+        }
+    }
+
 def redis_url():
     redis = env.get_service(label='redis28-swarm')
     if redis:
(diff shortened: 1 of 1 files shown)


jontours:
Ok, so I think the logic should be to point to the variable VCAP_APPLICATION.space_name

jontours:
Oops, looks like `settings` not found, just change that reference to `env.app` and either cast the string to uppercase or change ``feature`` to lower.

webservices/tasks/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 # Feature and dev are sharing the same RDS box so we only want dev to update
 schedule = {}
-if not os.getenv('FEATURE'):
+if settings.get('space_name') != 'FEATURE':
```
ccostino: Quick catch, the value is lower case in Cloud Foundry - it might be worth explicitly lower-casing the value pulled back just to be sure, e.g., `settings.get('space_name').lower() != 'feature'`, just in case that ever changes on us without us knowing.

jontours: @ccostino jinx!

webservices/tasks/__init__.py
```diff
@@ -13,7 +13,7 @@
 
 # Feature and dev are sharing the same RDS box so we only want dev to update
 schedule = {}
-if settings.get('space_name') != 'FEATURE':
+if env.app('space_name') != 'feature':
```
jontours: More nitpicking really, but since @ccostino had a good point on the potential for `feature` not being capitalized  it's probably best to add a .lower(), i.e. `if env.app('space_name').lower() != 'feature'.  Also, and this is really getting down into the weeds may be good to do a null check with an `and` in case the string is null.  Seems unlikely but I guess it could happen, in Java land I'd just do a 'feature'.equals().  Still don't miss Java though!

LindsayYoung: thanks for the tip @jontours and @ccostino, I missed it before!

jontours:
Looks good!

PR #1815, opened 2016-07-15
Add information to weekly refresh task email notifications

ccostino:
Part of #1716
Fixes https://github.com/18F/openFEC-web-app/issues/1357

This changeset adds a bit of extra information to our email notifications when the weekly refresh and rebuild task runs.  It adds support for calculating the projected total amount of records processed for a week's time in the itemized tables (schedule a and b), and provides this info along with the actual amount of processed records for a week's worth of time as a part of the email notification that is sent out.  This is done by adding an index on the `pg_date` column in our `ofec_schedule_a_*` and `ofec_schedule_b_*` tables to help with retrieving the records processed in the last week.

This changeset also modifies the weekly `refresh_and_rebuild` task to run the `update_all` command again to ensure everything is processed correctly.

/cc @LindsayYoung, @jontours


webservices/tasks/refresh.py
```diff
@@ -32,13 +32,31 @@ def refresh_and_rebuild():
...
+    itemized_schedules = ['a', 'b']
+
     with mail.CaptureLogs(manage.logger, buffer):
         try:
-            manage.rebuild_aggregates()
-            manage.refresh_materialized()
+            projected_weekly_totals = manage.get_projected_weekly_itemized_totals(
+                itemized_schedules
+            )
+            manage.update_all()
```
LindsayYoung: do we need to clear the old and new ques or is this taken care of in update_all()

webservices/tasks/refresh.py
```diff
@@ -32,13 +32,31 @@ def refresh_and_rebuild():
     As with regular updates, email logs are also sent.
     """
     buffer = io.StringIO()
+    itemized_schedules = ['a', 'b']
```
LindsayYoung: Maybe we should we add 'e' too, since it has been acting up

ccostino: Yes, you're right!

webservices/tasks/refresh.py
ccostino: Oh right, we did flag that - I will double check.

ccostino: Okay, we're good here.  `update_all()` runs all of the scripts in `data/sql_setup` when it hits the `update_itemized()` method calls.  The schedule A and B scripts still take care of dropping and recreating the queue tables in this case, so yes, they are taken care in this case already.

LindsayYoung: Thanks for explaining!

PR #1816, opened 2016-07-15
Feature/add subid filter to sched a b

jontours:
Filter sched_a and b tables.  @LindsayYoung initially brought up a valid concern to get this done before the release so that we can index on sub_id.  However that should be done by the partitioning logic.

But I went ahead and knocked this one out anyway.

@noahmanger @ccostino


LindsayYoung:
Awesome thank you!

Issue #1820, opened 2016-07-17
Consistently update schedule A and B data


LindsayYoung:
Updates failing because it is trying to add columns that already exist .

I think we just want to get rid of the things in the old que before adding the things in the new que, that way, replacements should run smoothly. 
https://github.com/18F/openFEC/blob/develop/data/sql_setup/prepare_schedule_a.sql#L37

Here is the error:
From Dev Sun, Jul 17, 2016 at 5:06 AM
```
2016-07-17 09:06:45 ERROR:manager:(psycopg2.IntegrityError) duplicate key value violates unique constraint "ofec_sched_a_2013_2014_tmp_pkey"
DETAIL:  Key (sub_id)=(2071520161306759407) already exists.
```

ccostino:
At least we're now seeing the primary key constraint violations we were expecting to see!

LindsayYoung:
I don't think the issue is what I thought it was, I am seeing a record identified by sub_id that is in the existing table and new queue table, but not the old table. 

Now I am looking closer at how our queues are made.

jontours:
Well, not that it helps or anything, but we know that must be happing in an insert statement... Just a question of why that insert even happens, or if it should happen.

LindsayYoung:
there are 78200 records in the schedule a new queue and it looks like 78185 of them have already been added

ccostino:
This goes back to the notion that the queue tables aren't being properly cleaned out.  Is it possible that this is the case due to the weekly update?  We fixed this in at least `dev` though by changing it back to `update_all`, which does properly clear out the queues, correct?

Based on the `fec-update-bot` notifications from last night it does look like `dev` is okay now, but staging is still affected.

ccostino:
@LindsayYoung, were those counts in `dev` or `staging`?

LindsayYoung:
Those counts were in dev.

I think the difference is that I ran rebuild_aggregates() on dev yesterday but not staging. The timing suggests that the issue was the weekly rebuild, but I remember we both checked to confirm that that cleared the queue. I can double check that again. 

I'll rebuild the aggregates on staging and see if I can get updates going on staging. Tonight we can try a nightly update on staging and a weekly rebuild on dev and see how it goes.

ccostino:
👍 to all of that, thanks! :-)

[...]

PR #1827, opened 2016-07-18
Reorder updates

LindsayYoung:
I think that grouping the updates with the respective queue deletion
might prevent an issue where the updates run half way and the things
that are successfully updated are not cleared from the queue.


ccostino:
Thanks, @LindsayYoung!  As noted, more investigation is likely needed into this (which we're following up on in #1820), but at least we get operations grouped together around the same tables.

PR #1829, opened 2016-07-19
Removed flask.ext imports

jontours:
Turns out there were some of the`flask.ext` import statements hanging around.  All of our imports are now current with Flask 0.11.  However there is still one deprecation warning, and I think that's originating from NPlusOne.  @jmcarp would it be all right if we open a issue with that and potentially submit a pull request?

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -8,8 +8,8 @@ import multiprocessing
 
 import networkx as nx
 import sqlalchemy as sa
-from flask.ext.script import Server
-from flask.ext.script import Manager
+from flask_script import Server
+from flask_script import Manager
 
 from webservices import flow
 from webservices.env import env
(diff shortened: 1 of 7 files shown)


jmcarp:
@jontours: sure, issues and pull requests against `nplusone` are always welcome. Where are you seeing the deprecated import? I noticed an old-style import in the `nplusone` tests, but I don't see anything that looks like `flask.ext` in the rest of the code.

jontours:
@jmcarp yeah that's the thing, I don't get a great warning statement from flask about which library the import originated from.  I'd be willing to say just ignore it for the time being.

jmcarp:
@jontours: I'd be happy to look at whatever output you have if you think it might be useful. Or you can `ag` through the packages in your environment and look for `flask.ext`.

jontours:
Ah, think I found it.  It looks like flask_sqlalchemy 2.0 used the old import statement, so that's probably where it originated from.  Thanks for the help.

PR #1831, opened 2016-07-19
before deploy update fixes

LindsayYoung:
Removing schedule e count for now and adding requirements for weekly updates


LindsayYoung:
close  #1820

LindsayYoung:
This was for partitioning, which we shouldn't do frequently.

PR #1836, opened 2016-07-20
New tests for sorting

jontours:
Currently added a new tests in test_utils to test hiding nulls for candidate totals.  This is to ensure that any changes to this endpoint doesn't not cause any regressions.

Still need to add a test for Elections, unfortunately I don't know of any easier ways of doing this, but it does work.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -89,6 +89,11 @@ class CommitteeHistoryFactory(BaseCommitteeFactory):
         model = models.CommitteeHistory
     cycle = 2016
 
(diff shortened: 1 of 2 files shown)


jontours:
This is good to go.  There may be some more tests to write but these two seem to be great edge cases.  Currently tests Election totals and candidate totals endpoint.

jontours:
Closes https://github.com/18F/openFEC/issues/1823

LindsayYoung:
Thank you!

jontours:
Thanks for taking a look, it turned out to be quite a challenge to write these two tests.  But, I learned a good deal about the election endpoint in the process.

PR #1842, opened 2016-07-26
Feature/docs tips

LindsayYoung:
I was trapped on a plane without wifi so I took a look at some docs improvements. One thing I wondered is if we showed headings instead of the url pieces to group like content. We can take that out if y'all don't like it- that is in its own commit.

@emileighoutlaw

### DIFF
--- a/webservices/resources/aggregates.py
+++ b/webservices/resources/aggregates.py
@@ -42,7 +42,7 @@ class AggregateResource(ApiResource):
 
 
 @doc(
-    tags=['schedules/schedule_a'],
+    tags=['receipts'],
     description=docs.SIZE_DESCRIPTION,
 )
 class ScheduleABySizeView(AggregateResource):
@@ -58,7 +58,7 @@ class ScheduleABySizeView(AggregateResource):
 
 
 @doc(
-    tags=['schedules/schedule_a'],
+    tags=['receipts'],
     description=(
         'Schedule A receipts aggregated by contributor state. To avoid double counting, '
         'memoed items are not included.'
@@ -83,7 +83,7 @@ class ScheduleAByStateView(AggregateResource):
 
 
 @doc(
-    tags=['schedules/schedule_a'],
+    tags=['receipts'],
     description=(
         'Schedule A receipts aggregated by contributor zip code. To avoid double '
         'counting, memoed items are not included.'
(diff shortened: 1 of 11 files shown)


emileighoutlaw:
This looks rad to me. I pushed a couple copyedits, but it's 👍 for a merge from my perspective.

webservices/resources/aggregates.py
```diff
@@ -42,7 +42,7 @@ def build_query(self, committee_id=None, **kwargs):
 
 
 @doc(
-    tags=['schedules/schedule_a'],
+    tags=['receipts'],
```
ccostino: What does changing the tags here do; asked another way, how will this change manifest itself?

webservices/spec.py
```diff
@@ -45,27 +45,27 @@
             'description': docs.FILINGS,
         },
         {
-            'name': 'schedules/schedule_a',
+            'name': 'receipts',
```
ccostino: Ah, they match up here?

ccostino:
This looks good to me, thanks for taking the time to do this @LindsayYoung!  I only had a quick question around the tags since I'm unfamiliar with them here.

webservices/resources/aggregates.py
LindsayYoung: I have heard from some users that they don't think a resource exists, such as line item receipts or independent expenditures. I think it is because they don't know the jargon. 

The tags are used for sorting and grouping the endpoints. 

Here, I am re-naming schedule/schedule to receipts in the top bar. I am thinking that will make it easier for people to find what they need.  I would like to get rid of the / / format to just use them as titles. 

We are already using these as titles and not literal url pieces for /financial/

I did add spaces to "independent expenditures" and other two word phrases, which does look weird with the slashes. I would rather remove the slashes but we can hyphenate those if you think that looks better for now.

screen shot 2016-08-03 at 12 57 07 pm

webservices/spec.py
LindsayYoung: That's right

webservices/resources/aggregates.py
ccostino: Oh, I see!  Thanks for taking the time to explain, I appreciate it!

Issue #1838, opened 2016-07-21
Add additional logging for partitioning


ccostino:
Our update process currently does not surface any information about the partitioning steps that may take place so we are unable to see where in the process it might be at any given time.  It would be good to add some logging statements around the following events:
- When partitioning for a particular table starts
- When a child table starts being processed
- When a child table finishes being processed
- When the partitioning for a table completely finishes

If anyone can think of any other events we should add logging for to increase visibility into what is happening during our update process, please feel free to add them!

LindsayYoung:
That sounds good. Timestamps would be nice for each step too- I think that is what you already had in mind.

jontours:
Maybe it would be possible to sense when indexes for a table have started and finished?  As that was particularly a problem yesterday.

ccostino:
Both good suggestions!  I believe the timestamps are already included as a part of the logging statements but we should double check.

LindsayYoung:
If it is not too hard to throw in, it would be nice to have a different email subject line when there is an error. 

maybe if there are no errors keep it he same:
FEC Update: dev | celery-worker
and if there is an error:
[ERROR] FEC Update: dev | celery-worker

PR #1844, opened 2016-07-27
Add support for renaming indexes

### LINKED ISSUE #1840: rename indexes
During partitioning all our indexes are being dropped and this is causing queries to be crazy slow and eat up storage space.

We want to add a function like rename tables to rename indexes, hopefully that will keep them from being dropped.

ccostino:
Part of #1838 
Fixes #1840 

When partitioning the itemized Schedule A and B tables, the indexes of the existing tables were being dropped and recreated, but this impacted the existing tables before they were replaced.  As partitioning these tables takes a long time with the full FEC data set, this has huge impact on database performance if anyone or anything is accessing the data in these tables.

This changeset alleviates this by renaming the indexes once the tables themselves are renamed so that they are not removed too early.

It also adds just a bit of logging information to help and get a head start there.

/cc @LindsayYoung, @jontours

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -228,8 +228,12 @@ def update_all(processes=1):
     update_itemized('a')
     update_itemized('b')
     update_itemized('e')
+    logger.info('Partitioning Schedule A...')
     partition.SchedAGroup.run()
+    logger.info('Finished partitioning Schedule A.')
+    logger.info('Partitioning Schedule B...')
     partition.SchedBGroup.run()
+    logger.info('Finished partitioning Schedule B.')
     rebuild_aggregates(processes=processes)
     update_schemas(processes=processes)
 
--- a/webservices/partition/base.py
+++ b/webservices/partition/base.py
@@ -131,15 +131,35 @@ class TableGroup:
 
     @classmethod
     def rename(cls):
+        # Rename master table
         cmds = [
             'drop table if exists {0}_master cascade',
             'alter table {0}_master_tmp rename to {0}_master',
         ]
+
         for cmd in cmds:
             db.engine.execute(cmd.format(cls.base_name))
+
+        # Rename child tables
         for cycle in get_cycles():
-            cmd = 'alter table {0}_tmp rename to {0}'.format(cls.get_child_name(cycle))
+            child_name = cls.get_child_name(cycle)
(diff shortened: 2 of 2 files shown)


jmcarp:
@ccostino: just curious--why does renaming indexes affect performance here? I didn't think index names mattered.

ccostino:
Hey @jmcarp!  Sorry, I should've been a little clearer above.  Renaming the indexes doesn't affect performance in and of itself, but the current flow of events does affect performance if folks are actively using the database while the partitioning is taking place.  Ideally we'd be able to run that when there's no activity going on, but we've run into situations where it gets backed up and is continuing during normal usage hours.  What happens as the partitioning takes place is that the indexes on the existing tables get dropped because they share the same name as the indexes being processed for the new tables being created.  If folks are using the site and querying against these tables, it brings the database to an absolute standstill and all available free space is used because it is it has to reconstruct everything for every query.

I would have figured that the indexes were tied to the tables they are created for and therefore even if they shared the same name as others in the database, it wouldn't matter because of the table association, but that's not the case (at least in Postgres).  Because the indexes are created at the time the tables still have `_tmp` in their names, they have `_tmp` in their names as well even after the table renames.  This change renames the indexes along with the tables themselves so that the existing indexes won't be dropped prematurely; they'll be dropped when the old tables are dropped at the start of the `rename` method instead.

I'm preserving the `create_index` method as is with the existing `index.drop()` call in case we ever have a situation where the update/partitioning process stops on us and leaves things in a half done state - that'll catch any that shouldn't be there from the unfinished processing.

FYI, we ran into this kind of situation last week because of a database shutdown on the RDS side due Amazon's systems verifying an invalid setting and resetting it automatically.  We started the partitioning again, not realizing that this issue with the indexes existed at the time, and it resulted in the database grinding to a halt performance-wise.  We discovered the issue when we took a look at the existing tables and seeing that there were no indexes on the tables (except for the GIN indexes, oddly enough - those didn't get dropped like the others).

ccostino:
*Note:* We're not going to merge this in as a part of the release going out now; it'll be a part of the next release, so I'll create a new PR against the `develop` branch instead.

LindsayYoung:
Thanks Carlo

PR #1847, opened 2016-07-29
Feature/load pdfs into s3

anthonygarvan:
Loading advisory opinion PDFs into S3. H/T to @annalee for getting this started. It takes a couple hours to load up all of them, so the code below looks for AOs that do not exist on S3 and only loads those. This assumes that the PDFs rarely (if ever) change, but are only added.


annalee:
:shipit:

PR #1849, opened 2016-07-29
Replace missing delete statements

ccostino:
Part of #1820 

This changeset replaces the delete statements that went missing from the schedule A queue processing we have.  This has caused an issue where multiple records with the same `sub_id` might exist in these tables and therefore might result in insert errors in the main `ofec_sched_a_*` tables due to existing primary keys from records already processed.

This changeset also addresses an issue where an index being created on the `fec_vsum_sched_b` table was being duplicated due to not being dropped first.

/cc @LindsayYoung, @jontours

### DIFF
--- a/data/sql_setup/prepare_schedule_a.sql
+++ b/data/sql_setup/prepare_schedule_a.sql
@@ -26,24 +26,31 @@ begin
         two_year_transaction_period_new = get_transaction_year(new.contb_receipt_dt, new.rpt_yr);
 
         if two_year_transaction_period_new >= start_year then
+            delete from ofec_sched_a_queue_new where sub_id = new.sub_id;
             insert into ofec_sched_a_queue_new values (new.*, timestamp, two_year_transaction_period_new);
         end if;
+
         return new;
     elsif tg_op = 'UPDATE' then
         two_year_transaction_period_new = get_transaction_year(new.contb_receipt_dt, new.rpt_yr);
         two_year_transaction_period_old = get_transaction_year(old.contb_receipt_dt, old.rpt_yr);
 
         if two_year_transaction_period_new >= start_year then
+            delete from ofec_sched_a_queue_new where sub_id = new.sub_id;
+            delete from ofec_sched_a_queue_old where sub_id = old.sub_id;
             insert into ofec_sched_a_queue_new values (new.*, timestamp, two_year_transaction_period_new);
             insert into ofec_sched_a_queue_old values (old.*, timestamp, two_year_transaction_period_old);
         end if;
+
         return new;
     elsif tg_op = 'DELETE' then
         two_year_transaction_period_old = get_transaction_year(old.contb_receipt_dt, old.rpt_yr);
 
         if two_year_transaction_period_old >= start_year then
+            delete from ofec_sched_a_queue_old where sub_id = old.sub_id;
             insert into ofec_sched_a_queue_old values (old.*, timestamp, two_year_transaction_period_old);
         end if;
+
         return old;
     end if;
 end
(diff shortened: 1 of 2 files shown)


data/sql_setup/prepare_schedule_a.sql
```diff
@@ -26,24 +26,31 @@ begin
...
             insert into ofec_sched_a_queue_new values (new.*, timestamp, two_year_transaction_period_new);
         end if;
+
         return new;
     elsif tg_op = 'UPDATE' then
         two_year_transaction_period_new = get_transaction_year(new.contb_receipt_dt, new.rpt_yr);
         two_year_transaction_period_old = get_transaction_year(old.contb_receipt_dt, old.rpt_yr);
 
         if two_year_transaction_period_new >= start_year then
+            delete from ofec_sched_a_queue_new where sub_id = new.sub_id;
```
jontours: In spirit of @LindsayYoung's idea, instead of maybe having to do a sort, maybe just add logic here that checks that the timestamp is later, i.e. `where sub_id = new.sub_id AND timestampFoo > new.timestampFoo`.

LindsayYoung: would that take care of cases where the same record was updated in the queue more than once?

jontours: Good point, and in that case would the timestamps be equal?  if so doing a `timestampFoo <= new.timestampFoo` could handle it.

LindsayYoung:
We might want to select distinct from the que by `sub_id` and sort decending by time stamp

data/sql_setup/prepare_schedule_a.sql
ccostino: I was thinking about this a bit more and am wondering if we're overthinking it or if I'm missing something still:  the `delete` statement as it is will remove all records that currently exist in the queue table(s) it is targeting that matches the `sub_id` of the record being inserted/updated/deleted in the table that caused the trigger to run.

Once these are in place there would only ever be 0 or 1 record(s) that could match the `sub_id` because anything existing in the queue table(s) would be removed by the `delete` statement before the new record to be processed would be inserted.

On the off chance that more than one somehow made it's way in though, we'd still want them all removed so we're only ever dealing with one record.  This is technically only an issue upon inserts into the `ofec_sched_*_*` tables due to the primary key constraint, but it's nice to know that this is the case across all operations from a consistency standpoint.

Furthermore, the `timestamp` doesn't exist on the record being handled by the trigger; that is calculated within the trigger itself and added to the copy of the record that is inserted into the queue tables.

I think I'm either seeing confusion about what's trying to be fixed or I'm still confused myself.  As I understand it, the issue is that multiple records with the same `sub_id` can sometimes exist in the schedule A queue tables we have, and if that's the case then when our nightly updates run that will cause unique key constraint errors with any `insert` operations into the `ofec_sched_a_*` tables. due to the primary key already existing in the table once the first one of those records with the same `sub_id` from the queue is processed.  As soon as the nightly update hits the second record to insert with the same `sub_id` that was processed before it will fail.  These `delete` statements in the trigger that helps manage the queue tables will prevent that from ever happening.

Does this help at all, or am I way out in left field? :-)

jontours: I have the same understanding.  @LindsayYoung when you mentioned doing a sort and grabbing the latest, what column did you have in mind to sort on?  It would have to be a column like contb_receipt_dt right?

LindsayYoung: I think we would want the day the record was received, since the amendments can talk about the same transaction and have the same sub_id.

ccostino: After a short chat we all determined that the fix as it stands is the appropriate way to go.  In the instance of the one edge case of processing a record with potentially older data than what was previously added to the queues for the nightly update, we decided to go ahead with the processing so that at least the data is consistent with what is in the source tables.

Should this ever be the case, the fix would be for the FEC folks to update the data in the source tables, which will trigger another update on our side to pull in the correct information.  By maintaining consistency, it'll be easier to spot such an issue if it were to ever happen (hopefully not!).

This is now ready for merge!

PR #1858, opened 2016-08-01
Takes unitemized totals out of state break downs

### LINKED ISSUE #1561: Sanders DC totals don't match totals on fec.gov presidential map
I think this came from the beta feedback system. The pres map only has about $212,000 for Sanders from DC. The beta site has over $10,000,000. The problem is Sanders is reporting a single transaction for UNITEMIZED TOTAL contributions and has it listed as being from DC.  This isn't uncommon. Most places where it would cause a problem we filter out contributions where the upper contributor name contains "UNITEM". 

HIllary does this too. Others may as well.

LindsayYoung:
This will fix totals where campaigns are adding a state to their unitemized contributions.

fix #1561


noahmanger:
Can we try to get this in this release?

ccostino:
@noahmanger when you say this release, do you mean the one that is currently live or the next one we're planning for?

I also took a look at this, I think it's good to go code-wise, but I'm not 100% sure where I'm looking for this info to compare (and I venture to guess that this needs to get into at least `dev` in order to see the full fix).

noahmanger:
I meant the one for this week, just because it's an important bug.

ccostino:
Gotcha - so this should be against the release branch instead it sounds like?

noahmanger:
If we're confident putting it into this one, then yeah.

ccostino:
Sounds good - @LindsayYoung if you want to make this PR against the open release branch, I'm happy to merge there instead!

LindsayYoung:
Since we are deploying tomorrow, I don't think that is enough time for FEC review. The individual logic is kind of complicated and while I think this is right, I want to give FEC a chance to look at it.

LindsayYoung:
closes  18F/openFEC-web-app#1494

ccostino:
Thanks @LindsayYoung!

PR #1864, opened 2016-08-02
Reports endpoint modification [WIP]

jontours:
The current progress of this feature can be found here: https://github.com/18F/openFEC/issues/1804

Most of the functionality is there, will need to iron out details about possible renaming some columns as discussed in the issue.


noahmanger:
Nice. I just tried this out locally and it seems to be working well. Here's the PR for the frontend if you want to test yourself https://github.com/18F/openFEC-web-app/pull/1338

jontours:
@noahmanger Awesome!  I'll have to test it out.  I just pushed a modification, a big change with it is to filter for multiple committee_ids the query parameter was changed to `committee_ids`.  So to filter by multiple committee ids the syntax would be `commitee_ids='C123'&committee_ids`.

This was necessary as that parameter was clashing with the `report\{committee_id}` endpoint.

noahmanger:
Thanks for the head up. Just made the fix. The response property is still `committee_id` singular though, right?

jontours:
Yes, the response has stayed the same.

jontours:
Yeah looks like the front-end works too, I just checkout out the /pac-party endpoint.  Some of the sorting functionality will not be testable until we update schemas on dev, as that builds the indexes that enables sorting.  But I've tested it locally so no worries there.

noahmanger:
Excellent. Reviewing the checklist on the related issue, it looks like the most important stuff is good to go. If you think the remaining items will take more than a day to implement, I'd say let's move forward with what we have and follow up with those later. Nice work!

jontours:
Thanks, I can probably get the filter for committee type by lunch.  The last filter, for principal committees, could actually be a little tricky so I'll see if I can get it done today but that one may have to wait.

webservices/resources/reports.py
```diff
@@ -105,11 +121,27 @@ def build_query(self, committee_id=None, committee_type=None, **kwargs):
...
+        query = filters.filter_range(query, kwargs, self.filter_range_fields)
+
         return query, reports_class, reports_schema
 
     def _resolve_committee_type(self, committee_id=None, committee_type=None, **kwargs):
-        if committee_id is not None:
-            query = models.CommitteeHistory.query.filter_by(committee_id=committee_id)
+        #This could be a potential pit fall, is there a better way to get
+        #committee type?  Was the reason this was done this way as a committee
+        #type can change over time?  Also take strong note
```
LindsayYoung: Committee type can change over time. Authorized committees frequently become PACs when the candidate is no longer in the race,

webservices/common/models/reports.py
```diff
@@ -49,11 +49,11 @@ class CommitteeReports(PdfMixin, BaseModel):
     refunded_other_political_committee_contributions_ytd = db.Column(db.Numeric(30, 2), doc=docs.add_ytd(docs.REFUNDED_OTHER_POLITICAL_COMMITTEE_CONTRIBUTIONS))
     refunded_political_party_committee_contributions_period = db.Column(db.Numeric(30, 2), doc=docs.add_period(docs.REFUNDED_POLITICAL_PARTY_COMMITTEE_CONTRIBUTIONS))
     refunded_political_party_committee_contributions_ytd = db.Column(db.Numeric(30, 2), doc=docs.add_ytd(docs.REFUNDED_POLITICAL_PARTY_COMMITTEE_CONTRIBUTIONS))
-    total_contributions_period = db.Column(db.Numeric(30, 2), doc=docs.add_period(docs.CONTRIBUTIONS))
+    total_contributions_period = db.Column('total_contributions_period', db.Numeric(30, 2), doc=docs.add_period(docs.CONTRIBUTIONS))
```
ccostino: Out of curiosity, why have these labels been added when we don't seem to have any where else?  Are they necessary and if so, should we consider adding it elsewhere going forward/

jontours: That's an excellent question actually.  When I was implementing the range filters for these sql_alchemy couldn't seem to find the column name with a default mapping.  I can circle back to that and see if these labels can actually be removed, because I had other issues at the time and I'm not entirely sure this was an issue.

jontours: No, they definitely have to be there, at least if you are piping your columns through filters in filters.py, like I did for these columns, here is the stack track when you remove the labels:

```
 
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/elements.py", line 494, in compile
    return self._compiler(dialect, bind=bind, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/elements.py", line 500, in _compiler
    return dialect.statement_compiler(dialect, self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 392, in __init__
    Compiled.__init__(self, dialect, statement, **kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 190, in __init__
    self.string = self.process(self.statement, **compile_kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 213, in process
    return obj._compiler_dispatch(self, **kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1602, in visit_select
    text, select, inner_columns, froms, byfrom, kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1694, in _compose_select_body
    for f in froms])
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1694, in <listcomp>
    for f in froms])
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1257, in visit_alias
    asfrom=True, **kwargs) + \
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/annotation.py", line 73, in _compiler_dispatch
    self, visitor, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1602, in visit_select
    text, select, inner_columns, froms, byfrom, kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 1699, in _compose_select_body
    t = select._whereclause._compiler_dispatch(self, **kwargs)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 921, in visit_binary
    return self._generate_generic_binary(binary, opstring, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 937, in _generate_generic_binary
    opstring + \
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/visitors.py", line 81, in _compiler_dispatch
    return meth(self, **kw)
  File "/Users/jonathancarmack/.pyenv/versions/3.5.1/lib/python3.5/site-packages/sqlalchemy/sql/compiler.py", line 624, in visit_column
    raise exc.CompileError("Cannot compile Column object until "
sqlalchemy.exc.CompileError: Cannot compile Column object until its 'name' is assigned.
```

ccostino: Oh wow, interesting.  It caught my eye because none of the other columns are defined like this.  I'm curious if we're handling other columns that are similar in nature in a different way, and if so how.  I'd have to dig deeper to see, I don't know off the top of my head.

webservices/resources/reports.py
```diff
@@ -105,11 +134,18 @@ def build_query(self, committee_id=None, committee_type=None, **kwargs):
...
 
     def _resolve_committee_type(self, committee_id=None, committee_type=None, **kwargs):
-        if committee_id is not None:
-            query = models.CommitteeHistory.query.filter_by(committee_id=committee_id)
+        #if kwargs.get('committee_id') is not None and len(kwargs.get('committee_id')) > 0:
+        if committee_id is not None or kwargs.get('committee_ids'):
+            comm_id = (
+                committee_id
+                if committee_id is not None
+                else kwargs.get('committee_ids')[0]
```
ccostino: We're just grabbing the first `committee_id` given and ignoring the rest?

jontours:
This has all the features implemented, the last item on the list was the hardest.

noahmanger:
Rad! What's the parameter name for candidate ID?

jontours:
It's `auth_committee_candidate_id`.  If the language seems off it's not big deal to change it.

webservices/common/models/reports.py
jontours: I've done some initial (but not exhaustive) research into this and all the fields that are being filtered via filters.py, have literal naming of the columns for their mapping.

webservices/resources/reports.py
jontours: Yeah, that's probably not the best strategy and I'm open to other ideas.  Maybe a popular vote algorithm?  Like whichever type most of the commitees have in common that's the one we return?

LindsayYoung: I think there will be some edge cases where this behaves oddly. I'll take a look

jontours: Yeah I think even if a user asked for mulitiple valid committee types within a specific committee type endpoint it could behave not as expected, for example some of the presidential commitees have a type of 'S'.  I'm guessing that's when a senator decided to run for president?  But it could definitely happen that one of those odd S types could pop up first in the list.

webservices/resources/reports.py
```diff
@@ -1,10 +1,12 @@
 import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
```
ccostino: Was this included for some debugging earlier? :-)  I don't see it being used.

jontours: It's to print out queries that error when you attempt to do a simple `print(query)` for debugging.  That reminds me, I've written this about three times so I'll take them method I used for that print statement and pull it up into the utils file.

webservices/common/models/reports.py
ccostino: I stand corrected!  I also found a similar implementation with the itemized models.

webservices/args.py
```diff
@@ -226,6 +226,15 @@ def make_seek_args(field=fields.Int, description=None):
...
     'is_amended': fields.Bool(description='Report has been amended'),
+    'committee_ids': fields.List(fields.Str, description=docs.COMMITTEE_ID),
+    'min_disbursements_amount': Currency(description='Filter for all amounts greater than a value.'),
+    'max_disbursements_amount': Currency(description='Filter for all amounts less than a value.'),
+    'min_receipts_amount': Currency(description='Filter for all amounts greater than a value.'),
+    'max_receipts_amount': Currency(description='Filter for all amounts less than a value.'),
+    'min_cash_on_hand_amount': Currency(description='Filter for all amounts greater than a value.'),
+    'max_cash_on_hand_amount': Currency(description='Filter for all amounts less than a value.'),
+    'type': fields.List(fields.Str, description=docs.COMMITTEE_TYPE),
+    'auth_committee_candidate_id': fields.List(fields.Str, description=docs.COMMITTEE_ID),
```
LindsayYoung: If this takes a committee_id argument, why would this be a separate field, rather than just using committee_ids?

LindsayYoung: I am reading the resource, and this looks like it is a candidate id. I know there were naming issues, was that why this is not called candidate_id? Either way, we will want the docs to point to CANDIDATE_ID.

webservices/resources/reports.py
LindsayYoung: Since we are only looking for principal or authorized committees, I think we can simplify this, we can look at the candidate id and see if it starts with a H, S or P and then type it appropriately.

webservices/args.py
jontours: As for why there is a separate `committee_ids` argument, I think you stated the reason.  This resource serves up two endpoints, and one of those endpoints is `reports\{committee_id}`.  I originally did just try using the committee_id and removed the committee_id path parameter, this of course caused the tests to fail.

So the approach to that is to either route `report\{committee_id}` to a different endpoint than this one, or use a separate `committee_ids` query param, as I did here.  Unless there is some way to do both with Flask, which didn't seem possible to me, but maybe there is a way?

jontours: For your second comment, I can name it `candidate_id`, I don't think there is any conflicts there.  I asked @noahmanger if we was ok with the naming, and I'm certainly open to renaming it, the reason I used that name is to make it somewhat explicit that there is a relation entailed in that parameter.  

I'll change the docs to point to `CANDIDATE_ID`.  Thanks for the catch.

webservices/resources/reports.py
jontours: Ah, that would be much simpler.  I wasn't part of the earlier discussions for this endpoint, but I didn't know that we were only concerned with authorized committees.  Is that only in the scenario that the user queries for committee reports based on multiple committee_ids?

LindsayYoung: @noahmanger We can't currently do bunch of committee_ids if they are different types.

So if someone wanted the reports for their house, senate and presidential race all at once, we can't currently deal with that because they are in 3 different tables and use 3 different schemas. In this proposal, we would only get one.

Our options are:
    a) don't accept multiple committee_ids 
    b) accept multiple ids only when they come from the same table.
    b) re-write the underlying data view and resource 

In my opinion, re-writing this now is more work than it is worth. Feel free to push back. I can see how this would be a useful feature.

noahmanger: So, the structure of the front-end matches the structure of the API. There's no top-level /reports/ page, there's just: `/reports/presidential` `/reports/house-senate/` `/reports/pac-party/` and `/reports/ie-only`

So yeah, if someone requested a `committee_id` of a committee that didn't match the current page, nothing would show up. And if they requested a `candidate` with authorized committees not of the current page, nothing would show up.

Whether or not we accept multiple IDs is kind of beside the point, since even accepting one ID of a non-matching type, won't return anything.

On the one hand, this is a pretty advanced feature and I think if we don't resolve this now, it's not a _huge_ deal. I think we can expect advanced users to recover.

But from a UX perspective, I think the best solution would be to just update the `names` endpoint to scope to candidates or committees of certain types. This is the endpoint that the typeahead hits, so if there was something like `/names/candidates/presidential/` that only returned presidential candidates, or `/names/committees/house-senate/` that only returned H & S type commmittees, that might work well.

noahmanger: Unless I'm misunderstanding the implications, it seems like if this approach would work, we should merge this in and then follow up with the name endpoint modification.

LindsayYoung:
I think we will want to remove the multiple committee_id option for now so we don't confuse API users.

noahmanger:
What's the point of removing multiple `committee_ids` if passing a single wrong committee ID has the same result?

webservices/resources/reports.py
jontours: I'm not adverse to adding it back in, as long as we have the impetus to implement the `names` endpoint as you suggested, as that would guarantee that only the proper committee ids could possible be passed to the API, at least via the web-app.  I think API users could still inject some uncertainty but as you said if a user comes up with some impossible type query, as in there are no results, they would just get back no results, not an error.

LindsayYoung: I think we can do candidate_id but we should take out committee_ids.

For candidate ids we can type them by looking at their ID, it starts with P, S or H. If they are running for multiple offices, they will have multiple ids. We won't return records when committees file incorrectly, which is common.

Committee_ids won't work if you add different types of committees. 

Making different typeaheads is a good follow up.

jontours: Currently the query parameter for `auth_committee_candidate_id` is a list.  Which if you think of it runs up to some of the same problems with multiple committees, i.e. multiple candidates with the possibility of being different types.  So should I make it truly `candidate_id` and just call it a day.  

With features like this the problems that a user could create through even somewhat sensical queries runs up into exponential numbers of possible combinations.  So I think for any given feature you can end game it to death, but, I think your points are valid.

With that being said I'll change `auth_committee_candidate_id` to `candidate_id`, and parse out it's type based on the first character.

jontours: @LindsayYoung I've made those changes if you want to give it a review.  Thanks!

noahmanger: Just to make sure I'm understanding, are you suggesting we don't have any way to filter reports by committee?

LindsayYoung:
added 2 minor fixes to get the filtering going for cash on hand and disbursements going.

webservices/resources/reports.py
LindsayYoung: For now, you can filter one committee at at time. 

We can add more filtering later, but it is a heavy lift.

noahmanger: I'm just confused why adding one committee at a time is any different. There's nothing stopping people from passing a PAC committee ID to the `/reports/presidential` endpoint as is, right?

LindsayYoung: My thinking what that getting no is much clearer than getting some results. You will assume you are getting all your results if it looks like it is working. 

I do think you have a good point about the url structure adding some clarity and expectations for data users. We will probably want to add something in the docs about all the committees being the same type to. 

I would be cool with adding multiple committees of the same type if we don't check the type here, and use [the url check](https://github.com/18F/openFEC/pull/1864/files/158a0b19d39e5b62aefb987497a3494bd0187bbb#diff-3e2e29f1fb0dc0c4c17dc617d5a140c8R154) to determine the type.

I merged this, so any changes would need to be a follow up PR

noahmanger: Ah I see. Thanks for explaining. Ordinarily I'd be cool with just doing the simpler solution for now, but in this case it would require making a new UI component in the form of a typeahead filter that only allows for one committee. 

👍  thanks for merging. I'll follow up with folks on Monday.

PR #1837, opened 2016-07-21
Fix individual calculations

LindsayYoung:
Remove 18J as an individual line item

close #1832

PR #1866, opened 2016-08-03
Feature/individual tweaks

### LINKED ISSUE #1779: Resolve individual vs. committee filtering issues
So that users always the data they expect, ensure that the individual vs committee contributor filter on receipts is behaving properly. 

**Criteria for completion**
- [ ] On `/receipts/`: selecting "individual" only shows contributions from non-committees
- [ ] On `/receipts/`: selecting "committee" only shows contributions from registered committees
- [ ] On `/receipts/individual-contributions/`: Only contributions from non-committees are shown
- [ ] All records from committees include a link to the committee page

**Notes**
Currently both on the Individual Contributions page and Receipts page, committee (or committee-like) contributors show up even if the Individual filter is applied. And conversely, non-committee contributors (i.e. those without committee IDs) show up if the Committee filter is applied. This was originally reported in https://github.com/18F/openFEC-web-app/issues/1185

In fixing this, we should also resolve any discrepancies or confusion between the "unique" filter (which currently passes `is_individual=true` and the committee vs individual filters, which pass `contributor_type=individual` and `contributor_type=committee`. 

cc @LindsayYoung @PaulClark2

LindsayYoung:
This is a methodology change that I proposed but wanted to run by @PaulClark2 and @jwchumley 

We are seeing a lot of entries that are coded as individuals when they have committee id numbers and are clearly not individuals. 

To reduce the errors, this checks if there is a committee id listed for the contributor. If there is a committee id and it is not the filing committee's id, we are not counting that record as an individual.

If this sounds good, I can update the documentation.

closes #1779


jwchumley:
When you say "Coded as Individuals" are you talking about the entity type or transaction code 15? In any case I think you can pretty reliably use other id != committee id as an indicator of a transaction from or to another committee on schedules A and B.

________________________________

PaulClark2:
`To reduce the errors, this checks if there is a committee id listed for the contributor. If there is a committee id and it is not the filing committee's id, we are not counting that record as an individual.`

earmarked (15E) and JF (15J) receipts and a few others will almost always have committee IDs in the other_id field and they are individual contributions.

LindsayYoung:
Thanks @PaulClark2 and @jwchumley 

- I have updated  the coded to include the additional types listed in #1837 
- I added an exception for line numbers that are expected to have committee id. Are there any other line numbers for individuals that are expected to include committee id.

here is the [updated function](https://github.com/18F/openFEC/blob/916834210cc3dd88539baf3f76bd77b97649a690/data/functions/individual.sql)

PaulClark2:
Form 3X Line 17 can include individual contributions that might have an other_id.  I think your fuction already included Line 17.  These would be coded transactions -- the codes are included in https://github.com/18F/openFEC/pull/1837

LindsayYoung:
Just got the OK from Paul and Jeff, we will look at the data it produces more closely when it is applied to dev.

data/functions/individual.sql
```diff
@@ -34,3 +45,28 @@ begin
   );
 end
 $$ language plpgsql immutable;
+
+
+-- unitemized contributions should not be included in the state breakdowns
+create or replace function is_unitemized(memo_text text) returns bool as $$
```
ccostino: It looks like this was included in the other PR and is now resulting in merge conflicts here.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_state.sql
```diff
@@ -9,9 +9,11 @@ select
     sum(contb_receipt_amt) as total,
     count(contb_receipt_amt) as count
 from fec_vsum_sched_a
-where rpt_yr >= :START_YEAR_AGGREGATE
+where
+    rpt_yr >= :START_YEAR_AGGREGATE and
+    is_unitemized(memo_text::text) = false
```
ccostino: Same here, I think this was also a part of the other PR already.

ccostino:
Thanks, @LindsayYoung!  I've addressed the merge conflicts, once the tests all pass in CI I'll merge and run the `update_functions`, `rebuild_aggregates`, and `update_schemas` tasks against `dev`.

Issue #1868, opened 2016-08-04
DB updates


LindsayYoung:
- [x] make sure we delete queue immediately after the update action (Carlo)
 - [x] add rollbacks on add statements sched A and B
 - [x] only update schedule E once (get rid of command) (Lindsay)
 - [x] clear bad data (Lindsay)
 - [x] re-run updates (Lindsay)

LindsayYoung:
did the deletes and updates are running

ccostino:
@LindsayYoung and @jontours:  I just completed my first pass of moving the queue deletes to happen after the processing takes place and it yielded an important realization around when they need to take place.  That `refresh_child` method where the actual deletes and inserts take place against a child table is run for each child table in the partition.  If we clear the queues out after those delete and insert steps that we were looking at earlier (starting here: https://github.com/18F/openFEC/blob/develop/webservices/partition/base.py#L177), then the queues will be cleared after each child is processed.

In other words, as soon as we start processing the queues will be cleared out entirely after the first child table is processed.  if that happens, we'll lose the data we want to process for all of the remaining child tables.

There are two approaches that come to mind in how to address this:

 * Proceed as we originally discussed, but modify the `delete` statements for the queues to only delete the records that match what was processed for the current child table.  At the end of refreshing all of the children, we could add an additional check to see if any records still remain in the queues and if so, add a bit of information to our nightly report letting us know some data was not processed.  This could be taken even further in the future if we figure out how we want to handle the data that got stuck in the queue (theoretically this shouldn't happen, but you never know!).
* Keep the statements clearing the queues where they are, but this is entirely dependent on adding some `try/excepts` with rollback support then.

I'd like to go with the first approach, but wanted to check with both of you to see if there were any concerns or if there were any other ideas first.

LindsayYoung:
I think we might want to go with try/excepts with rollback support. If there is a failure and we use the first method, we would sill want the work that was half finished to be rolled back.  So, if we are doing rollbacks anyway, approach 2 sounds simpler to implement. 

I also like the idea of breaking up this huge nightly process with try excepts because a lot can go wrong and I don't think it is worth halting everything.

ccostino:
Okay, that's definitely doable, thanks @LindsayYoung!  In addition to this then, it'd probably make sense to not clear out the queues either in the event of any error that results in a rollback.

Big +1 on the `try/excepts` to at least let independent pieces of the nightly updates finish as opposed to everything stopping.

[...]

PR #1870, opened 2016-08-05
Improve nightly aggregate updates

ccostino:
Part of #1838
Part of #1868

This changeset introduces a few improvements and error handling around our nightly update process for the itemized aggregate data.  We have been running into an issue where a record being added to one of the itemized child tables will already exist and therefore result in a primary key violation.  This causes the rest of the processing to immediately stop and leaves our processing in a half done state.

The goal here is to increase visibility into where the process is failing and rolling back any changes made so that we are back in a predictable state.  From there, we can troubleshoot to determine what is wrong and fix it so that the next nightly processing will not fail and we do not lose the updates that should have taken place previously.  In the future, we should be able to automate this troubleshooting and set up a retry mechanism.

/cc @LindsayYoung, @jontours


ccostino:
I think it might still be worth partially clearing out the queues if a child table is successfully processed.  It would be straightforward to add in, and the checks in the `update_aggregates` method can change to check to see if any records remain in either queue.  If that's the case, then we're at least only dealing with the records that couldn't be processed for a cycle(s) as opposed to all of them.

LindsayYoung:
That sounds like a good idea

ccostino:
Thanks, I'll get that worked in next!

ccostino:
@LindsayYoung:  I think I have the partial queue clearing support worked in now.  I'm curious if you think it's still straightforward to follow or if it was better to just keep it the other way.  @jontours please feel free to take a look, too!

At the very least this does provide us with much greater visibility into where an actual failure occurs and how many records remain in the queue tables once the processing is finished.

I'm also trying to figure out how best to work in some tests for this so that we can verify that the behavior and the transactions are working correctly.

manage.py
```diff
@@ -32,6 +32,37 @@
...
+def check_itemized_queues(schedule):
+    """Checks to see if the queues associated with an itemized schedule have
+    been successfully cleared out and sends the information to the logs.
+    """
+
+    remaining_new_queue = db.engine.execute(
+        'select count(*) from ofec_sched_{schedule}_queue_new'.format(
+            schedule=schedule
+        )
+    ).first()[0]
```
vrajmohan: Very minor - `scalar()` is equivalent to `first()[0]`

manage.py
```diff
@@ -32,6 +32,37 @@
...
+    remaining_old_queue = db.engine.execute(
+        'select count(*) from ofec_sched_{schedule}_queue_new'.format(
+            schedule=schedule
+        )
+    ).first()[0]
```
vrajmohan: ditto

webservices/partition/base.py
```diff
@@ -174,39 +180,67 @@ def refresh_child(cls, cycle, queue_old, queue_new):
...
-        # table. Because of this, we need to exclude the function call in the
-        # column factory normally used when populating the child tables during
-        # normal partitioning. Otherwise, an error is thrown due more values
-        # being specified than there are columns to accept them.
-        columns = [
-            column for column in cls.column_factory(queue_new)
-            if column.name != 'two_year_transaction_period'
-        ]
+        try:
+            select = sa.select([queue_old.c.get(cls.primary)]).where(
```
vrajmohan: WARNING - SQL bias ahead.

The 2 statements being executed, especially the INSERT on line 222, would read much better in plain SQL. I am a fan of SQLAlchemy in general, but this is a case where a PostgreSQL stored procedure (function) would be more readable.

vrajmohan:
The transaction boundaries look good to me.

manage.py
ccostino: I knew there was a better way but couldn't remember, thanks @vrajmohan!

ccostino:
@vrajmohan thank you so much for walking through this with me earlier and providing feedback, I appreciate it!  I'll incorporate some improvements based on your notes and/or file an issue to address them later on.

webservices/partition/base.py
ccostino: Thank you for the feedback here!  We're aware that this is a bit of a thornier part of the code to read through and have made some small improvements over time, but this is definitely a good point to bear in mind for the future.

ccostino:
@LindsayYoung and @jontours, this is ready for full review!

manage.py
```diff
@@ -32,6 +32,37 @@
...
+            schedule=schedule
+        )
+    ).scalar()
+    remaining_old_queue = db.engine.execute(
+        'select count(*) from ofec_sched_{schedule}_queue_old'.format(
+            schedule=schedule
+        )
+    ).scalar()
+
+    if remaining_new_queue == remaining_old_queue == 0:
```
LindsayYoung: Couldn't new things be added to the quue while the updates were running?

ccostino: Hrm, you're right, technically speaking they could...  Okay, I will think on this a bit more.  That raises another question though:  the existing process would update everything and then just wholesale delete everything when done (and this will still be the case for Schedule E), so in that event, we would've been losing updates there as well.

I'm going to assume table locks aren't desired and rule them out right now, but I'll mull this over/read up on some things to see if there's a better way.

Thanks for catching and flagging this!

ccostino: @LindsayYoung: I've been thinking about this more throughout the day and I wonder if we even want to do this.  The reason I question this now is that we're only clearing the records that we've processed for Schedule A and B*, not the entire queue.  Presumably at the end the queues would be clear, but theoretically speaking records could still be added to the queue during processing as you noted above.

I wonder if this is why we've seen some issues with the Schedule E records being out of sync - the queue processing takes place, but records are being added/modified/deleted while the processing is taking place, so at the end when the queues are cleared, we lose what was added to them during that time.

Would it just make more sense to focus on removing the records that we know we processed (or even as we process them - that might be too intensive and slow, though)?  If so, then there wouldn't be a need for this check anymore.

Alternatively, I wonder if we could modify the stored procedure to just take care of this entirely for us each time a record comes through.  It could be intensive at times, but if we have a read replica in place then only the master would be dealing with it.  When the replica is refreshed with the nightly backup it'll get the new changes, so the update cadence would remain the same as it does now.

*The current implementation in this PR deletes all records in the Schedule A and B queues based on the cycle we're currently in.  Perhaps that should strictly be the list of `sub_ids` that we successfully update instead.

jontours: I would opt for removing from the queues as they are processed.

LindsayYoung:
I really like the idea of getting a read replica going and doing the updates as we get them. Assuming that doesn't slow down the updates too much.

Also, the schedule e issues have have been that there are times when they don't seem to update at all, I do think that is a separate issue.

ccostino:
@LindsayYoung and @jontours:  Okay, thank you both for the feedback!  How about this for wrapping things up here:

 * We remove the checks to see if the queues were cleared out because the queue processing in this PR (for schedule A and B at least) will clear out the records that were processed during the nightly update.  If any of those updates fail, we'll still see error output in the logs and the emails and we won't lose the records because of the rollbacks.
 * We also change the code that clears out processed records to match based on IDs that were actually processed, not just blanket assuming the whole cycle can be thrown away (this accounts for records that went in after the fact, at least as best as possible).
 * We leave schedule E alone for now and focus on other tasks since it is likely a separate issue and a read replica would help improve this entire setup even more.

How does all of this sound to both of you?

LindsayYoung:
Sounds great

jontours:
Yeah I think that works.  Of course we could get into the whole duplicate rabbit hole thing again here.  As checking that you are only deleting a processed record I think we found can be tricky right?  Meaning that just checking on `sub_id` may not be enough.  Might need to check on more columns.

ccostino:
Ah right, but with the fixed triggers in all the tables we should only ever be dealing with one version of the record at any time in the table at this point.  At least, that's the intention!

jontours:
Oh yeah, I forgot we fixed that.

ccostino:
Okay, this should be ready for final review now.  I left the actual function to check the itemized queues in `manage.py` for now but removed the calls to it.  I can remove it if folks would prefer but I wonder if it'll be helpful in the near future when we come back to the remaining parts after having a replica in place.

manage.py
```diff
@@ -32,6 +32,37 @@
 manager.add_command('runserver', Server(use_debugger=True, use_reloader=True))
 
 
+def check_itemized_queues(schedule):
```
LindsayYoung: Do we still want this in here?

tests/test_integration.py
```diff
@@ -191,6 +191,119 @@ def test_sched_a_fulltext_trigger(self):
             1,
         )
 
+    def _get_sched_a_queue_new_count(self):
```
LindsayYoung: I don't think we have any of this data in the test dataset

LindsayYoung: it uses factory data, so that is good to go

LindsayYoung:
Thanks Carlo!

manage.py
ccostino: Likely not, but I left it in for the time being since I'm sure we'll be revisiting this soon.  The calls to it have been removed, though.

PR #1874, opened 2016-08-09
adding api for advisory opinion canonical page

anthonygarvan:
Note this branch is badly named now, it actually adds a new endpoint, not a new keyword to the existing search endpoint.

### DIFF
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -18,7 +18,7 @@ class AdvisoryOpinion(utils.Resource):
 
 class Search(utils.Resource):
     @use_kwargs(args.query)
-    def get(self, q=None, from_hit=0, hits_returned=20, _type='all', **kwargs):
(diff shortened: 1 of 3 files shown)


webservices/resources/legal.py
```diff
@@ -4,19 +4,31 @@
...
+        es_results = es.search(query)
+
+        results = {"docs": [r["_source"] for r in es_results["hits"]["hits"]]}
+        return results
+
 class Search(utils.Resource):
     @use_kwargs(args.query)
-    def get(self, q, from_hit=0, hits_returned=20, type='all', **kwargs):
-        if type == 'all':
+    def get(self, q=None, from_hit=0, hits_returned=20, _type='all', **kwargs):
```
adborden: Did you mean to include `q=None` here? I think eventually that's right, but we didn't change the rest of the code to handle `None` differently.

anthonygarvan: nope, good catch!

PR #1876, opened 2016-08-11
Bump python version for cloud foundry

adborden:
With the recent cloud.gov update, the python buildpack no longer supports python 3.4.3, so let's upgrade.
https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.5.8

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.4.3
+python-3.4.5


adborden:
Part of https://github.com/18F/openFEC/issues/1877

jontours:
Passes all tests, which is the best rubric we have!  Calling it good to go.

PR #1884, opened 2016-08-12
Feature/efile files

LindsayYoung:
We want to get this going so we have time to see it in action.

One thing to flag, is that they have receipt dates with times, I think that would be a good addition to this because that is something that multiple data users have asked for. When we get that we will want to adjust our column. We might want to have a `recipt_date` that is a date and a `reciept_date_time` for the date and time, that would keep some consistency between processed and e-filing data. The sorting should be on the date-time. 

This still needs indexes to be added on the tables and then we can turn the data validation back on. I am waiting for the receipt date time to add the indexes.

I will check in with Paul about this Monday.


noahmanger:
Beautiful 😻

webservices/common/models/filings.py
```diff
@@ -12,7 +12,7 @@ class Filings(db.Model):
     candidate_id = db.Column(db.String, index=True, doc=docs.CANDIDATE_ID)
     candidate_name = db.Column(db.String, doc=docs.CANDIDATE_NAME)
     cycle = db.Column(db.Integer, doc=docs.RECORD_CYCLE)
-    sub_id = db.Column(db.BigInteger, primary_key=True)
+    sub_id = db.Column(db.BigInteger, index=True, primary_key=True)
```
ccostino: Is this the index you were referring to in the PR description, or do you mean in one of the `real_efile_*` tables?  Or both? :-)

ccostino: Ah, I see them defined below too for one of the source tables.

webservices/docs.py
```diff
@@ -857,5 +857,29 @@ def add_ytd(var):
 SUMMARY = 'Longer description of event'
 EVENT_ID = 'An unique ID for an event. Useful for downloading a single event to your calendar. This ID is not a permanent, persistent ID.'
 
+# efiling
+EFILING_TAG = '''
+Efiling endpoints provide real-time campaign finance data for electronic filers.
+
+These endpoints are perfect for watching filings roll in when you want to know the latest information. Efiling endpoints
+only contain the most recent two years worth of data and don't contain the processed and coded data that
+you can find on the other endpoints. Those endpoints are better for in-depth analysis.
```
ccostino: Thank you for the clear explanation! :-)

webservices/docs.py
```diff
@@ -857,5 +857,29 @@ def add_ytd(var):
...
+'''
+
+EFILE_FILES = 'Basic information about electronic files coming into the FEC, posted as they are received.'
+FILE_NUMBER = 'Filing ID number'
+AMENDED_BY = '''
+If this report has been amended, this field gives the file_number of the report that should be used. For example,
+if a report is amended times, the first report and the amendment will have the file_number of the final amended
```
ccostino: >if a report is amended times

Looks like a word is missing here?

webservices/docs.py
```diff
@@ -857,5 +857,29 @@ def add_ytd(var):
...
+report in the ameded_by field and the final report will have no id in the amended_by field.
+'''
+AMENDS_FILE = '''
+For amendments, this file_number is the file_number of the previous report that is being amended. See ameded_by
```
ccostino: >See ameded_by

Looks like this should be `amended_by`?

webservices/resources/filings.py
```diff
@@ -76,3 +76,40 @@ class FilingsList(BaseFilings):
...
+
+    @property
+    def args(self):
+        return utils.extend(
+            args.paging,
+            args.efilings,
+            args.make_sort_args(
+                default='-receipt_date',
+                # TODO: add indexes and turn on validation
+                # validator=args.IndexValidator(models.EFilings),
```
ccostino: Not a problem here, just a note and question:  I took a look at the `IndexValidator` more closely as I hadn't seen it before, and it makes me wonder if we should have this on the views for the itemized schedules as well.  It's not currently there and if we ever run into a situation again where the indexes get dropped or aren't set properly, it appears this would also help alleviate the problem, right?

ccostino:
Just a couple of minor things, @LindsayYoung, but overall everything looks great, thank you!  I will hold off on merging until you're able to catch up with Paul about some of the marked items.

jontours:
Just realized I should have given this a review as well Friday, I'll review it early Monday.

noahmanger:
So I took some time to work on the front-end of this, which you can take for a spin over here: https://github.com/18F/openFEC-web-app/pull/1522 (note, you'll also need this branch of fec-style: https://github.com/18F/fec-style/pull/471) 

As I worked on it I found a few things the front-end needs in order to be useful. Essentially we need a few of the fields from the main `/filings/` response:
- [ ] `document_description`: the computed name of the document, e.g. `July Quarterly 2016`
- [ ] `pdf_url`. I tried just manually building it for some filings and it worked, but would be good for the API to provide this
- [ ] `amendment_indicator`: I don't think we have the same data here, but I verified that this can be derived from the last letter of the `form_type` field on this endpoint

Are those do-able?

webservices/docs.py
LindsayYoung: Yea, I was using `amended_by` for the final amendment, but I think you have a great point that this should be `amended_by` and the current `amended_by` should be `final_amendment`

webservices/common/models/filings.py
LindsayYoung: Yea, sub_id index is already there, we need to add the other indexes manually.

webservices/resources/filings.py
LindsayYoung: I am not opposed to adding that. 

We should double check it looks at the actual sql, and not just our sqlalchemy annotations.

jontours: I think I dug into that a bit and it looks at the actual postgres table.

webservices/docs.py
ccostino: Oh interesting!  I was calling out a spelling error here but if this leads to greater clarity then all the better. :-)

webservices/schemas.py
```diff
@@ -387,6 +387,17 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
 )
 augment_schemas(FilingsSchema)
 
+EFilingsSchema = make_schema(
+    models.EFilings,
+    fields={
+        'beginning_image_number': ma.fields.Str(),
+        'ending_image_number': ma.fields.Str(),
+        'pdf_url': ma.fields.Str(),
+        'is_amended': ma.fields.Boolean(),
```
jontours: Is there any reason these fields are declared here as extensions, I believe all of those should be able to be derived from the model itself, except maybe `is_amended`?

jontours:
This looks good to go.  I'll definitely want to merge dev into my PR after this gets merged.  The args I can take advantage of within my code for that PR as well.  Well done!

LindsayYoung:
The committee_id filtering is behaving oddly locally.

webservices/resources/filings.py
ccostino: Awesome!  @jontours I'll follow up with you separately about this soon to learn more.

LindsayYoung:
@noahmanger I want to add the document_discription and pdf_url in a new PR, to keep things moving. It should be really simple.
https://github.com/18F/openFEC/issues/1887

I am not sure it is a great idea to recreate the `amendment_indicator` because it is confusing and not as straight forward. When we switch off of the warehouse filings table, I would want to move towards this system of tracking amendments for the filings table.

noahmanger:
👍 for a separate PR. I have questions about amendment indicator, but will post in that issue.

webservices/common/models/filings.py
```diff
@@ -57,3 +57,35 @@ def document_description(self):
...
+    __tablename__ = 'real_efile_reps'
+
+    file_number = db.Column('repid', db.BigInteger, index=True, primary_key=True, doc=docs.FILE_NUMBER)
+    form_type = db.Column('form', db.String, doc=docs.FORM_TYPE)
+    committee_id = db.Column('comid', db.String, index=True, doc=docs.COMMITTEE_ID)
+    committee_name = db.Column('com_name', db.String, doc=docs.COMMITTEE_NAME)
+    # this will be a date time
+    receipt_date = db.Column('filed_date', db.Date, index=True, doc=docs.RECEIPT_DATE)
+    # add docs, confirm this is the receipt time
+    load_timestamp = db.Column('create_dt', db.DateTime, doc="This is the load date and will be deprecated when we have the receipt date time")
```
jontours: Is this true for the real_efile_f3* tables as well?

ccostino:
Thank you, @LindsayYoung!

PR #1890, opened 2016-08-16
Receipt date time

noahmanger:
This is just a proposal: using the `timestamp` db column for the `receipt_date`, but keeping it called `receipt_date` rather than `receipt_date_time`. 

From a (quasi-selfish) front-end implementation standpoint, this is better because it keeps the same field for sorting on the processed table as on the efiling table. The fact that one is a date stamp and one a datetime stamp I think is ok, but maybe there's other concerns I'm not aware of. You can always just format it however you want, which is what I did here:

Here's how it looks on efiling:
image

And on processed:
image

If we think this is bad for API users, I'm ok to change, but it makes sense to me.


webservices/args.py
```diff
@@ -223,6 +223,8 @@ def make_seek_args(field=fields.Int, description=None):
 efilings = {
     'file_number': fields.List(fields.Int, description=docs.FILE_NUMBER),
     'committee_id': fields.List(IStr, description=docs.COMMITTEE_ID),
+    'min_receipt_date': fields.DateTime(description='Selects all items received by FEC after this date and time'),
```
jontours: For people who might read these descriptions would it be better to say 'date and/or time', or somehow convey that it doesn't have to be a time?

LindsayYoung: good catch!

jontours:
Ok this looks good to go, after I get the all-clear from Travis I'll merge it.

PR #1891, opened 2016-08-17
Feature/loading statutes

anthonygarvan:
Creates manage.py tasks for loading title 52 subtitle III and title 26 subtitle H, also bug fix that was causing bad counts to be returned. Also updates API to return the statutes. Also adds tests for all legal work to date.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -17,7 +17,8 @@ from webservices.rest import app, db
 from webservices.config import SQL_CONFIG, check_config
 from webservices.common.util import get_full_path
 from webservices.load_legal_docs import (remove_legal_docs, index_statutes,
-    index_regulations, index_advisory_opinions, load_advisory_opinions_into_s3)
+    index_regulations, index_advisory_opinions, load_advisory_opinions_into_s3,
+    delete_advisory_opinions_from_s3)
 
 
 manager = Manager(app)
@@ -34,6 +35,7 @@ manager.command(index_statutes)
 manager.command(index_regulations)
 manager.command(index_advisory_opinions)
 manager.command(load_advisory_opinions_into_s3)
+manager.command(delete_advisory_opinions_from_s3)
 
 def get_projected_weekly_itemized_totals(schedules):
     """Calculates the weekly total of itemized records that should have been
--- a/manage.py
+++ b/manage.py
@@ -6,12 +6,15 @@ import logging
 import subprocess
 import multiprocessing
 import re
+from zipfile import ZipFile
+from io import BytesIO
 
 import networkx as nx
 import sqlalchemy as sa
 from flask_script import Server
 from flask_script import Manager
 import requests
+from xml.etree import ElementTree as ET
 
 from webservices import flow
 from webservices.env import env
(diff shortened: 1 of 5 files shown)


manage.py
```diff
@@ -367,6 +370,93 @@ def index_advisory_opinions():
                 print("%d docs loaded" % docs_loaded)
         print("%d docs loaded" % docs_loaded)
 
+def get_xml_tree_from_url(url):
+    r = requests.get(url, stream=True)
+
+    with open('temp.zip', 'wb') as f:
```
adborden: I think it would be better to dump this in a temp file/dir with [mkstemp()/mkdtemp()](https://docs.python.org/3.5/library/tempfile.html?highlight=mktemp#tempfile.mkstemp)  for best security practices and potentially to avoid writing to disk (tmp space is probably an in-memory file system).

manage.py
```diff
@@ -367,6 +370,93 @@ def index_advisory_opinions():
...
+     open('title.xml', 'wb') as title_xml:
+        title_xml.write(title.read())
+
+    return ET.parse('title.xml')
+
+def get_title_52_statutes():
+    es = utils.get_elasticsearch_connection()
+
+    title_parsed = get_xml_tree_from_url('http://uscode.house.gov/download/' +
+                    'releasepoints/us/pl/114/219/xml_usc52@114-219.zip')
```
adborden: Grr house.gov, Y U NO HTTPS??

manage.py
```diff
@@ -367,6 +370,93 @@ def index_advisory_opinions():
...
+    tag_name = '{{http://xml.house.gov/schemas/uslm/1.0}}{0}'
+    for subtitle in title_parsed.iter(tag_name.format('subtitle')):
+        if subtitle.attrib['identifier'] == '/us/usc/t52/stIII':
+            for subchapter in subtitle.iter(tag_name.format('subchapter')):
+                match = re.match("/us/usc/t52/stIII/ch([0-9]+)/sch([IVX]+)",
```
adborden: Not a huge deal, but I prefer to use symbolic group names in regex's for readability, as well as pre-compile the regexs for performance if I'm re-using them.

manage.py
```diff
@@ -6,12 +6,15 @@
 import subprocess
 import multiprocessing
 import re
+from zipfile import ZipFile
+from io import BytesIO
```
adborden: Is this used?

adborden:
Looking pretty good! I'm impressed that you fit it into three functions in `manage.py`, but I wonder if you'll have an easier time testing it if you split it up into a class or two?

Also, not so specific to this PR, but since we're touching the search endpoint, do we want to add documentation to the API docs?

adborden:
Note to self: @anthonygarvan and @annalee are working on tests, so we were going to hold of on merging this until those are in, too.

manage.py
```diff
@@ -367,6 +370,93 @@ def index_advisory_opinions():
...
+                              'title52/pdf/USCODE-2014-title52-subtitleIII-' +\
+                              'chap%s-subchap%s-sec%s.pdf'\
+                              % (chapter, subchapter_no, section_no)
+                    doc = {"doc_id": section.attrib['identifier'],
+                           "text": text,
+                           "name": heading,
+                           "no": section_no,
+                           "title": "52",
+                           "chapter": chapter,
+                           "subchapter": subchapter_no,
```
adborden: I noticed that `subchapter` does not exist for the title 26 statutes, is that by design?

tests/test_legal.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@patch('webservices.rest.legal.es.search', side_effect=es_advisory_opinion)
+class AdvisoryOpinionTest(unittest.TestCase):
+    def test_advisory_opinion_search(self, es_search):
+        app = rest.app.test_client()
+        response = app.get('/v1/legal/advisory_opinion/1993-02?api_key=1234')
+        self.assertEquals(response.status_code, 200)
+        result = json.loads(codecs.decode(response.data))
+        self.assertNotEqual(result, [], "Empty response!")
```
adborden: This is a nitpick for sure :D but I'm not a fan when my computer yells at me. Mind replacing the "!" with a period?

adborden:
Yay tests! Do you mind adding a test for the statutes search that we just added? Also, from the API consumer perspective, they care about the schema we're returning. It's a little funny because we're just returning the schema un-modified from ES, but wondering what tests we can add here to make sure we don't break the consumer's response parsing.

jmcarp:
Since tests are run with pytest, how about using the pretty assertion that it supports? http://doc.pytest.org/en/latest/assert.html#assert-with-the-assert-statement. Why `self.assertNotEqual(1, 2)` when you can `assert 1 != 2` 😉  ?

tests/test_load_legal_docs.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ElasticSearchMock:
+    def search():
+        pass
+
+def get_es():
+    return ElasticSearchMock()
+
+class StatuteTest(unittest.TestCase):
+    @patch('webservices.utils.get_elasticsearch_connection', side_effect=get_es)
```
adborden: Just out of curiosity, if you're using a `side_effect` that just returns a value, why not just use `return_value`?

anthonygarvan: I _thought_ return_value couldn't return functions.

PR #1892, opened 2016-08-17
Feature/additional reports endpoint work

jontours:
PR for additional reports endpoint work.  Tomorrow I'll put up some notes about how the new endpoints behave (they are very similar as before, but with some slightly different behavior, but more RESTful).


noahmanger:
Will this include multiple committee ID filtering?

jontours:
Yes it will.

noahmanger:
Working like a dream! That behavior for committee filtering is great. We can enhance later by either not showing non-matching committees in the typeahead or showing an error, but this is perfect for now.

Only thing I caught is the `/reports/ie-only/` response doesn't include the committee name. But other than that, seems to work great.

ccostino:
@jontours - looks like a very recent merge caused some conflicts, but once it's resolved I can take a final look and merge!

PR #1898, opened 2016-08-20
Legal phrase search

adborden:
@anthonygarvan I'd appreciate you taking a look at this. I'll set this up on `feature` so that folks can play around with it.

### DIFF
--- a/tests/test_legal.py
+++ b/tests/test_legal.py
@@ -4,6 +4,8 @@ import codecs
 import unittest
 from mock import patch
 
+from webservices.resources.legal import parse_query_string
+
 # TODO: integrate more with API Schema so that __API_VERSION__ is returned
 # self.assertEqual(result['api_version'], __API_VERSION__)
 
@@ -61,3 +63,17 @@ class SearchTest(unittest.TestCase):
             'total_advisory_opinions': 2,
             'advisory_opinions': [{'highlights': ['a', 'b']},
               {'highlights': ['c', 'd']}], 'total_all': 2}
+
+
+class LegalPhraseSearchTests(unittest.TestCase):
+    def test_parse_query_no_phrase(self):
+        parsed = parse_query_string('hello world')
+        self.assertEquals(parsed, dict(terms=['hello world'], phrases=[]))
+
+    def test_parse_query_with_phrase(self):
+        parsed = parse_query_string('require "electronic filing" 2016')
+        self.assertEquals(parsed, dict(terms=['require', '2016'], phrases=['electronic filing']))
+
+    def test_parse_query_with_many_phrases(self):
+        parsed = parse_query_string('require "electronic filing" 2016 "sans computer"')
+        self.assertEquals(parsed, dict(terms=['require', '2016'], phrases=['electronic filing', 'sans computer']))
--- a/tests/test_legal.py
+++ b/tests/test_legal.py
@@ -4,7 +4,7 @@ import codecs
 import unittest
 from mock import patch
 
-from webservices.resources.legal import parse_query_string
+from webservices.resources.legal import es, parse_query_string
 
 # TODO: integrate more with API Schema so that __API_VERSION__ is returned
 # self.assertEqual(result['api_version'], __API_VERSION__)
@@ -14,7 +14,7 @@ def es_advisory_opinion(query, size):
             {'_source': {'no': '123'}}]}}
 
 def es_search(q, index, size, es_from):
-    _type = q["query"]["bool"]["must"][1]["term"]["_type"]
+    _type = q["query"]["bool"]["must"][0]["term"]["_type"]
     if _type == 'regulations':
         return {'hits': {'hits': [{'highlight': {'text': ['a', 'b']},
                                     '_source': {}}], 'total': 1}}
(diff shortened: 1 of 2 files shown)


tests/test_legal.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+
+from webservices.resources.legal import parse_query_string
+
+class LegalPhraseSearchTests(unittest.TestCase):
```
anthonygarvan: Can you add test cases for `"phrase only"` and `"phrase followed by" term`.  Regexes do weird things sometimes with string starts.

webservices/resources/legal.py
```diff
@@ -22,6 +24,45 @@ def get(self, ao_no, **kwargs):
...
+
+    def _parse_query_string(query):
+        """Recursively pull out terms and phrases from query. Each pass pulls
+        out terms leading up to the phrase as well as the phrase itself. Then
+        it processes the remaining string."""
+
+        if not query:
+            return ([], [])
+
+        match = phrase_regex.search(query)
```
anthonygarvan: You've probably thought through this more than I have, but is recursion really necessary here? Why not just use `re.findall(regex, str)`?

webservices/resources/legal.py
```diff
@@ -30,21 +71,32 @@ def get(self, q, from_hit=0, hits_returned=20, _type='all', **kwargs):
...
                      }},
                 "highlight": {"fields": [{"text": {}},
                     {"name": {}}, {"number": {}}]},
                 "_source": {"exclude": "text"}}
 
+            if len(terms):
+                query['query']['bool']['must'].append({"match": {"_all": ' '.join(terms)}})
+
+            for phrase in phrases:
+                query['query']['bool']['must'].append({"match_phrase": {"text": phrase}})
```
anthonygarvan: I believe we want to use the `term` condition, not the `match_phrase` condition. Term looks for an exact match, match_phrase selects documents with all those terms in the same order (but not necessarily next to each other).

tests/test_legal.py
```diff
@@ -0,0 +1,16 @@
...
+        parsed = parse_query_string('hello world')
+        self.assertEquals(parsed, dict(terms=['hello world'], phrases=[]))
+
+    def test_parse_query_with_phrase(self):
+        parsed = parse_query_string('require "electronic filing" 2016')
+        self.assertEquals(parsed, dict(terms=['require', '2016'], phrases=['electronic filing']))
+
+    def test_parse_query_with_many_phrases(self):
+        parsed = parse_query_string('require "electronic filing" 2016 "sans computer"')
+        self.assertEquals(parsed, dict(terms=['require', '2016'], phrases=['electronic filing', 'sans computer']))
```
anthonygarvan: `assert parsed == dict(...)` syntax is preferred.

webservices/resources/legal.py
adborden: Hmm, I played around with `term`, but that behavior didn't seem right. For example, `"matters"` returned results, while `"Matters"` returned no results. `"electronic filings"` also returned no results for some reason.

As far as I can tell, `match_phrase` seems to be doing the right thing. The behavior I see matches the [elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/guide/1.x/_phrase_search.html)
> ["rock climbing"] will match only employee records that contain both “rock” and “climbing” and that display the words next to each other in the phrase “rock climbing.”

There is one quirk though, the highlights sometimes includes partial phrases which I suppose is expected given the explanation above. However, the documents returned still always match the entire phrase. In other words, searching for a phrase `"electronic filing"` returns a subset of the search `electronic filing`. Also, if you reverse the terms like `"filing electronic"`, you get no results as expected. So I'm not really sure how to proceed here. If there are configuration options for the highlights, then I feel better about the `match_phrase` approach.

Not sure if you saw my comment https://github.com/18F/fec-eregs/issues/192#issuecomment-241160861 but I wonder if the Simple Query String query will actually just do everything we want.

I pushed the current (with `match_phrase`) code up to the `feature` env, so you can play with it here https://fec-feature-proxy.18f.gov/data/legal/search/?search_type=all&search=%22electronic+filing%22

anthonygarvan: @adborden Nice work! After playing around with it, this looks right to me. I thought it would capture things like "electronic form filing" as well (i.e., terms in right order but which have other words in between), but it doesn't. So 🚀 .

tests/test_legal.py
adborden: :+1: will do.

webservices/resources/legal.py
adborden: Short answer is that it seemed like a complicated regex/looping problem thinking in recursion terms helped me reason through it. It's really just a loop that is chunking over the query string using regex to chunk it up.

`findall` doesn't return a list of matches, it returns strings or tuples, so you loose the nice api from regex matches. In my head, going with `for` and `findall`, you would have a long loop (in terms of lines of code within the block, my current solution is pushing what I would want in a `for` loop to give you an idea) to split the terms from the phrases. To do that you would have a variable with the remaining string, or be using start/end pointers to keep track of your work. OR you could have a slightly more complex regex, I was thinking `((?P<terms>[^\s"]+)\s+)?"(?P<phrase>[^"]*)"` and dealing with the groups as tuples wouldn't be so fun. The regex is hard here, because the query string is pretty free-form so I was hoping to keep it simple. When looping, if you've got more than one or two variables mutating, my head just goes to a function at that point with immutable variables to reduce the complexity. This is entirely subjective, and I didn't actually write the code in the `for`/`while` loop style, so I don't really know how it would've looked.

BUT, in general, I'm all for learning new APIs and getting simpler, more readable, efficient code in there, so feel free to post PR coments with code snippets or pseudo code if you have something in mind that would work better. I certainly do that to other folks as well, at least to get their opinion on my snippets :)

Thinking about it again, something like this might be better, but I would definitely want to add more test cases. IDK, what do you think?
```python
    def _parse_query_string(query):
        terms = []
        phrases = []

        matches = re.findall('((?P<term>[^\s"]+)*\s*)("(?P<phrase>[^"]*)")?', query)
        for match in matches:
            _, term, _, phrase = match
            if term:
                terms.append(term)
            if phrase:
                phrases.append(phrase)

        return (terms, phrases)
```

adborden:
:+1: @anthonygarvan and I walked through it together.

PR #1900, opened 2016-08-25
Fix transaction id in schedules and sort bug fixes

LindsayYoung:
The sort errors are fixed by adding a default sort order

For the transaction id, we had it going twice and the transaction_id was always null. Now we just have the string version going.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -1,5 +1,3 @@
-import datetime
-
 import sqlalchemy as sa
 
 import factory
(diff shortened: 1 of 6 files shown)


LindsayYoung:
Also the transaction ids look odd to me, they are not similar if you look at different committees but they come from `tran_id`- so I think they are right.

jontours:
@LindsayYoung I'll take a look at this, looks like an integration test is failing.

jontours:
Seems the problem with the test is failing is that by adding a default sort order for `sched_b` of `disbursement_date` causes the ordering of the JSON entries to be reversed, which is kind of funny since the values are null for that.  There are probably lots of ways to fix it, maybe initialize the dates in the factory to enforce the ordering?

LindsayYoung:
Thanks, adding dates to the factory makes sense to me. I'll do that now

LindsayYoung:
the test I am seeing fail is memoed_subtotal which we don't filter by

PR #1915, opened 2016-08-30
Schedule C endpoint

jontours:
Initial rough version, verifying and adding columns, filters, and sorting.


LindsayYoung:
Great start

jontours:
This just needs documentation added for the swagger-ui and should be ready for merge.

LindsayYoung:
I am having an issue with this example:
http://localhost:5000/v1/schedules/schedule_c/2080920101128213800/

webservices/resources/sched_c.py
```diff
@@ -0,0 +1,76 @@
...
+    def index_column(self):
+        return self.model.sub_id
+
+    filter_multi_fields = [
+        ('image_number', models.ScheduleC.image_number),
+        ('committee_id', models.ScheduleC.committee_id),
+    ]
+
+    filter_fulltext_fields = [
+        ('loaner_name', models.ScheduleC.loan_source_name),
```
LindsayYoung: We are storing this as an string but for name search we need a tsvector

jontours: @LindsayYoung adding a `TSVECTOR` for this columns won't be a simple thing to do, not how it's currently implemented.  For the other itemized tables we map to the `ofec...` tables, and in those tables we create a `TSVECTOR` column for fast full-text searching.

So to add that support for this we'd have to add a new TSVECTOR column and keep it update with a trigger, as documented in this reference: https://blog.lateral.io/2015/05/full-text-search-in-milliseconds-with-postgresql/

Not a hard thing to do at all, but won't be able to get in this release, so I'll just pull that filter for now and we can get it in for the release possible.

webservices/docs.py
```diff
@@ -399,11 +399,12 @@
...
 
 SCHEDULE_C_TAG = '''
-Schdule C tag stuff
+Schedule C shows all loans, endorsements and loan guarantees a committee
+receives or makes.
 '''
 
 SCHEDULE_C = SCHEDULE_C_TAG + '''
-Schedule C stuff
+The committee continues to report the loan until it is repaid.
```
jontours: @LindsayYoung Spelling looks good to go!!

LindsayYoung:
Thanks for all the updates to it! looks good

PR #1917, opened 2016-08-31
Phrase search highlight fixes

adborden:
Fixes https://github.com/18F/openFEC/issues/1916. Some extraneous terms were highlighted because they match the `should` queries in addition to the `match_phrase` query. We specify a [highlight query](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-request-highlighting.html#_highlight_query) explicitly to only highlight matches from the phrase search.

- Introduces [elasticsearch_dsl](https://github.com/elastic/elasticsearch-dsl-py) to make it easier to construct and manipulate quries.
- Replaces pyelasticsearch with elasticsearch-py
- Includes text fields from various document types to include in the highlights (description, summary, etc.)
- Phrase search on `_all` rather than just `text`.


requirements.txt
```diff
@@ -19,8 +19,9 @@ webargs==0.18.0
 slacker==0.8.6
 raven[flask]==5.8.1
 ujson==1.33
-pyelasticsearch==1.4
```
anthonygarvan: hrmm. The load_legal_docs still uses pyelasticsearch. Want to leave it in or update those as well?

webservices/resources/legal.py
```diff
@@ -74,42 +75,50 @@ def get(self, q, from_hit=0, hits_returned=20, type='all', **kwargs):
...
-            for hit in hits:
-                highlights = []
-                if 'highlight' in hit:
-                    for key in hit['highlight']:
-                        highlights.extend(hit['highlight'][key])
-                hit['_source']['highlights'] = highlights
-            count = es_results['hits']['total']
+                term_query = Q('match', _all=' '.join(terms))
+                must_query.append(term_query)
+                text_highlight_query = text_highlight_query & term_query
```
anthonygarvan: I'm confused by the bitwise operator here.

webservices/resources/legal.py
```diff
@@ -74,42 +75,50 @@ def get(self, q, from_hit=0, hits_returned=20, type='all', **kwargs):
...
+            if len(phrases):
+                phrase_queries = [Q('match_phrase', _all=phrase) for phrase in phrases]
+                must_query.extend(phrase_queries)
+                text_highlight_query = text_highlight_query & Q('bool', must=phrase_queries)
+
+            query = Search().using(es) \
+                .query(Q('bool',
+                         must=must_query,
+                         should=[Q('match', no=q), Q('match_phrase', _all={"query": q, "slop": 50})])) \
+                .highlight('description', 'name', 'no', 'summary', 'text') \
```
anthonygarvan: nice.

webservices/resources/legal.py
adborden: It's just some syntax sugar from the DSL http://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#query-combination

requirements.txt
adborden: Yikes not sure how I missed that. I'll convert it to elasticsearch-py

In taking a look, I'm seeing some duplicate code. It looks like the legal commands were never removed from `manage.py` https://github.com/18F/openFEC/blob/develop/manage.py#L315

adborden:
@anthonygarvan thanks for the review. I've pushed an update for the `load_legal_docs` and created https://github.com/18F/openFEC/pull/1931 to clean up the duplicate code https://github.com/18F/openFEC/pull/1931

PR #1918, opened 2016-08-31
Feature/load archived murs

anthonygarvan:
This covers work to load archived murs into elasticsearch and S3. Ready for review.


webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+
+    with NamedTemporaryFile('wb+') as pdf:
+        for chunk in response:
+            pdf.write(chunk)
+
+        pdf.seek(0)
+        try:
+            pdf_doc = slate.PDF(pdf)
+        except:
+            print('Could not parse pdf: %s' % pdf_key)
```
adborden: Do you mind printing this with `logger.error`?

webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+
+        pdf.seek(0)
+        bucket.put_object(Key=pdf_key, Body=pdf,
+                          ContentType='application/pdf', ACL='public-read')
+        return pdf_text, pdf_size, pdf_pages
+
+def get_subject_tree(html, tree=[]):
+    # get next token
+    root = re.match("([^<]+)(?:<br>)?(.*)", html, re.S)
+    list_item = re.match("<li>(.*?)</li>(.*)", html, re.S)
```
adborden: What does the question mark here do? I would think its redundant with the `.*`

adborden: ah, non-greedy match :+1:

webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+    regulations = []
+
+    for citation_text in citation_texts:
+        us_code_match = re.match("([0-9]+) U\.S\.C\. ([0-9]+)", citation_text)
+        regulation_match = re.match("([0-9]+) C\.F\.R\. ([0-9]+)(?:\.([0-9]+))?", citation_text)
+
+        if us_code_match:
+            url = 'http://api.fdsys.gov/link?collection=uscode&' +\
+                  'title=%s&year=mostrecent&section=%s'\
+                  % (us_code_match.group(1), us_code_match.group(2))
```
adborden: This might be a bit nit-picky, but to be on the defensive side we might want to url encode these parameters. The regex would prevent anything fishy, but just incase of future refactoring. We don't have an explicit test case around this.

webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+            us_codes.append({"text": citation_text, "url": url})
+        if regulation_match:
+            url = 'http://api.fdsys.gov/link?collection=cfr' +\
+                  '&titlenum=%s&partnum=%s&year=mostrecent'\
+                  % (regulation_match.group(1), regulation_match.group(2))
+            if regulation_match.group(3):
```
adborden: Using group names over numbers here might make this code a little more readable. I keep scrolling up to see the regex to remember what parameters we're using here.

requirements.txt
```diff
@@ -21,6 +21,7 @@ raven[flask]==5.8.1
 ujson==1.33
 pyelasticsearch==1.4
 requests==2.10.0
+git+https://github.com/timClicks/slate.git
```
adborden: It looks like slate is on [pypi](https://pypi.python.org/pypi/slate), is there a reason we want to use `master`?

webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+    mur_names = get_mur_names()
+    data = re.findall("<td[^>]*>(.*?)</td>", mur[2], re.S)
+    mur_no = re.search("/disclosure_data/mur/([0-9_A-Z]+)\.pdf", data[0]).group(1)
+    print("processing mur %s" % mur_no)
+    pdf_key = 'legal/murs/%s.pdf' % mur_no
+    if [k for k in bucket.objects.filter(Prefix=pdf_key)]:
+        print('already processed %s' % pdf_key)
+        return
+    text, pdf_size, pdf_pages = process_mur_pdf(mur_no, pdf_key, bucket)
+    # text, pdf_key, pdf_size, pdf_pages = ([None] * 4)
```
adborden: Did you mean to remove this? Looks like an old line to help debug.

webservices/load_legal_docs.py
```diff
@@ -247,3 +261,182 @@ def load_advisory_opinions_into_s3():
...
+        # shift
+        tree.append('ul_start')
+    elif list_item or root:
+        # shift
+        subject = re.sub('\s+', ' ', (list_item or root).group(1))\
+                    .strip().lower().capitalize()
+        tree.append({'text': subject})
+    else:
+        print(html)
+        raise Exception("Could not parse next token.")
```
adborden: This looks like a pretty standard parse switch, but if you're using a particular algorithm, or can describe the high-level idea of how the parsing works, I think that would be helpful to readers.

adborden:
This is looking pretty good. I had a bit of a hard time following along, though. I know that this kind of low-level tree parsing requires a lot of details and by nature is just kind of hard to follow. Do you think there's anything we can do to make it a little easier to read? There are two general ideas that come to mind:

- Including a snippet of an example text that you're parsing. That gives the reader a better idea about what needs to be parsed. Also things like `data[4]` would make much more sense if you could see that the 5th `td` was always the subject, for example.
- Including some code comments giving a little more context and any high-level details about the algorithm's approach. I'm sure there were some memory optimizations to get the code working well on CF, but I don't see any comments referring to that kind of thing. Do you think those would be relevant?

adborden:
Do you think it's worth adding a little documentation around how to load the legal docs? As a developer it's a bit of a pain that you can't load your ES instance without deleting some PDFs off of the dev S3, and there's not really any documentation around how to do it.

requirements.txt
anthonygarvan: the pypi version doesn't work :( Known issue that has been unresolved for some time.

anthonygarvan:
@adborden I'd prefer to keep the name csv in the repo. Adding it to S3 adds a lot of complexity and is not as good in terms of easily being able to audit and modify the data. It is really just static, small data, I agree it's a bit funny but I think the repo is where it belongs.

adborden:
Awesome, thanks for addressing my comments!

PR #1919, opened 2016-08-31
Added document_description and pdf_url to all efile reports summaries

jontours:
@noahmanger @ccostino @LindsayYoung, this PR adds the missing fields needed to the efiling summaries: document_description and pdf_url.  Verified that the urls exist, also fixed an ambiguous column name: qual -> qualified_muliticandidate_committee.  Gleaned that from the linked form!!  Yay information!

### DIFF
--- a/webservices/common/models/reports.py
+++ b/webservices/common/models/reports.py
@@ -4,6 +4,11 @@ from .base import db, BaseModel
 
 from sqlalchemy.ext.hybrid import hybrid_property
 
+from webservices.common.models.dates import ReportType
+from sqlalchemy.ext.declarative import declared_attr
+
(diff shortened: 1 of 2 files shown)


jontours:
Issue here: https://github.com/18F/openFEC/issues/1905

noahmanger:
Works great for efilings. My bad for not catching this earlier, but apparently we need `document_description` for the regular reports endpoints as well (they have `report_type_full` but not `document_description`. Is that trivial to add?

jontours:
I'd say it is trivial to add.  I can add it to this PR.

jontours:
Ok, this should be ready now that `document_description` is added to processed reports.

PR #1921, opened 2016-08-31
Sorting enabled for 'receipt_date', 'committee_id', and 'file_number'.

jontours:
Basically commented out sorting argument code and added correct columns. Is ready to go.


noahmanger:
Does something need to run before this works? Trying it out locally and it's not working.

webservices/args.py
```diff
@@ -223,6 +223,8 @@ def make_seek_args(field=fields.Int, description=None):
 efilings = {
     'file_number': fields.List(fields.Int, description=docs.FILE_NUMBER),
     'committee_id': fields.List(IStr, description=docs.COMMITTEE_ID),
+    'report_id': fields.Int(),
+    'receipt_date': fields.Date(),
```
LindsayYoung: we already have `min_reciept_date` and `max_reciept_date` so I don't think we want `receipt_date` too

jontours:
@noahmanger there will be errors on sorting for this until we add the indices.

webservices/args.py
jontours: Yes you are right!  I forgot that was is passed as the sort parameter determines the sorting column, with that being said the rest of the parameters can be pulled.

PR #1922, opened 2016-08-31
Some column name corrections for efile summaries

jontours:
@LindsayYoung @ccostino This is a very lightweight change, @noahmanger and I verified some columns, there is a better than even chance that there are still some lingering typos/errors, but most have been verified with the processed committee reports.


LindsayYoung:
Presidential
 -  summary lines - this might be a heading 
 - `report` should be `report_form`
- rpt_pgi, primary election, general election are all the primary general indicator

PAC/party
- `loan_repayments_received`_period should be `total_loan_repayments_made_period`
- `loan_repayments_received_ytd` should be `total_loan_repayments_made_ytd`
... there are a lot here that should have some name tweaks- can we just put a disclaimer on the API that we are going to rename things here? 

We are missing a lot of data for presidential and house/senate.  I thought it would be in real_efile_summary but it isn't - we can follow up on this later.

jontours:
So should `loan_repayments_received_[ytd|period]` be changed for the processed committee reports columns as well?  As those columns have the same nomenclature and are what I used as the guide for these column names.

ccostino:
Awesome, thanks @jontours!  The new language looks good.

PR #1923, opened 2016-08-31
update args

LindsayYoung:
we called repid file_number to keep consistent with the current reports endpoints.


jontours:
Good to go!

Issue #1897, opened 2016-08-19
Upgrade Postgres in Amazon RDS


ccostino:
We're currently running Postgres version 9.4.5 in our Amazon RDS instances, but version 9.5.2 has been out for some time now and is officially supported by Amazon RDS.  We would be able to benefit from the security improvements and feature updates (we have had our eye on `upserts` to improve our update processes further).

Furthermore, Paul and Rohan are experimenting with the next version of GoldenGate and working with Postgres 9.5.2 locally.  It would help to be running the same version on RDS.

In order to perform the upgrades, we would follow the steps here:  http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Upgrading.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion

Note that any replicas (we have one in `dev`) will need to be recreated as a part of this process.  Because of this, we'd like to take care of this before #1415.

We'll upgrade `dev` and `stage` first and verify the upgrade steps, note any issues, and formalize a plan for production before touching the `prod` instance.

ccostino:
In reading over the upgrade steps in Amazon's RDS guide, here is everything we need to do to upgrade our instances:

1. Check for prepared transactions that are still open and either commit them or roll them back - we can check for any with this query:

  ```sql
SELECT count(*) FROM pg_catalog.pg_prepared_xacts;`
  ```

2. Remove all `reg*` data types - we can check for any with this query:

  ```sql
SELECT count(*) FROM pg_catalog.pg_class c, pg_catalog.pg_namespace n, pg_catalog.pg_attribute a 
  WHERE c.oid = a.attrelid 
      AND NOT a.attisdropped 
      AND a.atttypid IN ('pg_catalog.regproc'::pg_catalog.regtype, 
                         'pg_catalog.regprocedure'::pg_catalog.regtype, 
                         'pg_catalog.regoper'::pg_catalog.regtype, 
                         'pg_catalog.regoperator'::pg_catalog.regtype, 
                         'pg_catalog.regconfig'::pg_catalog.regtype, 
                         'pg_catalog.regdictionary'::pg_catalog.regtype) 
      AND c.relnamespace = n.oid 
      AND n.nspname NOT IN ('pg_catalog', 'information_schema');
  ```

[...]

PR #1929, opened 2016-09-08
Modify mirror support to redirect all read traffic to followers

### LINKED ISSUE #1415: Add mirrors to production
So that our servers can handle exporting receipts and disbursements, make the necessary infrastructure upgrades.

- [x] Send proposal to FEC
- [x] Purchase

ccostino:
Depends on #1897 (finished and closed)
Closes #1415 (but also depends on it as there is a budget component to adding more replicas)

This changeset modifies the mirror support currently in place to redirect all of our read traffic to any configured follower (read replica) databases. The code checking for specific tasks might still be useful in the future, so I added a little extra support to toggle restricting traffic to tasks only that can be adjusted via an environment variable.

I opted to go this route instead of just having the setting be configured in the code directly because if we ever wanted/needed to change things on the fly, it's a lot quicker to modify the environment variable and `restage` the application vs. committing a code change and redeploying the app (especially given our `pandas` dependency at the moment).

I initially had some concerns about shifting over all read operations wholesale to the replica(s), but after some discussion (thanks, @jontours!) and more thinking I came to the conclusion that this should be okay.  We currently have all traffic going to the source database in each environment at the moment, and the read replicas in RDS are entirely separate instances with their own resources associated with them (in other words, they're not in a direct master/slave relationship).

One major upside to this is that it allows us to revisit the nightly update processes and possibly explore more adhoc/realtime updates in the source databases, and any such activities should no longer have a performance impact on the user-facing side since they shouldn't ever happen in the read replicas.  Another upside is the streaming replication support offered by Postgres in RDS: http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ReadRepl.html#USER_ReadRepl.PostgreSQL - we should see things flow through from the parent database right into the read replica(s).

There are several improvements and additions made to the README in order to support and document this change (as well as minor corrections our our workflow).

Lastly, this also updates SQLAlchemy to the latest patch level release of the 1.0 series.

/cc @LindsayYoung and @jontours


requirements.txt
```diff
@@ -38,4 +38,4 @@ redis==2.10.5
 
 #efiling
 pandas==0.16.2
-xlrd==0.9.4
\ No newline at end of file
+xlrd==0.9.4
```
ccostino: This is showing up as a change because of a missing new line at the end of the file currently.

jmcarp:
> The small downside to this is that the read replicas depend on the automated backup schedule to have the latest data found in the source databases

Are you saying that the read replicas are only updated nightly? Don't rds read replicas use streaming replication?

ccostino:
Yeah, that's what I thought, but I got confused by what I was reading in the documentation.  The Postgres documentation clearly talks about the streaming replication support: http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ReadRepl.html#USER_ReadRepl.PostgreSQL - I've seen the `WAL data` statements in the logs too, that should've reminded me as well.

jontours:
@ccostino This looks good to go for me.  It's not clear to me how all read request to the database get directed to the mirror though.  For example, when some hits the receipts endpoint and generates the http request, what is the sequence of operations that eventually passes this to the mirror?  Or is that not configured yet?

README.md
```diff
@@ -153,6 +161,14 @@ export SQLA_CONN=<psql:address-to-box>
    
    *Note: 18F team members should not set these environment variables. 18F and FEC team members will also have additional environment variables to set up. Please reach out to a team member for detailed information.*
 
+4. If you are using database replicas/mirrors you can also restrict connections to them to be asynchronous tasks only by running:
+
+   ```
+   export SQLA_RESTRICT_FOLLOWER_TRAFFIC_TO_TASKS=True
```
jmcarp: Using `True` here implies that the value of the environment variable might be evaluated literally, so that valid options would be `True` or `False`. But this is actually a string passed to `bool`, so any non-empty string will become `True`, including `False`. Might want to either change the parsing logic or change the example value for clarity.

ccostino:
@jontours:  Ah, no worries, great question!  The magic is here: https://github.com/18F/openFEC/pull/1929/files#diff-cf1769e032ea11237eb42fa0478e03a9R27

Specifically, that first check for `self._flushing` is what's looking for the read or write operations.  This is making use of some of the flag that SQLAlchemy sets on a session to determine whether or not a write is taking place.  Quote from the blog post referenced in the code:

>The SQLAlchemy Session makes all decisions about what Engine to use within the get_bind() method. This method is given pretty much the same information that db_for_read() and db_for_write() receive, where a Mapper is passed in, if available, and we can also check if the operation is a "write" just by checking if the Session is flushing.

More can be read in [the documentation on flushing](http://docs.sqlalchemy.org/en/rel_1_0/orm/session_basics.html?highlight=flushing#flushing).

Checking for the presence of this flag and if any followers have actually been set initially determines if it's a read operation that can be directed to a mirror.  The additional changes I made add a secondary check to see if we just want to shift Celery task traffic over instead of all read traffic.  If yes, then it'll also check to see if the request is tied to a task and if so, one that is supported.

In either case, this ultimately results in `True` or `False` to use a mirror.  If this is `True`, then the code in the `get_bind()` method randomly chooses a mirror/replica and returns its connection string to use for the connection.  If it's not, then we do what we've always done and rely on the default behavior, which is configured to use the `SQLA_CONN` connection string upon app startup.

Summary:

 * Any request that comes in for the database is ultimately part of a session and routed through the session object.
 * The session checks to see if the operation is a read or write by checking the `_flushing` flag.
 * If the operation is not currently flushing, it's a read and we can move forward with checking to see if any replicas/mirrors are available to use to route the request to.
 * (Optional) We also check to see if the operation is coming from a Celery task.

Did that answer your question?

README.md
```diff
@@ -372,27 +389,41 @@ We use git-flow for naming and versioning conventions. Both the API and web app
...
     ```
 
-* Reviewer merges hotfix branch into develop and master and pushes to origin
-* [auto] Develop is deployed to dev
-* [auto] Master is deployed to prod
+* Reviewer merges hotfix branch into `develop` and `master` and pushes to `origin`:
+
+    ```
+    git flow hotfix finish my-hotfix
+    git checkout master
```
jmcarp: Do you need these `checkout` lines?

webservices/common/models/base.py
```diff
@@ -16,18 +16,31 @@ def followers(self):
...
+    @property
+    def use_follower(self):
+        # Check for read operations and configured followers.
+        use_follower = (not self._flushing and self.followers)
+
+        # Optionally restrict traffic to followers for only supported tasks.
+        if use_follower and self.restrict_follower_traffic_to_tasks:
+            use_follower = (
+                celery.current_task and
+                celery.current_task.name in self.replica_tasks
```
jmcarp: Since `replica_tasks` got renamed to `follower_tasks`, does this work?

README.md
```diff
@@ -372,27 +389,41 @@ We use git-flow for naming and versioning conventions. Both the API and web app
...
+* `master` is deployed to `prod`
 
 ### Creating a release
-* Developer creates a release branch and pushes to origin:
+* Developer creates a release branch and pushes to `origin`:
 
     ```
     git flow release start my-release
-    git flow release publish my-release
+    git push origin release/my-release
```
jmcarp: Why not use the `publish` helper?

README.md
```diff
@@ -427,3 +458,8 @@ caching, and rate limiting.
...
+
+### Database mirrors/replicas
+Database mirrors/replicas are supported by the API if the `SQLA_FOLLOWERS` is set to one or more valid connection strings.  By default, setting this environment variable will shift all `read` operations to any mirrors/replicas that are available (and randomly choose one to target per request if there are more than one).
+
+You can optionally choose to restrict traffic that goes to the mirrors/replicas to be the asynchronous tasks only by setting the `SQLA_RESTRICT_FOLLOWER_TRAFFIC_TO_TASKS` environment variable to something that will evaluate to `True` in Python (simply using `True` as the value is fine).  If you do this, you can also restrict which tasks are supported on the mirrors/replicas.  Supported tasks [are configured here](https://github.com/18F/openFEC/blob/feature/modify-mirror-support/webservices/rest.py#L69) by adding their fully qualified names to the list in order to whitelist them.  By default, only the `download` task is enabled.
```
jmcarp: Looks like the relevant line might be 71, but this is a brittle way to describe the relevant config setting--how about referring to the name instead?

README.md
ccostino: Yes, you're right, and I'm just not quite sure how best to approach this given that an environment variable is just a string that needs to be cast if it's being used in any other capacity (and probably over thinking it).  What do you recommend setting this to instead?  Would using just a descriptive word such as `on` or `enabled` make more sense here?

README.md
ccostino: I ran into an issue with this last time I tried to use it, perhaps I had something misconfigured though.  I'll take another look.

README.md
ccostino: Yes, definitely a better idea, will change!

webservices/common/models/base.py
ccostino: D'oh, thanks for catching.  It does work, provided the names match obviously. :-)

ccostino:
Thanks @jontours, and thanks for the second pair of eyes @jmcarp!  I will get some changes in shortly to address the issues raised.

README.md
ccostino: Will `publish` take care of pushing multiple branches (e.g., `develop` and `master` after things have been merged from finishing, and will it also take care of ensuring that the tags are pushed as well in the case of `master`?  This was part of my initial question and issue with not being able to use it, or use it correctly at least.

ccostino: I'm not seeing mention of this in the [main documentation](https://github.com/nvie/gitflow#creating-featurereleasehotfixsupport-branches) and I see explicit references to doing a `git push` in [the cheatsheet](http://danielkummer.github.io/git-flow-cheatsheet/) for a release (sadly no direct link to the section).

ccostino:
I'll also attempt to use the full PR review tool next time, apparently I needed to refresh the page to get all of the links/buttons for it.

PR #1930, opened 2016-09-08
[merge with git flow] Hotfix: import all subparts from regulations

adborden:
This fixes the issue fec-partners were seeing when trying to search for regulations numbers and not seeing them as the top hit.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -10,6 +10,7 @@ import networkx as nx
 import sqlalchemy as sa
 from flask_script import Server
 from flask_script import Manager
+import requests
 
 from webservices import flow, partition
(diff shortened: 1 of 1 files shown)


adborden:
@anthonygarvan commented that a hotfix isn't required, we can can just run the reindex job for regulations.

manage.py
```diff
@@ -10,6 +10,7 @@
 import sqlalchemy as sa
 from flask_script import Server
 from flask_script import Manager
+import requests
```
ccostino: Thank you!  There's another import missing as well, the builtin `re` module (see around line 421).

PR #1938, opened 2016-09-12
sched_d endpoint (debts)

jontours:
Initial commit, still need to add documentation and verify sorting and filtering.


webservices/common/models/itemized.py
```diff
@@ -214,6 +214,59 @@ def pdf_url(self):
...
 
 
+class ScheduleD(PdfMixin,BaseItemized):
+    __tablename__ = 'fec_vsum_sched_d'
+
+    sub_id = db.Column(db.Integer, primary_key=True)
+    original_sub_id = db.Column('orig_sub_id', db.Integer)
+    committee_name = db.Column('cmte_nm', db.String, doc=docs.COMMITTEE_NAME)
+    creditor_debtor_id = db.Column('cred_dbtr_id', db.String)
+    creditor_debtor_name = db.Column('cred_dbtr_nm', db.String)
```
LindsayYoung: To be clear, we can add it later- but it will be fun!

jontours: Ok, I have an open PR on schedule_c that addresses adding a ts_vector field to an existing table and indexes it: https://github.com/18F/openFEC/pull/1939.  It was a bit different than what we've done with things in the past as we usually add the ts_vectors in the prepare_xyz sql statements.

To handle a table we aren't creating a trigger was needed to maintain the vectorized column.

webservices/spec.py
```diff
@@ -58,6 +58,11 @@
 
         },
         {
+            'name': 'debts',
+            'description': docs.SCHEDULE_D_TAG,
+
```
jontours: Oops do y'all catch those with a linter or just eagle eye?  Fixing now.

PR #1939, opened 2016-09-12
Enable search for candidate_name and loaner_name

jontours:
Sql needs to be added for this.  Which will basically be a new function and a trigger.


noahmanger:
Changed this PR to target `develop` and now there's conflicts.

jontours:
I can handle that, it's just the subset data.

data/sql_setup/prepare_schedule_c.sql
```diff
@@ -0,0 +1,22 @@
+-- Create trigger to maintain Schedule C search columns
+ALTER TABLE fec_vsum_sched_c DROP COLUMN if EXISTS candidate_name_text;
+ALTER TABLE fec_vsum_sched_c DROP COLUMN if EXISTS loan_source_name_text;
+
+ALTER TABLE fec_vsum_sched_c add column candidate_name_text tsvector;
+ALTER TABLE fec_vsum_sched_c add column loan_source_name_text tsvector;
+
+UPDATE fec_vsum_sched_c SET candidate_name_text = to_tsvector(cand_nm);
+UPDATE fec_vsum_sched_c SET loan_source_name_text = to_tsvector(loan_src_nm);
```
ccostino: What is the purpose of having these two `update` statements here given the trigger defined below?  I'm assuming for initializing but wanted to be sure.

webservices/args.py
```diff
@@ -456,8 +456,8 @@ def _deserialize(self, value, attr, data):
 
 schedule_c = {
     'committee_id': fields.List(IStr, description=docs.COMMITTEE_ID),
-    #'candidate_name': fields.List(fields.Str, description='Name of candidate'),
-    #'loaner_name': fields.List(fields.Str, description='Name of loaner'),
+    'candidate_name': fields.List(fields.Str, description='Name of candidate'),
+    'loaner_name': fields.List(fields.Str, description='Name of loaner'),
```
ccostino: Should we use `docs.CANDIDATE_NAME` for the description of `candidate_name` and create a `docs.LOANER_NAME` entry for the `loaner_name` description?

data/sql_setup/prepare_schedule_c.sql
jontours: These statements are needed to update all the vector columns with the input from their respective native column that already exists.  Note up above the columns are dropped if they already exist.  So essentially any time the prepare directory is ran through a manager task, these two columns are dropped and created, just for integrity's sake.  I do agree it's not needed, but if those two statements aren't here they should at least be commented out for the purposes of documentation, as someone will have to run these from the console.

ccostino: Gotcha, thanks!

webservices/args.py
ccostino: Ah, but what about using them here too for the schema descriptions?

ccostino:
Awesome, thanks @jontours!

PR #1941, opened 2016-09-13
Fix py.test warning

vrajmohan:
With py.test >= 3.0, the section name should be [tool:pytest]
Pin to latest versions of pytest and pytest-cov


LindsayYoung:
Thanks @vrajmohan!

PR #1942, opened 2016-09-13
Changing rate limit docs

noahmanger:
I think I made the change in the right place. @LindsayYoung is this right?

Resolves https://github.com/18F/openFEC/issues/1926


LindsayYoung:
Yes, thank you!

PR #1943, opened 2016-09-14
Feature/make efile parsing task

jontours:
This feature removes pandas from the API build, and adds a task to manage.py that generates json file for further parsing once the API loads.


LindsayYoung:
For testing this, we should run `./manage.py load_efile_sheets` and look at the endpoints- anything else?

LindsayYoung:
We will also want to move pandas from requiremens to the dev requirements, right?

jontours:
Yes, I can take care of that.  Also sorry for not following up to the previous question.  The task method `load_efile_sheets` only has to be run if any of the schema for the .csv provided by the FEC changes.  All it does is regenerate the json that is used to dereference the columns in the spreadsheet.  As is the most up to date json is provided with this PR in the `/data` directory of the API.

LindsayYoung:
Looks like we need pandas locally for the integration tests to pass. But I think they passed before when they were just in the dev requirements.

manage.py
```diff
@@ -296,9 +300,29 @@ def cf_startup():
         subprocess.Popen(['python', 'manage.py', 'update_schemas'])
 
 @manager.command
-def test_util():
-    df = efile_parser.get_dataframe(6)
-    efile_parser.parse_f3psummary_column_b(df)
+def load_efile_sheets():
```
LindsayYoung: Let's add some in-line documentation and maybe a reference to this in the wiki so that people in the future will be able to find this.

jontours:
Whoa, so I think you merged this before the tests were done.  oops.  Want me to reopen it?

PR #1945, opened 2016-09-16
Feature/enable additional downloads

### LINKED ISSUE #1925: Add downloads for new resources
Enable downloads for: 
 - reports 
 - efiling reports
 - efiling files

ccostino:
Closes #1925
Supports https://github.com/18F/openFEC-web-app/pull/1587

This changeset adds support for additional downloads/exports at other endpoints.  Specifically, it adds support for Filings, E-Filings, and Reports.

/cc @LindsayYoung, @jontours


tests/factories.py
```diff
@@ -193,10 +193,26 @@ def update_fulltext(obj, create, extracted, **kwargs):
...
+    file_number = factory.Sequence(lambda n: n)
+
+    class Meta:
+        model = models.EFilings
+
+
+class BaseFilingFactory(BaseFactory):
+    file_number = factory.Sequence(lambda n: n)
+
+class BaseF3PFilingFactory(BaseFilingFactory):
```
jontours: Might be good to add factories for the other base classes for efilings.  But we can take that one for the issue of adding tests for efilings.

ccostino: @jontours:  I completely agree with you on adding more of these, but think it would be okay to shift that work to the issue of adding more tests in general.

PR #1948, opened 2016-09-21
Feature/add mur no endpoint

anthonygarvan:
This makes the AdvisoryOpinions endpoint generic to fetch any doc by its `no`. Updates AO test and adds one for getting mur by `no`.


adborden:
:+1: Sweet, nice work re-using the advisory opinion API. Do you want to leave the old `/legal/advisory_opinions/:ao_no` up so that there's no downtime between updating the API and updating the web app?

adborden:
FWIW, [add_resource](http://flask-restful-cn.readthedocs.io/en/latest/api.html?highlight=add_resource#flask_restful.Api.add_resource) allows specifying multiple URLs, but because of the singular vs plural `advisory_opinions` the fix to keep the old route is more involved.

PR #1951, opened 2016-09-22
Adds committee names back into itemized schedule downloads

### LINKED ISSUE #1940: Include committee name in all schedule downloads
In order for users to understand the committees associated with the data they download, include the committee name in each row.

At some point, we removed the committee object from the schedule download files for the purposes of removing all the extraneous committee fields that came with it, but in so doing, we removed the actual committee name from the file, whether the spender or recipient. 

This needs to be implemented for:
- [x] Schedule A
- [x] Schedule B
- [x] Schedule E

Other issues referencing this bug:
https://github.com/18F/FEC/issues/467
https://github.com/18F/openFEC-web-app/issues/1438
https://github.com/18F/openFEC/issues/1894

ccostino:
Closes #1940 

This changeset addresses #1940 by adding support for related fields back into our exports on the itemized tables.  It does this by defining an extra attribute in our schema Meta objects that is ignored by everything but our export code.  This extra attribute allows you to set additional columns that are found in nested models related to the parent model you are trying to export.  If any are defined, the export functionality will follow the relationship to pull the information out and display it in the output with the specified label.  Documentation in the schema provides more information on how to do this.

This changeset also modifies our Schedule E model by removing the `committee_name` field set directly to the `cmte_nm` field in the table.  This field can sometimes be `null`, so we should be pulling it from the related `CommitteeHistory` model instead, just like the Schedule A and B models already do.

Lastly, this also supports an option to sort the columns in the output alphabetically.  This is disabled for now, but there is support to insert these additional conditional columns into specific positions as opposed to just the end.  This is used for the new `committee_name` column added back in for this PR.

/cc @LindsayYoung and @jontours

### DIFF
--- a/webservices/schemas.py
+++ b/webservices/schemas.py
@@ -2,6 +2,8 @@
 import re
 import functools
 
+from collections import namedtuple
+
 import marshmallow as ma
 from marshmallow_sqlalchemy import ModelSchema
 from marshmallow_pagination import schemas as paging_schemas
@@ -18,6 +20,19 @@ from sqlalchemy import func
 spec.definition('OffsetInfo', schema=paging_schemas.OffsetInfoSchema)
 spec.definition('SeekInfo', schema=paging_schemas.SeekInfoSchema)
 
+# A namedtuple used to help capture any additional columns that should be
+# included with exported data:
+# field: the field object definining the relationship on a model, e.g.,
+#   models.ScheduleA.committee (an object)
+# column: the column object found in the related model, e.g.,
+#   models.CommitteeHistory.name (an object)
+# label: the label to use for the column in the query that will appear in the
+# header row of the output, e.g.,
+#   'committee_name' (a string)
+
+# Usage:  Define a custom attribute in a schema's Meta options object called
+# 'relationships' and set to a list of one or more relationships.
+Relationship = namedtuple('Relationship', ['field', 'column', 'label', ])
 
 class BaseSchema(ModelSchema):
 
@@ -405,6 +420,13 @@ ScheduleASchema = make_schema(
             'contributor_employer_text',
             'contributor_occupation_text',
         ),
+        'relationships': [
+            Relationship(
+                models.ScheduleA.committee,
+                models.CommitteeHistory.name,
+                'committee_name',
+            ),
+        ],
     }
 )
 
@@ -501,6 +523,13 @@ ScheduleBSchema = make_schema(
             'recipient_name_text',
             'disbursement_description_text'
         ),
+        'relationships': [
+            Relationship(
+                models.ScheduleB.committee,
+                models.CommitteeHistory.name,
+                'committee_name',
+            ),
+        ],
     }
 )
 ScheduleBPageSchema = make_page_schema(ScheduleBSchema, page_type=paging_schemas.SeekPageSchema)
@@ -522,6 +551,13 @@ ScheduleESchema = make_schema(
         'exclude': (
             'payee_name_text',
         ),
+        'relationships': [
+            Relationship(
+                models.ScheduleE.committee,
+                models.CommitteeHistory.name,
+                'committee_name',
+            ),
+        ],
     }
 )
 ScheduleEPageSchema = make_page_schema(ScheduleESchema, page_type=paging_schemas.SeekPageSchema)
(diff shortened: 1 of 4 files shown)


ccostino:
I was thinking about this some more, would it make sense to add an index to the `name` column on the `ofec_committee_history_mv` view?

LindsayYoung:
That table is not that big so I definitely think it is worth adding indexes where they help. 

Also, have not taken a deep look yet, but thanks for expanding the docstrings!

ccostino:
Thanks, @LindsayYoung!  I can add that as a part of this PR, will take care of first thing next week.

LindsayYoung:
sorry I have not gotten to this yet

ccostino:
No worries at all @LindsayYoung!

webservices/tasks/download.py
```diff
@@ -71,20 +71,50 @@ def parse_kwargs(resource, qs):
...
     exclude = getattr(schema.Meta, 'exclude', ())
+    relationships = getattr(schema.Meta, 'relationships', [])
+    joins = []
     entities = [
         entity for entity in query_entities(query)
         if entity.key not in exclude
     ]
-    return query.with_entities(*entities)
+
+    for relationship in relationships:
```
jontours: this is clever!

webservices/schemas.py
```diff
@@ -18,6 +20,34 @@
...
+# at; defaults to -1 (end of the list), e.g.,
+#   1 (an integer, in this case the second spot in a list)
+
+# Usage:  Define a custom attribute in a schema's Meta options object called
+# 'relationships' and set to a list of one or more relationships.
+#
+# Note:  There is no clean way to provide default values for a namedtuple at
+# the moment. This wrapper is modeled after the following post:
+# https://ceasarjames.wordpress.com/2012/03/19/how-to-use-default-arguments-with-namedtuple/
+class Relationship(namedtuple('Relationship', 'field column label position')):
```
jontours: Since the schemas include the model it would seem to reason that it would recognize that relationship method defined on the model, where this seems duplicative.  But we discussed it and it seemed like you tried to get the downloads to work that way.  That's really the only thing I'm iffy on.

webservices/schemas.py
```diff
@@ -405,6 +435,14 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
             'contributor_employer_text',
             'contributor_occupation_text',
         ),
+        'relationships': [
```
jontours: And in that same vein of being duplicative, was it verified that the query supporting this relationship is only made once per request?  My concern is that if `relationship` is defined both here in the schema and on the model, that it may be doing the query twice?  Perhaps?  I don't know.  But with `schedule_a` and `schedule_b` for say the last cycle, that could be an extra join that is definitely not wanted.

webservices/schemas.py
ccostino: The schemas _do_ recognize the relationship, but it is a nested relationship, and while it is provided in the output when hitting the API endpoint directly, those relationships are _not_ followed in the export code.  When we reach that code, it does not follow those nested relationships and reforms the query to be against the primary model and all of its fields itself, nothing more.

webservices/schemas.py
ccostino: As mentioned above, the query gets reformed during the export code to scale back to just the model itself, and this is a field that is on a related/nested model.  Explicitly looking at this information (and in order to get it, we ultimately need the actual column to query against and the label in which we'd like it to be called).

At the point of where this is processed, we're only dealing with the schema `Meta` information and the model itself.  Adding this `relationships` component to the `Meta` object allows us to explicitly define these things on an as-needed basis without having it impact the existing infrastructure/API output.  `marshmallow` and `flask` no nothing about this object except in the export code, which explicitly looks for it.

Since the query is re-written to be just the fields on the primary model we're dealing with (in this case, the itemized schedule data), there are no joins to begin with.  Any relationship added here (and the export code checks to see if the relationship between tables has already been added so it doesn't add it more than once) will only add that specific join.

In the end, all of this code only adds a single join between the schedule table being exported and the committee history associated with a record (based on a `committee_id`).  The query ends up being a lot simpler then the API endpoint query. :-)

jontours: Thanks for the explanation!

PR #1961, opened 2016-10-03
Map archived citations to current statutes

adborden:
Fixes https://github.com/18F/fec-eregs/issues/263

In 2012, some of the FEC related statutes in title 2 and 42 were remapped to title 52. This table [shows the mapping](http://uscode.house.gov/editorialreclassification/t52/Reclassifications_Title_52.html). The table is included as a csv so old statute names can be looked up and remapped to new statutes. Both the citation text and the url reflect the new mapping.

This creates a new task `remap_archived_murs_citations` which will fetch archived MURs from elasticsearch, parse the title/section from the citation texts, remap them to the new title, and finally, save them back to elasticsearch.

@vrajmohan not sure how far you are with the regulations remapping, but you might be able to hook into this simply by changing how the the regulation urls are created.


webservices/load_legal_docs.py
```diff
@@ -368,6 +371,33 @@ def get_citations(data):
             raise Exception("Could not parse citation")
     return {"us_code": us_codes, "regulations": regulations}
 
+def map_pre2012_citation(title, section, archived_mur_citation_map={}):
```
vrajmohan: There is already a function that does this for current MURs - https://github.com/18F/openFEC/blob/develop/webservices/load_current_murs.py#L154. I suggest we refactor and use it here.

adborden:
@vrajmohan I just rebased this. I started [working on the refactor](https://github.com/18F/openFEC/compare/bug/map-archived-citations...adborden/refactor-reclassify-citations) you suggested, but I'm not sure it's correct. Happy to chat with you about it tomorrow. For current MURs, the citation is [assumed to be title 2](https://github.com/18F/openFEC/compare/bug/map-archived-citations...adborden/refactor-reclassify-citations#diff-93ebef09b4ba513c939dc8e5a2fd3fcaR121) but the archived MURs use both title 2 and 42, and if a citation has not been remapped, there's a fall back where it uses the unmodified title/section. This is the case when the statute has been repealed before 2012. The old citation is correct in that case. This also allows for idempotency when running the remap task.

Is that behavior okay for the current MUR case as well? I suspect it never happens in the data, but [we would fall back to title 2](https://github.com/18F/openFEC/compare/bug/map-archived-citations...adborden/refactor-reclassify-citations#diff-c191a9e2a03f75d373e3b8ba21b98de5L26) in that case.

adborden:
@vrajmohan this is ready to go.

FYI @anthonygarvan I chatted with @vrajmohan this morning and suggested we move all the legal loading code under a common directory, so I moved them under `webservices/legal_docs/`.

adborden:
After this is merged, but _before_ deleting the branch, you can merge the branch into `release/public-beta-20161012`.

data/archived_mur_citation_map.csv
```diff
@@ -0,0 +1,252 @@
+2:431,52:30101
```
vrajmohan: 1. We should eliminate the duplicated lines
2. We should not name the file with an _archived_ prefix
3. Better still, why not eliminate the CSV file, reading and caching it and have a map in Python like the old `reclassify_statutory_citation`?

tests/test_map_citations_murs.py
```diff
@@ -0,0 +1,164 @@
...
+        # spot check a few cases from the csv
+        assert reclassify_statutory_citation.reclassify_pre2012_citation('2', '431') == ('52', '30101')
+        assert reclassify_statutory_citation.reclassify_pre2012_citation('2', '437g') == ('52', '30109')
+        assert reclassify_statutory_citation.reclassify_pre2012_citation('42', '1973aa-1a') == ('52', '10503')
+
+        # and a fallback
+        assert reclassify_statutory_citation.reclassify_pre2012_citation('99', '12345') == ('99', '12345')
+
+
+    @mock.patch.object(reclassify_statutory_citation, 'reclassify_pre2012_citation')
```
vrajmohan: Why mock this? Isn't it a small and fast enough function that we can actually test it?

adborden: I'm a fan of testing single components at a time, so this test is for testing `get_citations` and assumes `reclassify_pre2012_citation` is tested elsewhere (it is). The tests below are at a "higher level", but simpler in the test coverage. There's also another test that does "the whole thing" by running the `load_archived_murs` job.

The purpose of the mock is to test that `get_citations` parses the title and section properly and then constructs the citation text and url properly. Overall this approach has these advantages
- it's clear to see inputs and outputs
- you don't have to think about (or debug) what `reclassify_pre2012_citation` is doing
- it's easy to add corner cases to this test

data/archived_mur_citation_map.csv
adborden: I think using a data file is the preferred way, to keep the data separate from the code. It's also the approach we use for other data loading tasks, like the [archive MUR names](https://github.com/18F/openFEC/blob/develop/webservices/load_legal_docs.py#L382).

Hmm, not sure how I didn't notice the duplicates, but I'll remove them and rename the file.

data/mur_citation_map.csv
```diff
@@ -0,0 +1,172 @@
+2:431,52:30101
```
vrajmohan: My eyes could be deceiving me, but I still see duplicates. Lines 1-4, for instance.

porta-antiporta: @adborden I think this is generated because on uscode it has a complete map for all of US2 to US52.  However, on http://www.fec.gov/law/feca/feca.pdf it focuses only on the FECA (parts of US 52 the FEC administers).

The duplicate arises because 2:431 note can either map to 52:10101 or 52:30101 when taken as a whole.  But FEC will only ever use 30101.

adborden: Ah, thanks for catching this folks. I was using the complete title 52 table, but it sounds like that's unnecessary and the original `reclassify_statutory_citation` method would've worked fine here. I'll simplify this once more.

adborden:
@porta-antiporta @vrajmohan thanks for catching that. Now that there's only ~30 citations to map, I've moved it into a python `dict` as per your suggestion.

PR #1962, opened 2016-10-04
Fixes date tests

jontours:
@ccostino @LindsayYoung This is the fix for the tests.  We talked about changing how filtering works on the endpoint to fix this.  But after looking at this in the morning it was unclear to me how to modify the endpoint without breaking previous functionality.  This change allows the tests to be year independent and still tests the logic in the same way as before, which to me is the most lightweight, low-impact way to support this test.

If there are still some changes that we think we need to make on the endpoint, I will have to talk to someone about it more, as I don't want to break anything.  Mostly it's unclear how to override default behavior as we have no arguments for `start_date` or `end_date`.  We have `min` and `max` for both of those, but I didn't feel comfortable adding what could potentially be complicated logic just to fix a couple of breaking tests.

### DIFF
--- a/tests/test_dates.py
+++ b/tests/test_dates.py
@@ -109,19 +109,20 @@ class TestCalendarDates(ApiBaseTest):
 
 
 class TestCalendarExport(ApiBaseTest):
-
+    today = datetime.datetime.now()
     def setUp(self):
         super().setUp()
+
         self.dates = [
             factories.CalendarDateFactory(
                 category='election',
-                start_date=datetime.datetime(2015, 10, 1),
+                start_date=datetime.datetime(self.today.year, 10, 1),
                 all_day=True,
             ),
             factories.CalendarDateFactory(
                 category='Roundtables',
-                start_date=datetime.datetime(2015, 10, 31, 2),
-                end_date=datetime.datetime(2015, 10, 31, 3)
+                start_date=datetime.datetime(self.today.year, 10, 31, 2),
+                end_date=datetime.datetime(self.today.year, 10, 31, 3)
             ),
         ]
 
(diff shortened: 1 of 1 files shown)


ccostino:
This seems straight forward and makes sense to me, thanks @jontours!  All good for you @LindsayYoung?

ccostino:
I wonder if this should also be setup as a hotfix so that we can deploy it to production more quickly, too.

LindsayYoung:
Good idea on the hotfix

Later, we might want to make the function use start date and end date if a user supplies one, I think that would be useful for a csv export but I think this is the right call for now

jontours:
Is there a way to make something an open feature PR a hotfix?  Or just start over?

ccostino:
I think given our ties to `git flow` it might make sense to start this one over.  There is some `git` magic you can perform to move/rebase a branch from one parent to another and rename it, but at that point you've performed just as many steps.

jontours:
Okay I'll go ahead and open a hotfix with this exact change implemented.  How does the hotfix process work.  I think I was about three weeks fresh on the job when I did one.  From what I'm seeing I open the hotfix, commit some changes, and then `git flow hotfix finish` is ran.  Is there a publish step to do though for review?

ccostino:
Ah, for a hotfix it's very close to how we work with a release.  Switch yourself to master via `git checkout master` and then do a `git pull` to make sure you have all of the latest updates.  Then run `git flow hotfix start <my-hotfix-name>` and everything else will be taken care of.  Make the commits in there and push the new branch to `origin` and create a new PR for review (ensuring it's against `master` for a correct diff to view).  However, we'll merge it with `git flow` so be sure to add a **MERGE WITH GIT FLOW** note to the title.

At that point, yeah, it's just a matter of someone else pulling it down and running `git flow hotfix finish <my-hotfix-name>` to wrap it all up (including providing and naming a tag!) and pushing `master` up to `origin` and Travis will take care of the rest with deployment.

LindsayYoung:
closing this since we merged with gitflow

PR #1963, opened 2016-10-04
[MERGE W/ GITFLOW] Added year to test model

jontours:
@ccostino @LindsayYoung hotfix for test fix.  Mostly identical to original PR.  Moved initialization of year variable to `setUp` method for convention's sake.


LindsayYoung:
Can get the merges ready for a post-meeting deploy. This deploy will also move read traffic, so we should check that

Issue #1953, opened 2016-09-26
Create API endpoint to serve overview data 


noahmanger:
So that users can see how much money has been raised and spent by all candidates, PACs, party committees and "other" groups, aggregated by each type, provide an API endpoint that shows this data.
 
---
## Background & details
This data will be used in these charts (and potentially elsewhere):
![image](https://cloud.githubusercontent.com/assets/1696495/18855008/58804cc4-8404-11e6-8f39-9d3bb49f3996.png)

I'll leave it to others to design what the response should look like, but if it's helpful, the current charts accept an array of objects (one array for raising, one for spending) that look like this:

```
  {
    'date': '1/31/2015',
    'candidates': -12682.75,
    'pacs': 37003916.07,
    'party': 25788128.68,
    'other': 0
  }
```

@PaulClark2 has written up elsewhere the SQL that is used to generate these numbers. 

**Criteria for completion:**
- [x] The API should return a response that contains the total amount of money raised or spent by each type of group from the beginning of the cycle to the last day of each month in that cycle. 
- [x] The aggregates should be calculated on a regular basis, as new filings always come in that can change an earlier month's totals
- [x] The results should be filterable by two-year period

PaulClark2:
Looking at Forms 3X, 3 and 3P, I think we can go back to 2007-2008. Prior to the 2007-2008 two-year period be didn't capture all of the data from the detailed summary page of reports.

I need to also look at the Form 5, 7 and 9 data.

noahmanger:
Here's the SQL from Paul:

For candidates:
```
select
extract(month from to_date(vs.cvg_end_dt, 'YYYYMMDD')) month_order,
to_char(to_date(vs.cvg_end_dt, 'YYYYMMDD'), 'Month') as month,
extract(year from to_date(vs.cvg_end_dt, 'YYYYMMDD')) as year,
sum(nvl(vs.ttl_receipts,0) - (nvl(vs.pty_cmte_contb ,0) + nvl(vs.oth_cmte_contb ,0) + nvl(vs.offsets_to_op_exp,0) + nvl(vs.ttl_loan_repymts,0) + nvl(vs.ttl_contb_ref,0))) as adj_tr,
sum(nvl(vs.ttl_disb, 0) - (nvl(vs.tranf_to_other_auth_cmte,0) + nvl(vs.ttl_loan_repymts,0) + nvl(vs.ttl_contb_ref,0) + nvl(vs.other_disb_per,0))) as adj_td
[...]

PR #1964, opened 2016-10-04
Chart endpoint API

xtine:
Massaging SQL.


LindsayYoung:
closes #1953

LindsayYoung:
closes #1953

webservices/resources/large_aggregates.py
```diff
@@ -0,0 +1,60 @@
...
+    @property
+    def index_column(self):
+        return self.model.idx
+
+
+@doc(
+    tags=['financial'],
+    description=docs.ENTITY_DISBURSEMENTS_TOTLAS,
+)
+class EntityDisbursementsTotalsView(ApiResource):
```
jontours: Is there any reason that these were broken out into two views?  If anything is expected to change with these endpoints (e.g. adding a new filter) it could be problematic as the changes will have to made to both views as opposed to using just one; we do this on the `reports\` endpoint by injecting the model, schema, and page schema at runtime based on what endpoint was requested.

data/sql_updates/large_aggregates.sql
```diff
@@ -0,0 +1,228 @@
...
+        pol_pty_cmte_contb_per as pty_cmte_contb,
+        other_pol_cmte_contb_per as oth_cmte_contb,
+        ttl_disb_per as ttl_disb,
+        -- double check var
+        tranf_to_other_auth_cmte_per as tranf_to_other_auth_cmte,
+        offsets_to_op_exp_per as offsets_to_op_exp,
+        ttl_loans_received_per as ttl_loan_repymts,
+        ttl_contb_ref_per as ttl_contb_ref,
+        other_disb_per
+    from fec_vsum_f3p
```
jontours: @LindsayYoung looking over the sql that Paul put together and it used disclosure.v_sum_and_det_sum_report to build this part (I think it's the same part).  He's seen this postgres sql so I'm assuming it's fine, but it is a difference.  Just kind of proofreading stuff.

PR #1965, opened 2016-10-05
Feature/link reg citations to eregs

vrajmohan:


anthonygarvan:
looks good, will merge once build passes

PR #1967, opened 2016-10-06
Feature/add csv url to reports

jontours:
csv work, I know @jwchumley mentioned that some of these aren't yet generated, but out of a random one I picked, it was 404.


jontours:
@noahmanger @LindsayYoung @ccostino this is ready to review.  The only tricky part with this PR was that some file_numbers are negative.  According to Jeff they are being processed and become positive once that is finished.  As far as that goes for the urls, no csvs exist for them until the are processed, and become positive file numbers.

ccostino:
@jontours:  Does this PR contain changes from other PRs?  It looks like a merge pulled in other things that were reviewed previously and those commits were treated as additions to this branch instead of a part of the original history.

jontours:
@ccostino I noticed that too, honestly have no idea why sometimes it does that and others it doesn't.  Because any time I merge in `develop` I run `git merge develop`.  I'd expect the same result every time, arghhh.

jontours:
I'll go ahead and open another PR against develop so y'all can actually filter to my work.

ccostino:
Argh, sorry about that!  There must be a setting or something that got flipped that might be causing that, I'm happy to help investigate to see.

PR #1975, opened 2016-10-12
fix pary to party

### LINKED ISSUE #1974: OpenFEC API Misspelling
Noticed a slight misspelling in the rad_branch field for the Party/Non Party designation type. It currently says "Party/Non Pary". Example: https://api.open.fec.gov/v1/rad-analyst/?rad_branch=unauthorized&api_key=DEMO_KEY&per_page=20&page=1

LindsayYoung:
There are spelling errors in the underlying data. This function fixes them. 

closes #1974

### DIFF
--- a/data/sql_updates/rad_analyst.sql
+++ b/data/sql_updates/rad_analyst.sql
@@ -1,3 +1,13 @@
+create or replace function fix_party_spelling(branch text)
+returns text as $$
+    begin
+        return case branch
+            when 'Party/Non Pary' then 'Party/Non Party'
+            else branch
+        end;
+    end
+$$ language plpgsql;
+
 drop materialized view if exists ofec_rad_mv_tmp;
 create materialized view ofec_rad_mv_tmp as
 select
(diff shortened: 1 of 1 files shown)


data/sql_updates/rad_analyst.sql
```diff
@@ -6,7 +16,7 @@ select
     -- I am not sure we want to filter on this, if the name is different here it will behave differently than the other committee searches by name
     cmte_nm as committee_name,
     anlyst_id as analyst_id,
-    rad_branch,
+    fix_party_spelling(rad_branch) as rad_branch,
```
jontours: Without background to this issue or this data table in general, was there a reason that the misspellings couldn't be corrected via a sql statement?

LindsayYoung: Do you mean moving the function into the select statement? I think that that would be fine here. I default to making clean up separate from the logic, because in lots of places there is so much cleanup that you can't see what the underlying logic is.

LindsayYoung: It is in the data a hundred times and this is quicker than asking them to change it

PR #1977, opened 2016-10-12
Feature/add csv links

jontours:
@ccostino @LindsayYoung this replaces the PR with the excess commits listed.


LindsayYoung:
In cases like efiling endpoints we aren't making views, we are using their views raw, so using a mixin is the only option. 

We can add it to the SQL for reports to be consistent with the pdf urls, but that can be a follow up task. I think @xtine is waiting on this

ccostino:
Ah okay, that makes sense, thanks @LindsayYoung!

PR #1981, opened 2016-10-14
Feature/search final opinion only

anthonygarvan:
Updates search logic to search only final opinions in universal search.

### DIFF
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -103,6 +103,9 @@ class UniversalSearch(utils.Resource):
                 .extra(size=hits_returned, from_=from_hit) \
                 .index('docs')
 
+            if type == 'advisory_opinions':
+                query = query.query("match", category="Final Opinion")
+
             if text_highlight_query:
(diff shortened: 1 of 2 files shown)


LindsayYoung:
looking now

webservices/resources/legal.py
```diff
@@ -103,6 +103,9 @@ def get(self, q, from_hit=0, hits_returned=20, type='all', **kwargs):
                 .extra(size=hits_returned, from_=from_hit) \
                 .index('docs')
 
+            if type == 'advisory_opinions':
```
LindsayYoung: I don't think it should hold this PR up, but maybe there should be a flag to over-ride the logic and search all the AOs? That would build in some flexibility for the future.

LindsayYoung:
One thing I noticed outside the scope of this review, was that the legal management commands don't have doc strings that explain them. They are named pretty well, but is helpful to have the extended discription of why you would want to run them when you type `./manage.py`

Overall, looks good. And I am glad I got the opportunity to get the statues and AOs going. Thanks

PR #1982, opened 2016-10-17
fixing refs, removing reg sync and adding mur sync

anthonygarvan:
This fixes the refresh error on dev.


LindsayYoung:
Thanks for getting to this so quickly!

PR #1986, opened 2016-10-19
Feature/implement fec urls

jontours:
I'm considering moving this to a sql function for the views we build.  Also would like to fix the .csv's not working (mostly for paper filings) with this pr.


jontours:
Waiting to hear back from Jeff on some of these but the URLs are good to go, we may want to move some of the url generation to a sql function but maybe make that a separate PR?  At least this way we get the functionality visible; and changing some of the endpoints to get the url would be an invisible process as far as the web-app is concerned.

webservices/resources/reports.py
```diff
@@ -114,20 +114,15 @@ def build_query(self, committee_type=None, **kwargs):
             default_schemas,
         )
         query = reports_class.query
-        # Eagerly load committees if applicable
         if hasattr(reports_class, 'committee'):
-            query = reports_class.query.join(reports_class.committee).options(sa.orm.joinedload(reports_class.committee))
```
LindsayYoung: Just curious why this was un-nested

jontours: The previous code I had made nested and after looking at it after some time away from it I had no idea really what I was trying to do.  It was tough finding a query that could make the relation properly without breaking the tests.  I'm fairly confident this is the write approach now.

LindsayYoung:
Work on this is good, but we are going to hold this since most of the underlying csv are not built. We at lease need the most recent ones to be populated.

noahmanger:
Aren't CSV links already merged in?

jontours:
Yeah, they are, that's what I referencing by opening up a PR against either develop or the release branch commenting out the new field.

jontours:
I believe this has been mostly reviewed, just needs to be merged.

webservices/schemas.py
```diff
@@ -375,7 +376,8 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
     make_schema,
     fields={
         'pdf_url': ma.fields.Str(),
-        #'csv_url': ma.fields.Str(),
+        'csv_url': ma.fields.Str(),
+        #'fec_url': ma.fields.Str(),
```
LindsayYoung: commented out fec instead of csv here

LindsayYoung:
Looks good will merge when tests pass

PR #1987, opened 2016-10-20
Feature/transition to disclosure schema

jontours:
This seems good to go, however is still failing some integration tests, not sure why.


LindsayYoung:
Did you update the schema in the test db?

PR #1988, opened 2016-10-20
Commenting out field from schema

jontours:
@ccostino @LindsayYoung comments out csv urls.


noahmanger:
Do we have feature flags going for the API? What about just hiding these behind a feature flag so it's easier to turn them on when we figure it out?

ccostino:
At the moment not really, and in most respects we don't need them for the type of work done.  In many cases with the schema work it's a matter of just not registering the schema before it's ready if we still wanted to get some preliminary work in, or not tying a resource to a URL.

This is a one-off case for this type of thing so I think we're okay for now.  But definitely good to think and ask about regardless!

Issue #1960, opened 2016-10-03
Issues with paging through nulls


LindsayYoung:
Since the data can have missing values, there needs to be a way to page through nulls.

discovered in #1927 

---

**Completion criteria:**
- [x] Schedule A can paginate through nulls
- [x] Schedule B can paginate through nulls
- [x] Schedule E can paginate through nulls

ccostino:
Upon some further investigation, here are some things to bear in mind about this:

* Whenever we supply a default column to sort by, whether we sort by ascending or descending has a bit of an impact due to what [Postgres does under the hood](https://www.postgresql.org/docs/9.5/static/queries-order.html).  TL;DR is that when we order something by descending order, Postgres will also set `NULLS FIRST` as a part of the query under the hood.  In the case of Schedule E, where `expenditure_date` is our default column to sort on in descending order, there are `null` values present in the table.
* When making requests directly to the API itself, the `sort_null_hide` flag is not automatically present in the request so records with `null` values in the sort column(s) set will appear unless it is explicitly provided.  `null` values do not on the web app because we set that flag to `True` within the app itself (due to this same issue causing problems with requests and empty results appearing).
* The paging mechanism already takes care of providing a way to page based on whatever field we define as the `index` (e.g., `sub_id` for the itemized records) simply by supplying a `last_index` parameter.  If you want the results to continue to be sorted, however, you must also supply the `last_x` (where `x` is the column you're sorting on).  This is a problem though if the value ends up being `null` for a couple of reasons:
  * Marshmallow will treat it as an invalid value and throw an error, e.g., `expenditure_date`, because it is not a valid value for the field type (`date` in this case).
  * Even if `null` is a valid value, that wouldn't give any indication as to which record is being referred to; we need to sort on another column, which we currently don't do.

@LindsayYoung, @jontours, and I all spoke the other day of possible solutions and one easy solution we came up with was possibly just setting another column to sort by that we know won't ever be `null`, such as the `sub_id` for the itemized models.  Some additional discussion revolved around supporting [keyset pagination](http://use-the-index-luke.com/no-offset), but this is already supported by our app via [marshmallow-pagination](https://github.com/jmcarp/marshmallow-pagination/blob/master/marshmallow_pagination/paginators.py#L59).

[...]

PR #1989, opened 2016-10-20
[WIP] Adds secondary sorting for itemized resources in the API

ccostino:
**[WIP:  DON'T MERGE]**

This changeset adds support for adding a secondary sort column for all itemized resources to help address the issue with paging through records that have `null` values for the initial sort column (e.g., `expenditure_date` for `ScheduleE`).  This does not fully fix the issue though as the instructions and usage still indicate that if you want to preserve ordering in your result you need to supply the `last_x` column (with `x` being the column you are sorting on) in addition to the `last_index` parameter.  This is still being hammered out and discussed in issue #1960.

### DIFF
--- a/webservices/sorting.py
+++ b/webservices/sorting.py
@@ -3,6 +3,15 @@ import sqlalchemy as sa
 from webservices.exceptions import ApiError
 from webservices.common.util import get_class_by_tablename
 
+ITEMIZED_MODELS = (
+    'ScheduleA',
+    'ScheduleB',
+    'ScheduleC',
+    'ScheduleD',
+    'ScheduleE',
+    'ScheduleF'
+)
+
 
 def parse_option(option, model=None, aliases=None, join_columns=None, query=None):
     """Parse sort option to SQLAlchemy order expression.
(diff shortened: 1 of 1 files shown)


jontours:
Or wait, that may not actually be a problem.  Nevermind!  I think that's more related to the issue of leaving out sort parameters.

ccostino:
Yup, without the `sort` parameters there will be no sorting! :-)

jontours:
Well, that's the thing, it actually _does_ still sort without sort parameters, but I think it's more closely related to the open issue about leaving out `last_xyz_date` in the request.  I'm not quite sure the best resolution for that besides either a) take out sorting by date by default or b) return an api exception if it's left out.

ccostino:
Oh right, I forgot about that being set by default. :thinking: yeah, I imagine we'll want to iterate on this a bit more.  I'm not sure of an appropriate full solution either at this point.

LindsayYoung:
The counts are estimates and can be very off To get all the results you should page through until the results stop. At that point I have done the query directly in postgres and compared the result there.

jontours:
I think the counts can be off when there are a large amount of rows.  For this there were only 245 rows.  And there is a difference in the number of actual rows returned if the `last_expenditure_date` is passed in or if it isn't.  The former returns more, the latter filters some out.

LindsayYoung:
womp womp

Do we need a new issue for that?

jontours:
No I think it's pretty much covered by this: https://github.com/18F/openFEC/issues/1978.

How we go about resolving may have evolved since I wrote the issue though.  At this point I'm leaning toward removing default sort by date.

LindsayYoung:
Sounds good, thanks!

PR #1994, opened 2016-10-21
Feature/disclosure transition

jontours:
This is ready to review.  Had some interesting git issues this morning (I can bore you with the details if need be), so opened this PR in lieu of the closed one.

This was mostly straightforward. There was an issue being that one of the models (`dates.py`) referenced a table directly that moved to the `disclosure` schema.  It wasn't hard figuring out how to map it to a different table, however the tests didn't recognize the syntax needed to point to a new table.  I created a mixin to solve this; essential for that one table the public schema is used for the tests and when deployed will use the disclosure schema.

/cc
@LindsayYoung @ccostino

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -22,6 +22,13 @@ def _reset_schema():
     rest.db.engine.execute('drop schema if exists public cascade;')
     rest.db.engine.execute('drop schema if exists disclosure cascade;')
     rest.db.engine.execute('create schema public;')
+    rest.db.engine.execute('create schema disclosure;')
+
+
+def _reset_schema_for_integration():
+    rest.db.engine.execute('drop schema if exists public cascade;')
+    rest.db.engine.execute('drop schema if exists disclosure cascade;')
+    rest.db.engine.execute('create schema public;')
 
 
 class BaseTestCase(unittest.TestCase):
@@ -102,7 +109,14 @@ class IntegrationTestCase(BaseTestCase):
 
     @classmethod
     def setUpClass(cls):
-        super(IntegrationTestCase, cls).setUpClass()
+        rest.app.config['TESTING'] = True
+        rest.app.config['SQLALCHEMY_DATABASE_URI'] = TEST_CONN
+        rest.app.config['PRESERVE_CONTEXT_ON_EXCEPTION'] = False
+        cls.app = rest.app.test_client()
+        cls.client = TestApp(rest.app)
+        cls.app_context = rest.app.app_context()
+        cls.app_context.push()
+        _reset_schema_for_integration()
         with open(os.devnull, 'w') as null:
             subprocess.check_call(
                 ['pg_restore', './data/subset.dump', '--dbname', TEST_CONN, '--no-acl', '--no-owner'],
--- a/webservices/common/models/dates.py
+++ b/webservices/common/models/dates.py
@@ -13,6 +13,10 @@ class ReportType(db.Model):
     report_type_full = db.Column('rpt_tp_desc', db.String, index=True, doc=docs.REPORT_TYPE)
 
 
+class DisclosureMixin(object):
+    __table__args = {"schema": "disclosure"}
+
+
 class ReportDate(db.Model):
     __tablename__ = 'trc_report_due_date'
 
@@ -28,6 +32,10 @@ class ReportDate(db.Model):
     @property
     def report_type_full(self):
         return clean_report_type(self.report.report_type_full)
+
+
+class ReportDateDisclosure(DisclosureMixin, ReportDate):
+    pass
 
 
 REPORT_TYPE_CLEAN = re.compile(r'{[^)]*}')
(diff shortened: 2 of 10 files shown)


tests/common.py
```diff
@@ -22,6 +22,13 @@ def _reset_schema():
     rest.db.engine.execute('drop schema if exists public cascade;')
     rest.db.engine.execute('drop schema if exists disclosure cascade;')
     rest.db.engine.execute('create schema public;')
+    rest.db.engine.execute('create schema disclosure;')
+
+
+def _reset_schema_for_integration():
+    rest.db.engine.execute('drop schema if exists public cascade;')
+    rest.db.engine.execute('drop schema if exists disclosure cascade;')
+    rest.db.engine.execute('create schema public;')
```
ccostino: Do we also need a line to create the `disclosure` schema here?

webservices/common/models/dates.py
```diff
@@ -13,6 +13,10 @@ class ReportType(db.Model):
     report_type_full = db.Column('rpt_tp_desc', db.String, index=True, doc=docs.REPORT_TYPE)
 
 
+class DisclosureMixin(object):
+    __table__args = {"schema": "disclosure"}
```
ccostino: Oh interesting, nice!  We are going to have to do this for other models and tables, I wonder if this is a better way to approach this going forward rather than making the change in each model.  One benefit I see is that if the schema changes again for a set of tables we only have to adjust in the one spot, though that may also necessitate a change to the Mixin name too, which would have to propagate through the associated models again and thus not actually give any benefit there.

Still, I like this idea. :-)

webservices/common/models/dates.py
```diff
@@ -30,6 +34,10 @@ def report_type_full(self):
         return clean_report_type(self.report.report_type_full)
 
 
+class ReportDateDisclosure(DisclosureMixin, ReportDate):
+    pass
```
ccostino: Okay, I see how you've accounted for the tests by doing this too.  Nice!

However, what do you think about it in the future if we generate an entirely new subset of data that _does_ account for all of the different schemas?  Should we think about this a little now or worry about it if that time comes?

tests/common.py
jontours: This PR had a couple of huge headaches, this is one of them.  I'll do my best to explain what I think is happening.  For the integration tests, we actually use the `subset.dump` to populate `cfdm_unit_test` using `pg_restore`.  For the public schema this worked fine.  However with `disclosure` this was causing an error.  And it seems that, and I'm postulating here, that pg_dump doesn't dump the public schema explicity.  But any other custom schema it does drump a `create schema` statement in the binary dump.  So when we attempt to run `create schema disclosure` from within the test set-up, it causes an error, stating that the schema already exists.  So this was a solution, maybe not the best.  The only thing I could think to do is filter out the `create schema disclosure` statement in the dump file, but that's an extra step, also there is no way to exclude these statements being exported via command line, as I hoped.

webservices/common/models/dates.py
jontours: That's the thing, the subset.dump does include the disclosure schema, but I was getting failures in `test_dates` specifically the test `test_reporting_date_filters` at https://github.com/18F/openFEC/blob/develop/tests/test_dates.py#L41.  This test hits the one model where the disclosure model is hit directly.  And was causing failures, stating that it couldn't find the table that backs that model (one of the tables that was migrated to the disclosure schema).  As far as I can tell the reason being is that the methodology the tests use don't recognize the syntax that points a model to a different schema.  Hence that's why I used a mixin.

tests/common.py
ccostino: Oh wow, interesting, thanks for explaining that.  That does seem odd that `pg_dump` wouldn't account for the `public` schema but would all others; I'll try digging a bit too to see if I can uncover anything, but if not or it takes to long, good to know this is a working solution for us!

ccostino:
Thanks again for providing all of the context and background info, @jontours!  It does seem like there's a little bit to investigate here but I wonder if it's worth pursuing those things at a later time and moving forward with what's here.  I'm okay with that if everyone else is; @LindsayYoung do you agree?

ccostino:
I'm also assuming this is ready to go despite the **[Don't merge!]** label in the title? :-)

jontours:
It's not ready to go at the moment.  I double checked that this solution does in fact serve up rows from the disclosure schema, and it's not!!  Currently investigating.

ccostino:
D'oh!  Let me know if there's anything I can do to help.

jontours:
Thanks @ccostino I think I have things working for now.  There were some slight changes that were needed so please re-review and feel free to ask questions, thanks for the input!

jontours:
Or wait, I commited to the wrong branch; just not my week so far!  Give me minute I'll be pushing more changes.

LindsayYoung:
Thanks for your work to figure out the tests, @jontours 

@ccostino you will want to update your test subset after this is merged

ccostino:
Awesome, thank you both!

jontours:
I'll have to talk to both of you about some of the _funny_ behavior with this code when y'all are both in the office; and like funny weird not funny ha-ha.

LindsayYoung:
So, the punchline is... I put a table from outside the schema on that list. Sorry Jon!

ccostino:
D'oh!

PR #1995, opened 2016-10-21
Removes amended records

LindsayYoung:
Better, party and other totals need work

### DIFF
--- a/data/sql_updates/large_aggregates.sql
+++ b/data/sql_updates/large_aggregates.sql
@@ -16,6 +16,7 @@ with cand as (
         ttl_contb_ref_per as ttl_contb_ref,
         other_disb_per
     from fec_vsum_f3p
+    where most_recent_filing_flag like 'Y'
     union all
     select
         extract(month from to_date(cast(cvg_end_dt as text), 'YYYY-MM-DD')) as month,
@@ -30,6 +31,7 @@ with cand as (
         ttl_contb_ref_col_ttl_per as ttl_contb_ref,
         other_disb_per
     from fec_vsum_f3
+    where most_recent_filing_flag like 'Y'
 ),
 cand_totals as (
     select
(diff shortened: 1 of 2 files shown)


LindsayYoung:
I already ran it on dev

LindsayYoung:
also, we probably want downloads for this if people wanted to use it for someting

LindsayYoung:
Ran on staging too

data/sql_updates/large_aggregates.sql
```diff
@@ -30,6 +31,7 @@ with cand as (
         ttl_contb_ref_col_ttl_per as ttl_contb_ref,
         other_disb_per
     from fec_vsum_f3
+    where most_recent_filing_flag like 'Y'
```
jontours: Ah, I was always wondering what this column was for.

data/sql_updates/large_aggregates.sql
```diff
@@ -142,11 +146,14 @@ party_totals as (
...
-        sum(independent_expenditures_period) as receipts,
-        sum(independent_contributions_period) as disbursements
-    from ofec_reports_ie_only_mv_tmp
+        extract(month from to_date(cast(cvg_end_dt as text), 'YYYY-MM-DD')) as month,
+        extract(year from to_date(cast(cvg_end_dt as text), 'YYYY-MM-DD')) as year,
+        sum(ttl_indt_exp) as receipts,
+        sum(ttl_indt_contb) as disbursements
+    from
+        fec_vsum_f5
+    where
```
jontours: I know it was probably mentioned somewhere so sorry if I'm slow on the uptake, but what is the rationale for recreating this logic here as opposed to just using the totals from `ofec_reports_ie_only_mv_tmp`?

LindsayYoung: Good catch. I think I changed the source of the table and then did then looked for duplicates and that made it better, but there was not a need to change tables. I can change it back, but I need to keep working on it anyway.

PR #2001, opened 2016-10-25
Add means_filed to reports and filings endpoints

jontours:
Also gonna add some logic to generate the fec urls via sql function.  We mentioned that before as a task and since it's so closely related knocking that out here as well.

### DIFF
--- a/data/functions/urls.sql
+++ b/data/functions/urls.sql
@@ -47,6 +47,16 @@ begin
 end
 $$ language plpgsql immutable;
 
+create or replace function means_filed(image_number text) returns text as $$
+begin
+    return case
+        when is_electronic(image_number) then 'e-file'
+        else 'paper'
+    end;
+end
+$$ language plpgsql immutable;
+
+
 create or replace function is_electronic(image_number text) returns boolean as $$
 begin
     return case
--- a/data/functions/urls.sql
+++ b/data/functions/urls.sql
@@ -29,3 +29,30 @@ begin
     end;
 end
 $$ language plpgsql immutable;
+
+create or replace function report_fec_url(image_number text, file_number integer) returns text as $$
+begin
+
+    return case
+        when file_number < 0 then null
+        when image_number is not null and not is_electronic(image_number) then format(
+            'http://docquery.fec.gov/paper/posted/%1$s.fec',
+            file_number
+        )
+        when image_number is not null and is_electronic(image_number) then format(
+            'http://docquery.fec.gov/dcdev/posted/%1$s.fec',
+            file_number
+        )
+    end;
+end
+$$ language plpgsql immutable;
+
+create or replace function is_electronic(image_number text) returns boolean as $$
+begin
+    return case
+        when char_length(image_number) = 18 and substring(image_number from 9 for 1) = '9' then true
+        when char_length(image_number) = 11 and substring(image_number from 3 for 1) = '9' then true
+        else false
+    end;
+end
+$$ language plpgsql immutable;
--- a/data/functions/urls.sql
+++ b/data/functions/urls.sql
@@ -56,12 +56,20 @@ begin
 end
 $$ language plpgsql immutable;
 
-
+--If the image number string is of length 18, and the 9th character is a 9, then it is electronic
+--if the image number string is of length 11, and the 3rd character is a 9, then also electronic
+--if imaage number is of length 11 and the 3rd and 4th characters are 02 and 03 then it is paper
+--all other combinations are paper filers. (this encoding is outlined at this GitHub issue:
+--https://github.com/18F/openFEC/issues/1882
 create or replace function is_electronic(image_number text) returns boolean as $$
 begin
     return case
         when char_length(image_number) = 18 and substring(image_number from 9 for 1) = '9' then true
         when char_length(image_number) = 11 and substring(image_number from 3 for 1) = '9' then true
+        --these last two cases aren't really needed, but good to add them to make this encoding
+        --explicit to the reader
+        when char_length(image_number) = 11 and substring(image_number from 3 for 2) = '02' then false
+        when char_length(image_number) = 11 and substring(image_number from 3 for 2) = '03' then false
         else false
     end;
 end
--- a/data/functions/urls.sql
+++ b/data/functions/urls.sql
@@ -34,7 +34,7 @@ create or replace function report_fec_url(image_number text, file_number integer
 begin
 
     return case
-        when file_number < 0 then null
+        when file_number < 1 then null
         when image_number is not null and not is_electronic(image_number) then format(
             'http://docquery.fec.gov/paper/posted/%1$s.fec',
             file_number
--- a/webservices/common/models/filings.py
+++ b/webservices/common/models/filings.py
@@ -3,12 +3,12 @@ import datetime
 from webservices import docs, utils
 from webservices.common.models.dates import ReportType
 from webservices.common.models.dates import clean_report_type
-from webservices.common.models.reports import CsvFecMixin
+from webservices.common.models.reports import CsvMixin, FecMixin
 
 from .base import db
 
 
-class Filings(db.Model, CsvFecMixin):
+class Filings(db.Model, CsvMixin):
     __tablename__ = 'ofec_filings_mv'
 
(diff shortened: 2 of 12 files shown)


webservices/common/models/filings.py
```diff
@@ -64,7 +66,7 @@ def document_description(self):
         )
 
 
-class EFilings(db.Model, CsvFecMixin):
+class EFilings(db.Model, CsvMixin, FecMixin):
```
ccostino: As I've been reviewing this and thinking back to another recent change, something triggered in my memory about mixins in Python.  It turns out that the way the Method Resolution Order (MRO) works in Python means you'll want to list the mixins first, before whatever the base class is in your object inheritance.  The reason for this is that the classes are read from right to left and the first one encountered is extended with the others that follow.

In this case it's not much of an issue because both of these mixins define a single method not found anywhere else.  But if you had a method defined in multiple classes and expected it to be overridden one way, it will quite likely end up behaving different!

More details can be found here:

* [Mixins and Python by Ian Lewis](https://www.ianlewis.org/en/mixins-and-python)
* [Multiple Inheritance is Hard by Ned Batchelder](http://nedbatchelder.com/blog/201210/multiple_inheritance_is_hard.html)
* [Stack Overflow Question](http://stackoverflow.com/questions/10018757/how-does-the-order-of-mixins-affect-the-derived-class)

Long story short, the object should be defined like this:

```python
class EFilings(CsvMixin, FecMixin, db.Model):
```

We'll probably want to adjust in other spots as well to prevent future headaches for ourselves, just in case any of these grow to have more methods that end up overriding things.

webservices/common/models/filings.py
```diff
@@ -3,12 +3,12 @@
...
 from webservices.common.models.dates import ReportType
 from webservices.common.models.dates import clean_report_type
-from webservices.common.models.reports import CsvFecMixin
+from webservices.common.models.reports import CsvMixin, FecMixin
 
 from .base import db
 
 
-class Filings(db.Model, CsvFecMixin):
+class Filings(db.Model, CsvMixin):
```
ccostino: See note below about mixins; order should be swapped like so:

```python
class Filings(CsvMixin, db.Model):
```

webservices/common/models/reports.py
```diff
@@ -60,7 +60,7 @@ def treasurer_name(self):
         return name
 
 
-class CommitteeReports(PdfMixin, CsvFecMixin, BaseModel):
+class CommitteeReports(PdfMixin, CsvMixin, BaseModel):
```
ccostino: This class definition is already fine. :-)

webservices/common/models/reports.py
```diff
@@ -315,7 +319,7 @@ class BaseFilingSummary(db.Model):
     column_a = db.Column('cola', db.Float)
     column_b = db.Column('colb', db.Float)
 
-class BaseFiling(PdfMixin, CsvFecMixin, db.Model):
+class BaseFiling(PdfMixin, FecMixin, db.Model):
```
ccostino: Okay here too.

webservices/common/models/reports.py
```diff
@@ -333,6 +337,8 @@ class BaseFiling(PdfMixin, CsvFecMixin, db.Model):
     election_state = db.Column('el_state', db.String)
     receipt_date = db.Column('create_dt', db.Date, index=True)
     sign_date = db.Column(db.Date)
+    means_filed = 'e-file'
```
ccostino: Seeing this again, I think we should define an `E_FILE_XXX` constant (variable, I know) and set it to `e-file`, then use that where all of these attributes are defined.

Related, is there some other way this should be defined in the model, such as a `hybrid_property` (e.g., [like this one in our itemized base model](https://github.com/18F/openFEC/blob/develop/webservices/common/models/itemized.py#L24))?

webservices/common/models/filings.py
```diff
@@ -81,6 +83,8 @@ class EFilings(db.Model, CsvFecMixin):
     amended_by = db.Column('superceded', db.BigInteger, doc=docs.AMENDED_BY)
     amends_file = db.Column('previd', db.BigInteger, doc=docs.AMENDS_FILE)
     amendment_number = db.Column('rptnum', db.Integer, doc=docs.AMENDMENT_NUMBER)
+    means_filed = 'e-file'
```
ccostino: I wonder if this should be a constant defined elsewhere?

webservices/resources/reports.py
```diff
@@ -73,6 +73,12 @@ def get_range_filters():
     ]
     return filter_range_fields
 
+def get_match_filters():
+    filter_match = [
+        ('filer_type', models.CommitteeReports.means_filed)
+    ]
+    return filter_match
```
ccostino: Shot in the dark for the issue I see below, but what if this were called `filter_match_fields`?  It shouldn't matter, but it is a convention used across all other filter definitions.

webservices/resources/reports.py
```diff
@@ -139,7 +142,8 @@ def build_query(self, committee_type=None, **kwargs):
             query = query.filter(reports_class.is_amended == kwargs['is_amended'])
 
         query = filters.filter_range(query, kwargs, get_range_filters())
-
+        #currently doesn't work, whyyyyy
+        #query = filters.filter_match(query, kwargs, get_match_filters())
```
ccostino: That's definitely curious, it doesn't work at all?  Is this why `filer_type` appears above?

webservices/resources/reports.py
```diff
@@ -186,6 +190,8 @@ def build_query(self, committee_id=None, committee_type=None, **kwargs):
 
         if committee_id is not None:
             query = query.filter_by(committee_id=committee_id)
+        if kwargs.get('filer_type'):
+            query = query.filter(reports_class.means_filed == kwargs['filer_type'])
```
ccostino: Same thing here as opposed to using `get_match_filters()`?

data/functions/urls.sql
```diff
@@ -29,3 +29,40 @@ begin
...
+    end;
+end
+$$ language plpgsql immutable;
+
+
+create or replace function is_electronic(image_number text) returns boolean as $$
+begin
+    return case
+        when char_length(image_number) = 18 and substring(image_number from 9 for 1) = '9' then true
+        when char_length(image_number) = 11 and substring(image_number from 3 for 1) = '9' then true
```
jontours: They were mentioned in a github issue but that can quickly get lost in the weeds.  I'll add a comment to `urls.sql`.

webservices/common/models/filings.py
jontours: Thanks for illuminating that, I would have no idea!

webservices/common/models/filings.py
```diff
@@ -81,6 +83,8 @@ class EFilings(db.Model, CsvFecMixin):
...
+    #fec_url = db.String(db.String)
```
jontours: There is; I may have gotten comment happy here, not sure what I was thinking.

webservices/common/models/reports.py
jontours: Well, circling back to this now I think a better thing to do would be yank them entirely, since they are serving no purpose (no filtering or anything).  And it's discernible from the endpoin itself that these resources were filed electronically.

webservices/resources/reports.py
jontours: I can take a look at this again.  Experience has taught me that time away from problems helps solve them.

ccostino:
Thanks, @jontours!

data/functions/urls.sql
ccostino: Thanks!  We should definitely document them here, I could see them getting lost and forgotten easily.

jontours:
@ccostino documentation and a couple of redundant cases added to the function.

webservices/common/models/reports.py
LindsayYoung: I think this should still be CsvFecMixin, right?

jontours: The fec url is now being added to the material view via a sql function, so the Mixin is not longer needed.

LindsayYoung: Thanks for clarifying!

ccostino:
Thanks, @jontours!

PR #2003, opened 2016-10-26
Removing reference to disclosure schema

jontours:
After talking to @LindsayYoung it was discovered that one of the tables is staying in the `public` schema for now.


ccostino:
Thanks, @jontours and @LindsayYoung!

PR #2005, opened 2016-10-26
Stopping mur update for now

LindsayYoung:
There is an error on import we can fix later


ccostino:
Thanks, @LindsayYoung and @anthonygarvan!

PR #2012, opened 2016-10-27
Relax candidate table construction

jontours:
I think this is a nice way to expose more candidates while keeping options to restrict them.  Some of the candidates will be `U` type candidates, however the `principal_committees` will still only be `P` type committees, so we can decorate any of the resources (how candidate search is done now) with more restrictions on a case-by-case basis.

@LindsayYoung @ccostino


jontours:
Thanks @LindsayYoung!

LindsayYoung:
Ran this on dev and staging, Looks like it works pretty well. Fixes the Lessig thing and I didn't find any side effects

Ran this on dev and running it on staging now

PR #2016, opened 2016-10-31
adding enable_stemming function

anthonygarvan:
This enables stemming (eg. "presidents" returns results for "president") under the english analyzer in elasticsearch.

### DIFF
--- a/webservices/legal_docs/load_legal_docs.py
+++ b/webservices/legal_docs/load_legal_docs.py
@@ -527,9 +527,9 @@ def enable_stemming():
         }
     }
 
-    docs_new = Index('docs_new', using=es)
-    if not docs_new.exists():
-        es.indices.create('docs_new', settings)
+    docs_backup = Index('docs_backup', using=es)
+    if not docs_backup.exists():
+        es.indices.create('docs_backup', settings)
 
     doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
     original_count = []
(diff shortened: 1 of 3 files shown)


webservices/legal_docs/load_legal_docs.py
```diff
@@ -492,3 +492,76 @@ def remap_citations(archived_murs):
...
+    doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
+    original_count = []
+    backed_up_count = []
+    reindexed_count = []
+    for _type in doc_types:
+        query = {"query": {"match_all": {}}}
+        total = Search().using(es).query(Q('term', _type=_type)).execute().hits.total
+        original_count.append(total)
+        all_docs = elasticsearch.helpers.scan(es, query, scroll='1m', index='docs', doc_type=_type, size=500)
+        all_docs = list(Result(doc).to_dict() for doc in all_docs)
```
vrajmohan: Why `list()` here and not on line 558?

webservices/legal_docs/load_legal_docs.py
```diff
@@ -492,3 +492,76 @@ def remap_citations(archived_murs):
...
+        count, _ = elasticsearch.helpers.bulk(es, all_docs, index='docs_new',
```
vrajmohan: Not a big deal, but the order of keyword args on line 559 is different.

webservices/legal_docs/load_legal_docs.py
```diff
@@ -492,3 +492,76 @@ def remap_citations(archived_murs):
...
+                    doc_type=_type, chunk_size=100, request_timeout=30)
+        backed_up_count.append(count)
+        docs_new.refresh()
+        print("Backed up %d %s." % (count, _type))
+
+    if original_count == backed_up_count and sum(original_count) > 0:
```
vrajmohan: For my understanding, under what circumstances would these counts be off? Wouldn't it be simpler to report the discrepancy as it happens, i.e. per _type?

webservices/legal_docs/load_legal_docs.py
```diff
@@ -492,3 +492,76 @@ def remap_citations(archived_murs):
...
+            }
+        },
+        "settings": {
+            "analysis": {"analyzer": {"default": {"type": "english"}}}
+        }
+    }
+
+    docs_new = Index('docs_new', using=es)
+    if not docs_new.exists():
+        es.indices.create('docs_new', settings)
```
vrajmohan: Are we stemming in the backup? Is that useful?

webservices/legal_docs/load_legal_docs.py
anthonygarvan: oops, that's a mistake! had it in there for debugging. Thanks!

webservices/legal_docs/load_legal_docs.py
anthonygarvan: I encountered issues with corruption when the settings were different. Sure docs_backup is a better name, thanks.

anthonygarvan:
We should keep the code in the code base-- it's nice to be able to audit and also to copy / paste from (as I did for this story).

PR #2017, opened 2016-10-31
[WIP] - IE expenditure by state for candidates

jontours:
Still need to add clean state data; filter out all non-IE pacs.


ccostino:
Note that the filter for non-IE PACs is essentially a filter on committee types; I will likely have to follow up for which specific ones.

ccostino:
Correction after some additional discussion: we should be okay as is with IEs, in fact we'll want to just remove the `group by` statements for `cmte_id` and `committee_type`.  This just leaves cleaning the state information.

lbeaufort:
Archiving this PR

PR #2023, opened 2016-11-02
prevent accidents 

LindsayYoung:
Since we need to make files to change the creds, I don't want those accidentally checked into source control. This will catch files that have the naming convention we outline in the wiki:
https://github.com/18F/openFEC/wiki/switch-out-cf-service-environment-variables


ccostino:
Thank you very much, @LindsayYoung!

PR #2024, opened 2016-11-03
Keyset pagination fix

jontours:
Addresses https://github.com/18F/openFEC/issues/2022.  The sort column is needed to ensure pagination can sort through pages and guarantee already seen records don't appear again, aka duplicates.

However, may employ a similar strategy to actually page through nulls, once they appear in the data set.  But strategy to employ will most likely be a little different.

cc @LindsayYoung @ccostino


LindsayYoung:
This looks good. Thank you again @jontours

PR #2027, opened 2016-11-03
[MERGE WITH GITFLOW] Release/public beta 20161109

LindsayYoung:


LindsayYoung:
Merged this with gitflow yesterday but it did not get closed, I think the last commit might have not gotten in

PR #2029, opened 2016-11-04
Pagination with null sort columns

jontours:
Still need to add logic for plugging in right min/max value depending on column data type (simple dict should do it)

### DIFF
--- a/webservices/utils.py
+++ b/webservices/utils.py
@@ -92,12 +92,7 @@ class SeekCoalescePaginator(paginators.SeekPaginator):
         lhs, rhs = (), ()
         if sort_index is not None:
             left_index = self.sort_column[0]
-            print(str(left_index.property.columns[0].type).lower())
-
-            if direction == sa.asc:
-                comparator = self.max_column_map.get(str(left_index.property.columns[0].type).lower())
-            else:
-                comparator = self.min_column_map.get(str(left_index.property.columns[0].type).lower())
+            comparator = self.max_column_map.get(str(left_index.property.columns[0].type).lower())
             left_index = sa.func.coalesce(left_index, comparator)
             lhs += (left_index,)
             rhs += (sort_index,)
@@ -110,7 +105,6 @@ class SeekCoalescePaginator(paginators.SeekPaginator):
             filter = lhs > rhs if direction == sa.asc else lhs < rhs
             cursor = cursor.filter(filter)
         query = cursor.order_by(direction(self.index_column)).limit(limit)
-        print(query)
         return query.all() if eager else query
 
     def _get_index_values(self, result):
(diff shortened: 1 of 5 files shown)


jontours:
Failing one of the new sort reversal tests, will look more into that next week.

jontours:
/cc @ccostino @LindsayYoung This addresses being able to page from non-null values to nulls, and to continue paging from there.  To support this some of the changes needed are:

 1) nulls always come last in an itemized endpoint 
2) `last_index` and `last_xyz` sort columns are going to be needed going forward to support proper pagination, i.e. don't filter out valid rows and don't include rows already seen (duplicates).

I looked at what it would take to support being able to paginate with only `last_index` and it would mean disabling default sort on the respective date for the endpoint.  I feel that's a fairly heavy handed change to add a limited increase in functionality (especially since the direct users of the api are just plugging in the indices via a script).  But, I've verified the endpoint now allows paging over from non-nulls to nulls, which for some committees will be a good feature as some committees have a large amount on incomplete data (nulls).

webservices/utils.py
```diff
@@ -62,16 +65,76 @@ def fetch_page(query, kwargs, model=None, aliases=None, join_columns=None, clear
...
+        if last_index is not None:
+            lhs += (self.index_column,)
+            rhs += (last_index,)
+        lhs = sa.tuple_(*lhs)
+        rhs = sa.tuple_(*rhs)
+        if rhs.clauses:
+            filter = lhs > rhs if direction == sa.asc else lhs < rhs
+            cursor = cursor.filter(filter)
+        query = cursor.order_by(direction(self.index_column)).limit(limit)
+        print(query)
```
LindsayYoung: Left a print in. Though, this was helpful for review

LindsayYoung:
It seems like we might want to add sorting columns with the coalesced values pre-computed in the sql doing it on the fly takes a huge performance hit. 

this doesn't complete:
http://localhost:5000/v1/schedules/schedule_a/?sort=-contribution_receipt_date&per_page=1&last%20index=4122820151260365597

in comparison .15 seconds:
https://api.open.fec.gov/v1/schedules/schedule_a/?sort=-contribution_receipt_date&per_page=1&last%20index=4122820151260365597&api_key=DEMO_KEY

(We already talked about it but documenting here with an example)

LindsayYoung:
I know you had some test scripts for these, would it be a big lift to add those as tests here? I think there was a separate issue for that, but we have had so many issues with pagination that would make me feel more confidant.

Thanks again and sorry for the review delay

jontours:
I don't think it would be a big lift, the tests would be very similar to the ones @ccostino wrote, might even be able to modify those a bit for the new behavior.

jontours:
Getting started on adding some tests for this now.

jontours:
One new test added and one slightly modified.  The tests demonstrate the ability to navigate from rows with null sort values to non-null values (ascending), and from non-null sort rows to rows with null sort values.  Ready to review, let me know if there are any questions about any of the logic, inheritance stuff added here.

LindsayYoung:
Thanks for this! I can take a look now

LindsayYoung:
Thank you

PR #2035, opened 2016-11-08
Feature/better citation text

vrajmohan:


anthonygarvan:
This is a total beast. Thanks for pushing on this, happy about not taking any shortcuts here.

PR #2038, opened 2016-11-10
[MERGE WITH GIT FLOW]: Update Travis CI encrypted variables

ccostino:
This changeset address a security issue reported by Travis earlier this morning in which a vulnerability was found with the enrypted environment variables.  We have cycled our credentials and used the newest version of the Travis CLI tool to generate new encrypted environment variables.

Doing this as a hotfix to ensure we are updated properly.

/cc @LindsayYoung

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ addons:
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


ccostino:
These ended up causing issues and are no longer valid.

PR #2040, opened 2016-11-14
[MERGE WITH GITFLOW] removed FEC url form model. I think it conflicted with the mixin

LindsayYoung:
should fix efile file endpoint


ccostino:
Thanks, @LindsayYoung!

PR #2052, opened 2016-11-22
[DON'T MERGE] Feature/db rename

LindsayYoung:
Adapting the updates to the new view tables, they are the same but don't have pg_update we need to make our own load date to grab the new records for nightly refresh on schedule e


LindsayYoung:
OK, this is working on dev, I will make new test data and then we can coordinate this with the FEC DBAs

LindsayYoung:
I think I got the merge fixed locally, I will re-run it on dev and make a new test set if it works.

LindsayYoung:
The code looks like it is ready to go, but we have to coordinate changes with DBAs to update code as we remake the DB.

If I could get a review now so that it is ready to go, that would be helpful.

data/sql_setup/prepare_schedule_e.sql
```diff
@@ -217,7 +217,7 @@ end
 $$ language plpgsql;
 
 drop trigger if exists ofec_sched_e_queue_trigger on fec_vsum_sched_e;
-create trigger ofec_sched_e_queue_trigger before insert or update or delete
+create trigger ofec_sched_e_queue_trigger instead of insert or update or delete
```
ccostino: We want all record processing to go directly against the queue tables instead of the source table?  My understanding of `instead of` means that any record change that is intended to happen against the `fec_vsum_sched_e` will happen only in the queue tables now with this change.  I just wanted to make sure that this is the desired behavior (and I think we talked about this before, but might be blanking on a detail from that conversation!).

data/sql_setup/prepare_schedule_e.sql
```diff
@@ -232,7 +232,7 @@ create trigger ofec_f57_trigger before insert or update or delete
 ;
 
 drop trigger if exists fec_vsum_f57_trigger on fec_vsum_f57;
-create trigger fec_vsum_f57_trigger before insert or update or delete
+create trigger fec_vsum_f57_trigger instead of insert or update or delete
```
ccostino: Same note as above but for the `fec_vsum_f57` table.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
```diff
@@ -33,6 +33,10 @@ create index on ofec_sched_a_aggregate_size_tmp (size);
 create index on ofec_sched_a_aggregate_size_tmp (total);
 create index on ofec_sched_a_aggregate_size_tmp (count);
 
+-- this drops totals during rebuild
+drop table if exists ofec_sched_a_aggregate_state cascade;
+drop table if exists ofec_sched_a_aggregate_size_old cascade;
```
ccostino: Would it make sense to remove the `alter table` statements below this if we're just going to drop them wholesale here?

webservices/common/models/dates.py
```diff
@@ -85,7 +85,7 @@ class ElectionClassDate(db.Model):
     senate_class = db.Column(db.Integer, index=True, doc=docs.SENATE_CLASS)
 
 
-class CalendarDate(BaseModel):
+class CalendarDate(db.Model):
```
ccostino: There is no more need of the `idx` column that the `BaseModel` provides?  I see the schema change below, so I'm assuming there's not; just wanted to double check!

data/sql_setup/prepare_schedule_e.sql
LindsayYoung: Maybe I am misunderstanding, but because it is going off of a view instead of a table, it needs the "instead of" rather than before.  I thought it should be fulfilling the same function that it did before. I remember reading up on it, but now I can't find where it came from, so I could be wrong.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
LindsayYoung: I added these statements so that you could do update_all more than once. If there is currently an aggregate_size_old table, it needs to dropped because it is two iterations ago, then the current aggregate_size table becomes the new aggregate_size_old table. 

If I am understanding you right, you saying that the cascade will get rid of the old old table which will get rid of the current table. (I was not sure of that dependency) In that case, renaming the current table is not helpful.

webservices/common/models/dates.py
LindsayYoung: We already have idx in the calendar date model as event_id. For the calendar table we needed a id so you could pull up a single event. This is where the warning about ids on  calendar date comes from when you run ./manage.py

data/sql_setup/prepare_schedule_e.sql
jontours: Since this was changed to a view how does the trigger work here? Sorry if that was covered already.

jontours: this == `fec_vsum_sched_e`

ccostino: Oh, I wasn't aware that there was a difference in the case of tables vs. views with triggers in that sense.  I should read up more too, but I had glanced at the beginning of the [Postgres Trigger documentation](https://www.postgresql.org/docs/9.5/static/sql-createtrigger.html) for that info.

webservices/common/models/dates.py
ccostino: Oh nice, so that warning will go away! :-)

data/sql_setup/prepare_schedule_e.sql
jontours: Yeah I just read up on it a bit and it seems as though it will work as expected if we are assuming that the scripts that the FEC is running are making inserts, updates, and deletes directly on the view ( take a look here for example: https://vibhorkumar.wordpress.com/2011/10/28/instead-of-trigger/).  So that would have to coordinated with the FEC, and sorry if that was what is meant with coordinating with the DBAs, have finally found some time to focus on this.  

Also, if I'm understanding correctly, if they are indeed doing inserts on the view, than it seems as though the insert, delete, etc. has to be ran on the underlying table as well, we aren't doing that so assuming the FEC is, things will be fine there too.

LindsayYoung:
Yea, this is the definition I was looking at earlier  https://www.postgresql.org/docs/9.5/static/trigger-definition.html

jontours:
In the section "Updatable Views" captures all the bullet points of what a "simple view" is.  If the view in question here meets those bullet points I think we'll be ok.

ccostino:
Ah nice, thanks for linking that too!

In this case, however, it's not an updatable view; `fec_vsum_sched_e` is a table.  There is a `fec_vsum_sched_e_vw` though (at least in the staging DB), but that has more than one entry in the `FROM` clause so it violates the first bullet in that list (it also violates the simple reference rule, there's a `CASE` statement in there).

ccostino:
Oh, but in `dev`, it is a view - should have checked there first.  In any case, it still doesn't meet the criteria of an updatable view for the same reasons.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
ccostino: Oh, no, what I meant was we have these two statements right below these new `drop table` statements:

```sql
alter table if exists ofec_sched_a_aggregate_size rename to ofec_sched_a_aggregate_size_old;
alter table ofec_sched_a_aggregate_size_tmp rename to ofec_sched_a_aggregate_size;
```

But in reading them again without tired eyes, I see what you mean and these are working against different tables.  Please disregard, sorry!

ccostino:
More notes on the updatable views:

>If the view is automatically updatable the system will convert any INSERT, UPDATE or DELETE statement on the view into the corresponding statement on the underlying base relation.

>If an automatically updatable view contains a WHERE condition, the condition restricts which rows of the base relation are available to be modified by UPDATE and DELETE statements on the view. However, an UPDATE is allowed to change a row so that it no longer satisfies the WHERE condition, and thus is no longer visible through the view. Similarly, an INSERT command can potentially insert base-relation rows that do not satisfy the WHERE condition and thus are not visible through the view.

>A more complex view that does not satisfy all these conditions is read-only by default: the system will not allow an insert, update, or delete on the view. You can get the effect of an updatable view by creating INSTEAD OF triggers on the view, which must convert attempted inserts, etc. on the view into appropriate actions on other tables. For more information see CREATE TRIGGER. Another possibility is to create rules (see CREATE RULE), but in practice triggers are easier to understand and use correctly.

>Note that the user performing the insert, update or delete on the view must have the corresponding insert, update or delete privilege on the view. In addition the view's owner must have the relevant privileges on the underlying base relations, but the user performing the update does not need any permissions on the underlying base relations (see Section 38.5).

That second to last paragraph applies to us in this situation I think, along with any necessary privileges.

LindsayYoung:
We will need the test data going again to make sure this works.

We could also try using the triggers directly and not doing an not using queue tables on this.

jontours:
In reference to the issue (https://github.com/18F/openFEC/issues/2072) with missing data for schedule_e, the update process that empties out queues and adds them to the `ofec` table relies on a join on the table `disclosure.nml_form_24`.  It seems that the FEC is now updating `disclosure.v_sum_and_det_sum_report`.  The function `ofec_sched_e_notice_update_from_f24` in `update_schedule_e.sql` does a join on the old table and hence can never add the new entries.

LindsayYoung:
Don't merge. We should keep this around because the dev database uses this naming scheme. We will talk to DBA team when they are back and get the names back in parody across the databases.

PR #2054, opened 2016-11-22
Migrating filings off data warehouse

jontours:
First cut at moving the tables.  Most have them been moved.  However discovered that `dimcmteproperties` most likely can't be replaced by `cmte_valid_fec_yr` in `sql_updates\committee_history.sql`, as many columns are missing.  It seems that `disclosure.nml_form_1` is a better candidate to replace that table.  Will verify with partners if that inituition is right.


data/sql_updates/committee_detail.sql
```diff
@@ -13,7 +13,7 @@ create index on ofec_committee_detail_mv_tmp(party);
 create index on ofec_committee_detail_mv_tmp(state);
 create index on ofec_committee_detail_mv_tmp(party_full);
 create index on ofec_committee_detail_mv_tmp(designation);
-create index on ofec_committee_detail_mv_tmp(expire_date);
+--create index on ofec_committee_detail_mv_tmp(expire_date);
```
ccostino: Should we just remove this line entirely?

data/sql_updates/committee_history.sql
```diff
@@ -32,53 +32,54 @@ select distinct on (fec_yr.cmte_id, fec_yr.fec_election_yr)
     fec_yr.cmte_nm as name,
     fec_yr.tres_nm as treasurer_name,
     to_tsvector(fec_yr.tres_nm) as treasurer_text,
-    dcp.org_tp as organization_type,
-    expand_organization_type(dcp.org_tp) as organization_type_full,
-    dcp.expire_date as expire_date,
+    f1.org_tp as organization_type,
+    expand_organization_type(f1.org_tp) as organization_type_full,
+    --verify, but this column is gone now
+    null as expire_date,
```
ccostino: Is this verified? :-)

ccostino: Also, is it used or referenced anywhere, or just purely supplied by the API as a part of a record?

data/sql_updates/filings.sql
```diff
@@ -4,57 +4,58 @@ select
     row_number() over () as idx,
     cand.candidate_id as candidate_id,
     cand.name as candidate_name,
-    fh.committee_id as committee_id,
+    --verify that cand_cmte_id is indeed just candidate's committee id--
+    filing_history.cand_cmte_id as committee_id,
```
ccostino: Is this also verified?

data/sql_updates/filings.sql
```diff
@@ -4,57 +4,58 @@ select
...
-left join ofec_candidate_history_mv_tmp cand on fh.committee_id = cand.candidate_id and get_cycle(fh.report_year) = cand.two_year_period
-left join dimreporttype report on fh.report_type = report.rpt_tp
-where report_year >= :START_YEAR
+    means_filed(begin_image_num) as means_filed,
+    report_fec_url(begin_image_num::text, filing_history.file_num::integer) as fec_url
+from disclosure.f_rpt_or_form_sub filing_history
+left join ofec_committee_history_mv_tmp com on filing_history.cand_cmte_id = com.committee_id and get_cycle(filing_history.rpt_yr) = com.cycle
+left join ofec_candidate_history_mv_tmp cand on filing_history.cand_cmte_id = cand.candidate_id and get_cycle(filing_history.rpt_yr) = cand.two_year_period
+left join staging.ref_rpt_tp report on filing_history.rpt_tp = report.rpt_tp_cd
+where rpt_yr >= :START_YEAR
```
ccostino: Should this still be referenced as `report_year`?

tasks.py
```diff
@@ -26,6 +26,8 @@
     'dimcandproperties',
     'f_rpt_or_form_sub',
     'f_item_receipt_or_exp',
+    #'staging.ref_pty',
+    #'staging.ref_rpt_tp',
```
ccostino: Should we just take these out?

tests/common.py
```diff
@@ -21,14 +21,19 @@
 def _reset_schema():
     rest.db.engine.execute('drop schema if exists public cascade;')
     rest.db.engine.execute('drop schema if exists disclosure cascade;')
+    rest.db.engine.execute('drop schema if exists staging cascade;')
     rest.db.engine.execute('create schema public;')
     rest.db.engine.execute('create schema disclosure;')
+    rest.db.engine.execute('create schema staging;')
```
ccostino: Why is this needed here and not below?

ccostino:
Also, have these SQL updates been run in `dev` already, or just locally?

data/sql_updates/committee_detail.sql
jontours: I'll remove it since it looks to be deprecated.

data/sql_updates/committee_history.sql
jontours: No use of it on the front-end.  We can ask Paul if that's ok; if so then maybe drop a line in the release notes that it is being deprecated.

data/sql_updates/filings.sql
jontours: Coming back to this now I'd say it's the correct column.

data/sql_updates/committee_history.sql
ccostino: Thanks!

data/sql_updates/filings.sql
jontours: Where `ofec_filings_mv` is built in `data\filings.sql` it is indeed referenced as `rpt_yr as report_year`.  So I'd say it's fine to leave this way here, and do all the renamings in `filings.sql`.

tests/common.py
jontours: This has to with the same problem I was having in the PR for migrating some tables to the `disclosure` schema.  The integration tests fail if those create schema statements are there. And the reason, or so I suspect, is in the subset dump the dump is exporting a statement to create those other schemas (but not the public schema!).  So having the create statements in the test logic was clashing and causing an error in the tests.  I did a little research to see if there is a way to exclude those create schema statements but didn't find anything.  Great documentation for leaving out tables; just not schemas.  Maybe can create an issue to get to later under `py.test grooming` from our quarterly planning.

data/sql_updates/filings.sql
ccostino: Ah okay - I was noting the change above and saw the column was renamed there, but this is referring to its original name.  Note that you do need to use the original column name for things like `ORDER BY` and `GROUP BY`, so it's confusing at times. :-)

tests/common.py
ccostino: Oh, right, I forgot about that!  Yes, that would probably be good to file a separate issue for this, even if only to document the behavior.  Thanks for explaining again, I appreciate it!

LindsayYoung:
updating schemas on dev and staging now

PR #2060, opened 2016-11-23
Feature/rad id

LindsayYoung:
This adds another ID for RAD employees, as requested by @jwchumley

### DIFF
--- a/data/sql_updates/rad_analyst.sql
+++ b/data/sql_updates/rad_analyst.sql
@@ -16,6 +16,7 @@ select
     -- I am not sure we want to filter on this, if the name is different here it will behave differently than the other committee searches by name
     cmte_nm as committee_name,
     anlyst_id as analyst_id,
+    cast(anlyst_short_id as numeric) as analyst_short_id,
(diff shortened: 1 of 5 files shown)


data/sql_updates/rad_analyst.sql
```diff
@@ -16,6 +16,7 @@ select
     -- I am not sure we want to filter on this, if the name is different here it will behave differently than the other committee searches by name
     cmte_nm as committee_name,
     anlyst_id as analyst_id,
+    cast(anlyst_short_id as numeric) as analyst_short_id,
```
jontours: Haha.  Lovin' it!

jontours: On serious note, is there a reason this was needed to cast to numeric, noted that it is type numeric(4,0) in the view.

LindsayYoung: Both IDs are numbers and the other ID is numeric typed, so I thought the display of numeric ids should be consistent.

jontours:
Tested filtering while pointing to dev, looks good!

PR #2061, opened 2016-11-25
Removing dimparty replacing with new table

### LINKED ISSUE #2051: Migrate off dimparty
Warehouse | Postgres
 ----------- | --------- 
 dimparty | staging.ref_pty 

- this should be easy

LindsayYoung:
Updated the task even though it is not working now
closes #2051


LindsayYoung:
I thought I could add the schema and table to the test data but it doesn't work for the tests, it says the schema already exists. I think @jontours ran into this before. Any ideas?

jontours:
Also covered here: https://github.com/18F/openFEC/pull/2054.  Sorry! I have to get better at linking issues to their PRs.

PR #2078, opened 2016-12-02
Feature/delete aos from es

vrajmohan:


anthonygarvan:
looks good! And thanks for the cleanup.

PR #2083, opened 2016-12-05
Casting columns to text, and then to dates

jontours:
Casting date columns to text (as they are integer columns) and then to date.


ccostino:
Thanks, @jontours!

PR #2087, opened 2016-12-06
Add RFAI filings to filings resource

jontours:
RFAIs were moved to their own table, this unions that table to bring in those records.  Most of the original columns were supported except for `pages`, can't find an obvious replacement for this column in the table `disclosure.nml_form_rfai`.

cc/ @ccostino @LindsayYoung


ccostino:
So far so good to me, thanks @jontours!  @LindsayYoung, you mentioned something about the question regarding the `pages` column still being useful, but was that in the context of this change or a different one?

data/sql_updates/filings.sql
```diff
@@ -1,60 +1,128 @@
...
+        'RFAI'::text as form_type,
+        rpt_yr as report_year,
+        get_cycle(rpt_yr) as CYCLE,
+        rpt_tp as report_type,
+        null::character varying(1) as document_type,
+        null::text as document_type_full,
+        begin_image_num::bigint as beginning_image_number,
+        end_image_num as ending_image_number,
+        0 as pages,
+        0.00 as total_receipts,
```
LindsayYoung: Can these bee null instead of 0?

jontours: They can be null, maybe that's best since we can't say for sure any of them are zero?  Or being zero wouldn't make sense.  I'll push that change shortly.

LindsayYoung:
Looks great

Ran on dev, will run on staging now

PR #2091, opened 2016-12-08
Feature/add ao advanced search api

### LINKED ISSUE #2056: Add API support for initial AO filters
In order to allow users the ability to further refine their search from a broad universe to a smaller one, add an initial set of filters to the AO search based on what we currently have data to support:

- AO number
- AO name
- AO issued date

This task is just to add the API support for these fields. I will make a separate issue for front-end implementation.

**Criteria for completion:**
- [ ] AO number accepts multiple values and returns all AOs with those numbers
- [ ] AO name accepts multiple values and returns all AOs with that name
- [ ] AO issued date accepts `min_date` and `max_date` and returns all AOs issued within the range

**Questions:**
For AO number and name, are we doing a fuzzy match or exact match? i.e. Does searching `197` in AO Number return `AO 1979-53: Ownership Campaign`? Does searching `Own` in AO Name return it?

For further background discussion, see https://github.com/18F/fec-eregs/issues/276.

anthonygarvan:
Closes #2056.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -142,8 +142,8 @@ query = {
     'type': fields.Str(required=False, description='Document type to refine search by'),
     'ao_no': fields.List(IStr, required=False, description='Force advisory opinion number'),
     'ao_name': fields.List(IStr, required=False, description='Force advisory opinion name'),
-    'ao_min_date': fields.Date(descripion="Minimum issue date of advisory opinion (inclusive)"),
-    'ao_max_date': fields.Date(description="Maximum issue date of advisory opinion (inclusive)")
+    'ao_min_date': fields.Date(description="Earliest issue date of advisory opinion"),
+    'ao_max_date': fields.Date(description="Latest issue date of advisory opinion")
 }
 
 candidate_detail = {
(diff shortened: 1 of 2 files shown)


webservices/args.py
```diff
@@ -136,10 +136,14 @@ def make_seek_args(field=fields.Int, description=None):
...
 query = {
-    'q': fields.Str(required=True, description='Text to search legal documents for.'),
+    'q': fields.Str(required=False, description='Text to search legal documents for.'),
     'from_hit': fields.Int(required=False, description='Get results starting from this index.'),
     'hits_returned': fields.Int(required=False, description='Number of results to return (max 10).'),
-    'type': fields.Str(required=False, description='Document type to refine search by')
+    'type': fields.Str(required=False, description='Document type to refine search by'),
+    'ao_no': fields.List(IStr, required=False, description='Force advisory opinion number'),
+    'ao_name': fields.List(IStr, required=False, description='Force advisory opinion name'),
+    'ao_min_date': fields.Date(descripion="Minimum issue date of advisory opinion (inclusive)"),
```
vrajmohan: Suggestion: for the description, earliest and latest are probably more apt than minimum and maximum. With this change, we could also get rid of the (rather technical) "(inclusive)".

PR #2093, opened 2016-12-08
Build amendment chain, and link to filings resources

jontours:
Initial sql done.  Getting the PR open as most of the sql is done.  Steps after this:
- [x] build MV(s) containing amendment chains for presidential, house-senate, and pac/party MVs.
- [x] link to `ofec_filings_mv`
- [x] link to presidential, house-senate, and pac/party MVs
- [x] link to individual committee views


jontours:
@ccostino this is ready for review.

data/sql_updates/filing_amendments_house_senate.sql
```diff
@@ -0,0 +1,105 @@
...
+    f3.rpt_tp,
+    f3.amndt_ind,
+    f3.receipt_dt,
+    f3.file_num,
+    f3.prev_file_num,
+    f3.mst_rct_file_num,
+    (oldest.amendment_chain || f3.file_num)::numeric(7,0)[],
+    (oldest.date_chain || f3.receipt_dt)::timestamp[],
+    oldest.depth + 1,
+    oldest.amendment_chain[1]
```
ccostino: This caught my eye and I was going to ask if this was intentional (I'm assuming it is, but wanted to verify), however I came across this in the [PostgreSQL Array documentation](https://www.postgresql.org/docs/9.5/static/arrays.html):

>The syntax for CREATE TABLE allows the exact size of arrays to be specified, for example:

   ```sql
   CREATE TABLE tictactoe (
      squares   integer[3][3]
   );
   ```

> However, the current implementation ignores any supplied array size limits, i.e., the behavior is the same as for arrays of unspecified length.

So I guess it's technically not needed?  No need to change, but just found it interesting!

data/sql_updates/filing_amendments_house_senate.sql
```diff
@@ -0,0 +1,105 @@
...
+  from oldest_filing_paper oldest, disclosure.nml_form_3 f3
+  where f3.amndt_ind = 'A' and f3.rpt_tp = oldest.rpt_tp and f3.rpt_yr = oldest.rpt_yr and f3.cmte_id = oldest.cmte_id and f3.file_num < 0 and f3.receipt_dt > date_chain[array_length(date_chain, 1)]
+), longest_path as
+ --here we find the longest paths
+  (SELECT b.*
+   FROM oldest_filing_paper a LEFT OUTER JOIN oldest_filing_paper b ON a.file_num = b.file_num
+   WHERE a.depth < b.depth
+), filtered_longest_path as
+  --filterng out erroneuou paths that skip intermediate dates
```
ccostino: Nit pick - spelling.

data/sql_updates/filing_amendments_house_senate.sql
```diff
@@ -0,0 +1,105 @@
...
+    SELECT cmte_id, rpt_yr, rpt_tp, amndt_ind, receipt_dt, file_num, prev_file_num, mst_rct_file_num, array[file_num]::numeric[] as amendment_chain, 1 as depth, file_num as last
+    FROM disclosure.nml_form_3
+    WHERE file_num = prev_file_num AND file_num = mst_rct_file_num and file_num > 0
+  )
+  union
+  select f3.cmte_id, f3.rpt_yr, f3.rpt_tp, f3.amndt_ind, f3.receipt_dt, f3.file_num, f3.prev_file_num, f3.mst_rct_file_num, (oldest.amendment_chain || f3.file_num)::numeric(7,0)[], oldest.depth + 1, oldest.amendment_chain[1]
+  from oldest_filing oldest, disclosure.nml_form_3 f3
+  where f3.prev_file_num = oldest.file_num and f3.rpt_tp = oldest.rpt_tp and f3.file_num <> f3.prev_file_num and f3.file_num > 0
+),
+--this joins the right sight to left having the effect that the max depth row will be null,
```
ccostino: Right `side`? :-)

data/sql_updates/filing_amendments_house_senate.sql
```diff
@@ -0,0 +1,105 @@
...
+    amndt_ind,
+    receipt_dt,
+    file_num,
+    prev_file_num,
+    mst_rct_file_num,
+    amendment_chain
+from filtered_longest_path;
+
+drop index if exists file_number_house_senate_paper_index;
+create unique index file_number_house_senate_paper_index on ofec_house_senate_paper_amendments_mv_tmp(idx);
```
ccostino: In thinking about this more and referencing the other scripts we have, we've created indexes with `_tmp` in the name before and haven't had to worry about dropping them (because they'd be dropped with the materialized view itself).  This seems to be the convention we've been following, though I don't feel strongly about changing this either.

data/sql_updates/filing_amendments_pac_party.sql
```diff
@@ -0,0 +1,108 @@
...
+  old_f.receipt_dt,
+  old_f.file_num,
+  old_f.prev_file_num,
+  mrf.file_num as mst_rct_file_num,
+  old_f.amendment_chain
+from oldest_filing old_f inner join most_recent_filing mrf on old_f.cmte_id = mrf.cmte_id and old_f.last = mrf.last
+) select row_number() over () as idx, * from electronic_filer_chain;
+
+drop index if exists file_number_pac_party_electronic_index;
+create unique index file_number_pac_party_electronic_index on ofec_pac_party_electronic_amendments_mv_tmp(idx);
```
ccostino: Same note as above regarding the indexes.

data/sql_updates/filing_amendments_pac_party.sql
```diff
@@ -0,0 +1,108 @@
...
+      flp.file_num,
+      flp.prev_file_num,
+      prf.file_num as mst_rct_file_num,
+      flp.amendment_chain
+      from filtered_longest_path flp inner join paper_recent_filing prf on flp.cmte_id = prf.cmte_id
+        and flp.last = prf.last)
+    select row_number() over () as idx, * from paper_filer_chain;
+
+drop index if exists file_number_pac_party_paper_index;
+create unique index file_number_pac_party_paper_index on ofec_pac_party_paper_amendments_mv_tmp(idx);
```
ccostino: Same note as above regarding the indexes.

ccostino: And so on for the others. :-)

data/sql_updates/reports_pac_party.sql
```diff
@@ -1,8 +1,8 @@
-drop materialized view if exists ofec_reports_pacs_parties_mv_tmp;
+drop materialized view if exists ofec_reports_pacs_parties_mv_tmp cascade;
```
ccostino: Just double checking: this `cascade` won't have any unintended side effects, will it?

webservices/common/models/filings.py
```diff
@@ -56,6 +57,10 @@ class Filings(CsvMixin, db.Model):
     fec_url = db.Column(db.String)
     means_filed = db.Column(db.String, doc=docs.MEANS_FILED)
 
+    amendment_chain = db.Column(ARRAY(db.Numeric))
```
ccostino: Nice!

data/sql_updates/filing_amendments_house_senate.sql
jontours: Ah this was tricky.  This is actually dereferencing the array and grabbing the oldest amendment in the chain, which is then used later in the process to join the table on itself to find the longest path.

data/sql_updates/filing_amendments_house_senate.sql
jontours: Ah that makes sense, will add that.

data/sql_updates/reports_pac_party.sql
jontours: Good catch!  Unintended...

webservices/common/models/filings.py
jontours: I'll add a doc that mimics what I slacked about the ordering here.

data/sql_updates/filing_amendments_house_senate.sql
ccostino: OH!  Right, there's no `ARRAY` casting going on, I missed that before, thanks!

ccostino:
Awesome, all of the changes look good, thanks @jontours!  There's only one other question I have; I just noticed that all of the other materialized views have a `cascade` in their drop statements.  I had called out the one but missed the others for some reason.  Should those also be removed?

jontours:
@ccostino The one you called out was indeed unintentional and didn't need to be there.  The other cascade statements do need to be there though.  The reason is all the `mv`s are collected into another `mv`, which you can see in `filings.sql`.  When that view drops it cascades to all the others, and postgres complains if you don't also have a cascade on the child tables.  The cascade statement is rather innocuous since the only thing it cascades to from there is dropping the indices for those tables.

ccostino:
Ah okay, great!  Thanks for clarifying that, I appreciate it.  This is good to go then, will merge!

PR #2098, opened 2016-12-14
Feature/use index aliases

vrajmohan:


data/sql_updates/committee_history.sql
```diff
@@ -34,7 +34,7 @@ select distinct on (fec_yr.cmte_id, fec_yr.fec_election_yr)
     to_tsvector(fec_yr.tres_nm) as treasurer_text,
     f1.org_tp as organization_type,
     expand_organization_type(f1.org_tp) as organization_type_full,
-    null as expire_date,
+    null::date as expire_date,
```
anthonygarvan: why are you making this change? Looks reasonable to me just wanted to make sure it was intentional.

anthonygarvan:
nice!

PR #2108, opened 2016-12-18
Protect against missing index

vrajmohan:
When initializing for the 1st time, the index may not be present


anthonygarvan:
looks good :)

PR #2113, opened 2016-12-20
Feature/add amended flags

jontours:
Only interesting thing here is that reports already have an `is_amended` flag, I left that flag alone and added `most_recent` to reports.  Also filings gained the flags `is_amended` and `most_recent`.


ccostino:
All looks good to me here, thanks @jontours!

PR #2115, opened 2016-12-21
Add commands to support index alias management

vrajmohan:
The commands are rather opinionated:
create_staging_index creates a new index `docs_staging` and moves `docs_index`
to point to it.
restore_from_staging_index moves documents from `docs_staging` back to `docs`
and restores the aliases `docs_index` and `docs_search`.


anthonygarvan:
looks good, nice work!

PR #2118, opened 2016-12-23
Add docstrings for management commands

vrajmohan:


LindsayYoung:
Thanks @vrajmohan this is great

PR #2120, opened 2016-12-26
Feature/separate text by document type

vrajmohan:


webservices/legal_docs/current_murs.py
```diff
@@ -244,16 +243,16 @@ def get_documents(case_id, bucket, bucket_name):
                 'category': row['category'],
                 'description': row['description'],
                 'length': row['length'],
+                'text': row['ocrtext'],
```
ccostino: Just making sure on this:  You don't need to have the extra space on the end here like the line below that this replaces?

ccostino:
Thanks, @vrajmohan!

Issue #2057, opened 2016-11-23
Add API support for additional AO filters


noahmanger:
In order to allow users the ability to further refine their search from a broad universe to a smaller one, add additional filters to the AO search beyond the [initial filters](https://github.com/18F/openFEC/issues/2056):

**Completion criteria:**
- [x]  **Pending AO:** Should be a boolean flag based on whether an AO has been closed with either a final opinion or withdrawn. 
- [x] **Document type:** Accepts the following options:
  - Final Opinion
  - Withdrawal of Request
  - AO Request, Supplemental Material and Extensions of Time
  - Comments and ex parte Communications
  - Draft Documents
  - Commissioner Statements
  - Votes
- [x] **Requester entity type:** Accepts one or more of the following options:
![image](https://cloud.githubusercontent.com/assets/1696495/21246866/2f4a631a-c2e0-11e6-8ba2-dffe98ac576f.png)
- [x] **Requester name:** Accepts multiple strings that return matches of the name of the entity who requested the AO

For further background, see [this issue](https://github.com/18F/fec-eregs/issues/276).

nickykrause:
description of [entity type](http://saos.fec.gov/saos/tips.html#_Entity_Roles_and)

noahmanger:
After talking this through with @nickykrause just now, I realized that there's a few more open questions around all of this stuff and that I lost somethings in translation. So to clarify:

First, I think the original determination [made here](https://github.com/18F/fec-eregs/issues/276#issuecomment-258505291) to treat "Pending AOs" as a possible value in "Document type" is actually incorrect. I believe what we want here is actually a boolean flag that's set on the AO _as a whole_ that says whether or not this AO has been resolved or if it's still pending.

This is different from a "Document type" filter which essentially is a refinement tool on the keyword search, telling the system to _only_ search within the selected documents _within_ an AO for matches. @AmyKort is this understanding correct? 

And if so, @anthonygarvan do we have the necessary data to support a `pending: true / false` type flag? Do we even have the pending AOs? I see there's [just a couple](http://saos.fec.gov/saos/searchao;jsessionid=51CB60DED462B492F42996DA17B6C74A?SUBMIT=pending).

As for the "Document type" filter, it accept one or more of the 7 possible values: 
  - Final Opinion	
  - Withdrawal of Request
  - AO Request, Supplemental Material and Extensions of Time	
  - Comments and ex parte Communications
  - Draft Documents	
  - Commissioner Statements
  - Votes

[...]

PR #2127, opened 2016-12-28
Feature/add additional ao fields

anthonygarvan:
this PR includes new indexing scheme and should include the extended API support for additional fields from #2057.


webservices/legal_docs/load_legal_docs.py
```diff
@@ -145,10 +145,17 @@ def index_advisory_opinions():
 
         es = utils.get_elasticsearch_connection()
 
-        result = db.engine.execute("""select DOCUMENT_ID, OCRTEXT, DESCRIPTION,
+        result = db.engine.execute("""SELECT DOCUMENT_ID, OCRTEXT, DESCRIPTION,
```
vrajmohan: Sorry to be a stick-in-the-mud, but I think it helps immensely if the SQL is formatted with uppercase keywords and lowercase everything else.

webservices/legal_docs/load_legal_docs.py
```diff
@@ -159,6 +166,20 @@ def index_advisory_opinions():
         for row in result:
             key = "legal/aos/%s.pdf" % row[0]
             pdf_url = "https://%s.s3.amazonaws.com/%s" % (bucket_name, key)
+
+            respondents = db.engine.execute("""select e.name, et.description
```
vrajmohan: ditto - SQL formatting

webservices/resources/legal.py
```diff
@@ -134,28 +134,66 @@ def get(self, q='', from_hit=0, hits_returned=20, type='all', **kwargs):
...
-            Q('match', documents__text=kwargs.get('document_text'))]
+            Q('match', documents__category=kwargs.get('mur_document_category')),
+            Q('match', documents__text=kwargs.get('mur_document_text'))]
         query = query.query("nested", path="documents", query=Q('bool', must=combined_query))
 
     return query
 
 def apply_ao_specific_query_params(query, **kwargs):
-    query = query.query('match', category='Final Opinion')
+    categories = {'F': 'Final Opinion',
```
vrajmohan: I may be wrong, but it seems to me that this mapping between abbreviations and values, and the corresponding one for `requestor_types` below may be duplicated somewhere else for the front-end. Can we ensure that this is present [Once and Only Once](http://c2.com/xp/OnceAndOnlyOnce.html)?

anthonygarvan: @vrajmohan probably the way to do that would be to expose it in the api. Seems strange to have an endpoint for a static value though. I guess it would be good in case it ever changes though? Ugh, I don't know.

vrajmohan:
Nice work!

PR #2128, opened 2016-12-28
Feature/raw sked e endpoint

jontours:
Close to ready, just need to add some valid test data.


jontours:
Adding some tests and this will be ready for review.

tests/test_itemized.py
```diff
@@ -425,3 +425,21 @@ def test_filters_sched_e(self):
...
+            ('support_oppose_indicator', ScheduleEEfile.support_oppose_indicator, ['S', 'O']),
+            ('is_notice', ScheduleEEfile.is_notice, [True, False]),
+            ('filing_form', ScheduleEEfile.form_type, ['F24N', 'F3XN'])
+        ]
+        for label, column, values in filters:
+            print(label, column, values)
+            [
+                factories.ScheduleEEfileFactory(**{column.key: value})
+                for value in values
+            ]
```
ccostino: Nice setup here!  May not need the `print` statement at this point, but not an issue.

webservices/args.py
```diff
@@ -615,6 +615,23 @@ def _deserialize(self, value, attr, data):
     'is_notice': fields.List(fields.Bool, description='Record filed as 24- or 48-hour notice'),
 }
 
+schedule_e_efile = {
+    #'cycle': fields.List(fields.Int, description=docs.RECORD_CYCLE),
```
ccostino: No need for cycle on this?  I noticed the `report_year` commented out in the factory in the tests above too, so just wanted to double check.

webservices/common/models/itemized.py
```diff
@@ -269,7 +297,7 @@ def pdf_url(self):
         return None
 
 
-class ScheduleE(BaseItemized):
+class ScheduleE(BaseItemized, PdfMixin):
```
ccostino: Remember about Python's MRO (Method Resolution Order) - you'll want the base class you intended to use as the last thing listed here and any mixins or other abstract classes before it:

```python
class ScheduleE(PdfMixin, BaseItemized)
```

jontours:
Adding one more column for this and will be ready to go.

webservices/common/models/itemized.py
```diff
@@ -26,6 +30,30 @@ def memoed_subtotal(self):
         return self.memo_code == 'X'
 
 
+class BaseRawItemized(db.Model):
```
ccostino: Is this different enough from the other `BaseItemized` class that it made more sense to create a new one?  Are there a bunch of different fields or just a handful of additional fields from the other?

Is there anticipation for other e-filing records to model in the future?  Would it be possible to inherit from the `BaseItemized` class in this one and go from there?

webservices/resources/sched_e.py
```diff
@@ -84,3 +93,45 @@ def filter_election(self, query, kwargs):
...
+        ('support_oppose_indicator', models.ScheduleEEfile.support_oppose_indicator),
+        ('filing_form', models.ScheduleEEfile.form_type),
+        ('is_notice', models.ScheduleEEfile.is_notice),
+    ]
+
+    @property
+    def args(self):
+        return utils.extend(
+            args.paging,
+            #args.itemized,
```
ccostino: If these are indeed separate from the other itemized models (see note above in the `BaseRawItemized` class), do we need to consider creating separate `args` for them?

webservices/common/models/itemized.py
jontours: I'd say that these new tables are different enough from the regular itemized resources to warrant a new base class (mostly because the processed itemized base class has more columns).  And I at least don't know of a way to use a schema base class that won't error out in SqlAlchemy given our current configuration, may be a way but not sure.

And yes, the new base class will be used for other raw itemized resources (a,b), all the columns in the base class are in common across the e-file tables.

jontours:
@ccostino addressed the concerns, thanks for the great input!

jontours:
argh, failing test, fixing this.

webservices/common/models/itemized.py
ccostino: Gotcha, thanks!  I just wanted to double check but this definitely makes sense.

ccostino:
Awesome, thank you @jontours!

PR #2134, opened 2017-01-03
Bumped dates for new cycle

jontours:
There may be a better way to resolve these problems but if there is it's certainly not obvious.  As the factories create the dates manually.  So not sure where the dates are getting messed up by our current year changing.  If there is a better fix go ahead and close this branch, may be a good opportunity to pair up.

/cc @ccostino @LindsayYoung

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -171,8 +171,8 @@ class ScheduleAFactory(BaseFactory):
     class Meta:
         model = models.ScheduleA
     sub_id = factory.Sequence(lambda n: n)
-    report_year = 2016
-    two_year_transaction_period = 2016
+    report_year = 2017
+    two_year_transaction_period = 2018
 
     @factory.post_generation
     def update_fulltext(obj, create, extracted, **kwargs):
@@ -185,8 +185,8 @@ class ScheduleBFactory(BaseFactory):
     class Meta:
         model = models.ScheduleB
     sub_id = factory.Sequence(lambda n: n)
-    report_year = 2016
-    two_year_transaction_period = 2016
+    report_year = 2017
+    two_year_transaction_period = 2018
 
     @factory.post_generation
     def update_fulltext(obj, create, extracted, **kwargs):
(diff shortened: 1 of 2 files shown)


ccostino:
You know what, I'd like to dive a bit more into this only because the tests shouldn't have broken for the reason you mentioned, @jontours - the dates are hardcoded in the factories.  If you'd like we could spend a bit of time earlier tomorrow taking a look!

jontours:
Agreed, sounds like a plan to me.

ccostino:
@jontours and I uncovered the underlying issue here and there are two things to bear in mind:

* [This line of config](https://github.com/18F/openFEC/blob/develop/webservices/config.py#L27) is used as the default value for itemized data API requests that do not specify a `two_year_transaction_period` value.  What is happening in the tests is that they are detecting `2018` as the cycle end given our current year of `2017` to use as the default value, but the test data is set for `2016` or earlier, so nothing is found and no results are returned in the simulated API requests.
* Consequently, this also means that we need to add child tables to our schedule A and B partitions for the 2017 and 2018 cycle.  It sounds like we can get away with just adding the child tables without having to worry about a full repartition as no data has come in yet.

In terms of the work here, I'm of two minds with it:  we can just leave things alone and fix the partitioning issue on the API side (there is a [separate issue in the web app](https://github.com/18F/openFEC-web-app/issues/1770) piece), or we can make a slight adjustment here and explicitly tell the API to use a specific year as the cycle end date for the itemized requests (instead of `CURRENT_YEAR`, something like `LAST_CYCLE_YEAR = 2016`).

I'm less of a fan of the latter though as it means having to update the code at various points without a real need to if we just provide a quick mechanism to add the underlying tables needed.  The interface should account for the new cycle but that does mean that no results may appear initially since it'll default to the latest cycle at the start of a new odd number year.

I will create a separate issue and PR for the partitioning work, I think I have a quick and easy way to address that piece.

Thanks for catching and looking into this, @jontours!  It turned out to encompass quite a bit more. :-)

LindsayYoung:
Just wanted to clarify that defaulting to the current cycle was intentional, you can always provide a different cycle if you want to. We also have an explanation of defaulting to the current cycle in the docs.

jontours:
We can maybe keep the default as being the current year but just change the test to pass in a filter to the api that specifies the year in the test.

ccostino:
That makes the most sense to me.

PR #2137, opened 2017-01-04
added kwargs for contribution period

jontours:
Defined map for arguments as a member of the test class and passed into all the url endpoints that need it.  Only one test need the year to be manually bumped, and that test explicitly tests without the two year contribution period.

/cc @ccostino @LindsayYoung

### DIFF
--- a/tests/test_itemized.py
+++ b/tests/test_itemized.py
@@ -6,7 +6,7 @@ from tests import factories
 from tests.common import ApiBaseTest
 
 from webservices.rest import api
-from webservices.common.models import ScheduleA, ScheduleE, ScheduleEEfile
+from webservices.common.models import ScheduleA, ScheduleB, ScheduleE, ScheduleEEfile
 from webservices.schemas import ScheduleASchema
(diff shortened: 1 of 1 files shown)


tests/test_itemized.py
```diff
@@ -15,15 +15,17 @@
 
 
 class TestItemized(ApiBaseTest):
+    kwargs = {'two_year_transaction_period': 2016}
```
ccostino: Nice!

tests/test_itemized.py
```diff
@@ -52,19 +54,20 @@ def test_sorting(self):
                 'last_contribution_receipt_date': receipts[0].contribution_receipt_date.isoformat(),
             }
         )
-
+    #This is the only test that the years will have to be bumped when in a new cycle
+    #maybe refactor to use some logic based on current year?
```
ccostino: At some point in the future, yes.  I might be able to work it in the update I'll do for the partitioned tables. :-)

PR #2139, opened 2017-01-05
[MERGE WITH GIT FLOW] Adds support for generating new itemized child tables

### LINKED ISSUE #2136: Add missing 2017 - 2018 cycle tables for schedule A and B
@jontours uncovered an issue in #2134 while trying to fix tests that started failing after the start of the new year, and it led both of us to realize an underlying problem:  we don't have child tables for the new, upcoming cycle of `2017 - 2018`.

I think this is doable relatively quickly and easily without needing to repartition completely unless we're missing data already, but I don't think we are (and I asked in the partners Slack channel to double check that).  In any case, here is what I believe this entails:

* Create a new management command, perhaps `add_itemized_partition_cycle` or something, and have it call the underlying `create_child` method in the partition objects directly with the specific cycle.
* Allow the user to specify a cycle, but by default have it just create one child table with the next cycle based off of last existing table and current calculated cycle (if the table already exists, then just print out a note to the user and exit the command immediately).
* Do this for both schedule A and schedule B together so they stay in sync.

Unfortunately we do need to take care of this now due to the fact that we are in 2017 and data will be coming soon.

ccostino:
**MERGE WITH GIT FLOW**

Closes #2136

This changeset adds a management command to support creating new child tables to an existing partition of Schedule A and Schedule B data. It will check first before just creating the new table(s) so existing data is not wiped out.

This also includes the beginning of a test suite for the partitioning code.

I modified the existing partitioning code a bit to break out some of the underlying child table processing code and ended up having to add a new flag to a couple of those methods to account for when we're doing a full partition vs. operating against already partitioned tables.

The new command is this:

```sh
./manage.py add_itemized_partition_cycle
```

Given the current year at the time of this PR (2017), this will attempt to create the 2017 - 2018 cycle tables.  If you haven't already repartitioned they'll be added, but if you have then it will skip over them.  You can also specify a cycle and/or a range:

```sh
./manage.py add_itemized_partition_cycle -c 2014 -a 5
```

This will start attempting to create child tables for the 2013 - 2014 cycle and work its way up to the 2021 - 2022 cycle (and again, skipping tables that already exist so they're not undone and wiped out).

The testing was proving to be fairly difficult to get going but I wanted to have something given the changes I had to make.  I'm definitely open to suggestions for improvement and/or pairing with someone to take them further!

/cc @LindsayYoung, @jontours

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -261,6 +261,28 @@ def update_aggregates():
     logger.info('Finished updating incremental aggregates.')
 
 @manager.command
+def add_itemized_partition_cycle(cycle=None, amount=1):
+    """Adds a new itemized cycle child table.
+    By default this will try to add just the current cycle to all partitioned
+    itemized schedule tables if it doesn't already exist.
+    If the child table already exists, skip over it.
+    """
+
+    amount = int(amount)
+
+    if not cycle:
+        cycle = SQL_CONFIG['CYCLE_END_YEAR_ITEMIZED']
+    else:
+        cycle = int(cycle)
+
+    logger.info('Adding Schedule A cycles...')
+    partition.SchedAGroup.add_cycles(cycle, amount)
+    logger.info('Finished adding Schedule A cycles.')
+    logger.info('Adding Schedule B cycles...')
+    partition.SchedBGroup.add_cycles(cycle, amount)
+    logger.info('Finished adding Schedule B cycles.')
+
+@manager.command
 def update_all(processes=1):
     """Update all derived data. Warning: Extremely slow on production data.
     """
--- a/webservices/partition/base.py
+++ b/webservices/partition/base.py
@@ -56,6 +56,37 @@ class TableGroup:
         cls.rename()
 
     @classmethod
+    def add_cycles(cls, cycle, amount):
+        """Adds new child tables to an existing partition.
+        Note:  Will not override existing tables.
+        """
+
+        parent = utils.load_table(cls.parent)
+
+        # Calculate all of the cycles to be added at once.
+        cycles = [ cycle + i for i in range(0, amount * 2, 2) ]
+
+        for cycle in cycles:
+            child_name = cls.get_child_name(cycle)
+
+            if utils.load_table(child_name) is None:
+                cls.create_child(parent, cycle, False)
+                cls.rename_child(cycle)
+                logger.info(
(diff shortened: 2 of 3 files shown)


manage.py
```diff
@@ -261,6 +261,28 @@ def update_aggregates():
...
 
 @manager.command
+def add_itemized_partition_cycle(cycle=None, amount=1):
+    """Adds a new itemized cycle child table.
+    By default this will try to add just the current cycle to all partitioned
+    itemized schedule tables if it doesn't already exist.
+    If the child table already exists, skip over it.
+    """
+
+    amount = int(amount)
```
jontours: is this because it can be a string in some instances?  via command line?

webservices/partition/base.py
```diff
@@ -145,27 +185,27 @@ def rename(cls):
         for cmd in cmds:
             db.engine.execute(cmd.format(cls.base_name))
 
-        # Rename child tables
```
jontours: nice refactor!

manage.py
ccostino: Yeah, this is because all of the arguments passed into the management commands come in via strings.  You need to explicitly type cast them (we had been doing this with the `processes` argument elsewhere).

webservices/partition/base.py
ccostino: Thanks!

manage.py
ccostino: I was trying to find more info and while looking up more details I think it might actually be better to revisit these at some point, given the documentation here: https://flask-script.readthedocs.io/en/latest/#adding-arguments-to-commands

I was basing the arguments off of what we already had and I noticed they were coming in as strings, but by setting the options as outlined in the `flask-script` docs, there might be another way of doing this.  That said, this is also simple. :-)

ccostino:
Thank you, @jontours!  Yes, once merged this would need to be run against all 3 environments.  I just took a look in `dev` and noticed that a repartitioning had already been started but stopped at some point.  I don't think it's currently going on, I don't see a partitioner app or any other thing that looks like a one-off app doing that.  So for `dev`, perhaps it makes sense to just repartition again?

`stage` and `prod` appear to be in good shape and should be okay to just run this command instead.  Also, no need to worry about the read replicas, those'll be taken care of automatically. :-)

PR #2141, opened 2017-01-05
Feature/search respondents

vrajmohan:


anthonygarvan:
omg this is brilliant. So happy you got kwargs to work for this!! I had been dreaming of this for some time but thought (I suppose mistakenly) that it would break marshmallow somehow. Great work 🎉 .

PR #2142, opened 2017-01-05
Update New Relic config

ccostino:
Related to https://github.com/18F/openFEC-web-app/issues/1701

This changeset updates the New Relic configuration and dependencies to the latest versions.  We recently ran into an issue with the web app failing to monitor correctly due to this, so we would like to ward off that problem here as well.

/cc @tadhg-ohiggins, @LindsayYoung, @jontours

### DIFF
--- a/newrelic.ini
+++ b/newrelic.ini
@@ -12,7 +12,7 @@
 #
 #    http://docs.python.org/library/configparser.html
 #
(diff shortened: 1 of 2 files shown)


ccostino:
Haha, that was due to the tool auto-generating all of the text, not me, but thanks for noting it! :-)

PR #2143, opened 2017-01-05
Feature/raw sked a endpoint

jontours:
This PR will subsume the changes made here https://github.com/18F/openFEC/pull/2138.  Turned out to properly implement this endpoint I had to solve a lot of problems in the other PR.  This PR also properly implemented eager loaded and silenced the N+1 complaints, a good thing!  This will be ready soon and B will be done shortly after (promise this time).

@ccostino @LindsayYoung


jontours:
This is done!

webservices/args.py
```diff
@@ -632,7 +644,11 @@ def _deserialize(self, value, attr, data):
         IStr(validate=validate.OneOf(['S', 'O'])),
         description='Support or opposition',
     ),
-    'is_notice': fields.List(fields.Bool, description='Record filed as 24- or 48-hour notice'),
+    'min_expenditure_date': fields.Date(description='Selects all items expended by this committee after this date'),
+    'max_expenditure_date': fields.Date(description='Selects all items expended by this committee before this date'),
+    'min_expenditure_amount': fields.Date(description='Selects all items expended by this committee greater than this amount'),
+    'max_expenditure_amount': fields.Date(description='Selects all items expended by this committee less than this date'),
```
ccostino: Would it make sense to move these descriptions into the `docs` module?  Also, should we remove the commented out lines?

webservices/common/models/itemized.py
```diff
@@ -47,9 +45,22 @@ class BaseRawItemized(db.Model):
...
+        return self.load_timestamp.year
+
+    @hybrid_property
     def memoed_subtotal(self):
         return self.memo_code == 'X'
 
+    @hybrid_property
+    def fec_election_type_desc(self):
+        election_map = {'P': 'PRIMARY', 'G': 'GENERAL', 'O': 'OTHER'}
+        return election_map.get(str(self.pgo).upper()[0])
```
ccostino: Is `self.pgo` (in the models that utilize this base class) guaranteed to always have a value that you can pull the first character out of?

webservices/common/models/itemized.py
```diff
@@ -122,6 +133,91 @@ class ScheduleA(BaseItemized):
...
+    """
+    contributor = db.relationship(
+        'CommitteeHistory',
+        primaryjoin='''and_(
+                foreign(ScheduleA.contributor_id) == CommitteeHistory.committee_id,
+                ScheduleA.report_year + ScheduleA.report_year % 2 == CommitteeHistory.cycle,
+            )'''
+    )
+    """
+    #contributor_name = db.Column('contbr_nm', db.String, doc=docs.CONTRIBUTOR_NAME)
```
ccostino: Do we not need these relationships for the efiling version of the Schedule A data?  What about enabling downloads on the data, wouldn't we want to see the committee information like we do for the regular Schedule A data?  Or if it's not needed, should this be removed?

webservices/schemas.py
```diff
@@ -461,14 +461,7 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
...
         ),
-        'relationships': [
-            Relationship(
-                models.ScheduleA.committee,
-                models.CommitteeHistory.name,
-                'committee_name',
-                1
-            ),
-        ],
+
```
ccostino: This related information is no longer needed?  That is going to have an impact on the downloads of Schedule A info, no more committee information will appear.  Was the intention to not have it included on the Schedule A E-Filing schema instead?

webservices/common/models/itemized.py
```diff
@@ -122,6 +133,91 @@ class ScheduleA(BaseItemized):
...
+            self.contributor_first_name,
+            self.contributor_middle_name,
+            self.contributor_suffix
+        )
+        name = (
+            name
+            if name
+            else None
+        )
+        return name
```
ccostino: Interesting!  I hadn't seen this `name_generator` before. :-)

webservices/common/models/itemized.py
jontours: It is not guaranteed to be non-null, but I didn't notice any null pointer references, I can add a nullity check though.

webservices/common/models/itemized.py
jontours: I can see if it's feasible to add the contributor info, but if I remember right the efile table was lacking the needed column to properly make this relation.

webservices/schemas.py
jontours: The committee relation is there, so I can simply add that back in.  When I was coding this I forgot what functionality this provided.  It should work just as the processed resources do.

webservices/common/models/itemized.py
ccostino: Ah gotcha, that makes sense then.

jontours:
@ccostino concerns addressed, also added the views to the whitelist for download ability.

webservices/schemas.py
ccostino: Ah, yeah, this will need to be added back in as well for the regular Schedule A itemized data too.  I just did a test and here's what you see without it:

screen shot 2017-01-13 at 2 42 29 pm

vs. this, with the relationship:

screen shot 2017-01-13 at 2 45 06 pm

It may be a bit hard to see, but the committee name column is there.

ccostino:
Thanks, @jontours!  The `Relationship` tuple needs to be added back to the `ScheduleASchema` as well though.  Also, the PR needs to be updated to go against the new release branch.  Hopefully those two things yield passing tests!

webservices/tasks/download.py
```diff
@@ -44,9 +44,11 @@
     reports.CommitteeReportsView,
     reports.EFilingSummaryView,
     sched_a.ScheduleAView,
+    sched_a.ScheduleAEfileView,
     sched_b.ScheduleBView,
     sched_d.ScheduleDView,
     sched_e.ScheduleEView,
+    sched_e.ScheduleEEfileView,
```
ccostino: Nice!

webservices/schemas.py
jontours: oops, damn ctl-x

PR #2145, opened 2017-01-07
Feature/search final disposition

vrajmohan:


webservices/resources/legal.py
```diff
@@ -133,17 +133,19 @@ def get(self, q='', from_hit=0, hits_returned=20, type='all', **kwargs):
...
+    if kwargs.get('mur_dispositions'):
+        query = query.query('match', disposition__data__disposition=kwargs.get('mur_dispositions'))
     if kwargs.get('mur_election_cycles'):
         query = query.query('term', election_cycles=kwargs.get('mur_election_cycles'))
-    if kwargs.get('mur_document_category') and kwargs.get('mur_document_text'):
+    if kwargs.get('mur_document_category'):
         combined_query = [
             Q('match', documents__category=kwargs.get('mur_document_category')),
-            Q('match', documents__text=kwargs.get('mur_document_text'))]
+            Q('match', documents__text=q)]
```
anthonygarvan: should we use a `term` or `terms` query here instead of match?

anthonygarvan: would also need to make the document category `not_analyzed` in the mapping

anthonygarvan: also can multiple categories be selected at same time?

anthonygarvan: sorry for all the comments. We can have multiple document types checked at the same time, which means it should be a fields.List(IStr...) argument type, and query should be against `terms`. Same thing with disposition (also a checkbox). See front end case: [1735](https://github.com/18F/openFEC-web-app/issues/1735). You can copy what I did for `ao_category`.

vrajmohan: `document_category` has indeed been set as `not_analyzed` in a previous commit.

anthonygarvan: oh great! Then you just need to change the args to accept a list and change the query to be a `terms` instead of `match`.

PR #2173, opened 2017-01-23
Filter rfai records marked for deletion

jontours:
@ccostino this should be a quick review.  `update_schemas` will need to be run after this is merged.


ccostino:
Thanks, @jontours!  Should we make this a hotfix though and branch off of `master` instead?

jontours:
@ccostino yeah that may make more sense

PR #2176, opened 2017-01-23
Remove fec_url from rfai_filings

jontours:
Resolves https://github.com/18F/openFEC/issues/2094.  `update_schemas` must be run for this.

### DIFF
--- a/webservices/schemas.py
+++ b/webservices/schemas.py
@@ -645,7 +645,7 @@ ElectioneeringPageSchema = make_page_schema(ElectioneeringSchema, page_type=pagi
 register_schema(ElectioneeringSchema)
 register_schema(ElectioneeringPageSchema)
 
-FilingsSchema = make_schema(
+BaseFilingsSchema = make_schema(
     models.Filings,
     fields={
         'document_description': ma.fields.Str(),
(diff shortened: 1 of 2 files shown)


ccostino:
Thanks, @jontours!

One question though after looking at the issue some in the context of this change: in addition to this, would it make sense to adjust the response for RFAIs to not include the `fec_url` property at all, or will providing `null` be enough to prevent it from showing up in the web app?

Also, is it all RFAIs that don't have `fec_url`s with them, or just some?  I'm assuming all based on the change, but just wanted to make sure I understood fully.

jontours:
Whoa never got an email for this comment, or missed it anyhow.  They will always be null.  I can ask @noahmanger how he prefers it.  Null or just removed entirely.

ccostino:
D'oh!  Not sure what happened there with the notifications - either way, sounds good, I will hold off on merging until we've had a chance to catch up.  Thanks!

jontours:
Ready for re-view @ccostino, turned out being slightly tricky as in some cases the url has to be removed and sometimes not.  Decorator pattern worked well here.

webservices/schemas.py
```diff
@@ -657,6 +657,12 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'], schemas['Ca
     },
     options={'exclude': ('committee', )},
 )
+class FilingsSchema(BaseFilingsSchema):
+    @post_dump
```
ccostino: Clever way of removing this, nice!

PR #2178, opened 2017-01-25
Link amendment chain to efilings

jontours:
Some work left to do for filters.  So far so good with this, it's performant as well.


jenniferthibault:
Holding for review until @noahmanger is back to check some questions with @jontours

PR #2181, opened 2017-01-30
Adds support to sort by date for entity charts

### LINKED ISSUE #2122: Add date sort to charts
Example url that need a date sort:

https://api.open.fec.gov/v1/totals/by_entity/?cycle=2016&page=1&per_page=20&api_key=DEMO_KEY

ccostino:
Closes #2122 

This changeset adds support for sorting by date in our entity charts.  It accomplishes this by deriving a date from the `month` and `year` columns that are created in the `ofec_entity_chart_mv` materialized view (and always assuming the first of the month as the day).  Note that we already have a `date` property on the Python side in the `EntityReceiptDisbursementTotals` model, so I did not want to disrupt anything currently using that.  Hence the name for this new column and field being `chart_date`.

h/t to @LindsayYoung for the implementation idea!

### DIFF
--- a/data/functions/calendar.sql
+++ b/data/functions/calendar.sql
@@ -382,3 +382,11 @@ returns text as $$
     end
 $$ language plpgsql;
 
+
+-- Retrieves the last day of the month for a given timestamp.
+create or replace function last_day_of_month(timestamp)
+returns timestamp as $$
+    begin
+        return date_trunc('month', $1) + (interval '1 month - 1 day');
+    end
+$$ language plpgsql;
--- a/data/sql_updates/large_aggregates.sql
+++ b/data/sql_updates/large_aggregates.sql
@@ -205,7 +205,7 @@ select
     month,
     year,
     cycle,
-    make_timestamp(cast(year as int), cast(month as int), 1, 0, 0, 0.0) as date,
+    last_day_of_month(make_timestamp(cast(year as int), cast(month as int), 1, 0, 0, 0.0)) as date,
     sum(candidate_receipts) OVER (PARTITION BY cycle order by year asc, month asc) as cumulative_candidate_receipts,
     candidate_receipts,
     sum(canidate_disbursements) OVER (PARTITION BY cycle order by year asc, month asc) as cumulative_candidate_disbursements,
--- a/data/sql_updates/large_aggregates.sql
+++ b/data/sql_updates/large_aggregates.sql
@@ -205,7 +205,7 @@ select
     month,
     year,
     cycle,
-    make_date(cast(year as int), cast(month as int), 01) as chart_date,
+    make_timestamp(cast(year as int), cast(month as int), 1, 0, 0, 0.0) as date,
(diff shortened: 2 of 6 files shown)


ccostino:
@LindsayYoung, @noahmanger, @xtine, this is ready for an initial review, but I want to make sure the approach is right (and the expected date values are correct!).  Could any of you take a quick look and let me know, and if things are good I'll finalize the PR, otherwise make adjustments as necessary.  Thanks!

LindsayYoung:
I think we can replace the current date with the date you generated- It is in the same format, right? 
I think that will be less confusing for API users not to have two date fields that are the same.

Also I think you want descending order for date that is our default elsewhere, but I will defer to @xtine for what the frontend needs

ccostino:
Ah okay, so remove that custom property on the class and have it use the date field being generated in the database now?  It should be in the same format but I will double check that, and at the very least modify the column name to be just `date` then in the API response and model.

Good catch on the descending order (though it gets ordered on the database side too in ascending order), I can take care of that too if @xtine would like.  Thank you!

xtine:
The existing line charts connect to Totals by entity (https://api.open.fec.gov/developers/#!/financial/get_totals_by_entity), so if they follow that date sort it should be good.

ccostino:
Ah okay - this work is connected to that endpoint and I think it was relying on the database sorting that was already happening, so maybe it does need to be left alone then?  We can catch up about that some more and review in a quick pairing session. :-)

xtine:
I think my last comment is a bit confusing -- you can ignore that. It's just that the date returned from that API should be sortable, so something like `sort=date` that could also be `sort=-date` ascending?

ccostino:
Ah okay, gotcha!  Yes, it would be `sort=date` for ascending and `sort=-date` for descending, just like in the other spots.

ccostino:
Okay, we should be good to go now!  If any further adjustments are needed we can certainly make them in a future PR.  Thanks @xtine for the pairing on the last bit!

dogweather:
I added a test which is failing, but for the wrong reason: I believe it's not executing the SQL code to create the `date`: `AssertionError: assert None == '1999-03-01'`

@ccostino Would you happen to know how to write a test which actually uses the db code?

ccostino:
@dogweather, I think what this will require is adding the `date` property explicitly to the factory as that record has to be generated with all of the expected data in it in order for the test to work properly.  To do this, we'd need to import `sqlalchemy`, which provides a hook to call a SQL function (either built-in or one you create yourself).

I think I have this worked out and will push a change or two shortly, but I'm running into a slightly different issue with it myself that I may have to pass back to you.

ccostino:
Also, I think there might be a way to get the last day of the month like the Python code was doing before; I'd feel a bit more comfortable if we can do this since it will mirror the code this replaces.  I'll try and work that in today too.

ccostino:
@xtine:  This is ready now, if all looks well to you then feel free to merge!

PR #2182, opened 2017-01-30
Feature/add missing columns amendments

jontours:
This adds some missing fields, `most_recent`.  And improves some others, specifically `is_amended`, `amended_by`.  In the past those fields relied on the superceded column in the database but there was a better than even chance that field was null.   Even though valid amendment chain information could be be gleaned from this.  Hence, this adds some changes that captures those fields.

### DIFF
--- a/webservices/common/models/filings.py
+++ b/webservices/common/models/filings.py
@@ -125,16 +125,15 @@ class EFilings(CsvMixin, FecMixin, db.Model):
 
     @property
     def amended_by(self):
-        if len(self.amendment.longest_chain) > 0:
+        if len(self.amendment.longest_chain) > 0 and self.is_amended:
             index = self.amendment.longest_chain.index(self.file_number)
             return self.amendment.longest_chain[index + 1]
         else:
             return self.superceded
+
     @property
     def is_amended(self):
-        if not self.superceded or not self.most_recent:
-            return True
-        return False
+        return self.superceded or not self.most_recent
 
     @property
     def pdf_url(self):
--- a/webservices/common/models/filings.py
+++ b/webservices/common/models/filings.py
@@ -4,7 +4,7 @@ from sqlalchemy.ext.hybrid import hybrid_property
 from webservices import docs, utils
 from webservices.common.models.dates import ReportType
 from webservices.common.models.dates import clean_report_type
-from webservices.common.models.reports import CsvMixin, FecMixin
+from webservices.common.models.reports import CsvMixin, FecMixin, AmendmentChainMixin
 
 
 
@@ -86,7 +86,7 @@ class EfilingsAmendments(db.Model):
     last = db.Column(db.Numeric)
     previous_file_number = db.Column('previd', db.Numeric)
 
-class EFilings(CsvMixin, FecMixin, db.Model):
+class EFilings(AmendmentChainMixin, CsvMixin, FecMixin, db.Model):
     __tablename__ = 'real_efile_reps'
 
     file_number = db.Column('repid', db.BigInteger, index=True, primary_key=True, doc=docs.FILE_NUMBER)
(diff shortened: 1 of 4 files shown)


webservices/common/models/filings.py
```diff
@@ -123,8 +124,15 @@ def document_description(self):
         )
 
     @property
+    def amended_by(self):
+        if not self.superceded and not self.most_recent and self.amendment and len(self.amendment.longest_chain) > 0:
```
dogweather: Can this `if` expression be extracted into a function? It's pretty long.

webservices/common/models/filings.py
```diff
@@ -123,8 +124,15 @@ def document_description(self):
...
+            index = self.amendment.longest_chain.index(self.file_number)
+            return self.amendment.longest_chain[index + 1]
```
dogweather: Also, can we extract these two lines? Their purpose isn't too clear to me.

webservices/common/models/filings.py
```diff
@@ -123,8 +124,15 @@ def document_description(self):
...
+        else:
+            return self.superceded
+    @property
     def is_amended(self):
-        if self.superceded is not None:
+        if self.superceded is not None or not self.most_recent:
```
dogweather: I'm not too familiar with the db schema; do we need to say `is not None`? Or is it not superfluous?

webservices/common/models/filings.py
jontours: By extract do you mean refactor?  Their purpose is to find the file_number of the filing that amends the current filing being referenced.  The `longest_chain` is the longest amend chain, if it exists, that contains this filing.

webservices/common/models/filings.py
jontours: Good catch!

webservices/common/models/filings.py
jontours: Agreed, after looking at this after some sleep it looks like this could be simplified to one conditional.  Commit forthcoming.

webservices/common/models/filings.py
```diff
@@ -125,14 +125,14 @@ def document_description(self):
...
-        if not self.superceded and not self.most_recent and self.amendment and len(self.amendment.longest_chain) > 0:
+        if len(self.amendment.longest_chain) > 0:
             index = self.amendment.longest_chain.index(self.file_number)
             return self.amendment.longest_chain[index + 1]
         else:
             return self.superceded
     @property
     def is_amended(self):
-        if self.superceded is not None or not self.most_recent:
+        if not self.superceded or not self.most_recent:
```
dogweather: Is this correct? Maybe I'm just tired, but this change seems to have introduced a bug? And could the function simply be:

```python
def is_amended(self):
  return self.superceded or not self.most_recent
```

jontours: Ah, yeah I'm thinking that work.

jontours: Change pushed.

webservices/common/models/reports.py
```diff
@@ -60,6 +60,32 @@ def treasurer_name(self):
...
+        return self.amendment.amendment_chain
+
+    @property
+    def is_amended(self):
+        return not self.most_recent
+
+    @property
+    def amended_by(self):
+        if len(self.amendment.longest_chain) > 0 and self.is_amended:
+            index = self.amendment.longest_chain.index(self.file_number)
```
dogweather: I'm wondering if these queries on `amendment` can be moved to the amendment model? It looks like the logic and structure of amendments is spread out into other classes.

dogweather: E.g., moving the logic into `EfilingsAmendment`:

```python
return self.amendment.next_in_chain(self.file_number)
```

Basically, clearing up some of these Law of Demeter things.

jontours: I'd agree that the logic is spread out to two other classes, and I accounted that by putting these behaviors in their own class and using a mixin to modify the behavior of the classes that need it.  I guess it could be done the other way, but I'm trying to match the common pattern of using mixins we've used throughout the api.

dogweather: I like the mixins; e.g., to say that the object can be amended. But I think it'd be best to do both: we shouldn't have this many 'dots' ... reaching this deep into an object:

```
self.amendment.longest_chain.index(...)
```

To me that's a clue that we're over-coupling them. I see that we've got that logic in a couple places, too.

jontours: Ah, it looks like you were commenting on that at the same time as me.  I can move some of the logic into the amendment class, I still think a mixin is valid as well, as it makes it easy to inject the behavior into the receiving class, and allows the class to either use it or override.

dogweather: Yes - I totally agree, the mixin is great, plus it makes sense to move this logic into `EFilingsAmendment`, I think.

jontours: Agree it's best to delegate behavior when possible.  I'm gonna argue I just squeaked in under the law of demeter.  ;)

dogweather:
Sweet! Looks great.

jontours:
Thanks for the thorough review @dogweather!

noahmanger:
Seeing that this is approved, I'm going to merge.

PR #2187, opened 2017-02-02
Implement amendment filter for reports

jontours:
This is not working but it is close.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -286,7 +286,20 @@ reports = {
     'max_total_contributions': Currency(description=docs.MAX_FILTER),
     'type': fields.List(fields.Str, description=docs.COMMITTEE_TYPE),
     'candidate_id': fields.Str(description=docs.CANDIDATE_ID),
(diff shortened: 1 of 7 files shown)


noahmanger:
Is this ready for review and merging if good?

jontours:
Yes, ready for review and merging.

webservices/args.py
```diff
@@ -286,7 +286,20 @@ def make_seek_args(field=fields.Int, description=None):
...
+        description='''
+        -N   new\n\
+        -A   amendment\n\
+        -T   terminated\n\
+        -C   consolidated\n\
+        -M   multi-candidate\n\
+        -S   secondary\n\
+
+        Null might be new or amendment.   If amendment indicator is null and the filings is the first or first in a chain treat it as if it was a new.  If it is not the first or first in a chain then treat the filing as an amendment.
+        '''
```
ccostino: Nit-pick thing:  should we move this to the `docs` module?

jontours: Yeah I think that's probably a good idea.  Once a string is used twice in the same file a good bet it will be used again somewhere else.

PR #2195, opened 2017-02-07
updates to legal doc load

anthonygarvan:
When reinitializing legal docs on prod I encountered some issues. Doesn't need to be a hotfix since this is an async operation.


vrajmohan:
Great!

PR #2203, opened 2017-02-09
Index AOs using 1 Elasticsearch doc per AO

vrajmohan:
In addition:
1. Load issue_date for AO
2. Collect citations per AO and not by AO and document category
3. Fix regex bug where parsing for citations assumed only at most 2
   digit serial number per year (ie, 2004-123 was not allowed)


anthonygarvan:
Looks great :) Two things:
- [ ] Please remove the dead `index_advisory_opinions` code from `load_legal_docs`
- [ ] Exclude the `text` of the documents when returning data from api. Some AOs are very big and would lead to a performance issue.
- [ ] Use `load_advisory_opinions` instead of `index_advisory_opinions` in the legal_docs `__init__.py`

PR #2204, opened 2017-02-10
Uncommenting csv urls

noahmanger:
This just uncomments the code that we had in place for csv downloads of filings.

@jontours is there anything else we need to do here? I tested it out and it all seems to work well.


LindsayYoung:
Working!

PR #2205, opened 2017-02-10
Feature/add filters to debts

jontours:
Filters for debts done.  Still need to take care of schedule_f.


data/sql_setup/prepare_schedule_d.sql
```diff
@@ -0,0 +1,19 @@
+-- Create trigger to maintain Schedule C search columns
+ALTER TABLE disclosure.nml_sched_d DROP COLUMN if EXISTS creditor_debtor_name_text;
+
+ALTER TABLE disclosure.nml_sched_d add column creditor_debtor_name_text tsvector;
```
ccostino: Is this safe to do, is there no data in it before dropping and renaming the column?  And are we doing this because the column is an incorrect data type?  If so, would a single statement like this work?

```sql
ALTER TABLE disclosure.nml_sched_d ALTER COLUMN creditor_debtor_name_text TYPE tsvector;
```

webservices/resources/sched_d.py
```diff
@@ -31,10 +31,18 @@ def index_column(self):
     filter_range_fields = [
         (('min_date', 'max_date'), models.ScheduleD.load_date),
         (('min_payment_period', 'max_payment_period'), models.ScheduleD.payment_period),
-        (('min_amount', 'max_amount'), models.ScheduleD.amount_incurred_period),
+        (('min_amount_incurred_period', 'max_amount_incurred_period'), models.ScheduleD.amount_incurred_period),
```
ccostino: Nice, so this is becoming more explicit. :-)

data/sql_setup/prepare_schedule_d.sql
jontours: I think it would be safe, because the column is built from another column on that table that will still exist.  That single statement may be better I'm not sure.

ccostino: Ah okay, I see what you mean with the `UPDATE` statement below - I'm not 100% sure myself, was just basing that off of the [PostgreSQL documentation on altering a table](https://www.postgresql.org/docs/9.5/static/ddl-alter.html).

noahmanger:
This this good to merge in to the release branch?

ccostino:
We're good to go; thanks, @jontours!

PR #2208, opened 2017-02-10
Add fec_file_id

jontours:
This is a string with the format of `FEC-{file_number}`.  Added to all efiling endpoints, reports, and filings.

### DIFF
--- a/webservices/common/models/reports.py
+++ b/webservices/common/models/reports.py
@@ -91,7 +91,7 @@ class AmendmentChainMixin(object):
 class FecFileNumberMixin(object):
     @property
     def fec_file_id(self):
-        if self.file_number > 0:
+        if self.file_number and self.file_number > 0:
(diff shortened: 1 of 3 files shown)


webservices/common/models/reports.py
```diff
@@ -87,7 +88,16 @@ def amended_by(self):
             return None
 
 
-class CommitteeReports(PdfMixin, CsvMixin, BaseModel):
+class FecFileNumberMixin(object):
+    @property
+    def fec_file_id(self):
+        if self.file_number > 0:
```
ccostino: I spoke a bit too soon, then saw the failing tests; it looks like we need an additional check to make sure that `self.file_number` exists before comparing its value, so this line should be:

```python
if self.file_number and self.file_number > 0:
```

Then things should be good to go!

PR #2211, opened 2017-02-13
Added test for multiple sort capability for filings

jontours:
Ready for review.


dogweather:
I see issues with the test — let me a commit or two to see what you think, instead of writing it out in English -

tests/test_filings.py
```diff
@@ -108,6 +108,21 @@ def test_sort_bad_column(self):
...
+    def test_sort_multi_column(self):
+        [
+            factories.FilingsFactory(beginning_image_number=2, coverage_end_date='2017-01-01'),
+            factories.FilingsFactory(beginning_image_number=1, coverage_end_date='2017-01-01'),
+            factories.FilingsFactory(beginning_image_number=0, coverage_end_date='2017-01-02'),
+        ]
+
+        results = self._results(api.url_for(FilingsList, sort=['coverage_end_date', 'beginning_image_number']))
+        self.assertEqual(results[0]['beginning_image_number'], '1')
+        self.assertEqual(results[2]['beginning_image_number'], '0')
```
dogweather: Ok, actually, I've got a couple of questions before I can contribute. (And this points to the clarity I think we can add with some edits.) 

What are these asserts testing for? Is it that when sorting by multiple fields, the secondary field is correctly ordered?

tests/test_filings.py
```diff
@@ -108,6 +108,21 @@ def test_sort_bad_column(self):
...
+        results = self._results(api.url_for(FilingsList, sort=['coverage_end_date', '-beginning_image_number']))
+        self.assertEqual(results[0]['beginning_image_number'], '2')
```
dogweather: This one is testing for reverse sort order?

tests/test_filings.py
```diff
@@ -108,6 +108,21 @@ def test_sort_bad_column(self):
...
+        results = self._results(api.url_for(FilingsList, sort=['-coverage_end_date', '-beginning_image_number']))
+        self.assertEqual(results[0]['beginning_image_number'], '0')
```
dogweather: And this final one ... I can't actually tell what this is testing for 🦊

tests/test_filings.py
jontours: Yes, that is indeed what this assert is for.

tests/test_filings.py
jontours: This test is that good old, regular sort still trumps the secondary sort.

dogweather: Ok, thanks. I'm on vacation now, so won't be submitting a commit. But my suggestion is to break this up into three tests with descriptive names containing what you said. E.g., 

`def test_old_regular_sort_still_trumps_secondary_sort`

I.e., test names which document the expected outcome.

jontours: Sure thing!

ccostino:
Awesome work, thanks @jontours and @dogweather!

PR #2215, opened 2017-02-17
Feature/add missing columns filings

jontours:
First cut.  Some columns still left to implement, but will be easy with the groundwork laid.


jontours:
Changed column `total_non_federal_transfers` to `total_transfers`, as that more accurately reflects what is desired on the front-end.

data/sql_updates/totals_house_senate.sql
```diff
@@ -7,54 +7,75 @@ with last as (
...
+        sum(hs.ttl_receipts_per) as receipts,
+        sum(hs.ref_indv_contb_per) as refunded_individual_contributions,
+        sum(hs.ref_other_pol_cmte_contb_per) as refunded_other_political_committee_contributions,
+        sum(hs.ref_pol_pty_cmte_contb_per) as refunded_political_party_committee_contributions,
+        sum(hs.tranf_from_other_auth_cmte_per) as transfers_from_other_authorized_committee,
+        sum(hs.tranf_to_other_auth_cmte_per) as transfers_to_other_authorized_committee,
+        max(hs.rpt_tp_desc) as last_report_type_full,
+        max(last.begin_image_num) as last_beginning_image_number,
+        max(greatest(last.coh_cop)) as last_cash_on_hand_end_period,
+        max(greatest(last.coh_bop)) as last_cash_on_hand_beginning_period,
```
LindsayYoung: Am I reading this wrong, or should `last.coh_bop` be `cash_beginning_period.coh_bop`

LindsayYoung: Yea, I am not sure I understanding what this value is trying to convey

noahmanger: We're trying to add the "Beginning cash on hand" from the two-year financial summary. Is that this value or should it be different?

LindsayYoung: This looks like the beginning cash on hand for the most recent report, rather than the period of the aggregate. Though, I may be misreading it.

jontours: Hey all, I logged in to do my tock and noticed these comments.  This line: https://github.com/18F/openFEC/blob/ffc761832daec447aff9d69b08888dedc218b8dd/data/sql_updates/totals_house_senate.sql#L76.  has the proper column.  I think this code change here was made before I fully understood what was needed.

jontours: @LindsayYoung Column is removed and this should be ready to review again.

jontours:
Hold off on this until I can get the tests to pass.

jontours:
This is ready now!  Greatly simplified by removing window functions.

data/sql_updates/totals_pac.sql
```diff
@@ -4,84 +4,106 @@ create materialized view ofec_totals_pacs_mv_tmp as
...
-    comm_dets.committee_type,
-    pnp.election_cycle
+), cash_beginning_period as (
+	  select distinct on (pac.cmte_id, pac.election_cycle)
+	      pac.cmte_id                           as committee_id,
+	      pac.election_cycle                    as cycle,
+	      pac.coh_bop as cash_on_hand
+	  from
+        fec_vsum_f3x_vw pac
+        inner join ofec_committee_detail_mv_tmp comm_dets on pac.cmte_id = comm_dets.committee_id
```
LindsayYoung: I am not sure why there is a join here. It seems like we just need the earliest cash on hand and we can do that by ordering by coverage end date or coverage start date. Maybe something along the lines of: ```
cash_beginning_period as (
   select distinct on (cmte_id, election_cycle)
       cmte_id as committee_id,
       election_cycle as cycle,
       coh_bop as cash_on_hand,
       cvg_end_dt
  from fec_vsum_f3x_vw 
  order by  cvg_end_dt  acc ```

LindsayYoung: Ok, now I see the purpose of the join is to limit the kind of committee types to just PACs. But if you are doing a left join to this information later, so I don't think you have to filter it here. Maybe there is there a performance consideration I am over looking.

data/sql_updates/totals_pac.sql
```diff
@@ -4,84 +4,106 @@ create materialized view ofec_totals_pacs_mv_tmp as
...
+        max(last.coh_cop) as last_cash_on_hand_end_period,
+        max(last.debts_owed_by_cmte) as last_debts_owed_by_committee,
+        max(last.debts_owed_to_cmte) as last_debts_owed_to_committee,
+        max(last.rpt_yr) as last_report_year
+    from
+        fec_vsum_f3x_vw pnp
+        inner join ofec_committee_detail_mv_tmp comm_dets on cmte_id = comm_dets.committee_id
+        left join last using (cmte_id, election_cycle)
+    where
+        pnp.most_recent_filing_flag like 'Y'
```
LindsayYoung: This was already in the code- and I was probably the one to put it there- but I we can use `=` instead of `like` because it is faster. (We could also address that detail later)

LindsayYoung:
Thanks for your work on this! I had some comments, but this is good to go. I will run on e/w stage now.

Issue #2196, opened 2017-02-08
Fix nightly update and database refresh tasks


ccostino:
We've recently run into a few issues with our nightly update tasks as well as the full-fledged refresh tasks and the underlying SQL scripts that power them, largely due to table name changes but also because of a few other reasons.

First, a few tables need to have their names adjusted and have `vw` attached to the end:

* `fec_vsum_sched_a`
* `fec_vsum_sched_b`
* Possibly a few others related to `fec_vsum_*` tables

Second, the nightly updates are running into two issues.  In `dev`, it's due to a primary key violation:

```
2017-02-08 09:00:00 ERROR:manager:(psycopg2.IntegrityError) duplicate key value violates unique constraint "ofec_sched_e_tmp_pkey"
DETAIL:  Key (sub_id)=(4062120141215616440) already exists.
```

In `stage` and `prod`, it's due to missing tables:

```
2017-02-08 09:00:00 ERROR:manager:(psycopg2.ProgrammingError) relation "fec_vsum_f57_queue_old" does not exist
```

Upon looking, the `fec_vsum_f57_queue_new` table doesn't exist either, but both tables are present in `dev`.

I haven't been able to tell if there are any other underlying issues at the moment, but that's the start of them.  We need to rectify these as soon as we can so we can get the nightly refreshes working again and run `update_all` on `dev`, so I'm placing it in this sprint as a bug and something to tackle before the next release.

LindsayYoung:
if you don't see fec_vsum_f57_queue_new probably need to run prepare schedule e

jontours:
Yeah running `prepare_schedule_e` should fix the issue with the missing table.  The duplicate key issue may be harder to find.

ccostino:
Okay great, thank you both!  I'll give that a shot - is it safe to run `prepare_schedule_e` in production during an off peak time to get that addressed?

Also flagging #2197 here as I suspect this data isn't showing up due to these tasks not running.

ccostino:
I just ran `prepare_schedule_e` in `stage` and `prod` after testing things out with locally to be sure.  Indexes appear to be in tact across the board for the itemized data and the tables that weren't in `stage` or `prod` when this started happening are back.  I will check in the morning to see if the nightly refresh worked or if another issue was lurking underneath.

ccostino:
Okay, we've had some progress, but unfortunately this exposed another issue:  `prod` and `stage` are now throwing this error:

```
[...]

PR #2218, opened 2017-02-27
Fixes update_all and underlying SQL

ccostino:
Part of #2196 

This changeset addresses a couple of remaining table/view renames that we had to account for with the last round of database updates.  The references to the schedule A and schedule B views have been updated as has the test subset data and partitioning code.

The changes in here also modify the behavior of the queue tables that are associated with the Schedule A and B updates.  The triggers that are associated with these tables now make use of `BEFORE` and `AFTER` triggers to account for record changes in the underlying source tables because we are now relying on traditional views to derive our `ofec_sched_x` tables from.  These views contain the data represented in a structure that we can easily process, but they are comprised of multiple source tables.

In order to accomplish this, we watch for changes on one of the source tables that contains the actual record we're concerned about, then reference how the record appears in the view so that we can account for data pulled in via joins with the other source tables.  Due to this setup, we need to make sure we capture the state of the view at the right moment depending on the operation.  In other words:

* **FOR INSERTS:** We need to capture the view as it appears `AFTER` a new record is inserted.
* **FOR DELETES:** We need to capture the view as it appears `BEFORE` a record is deleted.
* **FOR UPDATES:** We need to capture the view as it appears `BEFORE` and `AFTER` a record is updated.

In the case of updates, this means that we watch for those changes in both triggers we have setup.  For the `BEFORE` trigger, we add the existing record information prior to the update to the old queue table.  For the `AFTER` trigger, the record has now been updated so we capture the record information from the view after the update to add it into the new queue.  This ensures our processing continues to work as it did before.

A **HUGE** thanks to @jontours and @LindsayYoung for talking through the approach and validating it with me!

Additional notes:

* [Triggers on Views](https://www.depesz.com/2010/10/16/waiting-for-9-1-triggers-on-views/)
* [PostgreSQL Trigger documentation](https://www.postgresql.org/docs/current/static/sql-createtrigger.html)


LindsayYoung:
Took a quick look at the error messages and it reminded me that when I did the similar PR, I make the view by selecting all from the current table and giving the view the new view name.

ccostino:
Yeah, I took a look at the other view definitions and had done the same thing here.  The problem that seems to be happening now is that the `automap` extension for SQLAlchemy won't pick up on views, just tables.  Part of it has to do with no primary key according to what I've found, so I'll dig a bit deeper on this thread before seeing if there's an alternative way of accomplishing what's being done in the tests.

ccostino:
Actually, there are two problems here:

# Failing tests due to the use of views instead of tables
# Failing `update_all` due to an index being created on a view instead of a table

The second problem is caused by these [two lines](https://github.com/18F/openFEC/pull/2218/files#diff-5cf905b76958badfe6708f3445af7459R2):

```sql
drop index if exists fec_vsum_sched_b_vw_link_id_idx;
create index fec_vsum_sched_b_vw_link_id_idx on fec_vsum_sched_b_vw (link_id);
```

Long story short, we can't do that (indexes can only be created on tables and materialized views in PostgreSQL).  After doing a bit of reading it seems like it might be okay to add them to the source tables that underpin the view to achieve the same performance results, but I'm not 100% sure just yet.  Taking those two lines out yields a working `update_all` though.

As for the first issue, I'm trying to figure out the best way to represent a view with SQLAlchemy and use that instead of relying on the automapper, which doesn't map views (and also requires that a relation either have a primary key or that a primary key is at least modeled in the code).

jontours:
I'm reading the documentation a bit on SqlAlchemy to see if there is a way to point to the right primary key.

ccostino:
Yeah I agree with you @jontours, that is likely an underlying issue with this setup now given that the view is going to be the thing with any updates, and I'm not sure how the trigger would know if something new appeared.

We may very well need to reference the underlying source tables and not just for the tests, but for the update triggers too.

ccostino:
Okay, after some more deliberation around this potential issue and looking into previous changes we made to support the switch to a view for the Schedule E data, @LindsayYoung and I confirmed that those changes do work against the view as we intend them to and seem to be picking up the changes appropriately.

Given that, we should be okay going forward with the same modifications for Schedule A and B.  Those fixes will be incorporated into this changeset.

ccostino:
@LindsayYoung and @jontours, I still need to fix the tests, but at this point this PR is getting very close and I want to make sure the updates to the triggers for schedule A and B make sense.  If either you have a few moments to take a look and see if the changes make sense, I'd greatly appreciate it!

tests/test_integration.py
```diff
@@ -18,22 +18,31 @@
...
+    metadata = sa.MetaData(schema='disclosure')
+    automap = automap_base(bind=db.engine, metadata=metadata)
     automap.prepare(db.engine, reflect=True)
 
     class SchedAFactory(SQLAlchemyModelFactory):
         class Meta:
             sqlalchemy_session = db.session
-            model = automap.classes.fec_vsum_sched_a
-        filing_form = '11'
+            model = automap.classes.nml_sched_a
```
jontours: Ah, I was curious at to how to handle this.  I think I see what's happening here.  Although I'm a little wary about taking out any references to `disclosure.v_sum_and_det_sum_report`, as the tests are basically assuming that table is incidental to the updates.  Also, and we can talk about this, I can't seem to find at what point the rows are inserted into the partitions for `ofec_sched_a_master` and `ofec_sched_b_master`.

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -14,49 +14,86 @@ create index on ofec_sched_a_queue_old (timestamp);
...
 
         if two_year_transaction_period_old >= start_year then
-            delete from ofec_sched_a_queue_old where sub_id = old.sub_id;
-            insert into ofec_sched_a_queue_old values (old.*, timestamp, two_year_transaction_period_old);
+            delete from ofec_sched_a_queue_old where sub_id = view_row.sub_id;
+            insert into ofec_sched_a_queue_old values (view_row.*, timestamp, two_year_transaction_period_old);
         end if;
 
         return old;
+    elsif tg_op = 'UPDATE' then
```
ccostino: The `update` operation is defined in both trigger functions because one is associated with a `BEFORE` trigger and the other is associated with an `AFTER` trigger.  The reason behind this is because we're leveraging the state of the data as it appears in the view itself to make sure we're pulling all information.

In the case of an `update` operation, we want the state of the view before the change to the record takes place and after the fact in order to accurately capture the record information so that it is processed correctly (namely the `two_year_transaction_period` field, among other things).  The check for the `update` operation in the `BEFORE` trigger captures the state of the record as it exists prior to any change and inserts it into the old queue for deletion.  The check in the `AFTER` trigger captures the state of the record after the change has taken place and inserts it into the new queue for insertion.

I tried to simplify this as much as possible but the reliance on the view for the best representative state of information for all of this data made this a bit difficult.  That said, this allows us to not have to worry as much about what happens with the source data as long as these views continue to exist and provide us with the information we need in the structure we expect.

Please let me know if anything is still unclear!

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -18,49 +18,86 @@ create index on ofec_sched_b_queue_old (timestamp);
...
 begin
     if tg_op = 'INSERT' then
-        two_year_transaction_period_new = get_transaction_year(new.disb_dt, new.rpt_yr);
+        select into view_row * from fec_vsum_sched_b_vw where sub_id = new.sub_id;
+        two_year_transaction_period_new = get_transaction_year(new.disb_dt, view_row.rpt_yr);
 
         if two_year_transaction_period_new >= start_year then
-            delete from ofec_sched_b_queue_new where sub_id = new.sub_id;
-            insert into ofec_sched_b_queue_new values (new.*, timestamp, two_year_transaction_period_new);
+            delete from ofec_sched_b_queue_new where sub_id = view_row.sub_id;
```
ccostino: All good here? :-)  The setup hasn't changed from before, just how the data is supplied to the queue.

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -18,49 +18,86 @@ create index on ofec_sched_b_queue_old (timestamp);
...
+            insert into ofec_sched_b_queue_new values (view_row.*, timestamp, two_year_transaction_period_new);
```
ccostino: For this file, [line 31](https://github.com/18F/openFEC/pull/2218/files/473c77d571b9a3c4a49461fedc7dc8a19875da64#diff-5cf905b76958badfe6708f3445af7459R31) and [line 34](https://github.com/18F/openFEC/pull/2218/files/473c77d571b9a3c4a49461fedc7dc8a19875da64#diff-5cf905b76958badfe6708f3445af7459R34) are what take care of `view_row`.  Line 31 defines the variable and sets it to be a record of type `fec_vsum_sched_b_vw%ROWTYPE` - this means it carries the same column structure as what is in the view (could be a table or materialized view too).  Without this specific definition (for instance, if I had just gone with `view_row RECORD` instead), PostgreSQL would not know what to do with the data when utilizing it elsewhere, e.g, an insert into somewhere else, because it has no "type" and therefore can't be registered, [which results in an error like this](http://stackoverflow.com/questions/37815054/getting-an-error-record-type-has-not-been-registered-in-postgresql-what-is-wr).  By setting the type explicitly (which is awesome that you can do this), we can use the record more directly and map things very easily. :-)

Line 34 is what actually populates the record into the `view_row` variable.  Within a stored procedure/function like this (at least with PL/pgSQL), you can [select data from the database into a variable](https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-ASSIGNMENT) to be used elsewhere.  Since we want to take that `view_row` with the record to be processed pretty much as is but check a couple of things first, this makes it possible to do all of that.

As for what will actually be placed into the queues, yes, only records that are reflected in the views, and the triggers on the source table are what guarantee that.  None of this would happen without a change to a record in the source table, but as soon as that change is about to happen or just happened, accessing the view would reflect the current state of the data from the source tables.

jontours: Makes sense, thanks for the explanation.

tests/test_integration.py
jontours: I found the answer to my last question there, so don't worry about explaining it too much.

ccostino: Yeah, this was a bit tricky but simply required setting the `automapper` to a different schema.  What's a bit of a bummer about that you only have access to the one schema.  I suppose you could alleviate that with multiple `automapper` objects though, now that I think about it.

I'm with you on being wary about not having a completely accurate reflection of the databases as they actually exist, but I noticed that for these views, `disclosure.v_sum_and_det_sum_report` was only being referenced for 3 columns; the rest of the data came from `disclosure.nml_sched_x`.

However, those 3 columns represent `filing_form`, `rpt_tp`, and `rpt_yr`.  Would it be this table that actually receives more updates once a record exists in the corresponding `disclosure.nml_sched_x` table?  If so, then we either need to change the triggers again or have them run off of `disclosure.v_sum_and_det_sum_report` as well to ensure we always receive record updates when they occur.  This leads me to my own question:  is this how record amendments are managed?  If so, ugh, back to the drawing board a bit more...

As for when the rows are actually inserted into the `ofec_sched_x` tables (whichever child table they should actually go into), that is handled by the nightly update process:

1. First, [`manage.update_aggregates`](https://github.com/18F/openFEC/blob/develop/manage.py#L255) is what gets kicked off
1. [`partition.base.refresh_children`](https://github.com/18F/openFEC/blob/develop/webservices/partition/base.py#L210) is called for Schedule A and B as a part of that, which loops through all child tables
1. [`partition.base.refresh_child`](https://github.com/18F/openFEC/blob/develop/webservices/partition/base.py#L219) actually does the legwork for each child table, and clears out the processed records from the queues when done

At the end, both queues should be completely empty and our child tables updated with fresh data. :-)

ccostino: We're going to follow up with this question around `disclosure.v_sum_and_det_sum_report` with @PaulClark2 and @jwchumley to get clarification.  If we do need to support this, we'll do it in a new PR.

ccostino:
Thanks for the review and questions, @jontours!

ccostino:
@jontours and @LindsayYoung:  this is ready for merge now!  If something else needs to be done let me know and I'll take care of it ASAP so this can make it in.  Thanks!

jontours:
Those changes are good.  Thanks @ccostino!

PR #2229, opened 2017-03-04
This is a fix so records do not get duplicated if the committee

LindsayYoung:
has multiple entries in the committee history table.
I used the most up to date name, that is what we do elsewhere.
Removed idx because we were not using it since we have the unique
key sub_id.


jontours:
Thanks for catching that @LindsayYoung, I'll review and merge first thing Monday morning.

PR #2233, opened 2017-03-08
Fix/search aos by document category

vrajmohan:


webservices/resources/legal.py
```diff
@@ -105,7 +105,7 @@ def get(self, q='', from_hit=0, hits_returned=20, type='all', **kwargs):
                 .sort("sort1", "sort2")
 
             if type == 'advisory_opinions':
-                query = apply_ao_specific_query_params(query, **kwargs)
+                query = apply_ao_specific_query_params(query, q, **kwargs)
```
ccostino: Out of curiosity, what does `q` represent here?

vrajmohan: It is the standard keyword search parameter - http://example.com/foo?q=keyword. I agree that it is a mysterious abbreviation, but it appears to be as common as `i` for a loop counter.

ccostino: Ah okay, thanks!  I thought this was the case, but the `query` parameter was throwing me off.

PR #2235, opened 2017-03-08
Multi-sort fully enabled for reports

jontours:
This helps address the issue of amendments appearing lower in the list than the report they amended.

The params to pass into the url for proper sorting would be `sort=-receipt_date&sort=-beginning_image_number`.  It is best to do it this way I think as the time in the time stamp is all zeros for any equal dates.


jontours:
Resolves issue here: https://github.com/18F/FEC/issues/1156.

LindsayYoung:
Sorry this has been open so long!

LindsayYoung:
Yes, this works

PR #2252, opened 2017-03-17
Feature/move dev to cloud

LindsayYoung:


LindsayYoung:
- pull in /gov-cloud branch
- merge it

LindsayYoung:
check deploys in the tasks.py

tasks.py
```diff
@@ -157,17 +157,17 @@ def _detect_space(repo, branch=None, yes=False):
 
 
 SPACE_URLS = {
-    'dev': [('18f.gov', 'fec-dev-api')],
-    'stage': [('18f.gov', 'fec-stage-api')],
-    'prod': [('18f.gov', 'fec-prod-api')],
+    'dev': [('app.cloud.gov', 'fec-dev-api')],
+    'stage': [('app.cloud.gov', 'fec-stage-api')],
+    'prod': [('app.cloud.gov', 'fec-prod-api')],
```
ccostino: These will have to change in the future as we get the domain configuration set in GovCloud, right @LindsayYoung?

webservices/rest.py
```diff
@@ -58,7 +57,7 @@ def sqla_conn_string():
     if not sqla_conn_string:
         print("Environment variable SQLA_CONN is empty; running against " + "local `cfdm_test`")
         sqla_conn_string = 'postgresql://:@/cfdm_test'
-    print(sqla_conn_string)
+    # print(sqla_conn_string)
```
ccostino: We should probably just remove this line, right?

bin/run.sh
```diff
@@ -1,4 +1,3 @@
-# turn off slack for now!
-#invoke notify
+# invoke notify || echo 'failed to start slack'
```
ccostino: Do we want to get these working again, or hold off for now?

LindsayYoung: I vote for doing this later. We need new login from slack, so maybe just wait and maybe we will just want to move that to FEC slack.

tasks.py
LindsayYoung: I don't think so since we are still using the proxy app. We need a proxy anyway so I think keeping the proxy as is makes sense. We could change it later if there is some benefit down the road.

PR #2257, opened 2017-03-23
Add support and instructions for cf ssh

ccostino:
This changeset adds support and instructions for running `cf ssh` in the GovCloud environment.  While it is easy and straightforward to SSH into a running application, there is a little bit of environment setup to do once you are in.  This script makes it easy to take care of these things with one command.


ccostino:
Whoever is able to get to this for a review, please feel free!

PR #2260, opened 2017-03-24
Feature/gov cloud

LindsayYoung:
This is the code freeze branch. we will keep adding additional things needed for the ATO here.


test-requirements.txt
```diff
@@ -0,0 +1 @@
+bandit==1.4.0
```
ccostino: Do we want to move this into the `requirements-dev.txt` file or just keep it outside as is here?

tasks.py
```diff
@@ -152,8 +152,8 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    #('dev', lambda _, branch: branch == 'develop'),
-    ('dev', lambda _, branch: branch == 'feature/gov-cloud'),
+    ('dev', lambda _, branch: branch == 'develop'),
+    # ('dev', lambda _, branch: branch == 'feature/gov-cloud'),
```
ccostino: I'll readjust these while resolving the merge conflicts!

PR #2262, opened 2017-03-24
Candidate committee aggregates

jontours:
First cut; candidates for presidential is close to being done.  House/senate need to be added; and logic for `full_election` still needs to be worked out.


noahmanger:
Awesome. Have the updates been run on dev yet? I'd love to test it out to get the front-end working.

jontours:
I need to do some refactoring with some of the sql (mostly renaming for clarity), but this endpoint is functioning.

jontours:
@noahmanger Yes, the updates were ran on dev.  Should be ready to test from the web-app.  The url pattern is as discussed: so for McCain's 2016 senate full_election as an example it would be: 

```v1/candidate/S6AZ00019/totals/?per_page=20&sort=-cycle&page=1&full_election=true&cycle=2016```

**Comment by noahmanger** [comment]
Nice! 

image

**Comment by jontours** [comment]
I'm happy to see it coming together!!  I learned a lot about finance data doing this feature.

**Comment by noahmanger** [comment]
Just changed this to target the release branch.

File: tests/test_integration.py [hunk 1]
```diff
@@ -84,8 +84,16 @@ def setUpClass(cls):
         manage.update_all(processes=1)
 
     def test_update_schemas(self):
+        #adding this here: my rationale is that
+        #models.CaniddateCommitteeTotalsHouseSenate passed integration, so testing this specific
+        #model really isn't expanding code coverage.  I can try and get the model passing but it's proving
+        #to be difficult considering the joins needed (and our limited test subset)
+        whitelist = [models.CandidateCommitteeTotalsPresidential]
```
ccostino: Thank you for the note here, definitely appreciated!  We'll want to be mindful of this for the future, but I absolutely hear you about the test subset.

data/sql_updates/totals_candidate_committee_house_senate.sql
```diff
@@ -0,0 +1,199 @@
...
+        select * from final_combined_total
+;
+
+create unique index on ofec_totals_candidate_committees_house_senate_mv_tmp (candidate_id, cycle, full_election);
+
+create index on ofec_totals_candidate_committees_house_senate_mv_tmp (candidate_id);
+create index on ofec_totals_candidate_committees_house_senate_mv_tmp (election_year);
+create index on ofec_totals_candidate_committees_house_senate_mv_tmp (cycle);
+create index on ofec_totals_candidate_committees_house_senate_mv_tmp (receipts);
+create index on ofec_totals_candidate_committees_house_senate_mv_tmp (disbursements);
```
ccostino: 👏 Kudos on figuring out all of the sums and aggregates needed to piece this data together!

data/sql_updates/totals_candidate_committee_presidential.sql
```diff
@@ -0,0 +1,232 @@
...
+;
+--these columns should be a unique primary key when considered together
+create unique index on ofec_totals_candidate_committees_presidential_mv_tmp (candidate_id, cycle, full_election);
+
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (candidate_id);
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (election_year);
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (cycle);
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (receipts);
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (disbursements);
+create index on ofec_totals_candidate_committees_presidential_mv_tmp (federal_funds_flag);
```
ccostino: Ditto to above. :-)

webservices/common/models/totals.py
```diff
@@ -35,6 +35,42 @@ class CommitteeTotals(BaseModel):
     last_debts_owed_by_committee = db.Column(db.Numeric(30, 2))
     last_debts_owed_to_committee = db.Column(db.Numeric(30, 2))
 
+class CandidateCommitteeTotals(db.Model):
+    __abstract__ = True
+    #making this it's own model hieararchy until can figure out
+    #how to maybe use existing classes while removing primary
+    #key stuff on cycle
```
ccostino: I'm not sure I follow about removing the primary key stuff on cycle in the context of using existing classes, could you expand on this a bit?  Are you referring to other existing models we already have defined and leveraging those, or because of the `full_election` primary key defined on the models inheriting this base class of sorts?

While I'm a bit confused by this, I'm also not seeing anything wrong here in the structure as defined so this isn't a blocker!

webservices/resources/totals.py
```diff
@@ -78,6 +85,52 @@ def _resolve_committee_type(self, committee_id=None, committee_type=None, **kwar
...
+    },
+)
+class CandidateTotalsView(utils.Resource):
+
+    @use_kwargs(args.paging)
+    @use_kwargs(args.totals)
+    @use_kwargs(args.candidate_committee_totals)
+    @use_kwargs(args.make_sort_args(default='-cycle'))
+    @marshal_with(schemas.CommitteeTotalsPageSchema(), apply=False)
+    def get(self, candidate_id, **kwargs):
```
ccostino: I was looking through other code in our resources because I couldn't quite remember how specific URL parameters could come through as `None` (e.g., in `build_query()` below), but what I came across actually had the arguments defined in custom `get()` methods set to `None` by default as well.  Should that be the case here?

As an example, here's a resource a defined in [`webservices/resources/totals.py`](https://github.com/18F/openFEC/blob/develop/webservices/resources/totals.py#L42) that demonstrates what I'm referring to.

PR #2266, opened 2017-03-27
Update some of the backend tooling for the API

ccostino:
This changeset updates some of the tooling and infrastructure in the API to bring things up to date. It includes the following:

* Cleanup of .travis.yml
* Updates several Python dependencies
* Removal of all references to Sentry/Raven
* Updates cloud.gov Python runtime to 3.5.3
* Updates New Relic environment name in the manifest_feature.yml config
* A handful of documentation updates

The Travis cleanup includes updating nvm to make use of the latest stable release of Node.js, which I have been running locally for a while now with no issues.

I originally tried to get things running with Python 3.6 with these changes, but there are a couple of dependencies that are not playing nice with it (namely `slate` and the `distribute` dependency it has).  Things are working fine with Python 3.5 though, so this accounts for running things with the latest patch level release of it, 3.5.3.

A handful of the components updated are related to the downloads/exports, so to test them I ran both the updated web app (based on PR https://github.com/18F/openFEC-web-app/pull/1898) and a local copy of the Celery worker with this updated code as well, and pointed my local API to the dev GovCloud database.  There was a small change I needed to make to fix a failing test and account for a newer version of `sqlalchemy-postgres-copy`, but otherwise things should still be working just fine!

### DIFF
--- a/webservices/tasks/__init__.py
+++ b/webservices/tasks/__init__.py
@@ -4,8 +4,6 @@ import celery
 from celery import signals
 from celery.schedules import crontab
 
-from raven import Client
-from raven.contrib.celery import register_signal, register_logger_signal
 
 from webservices.env import env
 from webservices.tasks import utils
--- a/webservices/tasks/download.py
+++ b/webservices/tasks/download.py
@@ -194,8 +194,6 @@ def make_bundle(resource):
                 resource['query'],
                 resource['schema']
             )
-            import ipdb
-            ipdb.set_trace()
             copy_to(
                 query,
                 fp,
--- a/webservices/tasks/download.py
+++ b/webservices/tasks/download.py
@@ -194,10 +194,12 @@ def make_bundle(resource):
                 resource['query'],
                 resource['schema']
             )
+            import ipdb
+            ipdb.set_trace()
             copy_to(
                 query,
-                db.session.connection().engine,
                 fp,
+                db.session.connection().engine,
                 format='csv',
                 header=True
             )
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 sudo: false
 cache: pip
 python:
-  - "3.4"
+  - "3.5"
 addons:
   postgresql: "9.4"
 env:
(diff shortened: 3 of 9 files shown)


dogweather:
Excellent; how do you resolve the dependencies to upgrade the libraries?

ccostino:
In this case I'm only upgrading things that I either know what the upgrade path is going to be like (e.g., `invoke`, which had some breaking changes in recent updates) or that don't have any backward-incompatible changes with previous versions.  Larger pieces (e.g., SQLAlchemy), will take more time to test with newer versions.

As for things like `slate`, which haven't been updated in a while and therefore use older pinned dependencies, that's a slightly different story.  One or more of these options could be possible:

* File an issue with the project reporting the problem
* Submit a PR with updates/improvements for the author to merge
* Fork the library and fix it ourselves
* Find an alternative solution

Right now I'm more concerned with just getting what I can updated quickly and establishing more parity between the repos, Travis, and our cloud.gov environments as much as possible so that we're starting off on good ground with the GovCloud environment.

dogweather:
Thanks @ccostino yeah, I'm not familiar with a Python tool that does dependency resolution.

webservices/tasks/download.py
```diff
@@ -194,10 +194,12 @@ def make_bundle(resource):
                 resource['query'],
                 resource['schema']
             )
+            import ipdb
+            ipdb.set_trace()
```
ccostino: D'oh, thought I had removed this before the last commit.

ccostino:
It doesn't seem like there really is one at the moment.  As of now [`pip` itself does not have a formal one](https://github.com/pypa/pip/issues/988).  Taken from the beginning of the [`pip` User Guide](https://pip.pypa.io/en/stable/user_guide/):

>Requirements files are used to force pip to properly resolve dependencies. As it is now, pip doesn't have true dependency resolution, but instead simply uses the first specification it finds for a project. E.g if pkg1 requires pkg3>=1.0 and pkg2 requires pkg3>=1.0,<=2.0, and if pkg1 is resolved first, pip will only use pkg3>=1.0, and could easily end up installing a version of pkg3 that conflicts with the needs of pkg2. To solve this problem, you can place pkg3>=1.0,<=2.0 (i.e. the correct specification) into your requirements file directly along with the other top level requirements.

That said, if you're curious, there is a tool out there trying to shed more light on this, [`pipdeptree`](https://pypi.python.org/pypi/pipdeptree).

webservices/tasks/__init__.py
```diff
@@ -4,8 +4,6 @@
 from celery import signals
 from celery.schedules import crontab
 
-from raven import Client
```
LindsayYoung: Isn't this for New Relic?

LindsayYoung: Never mind me, this is right.

ccostino: Yeah it initially threw me off too, but that was the client library for Sentry!

ccostino:
Updated against latest changes in `develop` since the second GovCloud support branch was merged in!  This should still be good to go.

LindsayYoung:
Thanks for doing this!

PR #2272, opened 2017-03-29
Removing setuptools and wheel due to pip warning output

ccostino:
Turns out that including these two dependencies is something pip warns against - removing them to ensure things work as expected (or break as expected for that matter).


ccostino:
Whoever is able to review and merge first, please go for it! :-)

PR #2278, opened 2017-03-30
Remove unverified filers from candidate and committee endpoints

### LINKED ISSUE #2243: Remove unverified filers from candidate and committee endpoints
So that users are only exposed by default to verified filers, remove all unverified candidates and committees from default views.

---

The FEC has a table of "unverified filers". We should use this table to flag all candidates and committees that match. This should be done on the main `/candidates/` and `/committees/` endpoints as well as the `/names/` endpoint so matches aren't suggested in the typeahead. 

**Completion criteria:**
- [ ] All unverified candidates removed from candidates endpoints
- [ ] All unverified committees removed from committees endpoints
- [ ] All unverified options removed from `/names/` endpoint
- [ ] Remove all their filings

ccostino:
Fixes #2243

This changeset adds a check during our candidate and committee history creation to check for any unverified filers and exclude them from the data.  This will prevent the unverified filers from appearing in the candidate and committee endpoints.


ccostino:
Well, I got the tests passing, but that the expense of adding a bunch of data to the test subset.  This has ballooned the file size significantly, so I am going to have to figure out some way to cut it back down or an alternative solution to testing this.

ccostino:
I select ~30 random candidate and committee IDs to serve as test data and removed all the rest.  I also performed a `VACUUM FULL ANALYZE` in the subset database to cut down its size as much as possible, it is now down to ~18MB.  Much better!

data/sql_updates/candidate_history.sql
```diff
@@ -66,7 +66,8 @@ left join disclosure.cand_inactive inactive on
     fec_yr.cand_id = inactive.cand_id and
     fec_yr.fec_election_yr < inactive.election_yr
 left join staging.ref_pty ref_party on fec_yr.cand_pty_affiliation = ref_party.pty_cd
-where max_cycle >= :START_YEAR
+where max_cycle >= :START_YEAR and
+    fec_yr.cand_id not in (select distinct cmte_id from unverified_filers_vw where cmte_id similar to '(P|S|H)%')
```
jontours: should probably be ```select distinct cand_id ...  ```

ccostino: I did a double-take when I first saw this because I was like "d'oh, how could I miss that..." - turns out that view only has one ID column, `cmte_id`.  Counter-intuitive, I know!

There still appears to be an issue with these though, and I need to fix the test database again.  Might be worth pairing for a few minutes on this because by all accounts this SQL should do the trick, but with it in place no records end up appearing in these materialized views.

jontours: Ah, that makes sense actually.  Because I believe as written that where statement will always evaluate to false.  Because there is no chance to select a candidate_id that is equal to any committee id.  Yeah we can pair on it if you like.  But think you'll have to relate candiate_id's via `ofec_cand_cmte_link`.

ccostino: Ah okay, where should that link be made?  In the sub-clause with the `unverified_filers_vw` (so a join of sorts), or somewhere else?

jontours: False alarm.  I didn't realize that the `cmte_id` is also `cand_id` in some instances.  This is good as is!

ccostino: Ah okay, thanks!

ccostino:
I had to undo the vacuuming I did previously with the test subset as it broke a few other things, but the test data is now modeled correctly again.  I'd be very curious to get feedback on the tests themselves, too.

The basic premise for them is to do a full count of all records we currently have in our candidate and committee history materialized views, then do a count that excludes any IDs that appear in the unverified filer view and compare the two.  They should match.

Because the integration tests setup the full database for us, including running `update_schemas`, this ensures the materialized views are created with the updates contained in this PR which should exclude any unverified filer.

PR #2281, opened 2017-03-31
Fixing offsets

LindsayYoung:
variable fix


LindsayYoung:
This will require `manage.py update_schemas`

LindsayYoung:
Checking the data on this

ccostino:
Oh d'oh, looks like the test subset data needs to be fixed to account for this too. :-(

data/sql_updates/totals_pac.sql
```diff
@@ -60,7 +60,7 @@ with last as (
         sum(pnp.net_op_exp_per) as net_operating_expenditures,
         sum(pnp.non_alloc_fed_elect_actvy_per) as non_allocated_fed_election_activity,
         sum(pnp.ttl_nonfed_tranf_per) as total_transfers,
-        sum(pnp.ttl_op_exp_per) as offsets_to_operating_expenditures,
+        sum(pnp.offests_to_op_exp) as offsets_to_operating_expenditures,
```
ccostino: Ah I see one, `offsets` :-)

data/sql_updates/reports_pac_party.sql
```diff
@@ -45,8 +45,8 @@ select
     non_alloc_fed_elect_actvy_per as non_allocated_fed_election_activity_period,
     non_alloc_fed_elect_actvy_ytd as non_allocated_fed_election_activity_ytd,
     shared_nonfed_op_exp_per as nonfed_share_allocated_disbursements_period,
-    ttl_op_exp_per as offsets_to_operating_expenditures_period,
-    ttl_op_exp_ytd as offsets_to_operating_expenditures_ytd,
+    offests_to_op_exp as offsets_to_operating_expenditures_period,
```
ccostino: `offsets`

data/sql_updates/reports_pac_party.sql
```diff
@@ -45,8 +45,8 @@ select
...
+    offsets_to_op_exp_ytd_i as offsets_to_operating_expenditures_ytd,
```
ccostino: Is the `_i` at the end intended?

LindsayYoung: yea, and it means that we have a problem in the totals, I documented in the issue.

The problem is that on the forms they ask the same question more than once, but people don't always fill out things twice, so you need to take the greatest of both values.

That was the way on the original table, then we switched to a case where that wasn't true, Then we switched to a case where that was true again, but did not add the greatest back in.

ccostino:
Who would've thought the spelling things correctly would be an issue... 😆 ❤️ thanks @LindsayYoung!

Issue #2271, opened 2017-03-29
Introduce high-level Python version pinning


dogweather:
Soft Pinning with Python Package Dependency Resolution
----------------------------------------------------------
We'll create `requirements.in` files like [this one in OMB eRegs](https://github.com/18F/omb-eregs/blob/master/requirements.in#L2) which we'll write by hand, instead of manually editing `requirements.txt`. Then, we'll use [pip-tools](https://github.com/jazzband/pip-tools) to resolve Python package dependencies and fully manage `requirements.txt` for us. Pip-tools will also be how we install and upgrade the Python packages. This is also described in [Before you Ship](https://pages.18f.gov/before-you-ship/infrastructure/pinning-dependencies/).

The benefits of this, and why it's increasingly what projects do
---------------------------------------------------------------
This is SOP in Ruby, Elixir, and increasingly, Python, because;

* Packages can be upgraded with a much greater degree of safety and confidence.
* Teams upgrade more frequently because it's easy to, getting bug and security fixes faster, and reducing technical debt.
* Configuration is simpler because teams only need to list the high-level dependencies in `requirements.in`. This file is much shorter than `requirements.txt` and not cluttered with transitive dependencies
* Configuration is less error prone because there's a clear separation of concerns: people edit one file, and the software edits the other. Each file does its one job, better.

ccostino:
@dogweather, definitely good to keep in mind, but we do pin our dependencies already in the FEC repos. :-)

This one, for example: https://github.com/18F/openFEC/blob/develop/requirements.txt

(I have a PR open to remove setuptools and wheel now that we fixed the Slate issue

ccostino:
Ack, wrong button, didn't mean to close just yet!

dogweather:
I added a full description to show how this differs from our current practices.

ccostino:
Awesome, thank you!

cmc333333:
I think the big win here is around transitive dependencies. Here's a specific example: 

* `cfenv` (which is [included](https://github.com/18F/openFEC/blob/develop/requirements.txt#L4) in the openFEC requirements file) [depends on](https://github.com/jmcarp/py-cfenv/blob/master/setup.py#L8) `furl`, version >= 0.4.8
* `furl`, in turn, [depends on](https://github.com/gruns/furl/blob/master/setup.py#L98-L99) a minimum version of both `orderedmultidict` and `six`. These three libraries aren't currently part of openFEC's `requirements.txt`.

[...]

PR #2285, opened 2017-04-03
[WIP] High-level Python version pinning

dogweather:
Implements #2271


dogweather:
This [pip-tools bug](https://github.com/jazzband/pip-tools/issues/421) is blocking me; am working with @cmc333333 to resolve it.

dogweather:
I believe this specification is incorrect, and is causing the requirements to not be satisfiable: https://github.com/jmcarp/marshmallow-pagination/blob/master/setup.py#L10

I.e., we need marshmallow-sqlalchemy >= 0.8.0, and have been using it. But the lib above pins 0.4.1 in its setup.py. We'd been using it in this situation, apparently our process wasn't checking the requirements somehow.

`pip-compile` now refused to run, pointing out the problem:

```
Could not find a version that matches marshmallow-sqlalchemy==0.4.1,==0.8.0
```

PR #2288, opened 2017-04-03
Fix update_all processing

### LINKED ISSUE #2286: Fix update_all processing
Our `update_all` script is still not running at the moment and currently encountering two issues while processing:

* An execution error when attempting to run a subprocess, likely due to an unquoted/unescaped shell command.
* A database error: `sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column ofec_totals_house_senate_mv.last_debts_owed_to_committee does not exist`

These (and any other issues) need to be addressed ASAP so that we can properly refresh the new GovCloud database instances.

ccostino:
Fixes #2286 

This changeset addresses a couple of other issues that have cropped up with the processing in our `update_all` script.  There are two known issues thus far:

* An error while processing the new database connection strings, likely caused by unescaped/unquoted strings
* An error while processing one of the materialized views in which a column being referenced is not found

More items will be added as they are found until the script runs successfully again.  In addition to fixing these issues, this changeset also adds a few more logging statements to provide visibility into the state of things while running the command.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -222,10 +222,12 @@ def dump_districts(dest=None):
     """ Makes districts locally that you can then add as a table to the databases
     """
     source = db.engine.url
+
     if dest is None:
         dest = './data/districts.dump'
     else:
         dest = shlex.quote(dest)
+
     cmd = (
         'pg_dump "{source}" --format c --no-acl --no-owner -f {dest} '
         '-t ofec_fips_states -t ofec_zips_districts'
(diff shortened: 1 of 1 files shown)


manage.py
```diff
@@ -132,27 +132,57 @@ def load_nicknames():
...
+
+    load_table(
+        pd.read_excel('data/pacronyms.xlsx'),
+        'ofec_pacronyms',
+        if_exists='append'
+    )
+
+    logger.info('Finished loading pacronyms.')
 
 def load_table(frame, tablename, if_exists='replace', indexes=()):
```
LindsayYoung: Not something to be addressed in this PR but this reminds me we can delete this code.

manage.py
```diff
@@ -249,6 +292,20 @@ def update_itemized(schedule):
     logger.info('Finished Schedule {0} update.'.format(schedule))
 
 @manager.command
+def partition_itemized():
```
LindsayYoung: This is great, thanks!

PR #2293, opened 2017-04-05
[WIP] Fix queue processing to account for column type mis-match

### LINKED ISSUE #2286: Fix update_all processing

ccostino:
**WIP:  PLEASE DO NOT MERGE**

Fixes #2286 

This changeset addresses a few remaining issues with the SQL that references the queue tables and not accounting for the `unknown` type of the `schedule_type` column.  It also addresses the `refresh_child()` method in the partitioning for the same problem.  The previous work done to fix this issue did not take into account the nightly refresh component of the partition code, only the creation of the partitions.

Additional notes have been added to some of the comments as well to explain the purpose and usage of some methods as well.

### DIFF
--- a/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
+++ b/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
@@ -31,11 +31,11 @@ alter table ofec_sched_a_aggregate_employer_tmp rename to ofec_sched_a_aggregate
 create or replace function ofec_sched_a_update_aggregate_employer() returns void as $$
 begin
     with new as (
-        select 1 as multiplier, *
+        select 1 as multiplier, cmte_id, rpt_yr, contbr_employer, contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, contbr_id
         from ofec_sched_a_queue_new
     ),
     old as (
-        select -1 as multiplier, *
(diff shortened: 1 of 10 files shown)


ccostino:
SQLAlchemy will throw a warning when reflecting a table that has an `unknown` column type (or any type that it doesn't recognize), but it won't stop anything.  It's just a graceful handling of things it is unaware of (I traced it [back to here](https://bitbucket.org/zzzeek/sqlalchemy/issues/1327/postgressql-reflecting-tables-fails-if#comment-9010834)).  Not ideal, but also nice to know that things should continue processing as expected.

ccostino:
I think this is likely ready to go at this point, minus adding a note about that commented out line or creating some kind of improved way of specifying the cycle.  While we figure out what's best for that and get it cleaned up, I think the rest of the changes are ready for review now so this'll be ready to go later once the last change is made.  I don't anticipate more needing to be done.

Thank you so much for your help on this, @LindsayYoung, it is greatly appreciated!

PR #2297, opened 2017-04-06
Election search results fix

jontours:
Turns out to be a relatively simple solution here.  The `outerjoin` in the sql_alchmey code was causing nulled entries for Presidents and Senators, because all the code was doing was querying CandidateHistory, finding entries with a two-year period that matched the cycle passed in, and did an outerjoin on ElectionResults.  Since there are entries for numerous candidates in CandidateHistory for any given cycle there were nulled out entries in the join (null district and candidate_name).  The same problem for Senators.

The outer join was maybe necessary to address data inconsistencies as was discussed at https://github.com/18F/openFEC/issues/1271.  Because the endpoint was also allowing candidates with null districts.

This code is ready for review.  I checked the endpoint with a couple of states that were behaving badly (AK, NC).  A more thorough review may be needed in case this is filtering out some valid entries that I'm not aware of.

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -180,7 +180,7 @@ class ElectionView(utils.Resource):
         ).join(
             latest,
             aggregates.c.candidate_id == latest.c.candidate_id,
-        ).outerjoin(
+        ).join(
             outcomes,
             aggregates.c.candidate_id == outcomes.c.cand_id,
         )
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -67,14 +67,14 @@ class ElectionList(utils.Resource):
                 ],
                 else_=4,
             ).label('_office_status'),
-        ).outerjoin(
+        ).join(
             ElectionResult,
             sa.and_(
                 elections.c.state == ElectionResult.cand_office_st,
                 elections.c.office == ElectionResult.cand_office,
                 sa.func.coalesce(elections.c.district, '00') == ElectionResult.cand_office_district,
                 elections.c.two_year_period == ElectionResult.election_yr + cycle_length(elections),
-            ),
+            )
         ).order_by(
             '_office_status',
             ElectionResult.cand_office_district,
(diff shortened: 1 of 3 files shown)


jontours:
Ugh, failing tests so still some work to do here.

LindsayYoung:
taking a look now

webservices/resources/elections.py
```diff
@@ -91,8 +91,14 @@ def _get_elections(self, kwargs):
...
-                CandidateHistory.district == None,  # noqa
-                CandidateHistory.district != '99',
+                #This line was causing some semi-duplicate entries
+                #an example here: https://beta.fec.gov/data/elections/?cycle=2016&state=AK
+                #young shows up once. In one of those entries the district is null.  I don't know
+                #if this was to counter a bug or some incomplete data.  But I would say we don't
+                #want candidates with null districts showing up.  Thoughts?  The TODO directly
+                #above is also indicator of some prior inconsistent data, is this still the case?
+                #CandidateHistory.district == None,  # noqa
+                CandidateHistory.district != '99'
```
LindsayYoung: Correct me if I am wrong, but I think that since we wrote this we have switched the underlying data source I don't think we need this line, the 99 was a warehouse problem. We are using`disclosure.cand_valid_fec_yr` which is better there are some inconsistency in the numbers but it's things like 07 and 7, but that shouldn't be too bad for our purposes. 

As far as the Null values, @PaulClark2 @jwchumley there are 59 records, all house members that are missing districts. Is that something we can fix in `disclosure.cand_valid_fec_yr`? I think the more we can keep this logic simplified, the better.

webservices/resources/elections.py
```diff
@@ -67,14 +67,14 @@ def _get_records(self, kwargs):
                 ],
                 else_=4,
             ).label('_office_status'),
-        ).outerjoin(
+        ).join(
```
LindsayYoung: This is good, we should do this in the ElectionView too

LindsayYoung: Nope- looks like I was wrong about that

webservices/resources/elections.py
```diff
@@ -322,7 +322,7 @@ def filter_candidates(query, kwargs):
     query = query.filter(
         CandidateHistory.two_year_period <= kwargs['cycle'],
         CandidateHistory.two_year_period > (kwargs['cycle'] - duration),
-        CandidateHistory.election_years.any(kwargs['cycle']),
+        CandidateHistory.cycles.any(kwargs['cycle']),
```
jontours: This causes some null district entries in some of the results, been trying to figure out how to best filter those out.

LindsayYoung: @PaulClark2 @jwchumley can FEC fix the null district candidates? If I remember correctly, there are all house candidates and there are only about 50 of them.j

PR #2306, opened 2017-04-11
Improve logging and split out calls for nightly updates

ccostino:
This changeset adds a bit of extra support for logging when the nightly updates are run for the itemized schedule data.  It also splits the methods out a bit more from the one management command into one for each, just like the we did for the partitioning code.  This should provide us with more flexibility as well as more visibility into any problems that are currently being swallowed by the success messages.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -163,6 +163,7 @@ class TestViews(common.IntegrationTestCase):
         )
         db.session.commit()
         manage.update_aggregates()
+        manage.refresh_itemized()
         search = models.ScheduleA.query.filter(
             models.ScheduleA.sub_id == row.sub_id
         ).one()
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -163,6 +163,7 @@ def test_sched_a_fulltext_trigger(self):
         )
         db.session.commit()
         manage.update_aggregates()
+        manage.refresh_itemized()
```
LindsayYoung: Wondering why this was added to so many tests?

ccostino: It may not be needed - I did it as a safeguard since the code was originally being called as a part of `manage.update_aggregates()` but I can go back through and see if it's necessary in all of them.

ccostino: Oh I know why - most of these tests are against schedule A or B records, which expect the queue processing to have been done.  A couple don't though, so I will double check and remove as I can.

ccostino: Fixed!  Extraneous calls removed.

LindsayYoung:
Thanks!

PR #2316, opened 2017-04-13
Index commenters and representatives in AOs

vrajmohan:


anthonygarvan:
Are you sure we just need commenters & reps?

anthonygarvan:
(discussion in slack: there are only 4 roles, 2 for requestor and 1 for commenter + 1 for reps.

Issue #2319, opened 2017-04-17
Pac and Party reports returns nothing


jwchumley:
https://beta.fec.gov/data/reports/pac-party/?is_amended=false&data_type=processed

Presidential and House/Senate both return lists but if you choose PAC and Party it churns for a while and returns nothing.

noahmanger:
I'm seeing this too. Looks like a 502 error in the API call. Can anyone look into this? @ccostino @LindsayYoung @jontours ?

ccostino:
We were just discussing this in the FEC Slack #dev channel and the problem appears to be resolved.  The underlying cause was in the database with some of these queries:

```
canceling statement due to conflict with recovery
User query might have needed to see row versions that must be removed.
```

Which in turn was causing long-running queries to timeout in the web server resulting in worker timeouts (thus the 502 errors).  I'm not sure why these errors started happening but I didn't see any recent replication events or interruptions in RDS.  The production web site is using the read replica and that's where this activity was occurring, but again, no real indication as to why.

@jwchumley and @noahmanger, if you try again (you may need to hard refresh your browser), do you still see the error?

ccostino:
For more details and things to look at more closely:

* http://stackoverflow.com/questions/14592436/postgresql-error-canceling-statement-due-to-conflict-with-recovery
* https://www.postgresql.org/docs/current/static/hot-standby.html#HOT-STANDBY-CONFLICT

noahmanger:
This is working now, but it does seem to be particularly slow. I'm closing though since it's no longer broken.

LindsayYoung:
This one is not working, the call times out (there are over a million records for this.) I double checked and the indexes look right. 

We might need to change pagination to what we have itemized data. 

The fastest frontend fix is to enforce filtering by cycle. Could we get that going @noahmanger?

noahmanger:
That was an easy add: https://github.com/18F/openFEC-web-app/pull/1979

And it seems to help. Is the idea this is a temporary fix pre-flip and then we'll address the underlying issues later?

LindsayYoung:
Thanks @noahmanger. I can add something to the API side too. 

I think that will buy us a lot of time. If this comes up as a request again, we can see if a change in the pagination would allow searches across all the records.

noahmanger:
Fixed by https://github.com/18F/openFEC-web-app/pull/1979

PR #2330, opened 2017-04-20
Feature/restore efile reports

LindsayYoung:
This fixes the issue for now, we want to turn on the index validator again but it is not passing schemas

### DIFF
--- a/webservices/resources/reports.py
+++ b/webservices/resources/reports.py
@@ -263,7 +263,7 @@ class EFilingSummaryView(views.ApiResource):
             args.efilings,
             args.make_sort_args(
                 default='-receipt_date',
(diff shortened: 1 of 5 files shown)


LindsayYoung:
closes #2319

webservices/resources/reports.py
```diff
@@ -263,7 +263,7 @@ def args(self):
             args.efilings,
             args.make_sort_args(
                 default='-receipt_date',
-                validator=args.IndexValidator(self.model),
+                # validator=args.IndexValidator(self.model),
```
ccostino: We've noted that this will have to be adjusted and uncommented out in #2334.

PR #2348, opened 2017-04-25
Move contributor aggregate after single receipt amount

noahmanger:
This re-orders the schedule A model so that the contributor aggregate amount shows up after the individual receipt amount for greater clarity when downloading and viewing as a CSV.

cc @PaulClark2 @ChristianHilland


LindsayYoung:
Thanks!

PR #2351, opened 2017-04-26
Appends FRQ so as not to wipe out other arguments if they exist

jontours:
Other form types were being erased by the logic, this preserves them adds FRQ.

### DIFF
--- a/webservices/resources/filings.py
+++ b/webservices/resources/filings.py
@@ -61,8 +61,7 @@ class BaseFilings(views.ApiResource):
 
     def get(self, **kwargs):
         if kwargs.get('form_type') and 'RFAI' in kwargs.get('form_type'):
-            #If user happened to also pass in 'FRQ', remove it to prevent unnecessary filtering
-            #by appending to the list we don't wipe out the other form_type(s) passed in
+            #Adds FRQ types if RFAI was requested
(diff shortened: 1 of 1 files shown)


webservices/resources/filings.py
```diff
@@ -61,7 +61,10 @@ def args(self):
 
     def get(self, **kwargs):
         if kwargs.get('form_type') and 'RFAI' in kwargs.get('form_type'):
-            kwargs['form_type'] = ['RFAI', 'FRQ']
+            #If user happened to also pass in 'FRQ', remove it to prevent unnecessary filtering
+            #by appending to the list we don't wipe out the other form_type(s) passed in
+            kwargs.get('form_type').remove('FRQ')
```
LindsayYoung: Seems like we should leave this if that is what someone is explicitly looking for

jontours: If somebody is explicitly looking for `FRQ` files this will not alter that behavior.

LindsayYoung: OK, thanks for explaining!

jontours: It's much simpler to just remove it, whatever performance hit on database is minimal (if any at all) is acceptable if we keep the logic simple.

PR #2355, opened 2017-04-27
Update env var references

ccostino:
This changeset updates most env var references to use `env.get_credential` to ensure that an environment variable is always found.


LindsayYoung:
Thanks!

PR #2361, opened 2017-05-01
[WIP] - Fixes null versioning on some efiled records

jontours:
Coding is done for this but need to run update schemas to verify this doesn't cause unintended consequences, as adding that null check could cause some weird things, given that it's in the base case of the recursion.  After verification will update the label on the PR.


lbeaufort:
Archiving this PR

PR #2363, opened 2017-05-03
Line number filter

jontours:
Ready for review.  Partitioning has to be done for this as it adds a column to the itemized endpoints.  Since we have to do that again here soon I didn't see that as a problem.


noahmanger:
Not counting as a review, but just wanted to say thanks for including the `line_number_label` in there as well. Smart!

jontours:
Gonna add some inline documentation for the line number filter and add a filter test for this.  It will be ready for sure then.

PR #2372, opened 2017-05-08
Fixes missing cycles for presidential candidates

jontours:
Fixed for Presidential, going to try and find a good test case for senatorial candidates.


jontours:
Thanks! @ccostino

PR #2374, opened 2017-05-08
[WIP] page count issue

LindsayYoung:
Want to test and see if we can fix the API by getting rid of the itemized counts for now. That part of the query is timing out.


LindsayYoung:
This isn't helpful after all

PR #2391, opened 2017-05-12
[WIP] Adding compound indexes on ids and cycle

LindsayYoung:
Adding compound indexes on ids and cycle, since those are almost always called together

also got rid of change log- I thought that already happened


LindsayYoung:
running this on dev now

Issue #2390, opened 2017-05-12
Interim fix for the nightly itemized schedule updates


ccostino:
@LindsayYoung and I were discussing the impact of the nightly updates routinely failing and putting us in a perpetual state of requiring partitioning of the itemized schedule A and B data, and we came to the realization that there is a solution for a stop-gap fix until we're able to address the root cause of the issue (detailed in issues #2309 and #2227; we have a meeting setup with FEC folks next week).

The idea here is that we check for the presence of an empty recordset from the view before we process it and populate it in the designated queue tables.  [PostgreSQL provides a way to check for this easily](https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS), so we'll implement a hot fix to address it and at least get our nightly updates cleaned up now.

There's an additional task that we can do as a separate hot fix to keep track of which `sub_id`s fall into this state in a separate table and retry them at another time (e.g., a day later).  They'd remain in this table until the record is successfully processed, and it'll give us a bit more of a window into which records are falling into this state so we can understand why better.  Is it an expected thing (e.g., they don't exist in the corresponding `disclosure.f_item_receipt_or_exp` table) or something else?  It'll provide us with information to go on.

To recap:

* [x] Fix the nightly updates of itemized data by checking for empty resultsets before processing (#2392)
* [x] Add support for keeping track of and retrying failed `sub_id`s (#2398)

ccostino:
For the second piece of these, we anticipate the parts/tasks to be as follows:

* Create two tables to track `sub_ids` not found for schedule A and B respectively
* Add support to the trigger methods (via an `else` clause to the empty resultset check) to put a `sub_id` not found into the corresponding table - might have to track the operation as well
* Add an additional nightly update task to process the `sub_id`s not found the previous day to see if they're now available in the view

noahmanger:
Can this be closed now that #2398 is in?

ccostino:
Yes, I think so.  If something else comes up with this it's going to be a new issue and require new work.  Thanks, @noahmanger!

PR #2392, opened 2017-05-12
[MERGE WITH GIT FLOW] Fix itemized schedule trigger errors

ccostino:
**MERGE WITH GIT FLOW**

**NOTE:**  Requires running `update_itemized` (for `a` and `b`), which may require additional work as those will wipe out the existing schedule A and B queues.

Part of #2390

This changeset addresses an issue we are currently having with our itemized schedule triggers.  The triggers are setup to select rows from views currently, but they often run into a situation where the selection statement actually returns an empty resultset.  This in turn causes what amounts to an empty record (excluding what we ourselves add) being added to the itemized schedule queues to be processed later, only to fail because of trying to insert records that do not comply with table constraints (e.g., inserting a `NULL` value into a primary key column).

This puts us in a state of having to perpetually refresh data manually, which requires repartitioning the itemized schedule tables.  There is more work to be done to fix the root cause(s) of this problem, but this will at least prevent errors in our nightly refresh process for the time being.

I have also fixed a couple of issues with the integration tests in this.

**NOTE:** This now includes the index fixes @LindsayYoung worked on so they all go in together without necessitating multiple hot fixes.


data/sql_setup/prepare_schedule_a.sql
```diff
@@ -28,24 +28,38 @@ declare
...
-            insert into ofec_sched_a_queue_new values (view_row.*, timestamp, two_year_transaction_period_new);
-        end if;
+        -- Check to see if the resultset returned anything from the view.  If
+        -- it did not, skip the processing of the record, otherwise we'll end
+        -- up with a record full of NULL values.
+        -- "FOUND" is a PL/pgSQL boolean variable set to false initially in
+        -- any PL/pgSQL function and reset whenever certain statements are
+        -- run, e.g., a "SELECT INTO..." statement.  For more information,
+        -- visit here:
+        -- https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS
```
LindsayYoung: This is great, thanks

ccostino: Thank you, you're welcome! :-)

PR #2393, opened 2017-05-12
[MERGE WITH GITFLOW] hotfix/more celery in prod

LindsayYoung:
moving celery to the inherited manifests so we can specify instances


LindsayYoung:
still working on this syntactically

manifest_base.yml
```diff
@@ -18,8 +18,6 @@ applications:
   health-check-type: process
   command: celery beat --app webservices.tasks
 - name: celery-worker
-  instances: 1
-  memory: 512M
```
ccostino: Do the apps default to these values when pushed to cloud.gov if they're not specified?

ccostino: Note that this was accounted for in the subsequent commits - thanks @LindsayYoung! :-)

PR #2398, opened 2017-05-17
[MERGE WITH GIT FLOW] Fix nightly update trigger improvements

### LINKED ISSUE #2390: Interim fix for the nightly itemized schedule updates

ccostino:
**MERGE WITH GIT FLOW**

Fixes #2390

This changeset addresses a couple of issues we uncovered with our last nightly update fixes.  First, the record was not being returned properly in the trigger function due to incorrect shuffling around of return statements.  Second, we now also track which sub_ids are not found and place them in a secondary queue.  This queue is processed separately and mirrors the functionality of the main queues with the exception that sub_ids will remain in here until the record can be processed successfully from the itemized views we have to process from.

These improvements are a stop gap measure to keep our nightly updates running smoothly until we can improve the overall processing of the itemized schedule data.


data/sql_setup/prepare_schedule_a.sql
```diff
@@ -1,3 +1,15 @@
+-- Create table to hold sub_ids of records that fail during the nightly
+-- processing so that they can be tried at a later time.
+-- "sub_id" is not unique as we may have to process it multiple times.
+-- The "action" column denotes what should happen with the record:
+--    insert, update, or delete
+drop table if exists ofec_sched_a_nightly_retries;
+create table ofec_sched_a_nightly_retries (
+    sub_id numeric(19,0) not null primary key,
```
vrajmohan: `primary key` appears to violate comment on line 3.

ccostino: D'oh!  When I had worked through this more I realized we shouldn't ever have a `sub_id` in this table more than once, so I figured the `primary key` would be okay.  I'll update the comment to reflect that, thanks!

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -87,33 +164,52 @@ begin
...
+            -- so keep track of this sub_id if we haven't already so we can
+            -- try processing it again each night until we're able to
+            -- successfully process it.
+            raise notice 'sub_id % not found, adding to secondary queue.', old.sub_id;
+
+            delete from ofec_sched_a_nightly_retries where sub_id = old.sub_id;
+            insert into ofec_sched_a_nightly_retries values (old.sub_id, 'update');
         end if;
+
+        return old;
```
vrajmohan: We should be returning `new` here.

ccostino: *facepalm* you're right!

ccostino:
@vrajmohan, @LindsayYoung, and @jontours - this is ready for a full review now, but I'd also like to give it one more look too before merging, I'll take a look later tonight!

ccostino:
Ready for review and merge. :-)

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -14,6 +25,54 @@ create index on ofec_sched_a_queue_old (timestamp);
...
+
+            if two_year_transaction_period >= start_year then
+                -- Determine which queue(s) the found record should go into.
+                case schedule_a_record.action
+                    when 'insert' then
+                        delete from ofec_sched_a_queue_new where sub_id = view_row.sub_id;
+                        insert into ofec_sched_a_queue_new values (view_row.*, timestamp, two_year_transaction_period);
+
+                        delete from ofec_sched_a_nightly_retries where sub_id = schedule_a_record.sub_id;
+                    when 'delete' then
```
LindsayYoung: Do we want to get rid of the view check and delete the record directly by sub id? That can be a follow up task if we want to get this moving sooner

ccostino: That makes sense to me, but I think that should be a follow up task because there are a couple of ways to go about that. :-)

PR #2403, opened 2017-05-20
Feature/restrict ao search fields

vrajmohan:


tests/test_legal.py
```diff
@@ -210,7 +210,12 @@ def test_query_dsl_with_ao_category_filter(self, es_search):
                 'bool': {
                     'must': [
                         {'term': {'_type': 'advisory_opinions'}},
-                        {'match': {'_all': 'president'}},
+                        {'bool': {'minimum_should_match': 1}}
```
anthonygarvan: I don't understand this line, isn't the query covered by the minimum_should_match from below?

vrajmohan: I was puzzled by it too and this was one of the reasons this PR took so long. This one seems to be generated by `elasticsearch_dsl` and is (redundantly) inside the `must`. The one needed by the `should` is on line 241.

anthonygarvan: wow yeah, that is super confusing.

PR #2406, opened 2017-05-22
Remove RAISE statements from triggers

ccostino:
This changeset removes the `RAISE` statements in the itemized schedule triggers as they appear to be causing GoldenGate to malfunction when they are called.  They are only supposed to write statements to the PostgreSQL log, not throw exceptions, but it seems that GoldenGate is treating them as exceptions anyway and halting processing.


LindsayYoung:
Let's hope this is the issue! Thanks again!

Issue #2309, opened 2017-04-11
Investigate issue with itemized schedule queues not being processed


ccostino:
Since moving and deploying fully to GovCloud, we are noticing issues in the cloud.gov logs around the nightly update processing of the itemized schedule data.  The updates are actually failing behind the scenes because the process is reporting errors when processing the queue tables (records with `null` values, including in the `sub_id` field, which is the primary key and doesn't allow `null` values!).

One problem with this is that our logging and email reports isn't providing enough information, but that is already in the works to be fixed in #2306.  The remaining issue is with why this is happening with the data in the first place and why the queue processing is still failing.

ccostino:
@jontours shared a great insight as to why this might be occurring and @jwchumley was able to at least partially confirm our suspicions, if not give us a clear path forward.  Thank you both!

Since the switch from using the older `sched_a` and `sched_b` tables (and their successors, `fec_vsum_sched_a` and `fec_vsum_sched_b`) to the new views, `fec_vsum_sched_a_vw` and `fec_vsum_sched_b_vw`, we've struggled with getting our nightly refreshes for the itemized schedule data back on track.  [My](https://github.com/18F/openFEC/pull/2218) [previous](https://github.com/18F/openFEC/pull/2293) [attempts](https://github.com/18F/openFEC/pull/2306) to fix and shed more light on the issue have gotten us close, but not quite there yet.  I made the assumption that the underlying `disclosure.nml_sched_a` and `disclosure.nml_sched_b` tables where the primary driving factors to changes behind those views, but this is incorrect.

@jontours determined that it was likely because the records were also not found in `disclosure.v_sum_and_det_sum_report` and @jwchumley confirmed that this could be the case, though with some potential caveats.

I did a bit more digging just to see why we'd ultimately end up with mostly empty records in the queues and discovered that if I tried to select a record with a `sub_id` that was not also found in `disclosure.v_sum_and_det_sum_report`, that'd return an empty result set.  This is not surprising, but what I failed to account was the fact that we're adding metadata to the record and then storing it, so the functions tied to the triggers we have setup are appending a couple of pieces of information and then storing the record in the corresponding queue table.  Since these records have nothing else in them, the remaining columns are filled with `nulls`.

At this point it's now a matter of finding the correct source table(s) to watch for changes on before checking the views for all of the information and accounting for any other caveats that might be a part of this.

[...]

PR #2410, opened 2017-05-22
[WIP] Refactor itemized schedule A/B processing to use "upsert"

### LINKED ISSUE #2446: Refactor itemized schedule A and B setup
We've had several issues come up in the last few months around schedule A and B and how to better handle that data:

* [Investigate PostgreSQL's "upsert" capabilities](https://github.com/18F/openFEC/issues/2227)
* [Queues not being processed](https://github.com/18F/openFEC/issues/2309)
* [Database performance and tuning research](https://github.com/18F/openFEC/issues/2423)
* [Research spike for ElasticSearch support](https://github.com/18F/openFEC/issues/2373)

There have also been numerous PRs worked on that attempted to address numerous issues we've had with the recent switch to views as the data source for the `ofec_*` tables.  We've only just made this work again prior to the site launch, but we're already facing the impacts of trying to build more on this data and offer improvements, most notably being able to search/filter/browse across cycles.

We had a great kick-off conversation with some of the folks on FEC's DBA team around the schedule A and B data, our current setup, and where we'd like to go from here.  We'll be having more of these in the future but for the moment there's some work we can take care of on the API side now to set ourselves up for the future.  At a high level this looks like the following:

* Remove the queue processing for schedule A and B data and leverage PostgreSQL's support of [`INSERT INTO ... ON CONFLICT DO ...`](https://www.postgresql.org/docs/current/static/sql-insert.html#SQL-ON-CONFLICT) statement and clause
* Shift the partitioning code to the SQL side
* Tune our indexes and queries against this data

---

## Additional details and TODO items

We already have a PR open (#2410) with this work semi-started (it will be closed out in favor of newer work though informed by some recent lessons-learned), but here is the breakdown of everything that needs to take place as a part of this refactoring (some of these `TODO`s are found in the code as well):

* [x] Shift most of the partitioning code to the SQL side (maintain some of the Python code for management and cleanup):
   * [x] Shift master table creation to the `sql_prepare` scripts
   * [x] Figure out how to get data in the right child table automatically when inserting into a master table - this requires defining a trigger on the master tables like the found at the end of section 5.10.2 in the [PostgreSQL Partitioning documentation](https://www.postgresql.org/docs/current/static/ddl-partitioning.html#DDL-PARTITIONING-IMPLEMENTATION) and leveraging the [`INSERT INTO ... ON CONFLICT DO ...`](https://www.postgresql.org/docs/current/static/sql-insert.html#SQL-ON-CONFLICT) statement and clause - this would account for `INSERT`s and `UPDATE`s in a single `BEFORE INSERT` trigger, but we'd also need a `BEFORE DELETE` trigger to delete rows directly (also note that all triggers will need to key off the `two_year_transaction_period` in addition to the `sub_id` to ensure the correct child table is used)
   * [x] Adjust the `BEFORE INSERT` trigger to process some of the fields added on the API side before inserting or updating rows, thus removing the need to repeat this logic elsewhere (see the current [queue and nightly retry processing](https://github.com/18F/openFEC/blob/develop/data/sql_setup/prepare_schedule_a.sql) and [partitioning code](https://github.com/18F/openFEC/blob/develop/webservices/partition/sched_a.py) for examples of these processed fields - all reference schedule A but the same applies for schedule B)
   * [x] Shift child table management to scripts in a new `sql_partitioning` directory, including table adjustments, index creation, adding all constraints, and any other updates (e.g., `SET STATISTICS`)
   * [x] Leverage the trigger being added to the master table to just do a wholesale insert into the master table and have it deal with putting the data in the correct child tables
* [x] Remove all itemized schedule nightly refresh code/queue processing as it should no longer needed
* [x] Refactor the aggregate processing/updating for itemized schedule A and B as these still reference the queue tables
* [x] Update tests to reflect the changes

## Caveats

There are a couple of caveats that we'll have to account for with these changes:

* [x] [PostgreSQL's Partitioning documentation](https://www.postgresql.org/docs/current/static/ddl-partitioning.html#DDL-PARTITIONING-CAVEATS) mentions a specific caveat about using `INSERT` statements with the `ON CONFLICT` clause and that it may not work correctly.  We'll have to investigate this and make adjustments as necessary. (VM to investigate)
* [x] The `ON CONFLICT DO UPDATE...` part of the clause **requires** that you explicitly set values.  From the [PostgreSQL `INSERT` docs](https://www.postgresql.org/docs/current/static/sql-insert.html#SQL-ON-CONFLICT):  "... `DO UPDATE` clause specifying the exact details of the UPDATE action to be performed in case of a conflict."

ccostino:
**WIP:  PLEASE DO NOT MERGE**

Part of #2309
Part of #2423
Closes #2446

This changeset introduces a major refactoring in how we keep track of and process the itemized schedule data.  Instead of using the queue tables like before, we are now leveraging the new `INSERT INTO ... ON CONFLICT ...` functionality that PostgreSQL provides as of 9.5+.  This has allowed us to greatly simplify the triggers we have associated with the itemized schedule A and B data and remove a lot of complicated data processing and management code with the table partitioning we have setup for the API.  It also provides us with updates to the table as the changes happen instead of during a nightly process.

These changes are only for schedule A and B right now, but at least schedule E could also benefit from a similar refactor in the future.


ccostino:
@vrajmohan I know you're already diving in with me on this, thanks!  If @anthonygarvan, @jontours, and/or @LindsayYoung want to hop in too please feel free to do so, even if just for pairing sessions at times.  There's plenty to do! :-)

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -31,7 +32,269 @@ begin
...
+                    view_row.image_num,
+                    view_row.file_num,
+                    view_row.link_id,
+                    view_row.orig_sub_id,
+                    view_row.sub_id,
+                    view_row.filing_form,
+                    view_row.rpt_tp,
+                    view_row.rpt_yr,
+                    view_row.election_cycle,
+                    view_row.timestamp,
```
vrajmohan: view_row does not have `timestamp` column

ccostino: That's what you get for staring at code too long... 😆

ccostino: FYI that I still need to fix this.

ccostino:
This PR is going to be deprecated in favor of a fresh start based on lessons learned this past week.

ccostino:
At this point I think it's safe to close this PR - schedule E received a complete overhaul and @vrajmohan has been making a lot of great progress with a cleaner implementation in #2498!

PR #2414, opened 2017-05-23
fixing index name

LindsayYoung:


LindsayYoung:
Thanks @noahmanger for finding that!

PR #2424, opened 2017-05-24
Adding a refresh command and task that runs every 15 minutes

### LINKED ISSUE #2422: refresh task for calendar
refresh it every 15 minutes 

I think we can make a management command, and add a cron job based on that, that way we could run it as a task too if there was a calendar emergency or something

LindsayYoung:
closes #2422

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -475,6 +475,13 @@ def load_efile_sheets():
         df.drop(columns_to_drop, axis=1, inplace=True)
         df.to_json(path_or_buf="data/" + table + ".json", orient='values')
 
+@manager.command
+def refresh_calendar():
+    """ Refreshes calendar data
+    """
+    with db.engine.begin() as connection:
+        connection.execute('REFRESH MATERIALIZED VIEW CONCURRENTLY ofec_omnibus_dates_mv')
+
 
 if __name__ == '__main__':
     manager.run()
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -475,6 +475,13 @@ def load_efile_sheets():
         df.drop(columns_to_drop, axis=1, inplace=True)
         df.to_json(path_or_buf="data/" + table + ".json", orient='values')
 
+@manager.command
+def refresh_calendar():
+    """ Refreshes calendar data
+    """
+    with db.engine.begin() as connection:
+        connection.execute('REFRESH MATERIALIZED VIEW CONCURRENTLY ofec_omnibus_dates_mv')
```
ccostino: This will refresh without issue, right?  Meaning there are no dependencies we need to be aware of?

webservices/tasks/__init__.py
```diff
@@ -17,6 +17,10 @@
             'task': 'webservices.tasks.refresh.refresh',
             'schedule': crontab(minute=0, hour=9),
         },
+        'calandar': {
+            'task': 'manage.refresh_calendar',
+            'schedule': crontab(minute='0,5,30,45'),
```
ccostino: Should this be `0,15,30,45`?

manage.py
LindsayYoung: no ofec dependencies

webservices/tasks/__init__.py
LindsayYoung: yes

PR #2425, opened 2017-05-24
Refresh AOs every 15m

vrajmohan:
Check every 15 minutes for AOs modified in the last day. Use the
first AO in this set as the starting AO for loading advisory opinions.
This may be run multiple times with no harm as the loading process is
idempotent.


anthonygarvan:
this looks great, nice work :)

PR #2432, opened 2017-05-25
Feature/zero down deploy

### LINKED ISSUE #2388: zero downtime celery
Modify the deploy script so that the redis and celery infrastructure also work in a zero downtime manner.

LindsayYoung:
This should prevent download errors during redeployment. It creates 3 separate zero downtime deploys for each app. 

This creates more manifests than is ideal, but chained inheritance should make it so most things only need to be changed in one place and I tucked them in a folder. The checks on deploy require a manifest so I put a place holder manifest in the root. 

closes #2388


LindsayYoung:
I could make this a hotfix if we have more hotfixes going out, if not it would be nice to do a normal release

LindsayYoung:
Ok, I might be able to do this without an insane number of manifests

LindsayYoung:
I would love to reduce the number of manifests but I think the cf push from a multiple app manifest is a bit buggy. We kept seeing the celery apps deploy twice for a while now and I can see when we do a deploy, there is a long spike of downtime for downloads when the API deploys and then another spike of downtime when the celery worker rebuilds again

LindsayYoung:
Confirmed, cf push is buggy, but we can make travis install a new one! I am going to make a new PR

PR #2433, opened 2017-05-26
Move schedule F to `MV`

jontours:
Almost ready for review.


jontours:
Test is failing but think the new subset data will clear that.

jontours:
Argh, looks like I built this off an old master, gonna close this in favor of a true hotfix.  ooops.

PR #2434, opened 2017-05-26
[MERGE WITH GIT FLOW] Change Schedule C to a materialized view

ccostino:
**MERGE WITH GIT FLOW**

This changeset readjusts our schedule C data to be available via a materialized view as opposed to accessing the fec_vsum_sched_c_vw view directly.  As we cannot setup triggers to watch the view, this is an easier way to keep the information updated.

This also fixes an issue with the site currently where loan information is not visible because of a reference to the pg_date column, which was not in the original view.


data/sql_updates/sched_c.sql
```diff
@@ -0,0 +1,31 @@
...
+    *,
+    to_tsvector(cand_nm) as candidate_name_text,
+    to_tsvector(loan_src_nm) as loan_source_name_text,
+    now() as pg_date
+from fec_vsum_sched_c_vw;
+
+create unique index on ofec_sched_c_mv_tmp (sub_id);
+
+-- Create simple indexes on filtered columns
+create index on ofec_sched_c_mv_tmp (cmte_id);
```
jontours: I might borrow this indexing strategy for schedule_c.

PR #2439, opened 2017-05-26
Hotfix/deploy fix

LindsayYoung:
sorry counted wrong

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -197,7 +197,7 @@ def deploy(ctx, space=None, branch=None, login=None, yes=False):
     for app in ('api', 'celery-worker', 'celery-beat'):
         deployed = ctx.run('cf app {0}'.format(app), echo=True, warn=True)
         cmd = 'zero-downtime-push' if deployed.ok else 'push'
-        ctx.run('cf {0} {1} -f manifests/manifest_{1}_{2}.yml'.format(cmd, appstr.replace('-','_'), space), echo=True)
+        ctx.run('cf {0} {1} -f manifests/manifest_{1}_{2}.yml'.format(cmd, app.replace('-','_'), space), echo=True)
 
 
 @task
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -197,7 +197,7 @@ def deploy(ctx, space=None, branch=None, login=None, yes=False):
     for app in ('api', 'celery-worker', 'celery-beat'):
         deployed = ctx.run('cf app {0}'.format(app), echo=True, warn=True)
         cmd = 'zero-downtime-push' if deployed.ok else 'push'
-        ctx.run('cf {0} {1} -f manifests/manifest_{2}_{3}.yml'.format(cmd, app, space), echo=True)
+        ctx.run('cf {0} {1} -f manifests/manifest_{1}_{2}.yml'.format(cmd, app, space), echo=True)
```
ccostino: This still won't work because `celery-worker` and `celery-beat` don't match the file names, which have `celery_worker` and `celery_beat` in them.  I think if we just change the dash to an underscore in the app list in the `for in` loop that should do the trick!

LindsayYoung: good catch

tasks.py
```diff
@@ -197,7 +197,7 @@ def deploy(ctx, space=None, branch=None, login=None, yes=False):
     for app in ('api', 'celery-worker', 'celery-beat'):
         deployed = ctx.run('cf app {0}'.format(app), echo=True, warn=True)
         cmd = 'zero-downtime-push' if deployed.ok else 'push'
-        ctx.run('cf {0} {1} -f manifests/manifest_{2}_{3}.yml'.format(cmd, app, space), echo=True)
+        ctx.run('cf {0} {1} -f manifests/manifest_{1}_{2}.yml'.format(cmd, appstr.replace('-','_'), space), echo=True)
```
ccostino: D'oh, appstr isn't defined, should still be just app. :-)

PR #2441, opened 2017-05-26
[MERGE WITH GIT FLOW] Fix manifest file formatting

ccostino:
**MERGE WITH GIT FLOW**

Several of the manifest files were not formatted correctly, thus causing YAML parsing errors in environments other than development.  This changeset fixes the remaining files.


vrajmohan:
Looks good to me

PR #2443, opened 2017-05-30
hotfix/end-whitelist

LindsayYoung:
Removing the whitelist so all resources can be downloaded through the API, via web app

- want to test this on dev now


LindsayYoung:
This works but doesn't really solve the issue. For the downloads, the views need to be modified so the download script can call get_query an get the un paginated information. It is not a big lift, but we will need to prioritize that later.

PR #2447, opened 2017-05-31
returning 404 if no mur docs are found

anthonygarvan:


webservices/resources/legal.py
```diff
@@ -58,8 +59,11 @@ def get(self, doc_type, no, **kwargs):
             .execute()
 
         results = {"docs": [hit.to_dict() for hit in es_results]}
-        return results
 
+        if len(results['docs']) > 0:
```
vrajmohan: It may read better to check `es_results` _before_ constructing the `results` dict.

PR #2448, opened 2017-05-31
Naming the indexes explicitly 

LindsayYoung:
on schedule A and B because sqlalchemy's auto naming was over writing them

This will require repartitioning.


ccostino:
Indexes look good, I just tested locally!  I just want to check if we should switch the target branch.

PR #2449, opened 2017-05-31
fix name

LindsayYoung:
The prod alerts were labeled staging 

Sorry, my bad.


ccostino:
D'oh, I thought I fixed those last time around when I noticed them after helping fix the celery worker issues!  Sorry about that.  Should we target the release branch now that those are created?

PR #2453, opened 2017-06-02
Hotfix/calandar update

LindsayYoung:
This is really minor, but will make content managers happy. I would like to get it in. I can deploy it later if someone approves

### DIFF
--- a/webservices/tasks/refresh.py
+++ b/webservices/tasks/refresh.py
@@ -30,6 +30,6 @@ def refresh():
 
 
 @app.task
-def refresh_calandar():
+def refresh_calendar():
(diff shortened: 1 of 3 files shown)


webservices/tasks/refresh.py
```diff
@@ -27,3 +27,9 @@ def refresh():
         mail.send_mail(buffer)
     except Exception as error:
         logger.exception(error)
+
+
+@app.task
+def refresh_calandar():
```
ccostino: Spelling!  `calendar` :-)

LindsayYoung: Thank you!

PR #2462, opened 2017-06-08
[WIP] Feature/use trigger for partition updates

vrajmohan:
The tests are failing because the parent and child tables do not have columns in the same order. I will be attempting to fix this by creating the child table using an `INHERITS` clause so that the column orders are identical. We could name the columns explicitly in the trigger instead - I would prefer not to do this as the trigger becomes verbose with no other benefit.


data/sql_setup/prepare_schedule_a.sql
```diff
@@ -214,3 +214,16 @@ create trigger nml_sched_a_after_trigger after insert or update
 drop trigger if exists nml_sched_a_before_trigger on disclosure.nml_sched_a;
 create trigger nml_sched_a_before_trigger before delete or update
     on disclosure.nml_sched_a for each row execute procedure ofec_sched_a_delete_update_queues(:START_YEAR_AGGREGATE);
+
+CREATE OR REPLACE FUNCTION insert_sched_master() RETURNS TRIGGER AS $$
+DECLARE
+    child_table_prefix TEXT = TG_ARGV[0];
+    child_table_name TEXT;
+BEGIN
+    child_table_name = child_table_prefix || get_partition_suffix(new.two_year_transaction_period);
```
ccostino: We'll need to derive both the `two_year_transaction_period` and `timestamp` fields (though potentially not `timestamp` - need to double check) before we can just access them an insert them with all of the other values.

I imagine that before this line, we'll need to still declare these like so:

```sql
timestamp TIMESTAMP = CURRENT_TIMESTAMP;
two_year_transaction_period SMALLINT;
```

And then set `two_year_transaction_period` like so:

```sql
two_year_transaction_period = get_transaction_year(view_row.contb_receipt_dt, view_row.rpt_yr);
```

Furthermore, the nature of deriving the value for `two_year_transaction_period` has come up in discussion recently with the FEC DBA folks, so we should also verify this formula (or see if we can/need to just stick solely with `rpt_yr`.  @LindsayYoung, do you remember if there was anything specific said in the last meeting we had with them around this or if we'd still have to investigate further?

webservices/common/models/itemized.py
```diff
@@ -94,9 +97,8 @@ class ScheduleA(BaseItemized):
     contributor_middle_name = db.Column('contbr_m_nm', db.String)
     contributor_last_name = db.Column('contbr_nm_last', db.String)
     contributor_suffix = db.Column('contbr_suffix', db.String)
-    # Street address omitted per FEC policy
-    # contributor_street_1 = db.Column('contbr_st1', db.String)
-    # contributor_street_2 = db.Column('contbr_st2', db.String)
+    contributor_street_1 = db.Column('contbr_st1', db.String)
+    contributor_street_2 = db.Column('contbr_st2', db.String)
```
ccostino: Are we sure these need to go back in?  They were omitted previously at FEC's request.

webservices/common/models/itemized.py
```diff
@@ -76,9 +76,12 @@ def pdf_url(self):
 class ScheduleA(BaseItemized):
     __tablename__ = 'ofec_sched_a_master'
 
+    committee_name = db.Column('cmte_nm', db.String, doc=docs.COMMITTEE_NAME)
```
ccostino: I thought this was already defined in this model?

webservices/common/models/itemized.py
```diff
@@ -76,9 +76,12 @@ def pdf_url(self):
...
+
     # Contributor info
     entity_type = db.Column('entity_tp', db.String)
     entity_type_desc = db.Column('entity_tp_desc', db.String)
+    unused_contbr_id = db.Column('contbr_id', db.String)
```
ccostino: What is this field for?

webservices/common/models/itemized.py
```diff
@@ -118,11 +121,36 @@ class ScheduleA(BaseItemized):
...
+    conduit_committee_street1 = db.Column('conduit_cmte_st1', db.String)
+    conduit_committee_street2 = db.Column('conduit_cmte_st2', db.String)
+    conduit_committee_city = db.Column('conduit_cmte_city', db.String)
+    conduit_committee_state = db.Column('conduit_cmte_st', db.String)
+    conduit_committee_zip = db.Column('conduit_cmte_zip', db.Integer)
+
+    donor_committee_name = db.Column('donor_cmte_nm', db.String)
+    national_committee_nonfederal_account = db.Column('national_cmte_nonfed_acct', db.String)
+    cycle = db.Column('election_cycle', db.Integer)
+    timestamp = db.Column('timestamp', db.DateTime)
```
ccostino: There are a lot of field additions and some changes here; it's totally fine if we need them, but I'm curious why they're now included.  The `conduit info` in particular, what is that referring to?

Are these pulled from the `BaseItemized` class?  If so, we should review these with @jontours and @LindsayYoung as there were some things we had to leave there and others that we placed in the specific itemized classes to account for API needs and requirements.

webservices/common/models/itemized.py
```diff
@@ -118,11 +121,36 @@ class ScheduleA(BaseItemized):
...
 
     # Transaction meta info
+    election_type = db.Column('election_tp', db.String) # ? election_type looks like it's included in BaseItemized already
```
ccostino: No, I don't think it is after looking above.  I'm curious as to why it was omitted and if we should include it now.

webservices/common/models/itemized.py
```diff
@@ -118,11 +121,36 @@ class ScheduleA(BaseItemized):
...
+    election_type_full = db.Column('election_tp_desc', db.String)
```
ccostino: This is probably okay to have, but I just wanted to point out that the `BaseRawItemized` class (not inherited from here) has a `fec_election_type_desc` property defined on it with SQLAlchemy's `@hybrid_property` decorator.  Should we have a similar setup for the `BaseItemized` class?

/cc @LindsayYoung and @jontours on this one too

webservices/partition/base.py
```diff
@@ -266,46 +267,20 @@ def rename_child(cls, cycle):
...
 
         try:
-            delete_select = sa.select([queue_old.c.get(cls.primary)]).where(
-                queue_old.c.two_year_transaction_period.in_([start, stop])
-            )
-            delete = sa.delete(child).where(
-                child.c.get(cls.primary).in_(delete_select)
+            delete_select = sa.select([queue_old.c.get(cls.primary)])
+            delete = sa.delete(master_table).where(
+                master_table.c.get(cls.primary).in_(delete_select)
```
ccostino: Interesting!  So you're removing the specific processing for each child table and shifting the updates from the queues to the master table itself?

webservices/partition/base.py
```diff
@@ -316,6 +291,7 @@ def refresh_child(cls, cycle, queue_old, queue_new):
             # populating the child tables during normal partitioning.
             # Otherwise, an error is thrown due more values being specified
             # than there are columns to accept them.
+            # VM TODO - Is this still needed?
```
ccostino: Given the other changes here and setup of the master table, it may not be.  We'll have to review the flow of the data through the processing and see if it's still an issue.  It'd be nice if it weren't!

webservices/partition/sched_a.py
```diff
@@ -89,3 +89,10 @@ def index_factory(cls, child):
     def update_child(cls, child):
         cmd = 'alter table {0} alter column contbr_st set statistics 1000'.format(child.name)
         db.engine.execute(cmd)
+
+    @classmethod
+    def create_trigger(cls):
+        db.engine.execute('DROP TRIGGER IF EXISTS insert_sched_a_trigger ON ofec_sched_a_master')
```
ccostino: The trigger name should be `insert_sched_trigger` based on definition above, right?

webservices/partition/sched_a.py
```diff
@@ -89,3 +89,10 @@ def index_factory(cls, child):
...
+        db.engine.execute('''
+            CREATE trigger insert_sched_a_trigger BEFORE INSERT ON ofec_sched_a_master FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_a_');
```
ccostino: Same note as above:  The trigger name should be `insert_sched_trigger` based on definition above, right?

webservices/partition/sched_b.py
```diff
@@ -79,3 +79,10 @@ def index_factory(cls, child):
     def update_child(cls, child):
         cmd = 'alter table {0} alter column recipient_st set statistics 1000'.format(child.name)
         db.engine.execute(cmd)
+
+    @classmethod
+    def create_trigger(cls):
+        db.engine.execute('DROP TRIGGER IF EXISTS insert_sched_a_trigger ON ofec_sched_b_master')
+        db.engine.execute('''
+            CREATE trigger insert_sched_a_trigger BEFORE INSERT ON ofec_sched_b_master FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_b_');
+            ''')
```
ccostino: Same note as above for both of these lines:  The trigger name should be `insert_sched_trigger` based on definition above, right?

webservices/common/models/itemized.py
LindsayYoung: @vrajmohan found some differences in the base table and the child tables, so it does make sense to get our tables in line.

If this table acts like the other tables I have seen candidate_name is consistent, but first, last, middle, prefix and suffix is only populated for some filers, so we generally just go with name to make the data less confusing for users. 

For the conduit information that is valuable if you are looking at earmarks, but I think we will just want ID and name, people can look up the address elsewhere if they need it. 

We will want to look at what these look like as a download, we will want to include all useful things, but try not to overwhelm people.

This might be a fun opportunity for usability testing on the downloads and API

ccostino: Ah nice, thanks @LindsayYoung!  That makes total sense then with the field changes.

And yes, we'll have to check what the downloads look like (and the API responses themselves).  It would be a good opportunity for usability testing on these features! :-)

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -214,3 +214,16 @@ create trigger nml_sched_a_after_trigger after insert or update
...
+    EXECUTE format('INSERT INTO %I VALUES ($1.*)', child_table_name)
+        USING new;
```
ccostino: Do we want to incorporate the `ON CONFLICT` clause here too to account for updates?  Do you think we'll need a separate `BEFORE UPDATE` trigger instead or in addition to that?

data/sql_setup/prepare_schedule_a.sql
ccostino: Another thing about this trigger: we still need to reference the view row corresponding to the record that was inserted, and make sure it is present in the view to begin with.  Otherwise, the row will either be based only on one of the underlying source tables, or potentially made up of almost all `NULL` values if we're not careful when pulling from the view.

webservices/partition/sched_b.py
ccostino: This is what I get for reviewing before I've had ☕️ - I see this is correct now and the trigger names are as they should be, and the function name to execute when they're fired matches as well.

webservices/partition/sched_a.py
ccostino: Disregard!

webservices/partition/sched_a.py
ccostino: Disregard!

ccostino:
I'm going to try and fix up the initial things I found and then get the tests to pass before continuing on with this first round of improvements.

vrajmohan:
Opening a different PR.

PR #2464, opened 2017-06-08
Reduces chance of null coverage end dates in committee totals

jontours:
Places nulls last for `cvg_end_dt`.  Then when we find distinct entries, we aren't left with null coverage end dates.  Addresses https://github.com/18F/FEC/issues/4201.


jontours:
Agreed, there was only one record for this committee where cvg_end_dt wasn't null, so could be a one-off issue, or indicator of some larger problem.

PR #2469, opened 2017-06-09
Add F5 filings to raw IE endpoint

jontours:
This needs some rethinking, it added about 300000 records which is proving problematic given the join on `ofec_filings_mv` and efile reports.  Will see if some indices can help, but currently the query takes about ~7 seconds.


jontours:
Weird; so I check out this PR this morning, run the query and it runs under a second.  So......good to go I guess?

LindsayYoung:
It is probably fine for now, if you are not seeing awful delays in the search. We have to switch to f item soon anyways.

LindsayYoung:
Glad I am not the only one who does that

jontours:
Ok, I've got the query down to about 2-3 seconds, which when using isn't all that noticeable.  Gonna push what I have and call it good, don't want to burn anymore time on this unless the performance is noticeably bad by users.

jontours:
Will someone rid me of these meddlesome tests?  working it.

PR #2470, opened 2017-06-09
Reference correct column for loans

jontours:
Ready for review after update schemas is finished and totals verified.


jontours:
Totals are verified correct on dev. https://fec-dev-proxy.app.cloud.gov/data/committee/C00610113/?cycle=2016

LindsayYoung:
Thanks Jon! 
I double-verified with the crosswalk an history that it had been that way in the previous table.

jontours:
Thanks @LindsayYoung, these loans are difficult.  If was to game the system...

PR #2471, opened 2017-06-09
Upgrade Celery components and settings/configuration

LindsayYoung:
We've been having a few issues with the downloads recently and have pinned down the issue to be a couple of things:

* Not enough disk space allocated in the Celery apps that support downloads
* Out of date dependencies

We need to upgrade Celery and its related dependencies and allocate more disk space to the apps to provide an immediate fix to the problem.  A more long-term solution would be looking into writing download output directly to the S3 buckets if possible (this would be a separate issue and PR).


LindsayYoung:
Related to #2471

LindsayYoung:
This upgrades celery once and doubles disk space in prod.

manifests/manifest_celery_worker_prod.yml
```diff
@@ -6,6 +6,7 @@ applications:
   - name: celery-worker
     instances: 1
     memory: 4G
+    disk_quota: 2G
```
ccostino: Is this a bump up from what we have now?  It defaults to 1G, right?

webservices/tasks/utils.py
```diff
@@ -12,7 +12,7 @@ def get_bucket():
     session = boto3.Session(
         aws_access_key_id=env.get_credential('access_key_id'),
         aws_secret_access_key=env.get_credential('secret_access_key'),
-        region_name=env.get_credential('region')
+        region_name=env.get_credential('region'),
```
ccostino: Nit-pic:  extra comma at the end of function argument list?

webservices/tasks/__init__.py
```diff
@@ -45,8 +46,17 @@ def redis_url():
...
+    acks_late=True,
 )
 
+app.conf.ONCE = {
+    'backend': 'celery_once.backends.Redis',
+    'settings': {
+        'url': redis_url(),
+        'default_timeout': 60 * 60
+    }
+}
```
ccostino: Just to make sure I understand correctly, this is different from the `app.conf.update` call above, right?  I'm assuming this is for Celery Once specifically, but it also appears that the original update above is making referencing to Celery Once defaults, which is part of what's confusing me.

LindsayYoung: That was because of a syntax change in upgrading celery once, I read the docs a while ago so I am not remembering the details.

webservices/tasks/__init__.py
```diff
@@ -45,8 +46,17 @@ def redis_url():
         'webservices.tasks.legal_docs',
     ),
     CELERYBEAT_SCHEDULE=schedule,
+    acks_late=True,
```
ccostino: Should this setting be [`tasks_acks_late`](http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-task_acks_late) instead?

LindsayYoung: Yes, great catch

manifests/manifest_celery_worker_prod.yml
LindsayYoung: yea it defaulted to 1G

LindsayYoung:
We are going to do this in 2 phases, this first and a full upgrade of celery after

webservices/tasks/utils.py
LindsayYoung: Style comments always welcome. Thanks

LindsayYoung:
bre

PR #2483, opened 2017-06-15
Swap out old fec_vsum views for new fec_fitem views

ccostino:
This changeset swaps out our references to the older `fec_vsum`-based views to the new `fec_fitem`-based views.  We are doing this to help improve the accuracy of the reported data based on research into the results currently shown on the site.

**NOTE:**

These views are currently available only in the `dev` database instance, so we'll have to be careful of timing and make sure they are made available in `stage` and `prod` before releasing.

Furthermore, because these have an impact on all tables and materialized views that touch the itemized schedule data (not just A and B), we'll have to update aggregates, repartition, and update schemas at the least, if not just run a full `update_all`.


data/sql_setup/prepare_schedule_e.sql
```diff
@@ -195,30 +195,30 @@ end
...
         return old;
     end if;
 end
 $$ language plpgsql;
 
-drop trigger if exists ofec_sched_e_queue_trigger on fec_vsum_sched_e_vw;
+drop trigger if exists ofec_sched_e_queue_trigger on fec_fitem_sched_e_vw;
 create trigger ofec_sched_e_queue_trigger instead of insert or update or delete
-    on fec_vsum_sched_e_vw for each row execute procedure ofec_sched_e_update_queues()
+    on fec_fitem_sched_e_vw for each row execute procedure ofec_sched_e_update_queues()
```
LindsayYoung: Noting this fix is coming

Issue #2476, opened 2017-06-13
Bug: Alabama senate special election missing


noahmanger:
Currently the Alabama special senate election is missing from the election lookup data: https://www.fec.gov/data/elections/?cycle=2018&state=AL

noahmanger:
I'm pretty sure this is an API/data issue. The API call that the tool makes is this: https://api-stage.open.fec.gov/v1/elections/search/?api_key=DEMO_KEY&cycle=2018&state=AL

There are no senate results in that.

On the client side we also call the `/dates/` endpoint to look for any upcoming election dates, and if there are any for state/office/district combinations that match those provided by the `/elections/` endpoint, then we use the new date, but this requires that seat showing up as having an election in the first place.

cc @LindsayYoung

LindsayYoung:
The [election date model](https://github.com/18F/openFEC/blob/f234a1833b3561da525017ad49ac657c55335780/webservices/common/models/dates.py#L46) takes the data directly from `fecapp.trc_election` 

The data is not in the table. The following query yields no results:
```
select * from fecapp.trc_election where election_state='AL' and election_yr=2017 order by election_date desc;
```
These are the elections we have for the year in that table:

 election_state | office_sought | election_district |       election_notes       | election_date 
--|--|--|--|--
 GA             | H             | 06                | Rep. Tom Price's Seat      | 2017-06-20
 SC             | H             | 05                | Rep. Mick Mulvaney's Seat. | 2017-06-20
 CA             | H             | 34                | Rep. Xavier Becerra's Seat | 2017-06-06
 MT             | H             | 00                | Rep. Ryan Zinke's Seat     | 2017-05-25
 SC             | H             | 05                | Rep. Mick Mulvaney's Seat. | 2017-05-16
 SC             | H             | 05                | Rep. Mick Mulvaney's Seat. | 2017-05-02
 GA             | H             | 06                | Rep. Tom Price's Seat      | 2017-04-18
 KS             | H             | 04                | Rep. Mike Pompeo's Seat    | 2017-04-11
 CA             | H             | 34                | Rep. Xavier Becerra's Seat | 2017-04-04
 KS             | H             | 04                | Rep. Mike Pompeo's Seat    | 2017-02-11
 KS             |               | 04                | Rep. Mike Pompeo's Seat    | 2017-02-11
 KS             | H             | 04                | Rep. Mike Pompeo's Seat    | 2017-02-09
 GA             | S             |                   |                            | 2017-01-10
[...]

PR #2485, opened 2017-06-15
Create elections list and view from election dates

LindsayYoung:
Bases the election year off of dates.
We were missing special elections before.
We can't use the candidate history directly because people file for the wrong election.
The date data starts in 2003.


LindsayYoung:
closes #2476

noahmanger:
Will we need to make any front-end changes for this?

LindsayYoung:
The only change should be limiting the year select, I will test it with the webapp tomorrow

LindsayYoung:
I am making better tables, there is still a disconnect between that and the results, that I am working on

LindsayYoung:
We don't have Senate class data in the elections data so that makes it so that we can't  figure out the senate incumbent for an upcoming special election with this data

LindsayYoung:
Got it working better for Senate but need to circle back on presidential-  seeing duplicates

data/sql_updates/election_outcome.sql
```diff
@@ -1,45 +1,160 @@
...
+    where office_sought != 'P'
+        group by
+        election_state,
+        office_sought,
+        election_district,
+        election_yr
+),
+election_dates_president as(
+    select distinct on (election_yr)
+        'US'::char(2) as election_state,
```
ccostino: I'm sure this is okay, but just wanted to double check - substituting in `US` here as a state abbreviation won't have an adverse effect on the data returned or any processing that takes place leveraging this view as the source, right?

data/sql_updates/election_outcome.sql
```diff
@@ -1,45 +1,160 @@
...
-union all
-(
-    select distinct on (cand_office_st, cand_office_district, cand_election_yr)
-        cand_election_yr - 2 as election_yr, *
-    from disclosure.cand_valid_fec_yr
+),
+-- This creates election records and the incumbent info if applicable.
+-- We need to do this because a null in the elections date district data
+-- means all the districts in the state but the district is listed,
+-- it only applies to one district.
```
ccostino: Thank you for all of these comments describing what each subset of data represents, it's a huge help! :-)

data/sql_updates/election_outcome.sql
```diff
@@ -1,45 +1,160 @@
...
+        ed.election_district as cand_office_district,
+        fec.cand_pty_affiliation,
+        fec.cand_name,
+        ed.election_yr::numeric as election_yr
+    from election_dates_house_senate ed
+    left join
+        incumbent_info fec on
+        ed.election_state = fec.cand_office_st and
+        ed.office_sought = fec.cand_office and
+        -- in the date table the district is 3 characters
```
ccostino: And thank you for noting data oddities/inconsistencies as well!

data/sql_updates/election_outcome.sql
ccostino: We just caught up with an in-person chat and this mirrors the output from before, so not only is it okay it's expected and other code is set to support it already.  We're good!

Issue #2492, opened 2017-06-16
24- and 48-hour Independent Expenditure transactions seem to be missing


PaulClark2:
24- and 48-hour Independent Expenditure transactions seem to be missing from web-app results

Returns no results:
https://www.fec.gov/data/independent-expenditures/?data_type=processed&committee_id=C00053553&cycle=2018&is_notice=true&candidate_id=H8GA06195&max_date=06%2F16%2F2017

Returns 6 rows:
`select cmte_id, pye_nm, s_o_cand_id, s_o_cand_nm, exp_amt, exp_dt, dissem_dt from public.fec_sched_e_notice_vw where cmte_id = 'C00053553' and (exp_dt between '01-JAN-2017' and '16-JUN-2017' or dissem_dt between '01-JAN-2017' and '16-JUN-2017') and s_o_cand_nm like '%OSSOFF%';`

noahmanger:
Could this be related to https://github.com/18F/FEC/issues/4246 ?

PaulClark2:
Yes. I was just going to send you a message on Slack.

jontours:
As far as these go the root problem is that none of the records in this queue are being written out to the ofec_sched_e table: 

This query,  select count(*) from ofec_nml_24_queue_new where s_o_cand_id = 'H8GA06195'; 

Has 170 rows.  Can investigate why those aren't getting processed.

ccostino:
Hmm, so could be two issues then - making sure that the records in the queues are being processed properly and that they're being added to the queues in the first place (as we've noted, the triggers are currently running against the views, which wouldn't work).

jontours:
Disregard; there is a problem with the queues but it seems to not be related, as the 6 records from up above aren't in the queues.  Which, is weird.

ccostino:
Okay, @jontours and I think we have this worked out via the trigger updates we'll need to make to get the queue processing working correctly again.

ccostino:
@LindsayYoung created a hotfix to rebuild schedule E on a nightly basis while we work on fixing the update mechanisms, especially since we need the `fitem`-based views in staging and production before they can be released.

ccostino:
We've also addressed this with another hotfix in the web app: https://github.com/18F/openFEC-web-app/pull/2140

noahmanger:
[...]

PR #2495, opened 2017-06-16
Update Schedule A, B, and E setup to have proper triggers

ccostino:
Addresses #2492 

This changeset fixes several issues with the triggers we currently have on the Schedule A, B, and E tables (Schedule E in particular).  We need to be watching for updates against the `disclosure.f_item_receipt_or_exp` table in addition to the corresponding `disclosure.nml_sched_x` tables, and we also need to make sure that we are pulling the record information from the corresponding `fec_fitem_sched_x_vw` views as well.

The checks for Schedule A and B have been updated to account for new triggers on the `disclosure.f_item_receipt_or_exp` table by checking just for the `rpt_yr` now, which was a recommendation given by the FEC DB team.  This column is now coming from a single source of truth, the `disclosure.f_item_receipt_or_exp` table (which is what is driving the `rpt_yr` column in all of the `fec_fitem_sched_x_vw` views), so it has allowed us to simplify these calculations significantly.

In addition to ensuring no missed Schedule A and B updates, this should also take care of the remaining issues we have had around Schedule E.

Lastly, this PR also accounts for index renaming on Schedule E so in the event of recreating everything; it won't wipe out what is there while things are running (same issue we ran into with Schedule A and B before).

**TODO**

* [x] Update test subset data to include some semblance of `disclosure.f_item_receipt_or_exp`
* [x] Fix any other remaining issues
* [ ] Test in `dev` environment (run a headless app with this branch and perform an `update_all`)
    (app is running but testing has not started)

Absolutely enormous thank you to @jontours and @LindsayYoung for all of the help and extra 👀 all over this!


data/sql_setup/prepare_schedule_e.sql
```diff
@@ -10,113 +10,46 @@ select
...
-select cmte_id, pye_nm, payee_l_nm, payee_f_nm, payee_m_nm, payee_prefix, payee_suffix,pye_st1, pye_st2, pye_city, pye_st,
-    pye_zip, entity_tp, entity_tp_desc, exp_desc, catg_cd, catg_cd_desc, s_o_cand_id, s_o_cand_nm, s_o_cand_nm_first,
-    s_o_cand_nm_last, s_o_cand_m_nm, s_o_cand_prefix, s_o_cand_suffix, s_o_cand_office, s_o_cand_office_desc,
-    s_o_cand_office_st, s_o_cand_office_st_desc, s_o_cand_office_district, s_o_ind, s_o_ind_desc, election_tp,
-    fec_election_tp_desc, cal_ytd_ofc_sought, exp_amt, exp_dt, exp_tp, exp_tp_desc, conduit_cmte_id, conduit_cmte_nm,
-    conduit_cmte_st1, conduit_cmte_st2, conduit_cmte_city, conduit_cmte_st, conduit_cmte_zip, action_cd, action_cd_desc,
-    tran_id, schedule_type, schedule_type_desc, image_num, file_num, link_id, orig_sub_id, sub_id, filing_form,
-    rpt_tp, rpt_yr, cycle, cast(null as timestamp) as TIMESTAMP, image_pdf_url(image_num) as pdf_url, True,
-    to_tsvector(pye_nm)
-from fec_sched_e_notice_vw;
```
jontours: I think we may have gotten a little delete happy; this view has all the `F24` form types, so I think this code has to stay in (all code related to it).  Or else we will miss the `F24` updates (the only form types not covered by the f_item table).  I'll kick this back to you and then review more from there.

ccostino: Ah okay, thanks for flagging this!  I'll add it back in and we'll continue. :-)

ccostino:
Thank you @jontours and @LindsayYoung for the additional help and guidance on this!

data/sql_incremental_aggregates/update_schedule_e.sql
```diff
@@ -1,274 +0,0 @@
-create or replace function ofec_sched_e_update() returns void as $$
```
LindsayYoung: A+ for the idea of ditching this process

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -96,7 +96,14 @@ begin
         -- visit here:
         -- https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS
         if FOUND then
-            two_year_transaction_period = get_transaction_year(new.contb_receipt_dt, view_row.rpt_yr);
+            -- Requires that the hstore extension is installed.
+            if hstore(new) ? 'contb_receipt_dt' then
+                contb_receipt_dt = new.contb_receipt_dt;
+            else
+                contb_receipt_dt = cast(null as timestamp);
```
ccostino: @LindsayYoung:  Instead of just ignoring the value here (and in the other areas where this fix has been made), should we leverage the `rpt_yr` value in the `disclosure.f_item_receipt_or_exp` table instead and use that before falling back to the row in the view?

LindsayYoung: Let' s do a quick query to make sure it is dependable- it should be, but I would still confirm that in the data

ccostino: Well one thing is it's a `numeric` type and not a `timestamp`, d'oh.

jontours: One day we can sit around the campfire and I can tell you a yarn about how much heartache this issue gave me with the totals PR.

LindsayYoung: giphy-tumblr

ccostino: Actually this may not be an issue - the view pulls `rpt_yr` from the underlying `f_item` table anyway, so passing `null` in for the first argument will just force it to use that value regardless.  This should be fine then and keeps things simple. :-)

LindsayYoung:
Thanks for all the work that went into this @ccostino

PR #2497, opened 2017-06-19
Feature/sched f column refactor

jontours:
Fixes the column conflict with `update_schemas` fo schedule f.


LindsayYoung:
Thanks for fixing this!

PR #2498, opened 2017-06-21
Feature/use trigger for partition updates with clean history

vrajmohan:


data/functions/itemized.sql
```diff
@@ -16,3 +124,22 @@ begin
...
 $$ language plpgsql;
+
+CREATE OR REPLACE FUNCTION rename_table_cascade(table_name TEXT) RETURNS VOID AS $$
+DECLARE
+    child_tmp_table_name TEXT;
+    child_tables_cursor CURSOR (parent TEXT) FOR
+        SELECT c.oid::pg_catalog.regclass AS name
+        FROM pg_catalog.pg_class c, pg_catalog.pg_inherits i
+        WHERE c.oid=i.inhrelid
+            AND i.inhparent = (SELECT oid from pg_catalog.pg_class rc where relname = parent);
```
ccostino: @vrajmohan I have a sense of what's going on here but am not as familiar with Postgres' system tables/objects.  The general setup here to grab all child tables that share the same base name as the given parent table and setup the result of that as a cursor to then be looped through below, right?

data/functions/itemized.sql
```diff
@@ -16,3 +124,22 @@ begin
...
+BEGIN
+    EXECUTE format('DROP TABLE IF EXISTS %I CASCADE', table_name);
+    EXECUTE format('ALTER TABLE %1$I_tmp RENAME TO %1$I', table_name);
```
ccostino: In the first statement there is `%I`, which means "treat the interpolated value as a SQL identifier." [[source](https://www.postgresql.org/docs/current/static/functions-string.html#FUNCTIONS-STRING-FORMAT)]

In the second statement, you have `%1$I` and only one argument to the `format` function - what is this interpolating?  It appears to be the same thing, which is why I'm confused that it's not just `%I`. :-)

Based on the lines just below this, it looks like you're just referring to the positional argument explicitly but telling it to still treat it as a SQL identifier?  If so that's fine, I just want to make sure I understand!

data/functions/itemized.sql
```diff
@@ -16,3 +124,22 @@ begin
...
+
+    FOR child_tmp_table IN child_tables_cursor(table_name) LOOP
+        child_tmp_table_name = child_tmp_table.name;
+        EXECUTE format('ALTER TABLE %1$I RENAME TO %2$I', child_tmp_table_name, regexp_replace(child_tmp_table_name, '_tmp$', ''));
+    END LOOP;
+END
+$$ LANGUAGE plpgsql;
```
ccostino: I think I had originally envisioned this as being a simpler implementation - we more or less just hard code the master table and child table names in the creation of the partition and just substitute the "a" and "b" portions in as needed, and I saw the renaming of them at the end working in a similar way in my mind's eye.

That said, I see that this can be used for any partition setup that requires the renaming to support a zero-down-time update, which is awesome! :-)

data/functions/utils.sql
```diff
@@ -72,3 +72,14 @@ begin
...
+CREATE OR REPLACE FUNCTION get_partition_suffix(year NUMERIC)
+RETURNS TEXT AS $$
+BEGIN
+    IF year % 2 = 0 THEN
+        RETURN (year - 1)::TEXT || '_' || year::TEXT;
+    ELSE
+        RETURN year::TEXT || '_' || (year + 1)::TEXT;
+    END IF;
+END
+$$ LANGUAGE PLPGSQL IMMUTABLE;
```
ccostino: I've always wondered about this:  what is the purpose of the `IMMUTABLE` keyword at the end and why use it here but not elsewhere?

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,177 @@
...
+    contributor_employer_text   TSVECTOR,
+    contributor_occupation_text TSVECTOR,
+    is_individual               BOOLEAN,
+    clean_contbr_id             VARCHAR(9),
+    two_year_transaction_period SMALLINT
+);
+
+-- Create the child tables.
+SELECT create_itemized_schedule_partition('a', 1978, 2018);
+SELECT finalize_itemized_schedule_a_tables(1978, 2018);
```
ccostino: I forgot to make a note of this and change it before.  We may want to shift these to environment variables that are made available to the SQL environment, much like `START_YEAR` and `START_YEAR_AGGREGATE` are.  These are managed in the [application config](https://github.com/18F/openFEC/blob/develop/webservices/config.py#L23).

I can also see how this would be confusing though, as it took me a while to remember how these were setup.  Perhaps there's a better way?  Totally fine if that's a separate story for the future, too, and we just add these values in for now.

data/sql_partition/partition_schedule_b.sql
```diff
@@ -0,0 +1,94 @@
...
+    recipient_name_text           TSVECTOR,
+    disbursement_description_text TSVECTOR,
+    disbursement_purpose_category TEXT,
+    clean_recipient_cmte_id       VARCHAR(9),
+    two_year_transaction_period   SMALLINT
+);
+
+-- Create the child tables.
+SELECT create_itemized_schedule_partition('b', 1978, 2018);
+SELECT finalize_itemized_schedule_b_tables(1978, 2018);
```
ccostino: Same note about the start and end dates here as in schedule A.

manage.py
```diff
@@ -290,33 +290,13 @@ def update_itemized(schedule):
     logger.info('Finished Schedule {0} update.'.format(schedule))
 
 @manager.command
-def partition_itemized():
-    """This command runs the partitioning against the larger itemized
-    schedule tables.
+def partition_itemized(schedule):
+    """This command runs the partitioning against the specified itemized
+    schedule table.
```
ccostino: Thank you for cleaning this up! :-)

tests/test_integration.py
```diff
@@ -88,15 +88,17 @@ def test_update_schemas(self):
         #models.CaniddateCommitteeTotalsHouseSenate passed integration, so testing this specific
         #model really isn't expanding code coverage.  I can try and get the model passing but it's proving
         #to be difficult considering the joins needed (and our limited test subset)
-        whitelist = [models.EntityReceiptDisbursementTotals]
+        whitelist = [
+            models.EntityReceiptDisbursementTotals,
+            models.itemized.ScheduleA,
+            models.itemized.ScheduleB,
+        ]
```
ccostino: As discussed, totally appropriate.  These models are not managed by the `update_schema` process.

webservices/common/models/itemized.py
```diff
@@ -44,11 +40,6 @@ class BaseRawItemized(db.Model):
     back_reference_schedule_name = db.Column('br_sname', db.String)
     load_timestamp = db.Column('create_dt', db.DateTime)
 
-
-    @hybrid_property
-    def report_type(self):
-        return self.filing.form_type
-
```
ccostino: Oh wow, I didn't realize the method was defined twice in there, d'oh!  Thanks for fixing!

webservices/partition/base.py
```diff
@@ -101,6 +101,7 @@ def run(cls):
             cls.create_child(parent, cycle)
 
         cls.rename()
+        cls.create_trigger()
```
ccostino: Should this method (meaning the `run` method this line is a part of) change more?  If we're now creating the partition in the SQL side, there's no need to create it here.  In other words, by the time this code is run, the master and child tables in the partition should have already been created and setup.  The only thing not done is adding these new triggers, which perhaps we could also shift to the SQL side.

This would mean that other methods in this class, e.g., `create_master`, `create_child`, and `create_indexes`, would go away as well.

What do you think?

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,177 @@
...
+
+-- Create the insert trigger so that records go into the proper child table.
+DROP TRIGGER IF EXISTS insert_sched_a_trigger_tmp ON ofec_sched_a_master_tmp;
+CREATE trigger insert_sched_a_trigger_tmp BEFORE INSERT ON ofec_sched_a_master_tmp FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_a_');
+
+SELECT rename_table_cascade('ofec_sched_a_master');
```
ccostino: We may not want to do this rename just yet because the partition is still fresh and hasn't been populated with data yet.

At this point we'd I think we want to do the following:

* Insert all of the records into the appropriate child tables
* Apply all of the indexes by calling to the `finalize_itemized_schedule_a_tables ` method (because having the indexes on before the records are added will slow down insertion)
* Rename everything in the partition:  master table, child tables, and indexes, which there is currently no method for

Does this sound right to you?

data/sql_partition/partition_schedule_b.sql
```diff
@@ -0,0 +1,94 @@
...
+
+-- Create the insert trigger so that records go into the proper child table.
+DROP TRIGGER IF EXISTS insert_sched_b_trigger_tmp ON ofec_sched_b_master_tmp;
+CREATE trigger insert_sched_b_trigger_tmp BEFORE INSERT ON ofec_sched_b_master_tmp FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_b_');
+
+SELECT rename_table_cascade('ofec_sched_b_master');
```
ccostino: Same note about this step in the process as schedule A.

webservices/partition/base.py
```diff
@@ -203,8 +195,7 @@ def create_constraints(cls, child, cycle, temp=True):
             'alter table {child} alter column {primary} set not null',
             'alter table {child} add primary key ({primary})',
             'alter table {child} alter column filing_form set not null',
-            'alter table {child} add constraint check_two_year_transaction_period check (two_year_transaction_period in ({start}, {stop}))',  # noqa
-            'alter table {child} inherit {master}'
+            'alter table {child} add constraint check_two_year_transaction_period check (two_year_transaction_period in ({start}, {stop}))'  # noqa
```
ccostino: The constraint shouldn't need to be defined here as it should already been on the child table given the SQL changes made above.  Also, see note above about removing this entire method (`create_child`).

webservices/partition/base.py
```diff
@@ -316,6 +277,7 @@ def refresh_child(cls, cycle, queue_old, queue_new):
             # populating the child tables during normal partitioning.
             # Otherwise, an error is thrown due more values being specified
             # than there are columns to accept them.
+            # VM TODO - Is this still needed?
```
ccostino: I thought about this some more and I think it is still needed.  The queue tables aren't going away with these changes, and they're setup with the `two_year_transaction_period` column in them.  By having that column be a part of the record, the column processing that takes place in here will run into a conflict and halt because it will expect it to not be there.

I'm sorry if this is confusing, I'm happy to cover in more detail if you'd like!

webservices/partition/sched_a.py
```diff
@@ -58,31 +58,38 @@ def column_factory(cls, parent):
...
+#            sa.Index('ix_{0}_sub_id_date_tmp'.format(child.name[:-4]), 'contb_receipt_dt', 'sub_id'),
+#
+#            sa.Index('ix_{0}_cmte_id_tmp'.format(child.name[:-4]), 'cmte_id', 'sub_id'),
+#            sa.Index('ix_{0}_cmte_id_amount_tmp'.format(child.name[:-4]), 'cmte_id', 'contb_receipt_amt', 'sub_id'),
+#            sa.Index('ix_{0}_cmte_id_date_tmp'.format(child.name[:-4]), 'cmte_id', 'contb_receipt_dt', 'sub_id'),
+#
+#            sa.Index(None, 'contributor_name_text', postgresql_using='gin'),
+#            sa.Index(None, 'contributor_employer_text', postgresql_using='gin'),
+#            sa.Index(None, 'contributor_occupation_text', postgresql_using='gin'),
+#        ]
```
ccostino: Oh, I see what you did here.  Okay, this makes sense for the interim, but I still think we can go further based on my notes above. :-)

webservices/partition/sched_a.py
```diff
@@ -58,31 +58,38 @@ def column_factory(cls, parent):
...
     def update_child(cls, child):
         cmd = 'alter table {0} alter column contbr_st set statistics 1000'.format(child.name)
         db.engine.execute(cmd)
+
+    @classmethod
+    def create_trigger(cls):
+        db.engine.execute('DROP TRIGGER IF EXISTS insert_sched_a_trigger ON ofec_sched_a_master')
+        db.engine.execute('''
+            CREATE trigger insert_sched_a_trigger BEFORE INSERT ON ofec_sched_a_master FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_a_');
+            ''')
```
ccostino: As noted above, if we're can shift the other parts of setting up the partitions on the SQL side, could we also move this to the SQL side as one of the final steps?

webservices/partition/sched_b.py
```diff
@@ -53,28 +53,36 @@ def column_factory(cls, parent):
...
         cmd = 'alter table {0} alter column recipient_st set statistics 1000'.format(child.name)
         db.engine.execute(cmd)
+
+    @classmethod
+    def create_trigger(cls):
+        db.engine.execute('DROP TRIGGER IF EXISTS insert_sched_b_trigger ON ofec_sched_b_master')
+        db.engine.execute('''
+            CREATE trigger insert_sched_b_trigger BEFORE INSERT ON ofec_sched_b_master
+                FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_b_');
+            ''')
```
ccostino: Same notes in this file as in the Schedule A version.

data/functions/itemized.sql
vrajmohan: Correct.

data/functions/itemized.sql
vrajmohan: `%1$I` allows you to use the same positional parameter (in this case, the first) for multiple substitutions. Otherwise, I would have had to write:
```
EXECUTE format('ALTER TABLE %I_tmp RENAME TO %I', table_name, table_name);
```

data/functions/utils.sql
vrajmohan: It marks it as a `pure` function - that it will always give the same result for the same input, allowing it to be cached/memoized.

data/sql_partition/partition_schedule_a.sql
vrajmohan: Agreed, we should use a similar strategy to `START_YEAR and START_YEAR_AGGREGATE`

data/sql_partition/partition_schedule_a.sql
vrajmohan: That sounds right to me. I commented out the population as it was failing and was not needed  in the `test` environment.

webservices/partition/base.py
vrajmohan: Agreed, I was going to add these as TODOs, but it slipped my mind.

webservices/partition/base.py
vrajmohan: Right, most of this module may go away.

webservices/partition/sched_a.py
vrajmohan: Agreed.

data/functions/itemized.sql
ccostino: Ah, I see, thank you!  TIL. :-)

data/functions/utils.sql
ccostino: Thank you!  TIL two things!

Now I wonder if there are other spots where we should be using this and are not currently. :-)

webservices/partition/base.py
vrajmohan: Why not remove `two_year_transaction_period`from the output of `column_factory`?

ccostino: Remove the `two_year_transaction_period` column from the `column_factory`...  You may be right!  With the partition creation happening on the SQL side now, we don't need to keep track of it anymore.  Just the ones that the queues don't already have we still need for the time being.

data/functions/itemized.sql
```diff
@@ -1,3 +1,111 @@
...
+        EXECUTE format('CREATE INDEX idx_%s_is_individual_tmp ON %I (is_individual)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_clean_contbr_id_tmp ON %I (clean_contbr_id)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_two_year_transaction_period_tmp ON %I (two_year_transaction_period)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_sub_id_amount_tmp ON %I (contb_receipt_amt, sub_id)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_sub_id_date_tmp ON %I (contb_receipt_dt, sub_id)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_sub_id_cmte_id_tmp ON %I (cmte_id, sub_id)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_sub_id_cmte_id_amount_tmp ON %I (cmte_id, contb_receipt_amt, sub_id)', child_table_root, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_sub_id_cmte_id_date_tmp ON %I (cmte_id, contb_receipt_dt, sub_id)', child_table_root, child_table_name);
```
ccostino: @anthonygarvan and @vrajmohan:  In thinking about the DB performance research that was done and what the next steps will be, do you think it makes sense to add other multi-column indexes here (and for schedule B) to support searching across child tables in a partition?  I ask because it'll reduce the amount of times we have to repartition if we can work this in now, but feel free to push back if that's biting off more than we should chew here.

Regardless, is it just a matter of setting up all permutations of the columns we allow folks to filter or search on combined with the columns we allow them to sort by, lastly (and always) followed by `sub_id`?  And how do the `TSVECTOR` columns fit into this?

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,259 @@
...
+
+-- Create the child tables.
+SELECT create_itemized_schedule_partition('a', :PARTITION_START_YEAR, :PARTITION_END_YEAR);
+SELECT finalize_itemized_schedule_a_tables(:PARTITION_START_YEAR, :PARTITION_END_YEAR);
+
+-- Create the insert trigger so that records go into the proper child table.
+DROP TRIGGER IF EXISTS insert_sched_a_trigger_tmp ON ofec_sched_a_master_tmp;
+CREATE trigger insert_sched_a_trigger_tmp BEFORE INSERT ON ofec_sched_a_master_tmp FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_a_');
+
+CREATE VIEW fec_vsum_sched_a_vw AS
```
ccostino: Ah, we shouldn't be creating these here, nor do we need this view anymore.

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,259 @@
...
+    line_number_label           TEXT
+);
+
+-- Create the child tables.
+SELECT create_itemized_schedule_partition('a', :PARTITION_START_YEAR, :PARTITION_END_YEAR);
+SELECT finalize_itemized_schedule_a_tables(:PARTITION_START_YEAR, :PARTITION_END_YEAR);
+
+-- Create the insert trigger so that records go into the proper child table.
+DROP TRIGGER IF EXISTS insert_sched_a_trigger_tmp ON ofec_sched_a_master_tmp;
+CREATE trigger insert_sched_a_trigger_tmp BEFORE INSERT ON ofec_sched_a_master_tmp FOR EACH ROW EXECUTE PROCEDURE insert_sched_master('ofec_sched_a_');
```
ccostino: Shouldn't we do the full insert into the tables before the indexes are created to help speed up inserting the records into the partition and ensure the indexes are fresh with the new data?

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,259 @@
...
+    CURRENT_TIMESTAMP as timestamp,
+    CURRENT_TIMESTAMP as pg_date,
+    image_pdf_url(image_num) AS pdf_url,
+    to_tsvector(concat(contbr_nm, ' ', contbr_id)) AS contributor_name_text,
+    to_tsvector(contbr_employer) AS contributor_employer_text,
+    to_tsvector(contbr_occupation) AS contributor_occupation_text,
+    is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text) AS is_individual,
+    clean_repeated(contbr_id, cmte_id) AS clean_contbr_id,
+    get_transaction_year(contb_receipt_dt, rpt_yr) AS two_year_transaction_period
+FROM fec_vsum_sched_a_vw;
```
ccostino: We want to insert from the existing `fec_fitem_sched_a_vw` view, which is created and managed by the FEC.  I'm sorry for the confusion!

data/sql_partition/partition_schedule_a.sql
```diff
@@ -0,0 +1,259 @@
...
+
+SELECT rename_table_cascade('ofec_sched_a_master');
```
ccostino: Building on a note from before - should we finalize the tables (create the indexes) here, just before doing the renaming?

data/sql_partition/partition_schedule_b.sql
ccostino: Flagging this again for the same remarks as schedule A as I know this part is still being worked on.

vrajmohan:
Thanks, @ccostino. All review comments addressed.

data/functions/itemized.sql
```diff
@@ -15,3 +121,22 @@ begin
     perform retry_processing_schedule_b_records(:START_YEAR_AGGREGATE);
 end
 $$ language plpgsql;
+
+CREATE OR REPLACE FUNCTION rename_table_cascade(table_name TEXT) RETURNS VOID AS $$
```
ccostino: I realized we are missing one more thing in this - renaming the indexes and constraints after the tables are renamed.  I've made another note below where this happened on the Python side for reference.

Other constraints, such as the `CHECK` constraint for the child tables and any primary key constraints, should also be taken into account to ensure that the exist if and when a repartition has to be run so we still maintain zero down-time and disruption in API availability and performance.

webservices/partition/base.py
```diff
@@ -142,186 +127,25 @@ def get_child_name(cls, cycle):
...
-            child.primary_key.name.replace('_tmp', '')
-        )
-        db.engine.execute(cmd)
-
-        # Rename child table indexes
-        for index in child.indexes:
-            cmd = 'alter index {0} rename to {1}'.format(
-                index.name, index.name.replace('_tmp', '')
-            )
-            db.engine.execute(cmd)
```
ccostino: This code is what took care of renaming the indexes after all of the tables themselves were renamed - all it does is drop the `_tmp` part at the end of them, but it has to be done for each child table.

vrajmohan: The db dump changes were to add delete_ind column and to add explicit casts for some view columns to reduce `pg_restore` warnings. The view creation was all in SQL, that has been now removed.

data/functions/itemized.sql
vrajmohan: I would recommend _naming_ the indexes and constraints so that _renaming_ is not required - after all these are ephemeral names.

ccostino: I totally agree with you on that!  However, we noticed that when we had the "proper" names to begin with and used the same names when creating the new temp tables in prep for an update, the indexes would get dropped/overwritten during updates.  This in turn would cause massive performance issues due to backed up queries and filled connection pools because the API calls on the schedule data would grind to a halt without the indexes there.

We found that if we named the indexes with `_tmp` in them at first when creating the new tables/materialized views during an update and then renaming them preventing this from happening.  If this is still confusing or unclear let's hop on a chat to talk further!

vrajmohan: Ah, I did not realize that index names need to unique in the schema, not just in the table.

ccostino: Unfortunately yes, lesson learned the hard way... 😆

vrajmohan:
Indexes renamed. I don't think the check constraint (only one of these) needs to be renamed.

ccostino:
Awesome, thanks @vrajmohan!  I wasn't sure on the other constraints but come to think of it we don't rename them now so probably okay. :-)

Looking forward to diving into this tomorrow for final testing!

data/functions/itemized.sql
ccostino: Disregard, we're going to follow up with this separately and not as a part of this PR!

PR #2499, opened 2017-06-22
Add html links to \reports and \filings

jontours:
Just need to hook up api.

### DIFF
--- a/data/functions/urls.sql
+++ b/data/functions/urls.sql
@@ -59,16 +59,17 @@ $$ language plpgsql immutable;
 create or replace function report_html_url(means_filed text, cmte_id text, filing_id text) returns text as $$
 BEGIN
     return CASE
-       when means_filed = 'paper' then format (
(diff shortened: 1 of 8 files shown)


data/functions/urls.sql
```diff
@@ -56,6 +56,24 @@ begin
...
+           'http://docquery.fec.gov/cgi-bin/paper_forms/%1$s/%2$s/',
+            cmte_id,
+            filing_id
+       )
+       when means_filed = 'e-file' then format (
+           'http://docquery.fec.gov/cgi-bin/forms/%1$s/%2$s/',
+            cmte_id,
+            filing_id
+       )
+       else null
```
ccostino: Nice that you were able to incorporate both paper and efiling URLs into here!

jontours: Yup functions really are becoming a fun thing for me here lately.

PR #2502, opened 2017-06-26
Union simplification for Schedule E

jontours:
Argh, test is failing when it really shouldn't.  I'm thinking it's in the test data.   Will update soon.


jontours:
@ccostino we can merge this now if you like, as it's ready to go.

PR #2524, opened 2017-07-12
Feature/add entities with types to ao doc

anthonygarvan:


vrajmohan:
URLs are nice now :-)

PR #2530, opened 2017-07-17
first pass at compound indexes for sched_a and sched_b

anthonygarvan:


ccostino:
All right, let's give this a shot and see how this goes! :-)

PR #2540, opened 2017-07-18
Move add_itemized_partition_cycle feature to SQL

vrajmohan:
Also remove unused code.

### DIFF
--- a/data/functions/itemized.sql
+++ b/data/functions/itemized.sql
@@ -16,6 +16,28 @@ END
 $$ LANGUAGE plpgsql;
 
 
+CREATE OR REPLACE FUNCTION add_partition_cycles(start_year NUMERIC, amount NUMERIC) RETURNS VOID AS $$
+DECLARE
+    end_year NUMERIC = start_year + (amount - 1) * 2;
+    master_table_name TEXT;
+    child_table_name TEXT;
+    schedule TEXT;
+BEGIN
+    FOR cycle IN start_year..end_year BY 2 LOOP
+        FOREACH schedule IN ARRAY ARRAY['a', 'b'] LOOP
+            master_table_name = format('ofec_sched_%s_master', schedule);
+            child_table_name = format('ofec_sched_%s_%s_%s', schedule, cycle - 1, cycle);
+            EXECUTE format('CREATE TABLE %I (
+                CHECK ( two_year_transaction_period in (%s, %s) )
+            ) INHERITS (%I)', child_table_name, cycle - 1, cycle, master_table_name);
+
+        END LOOP;
+    END LOOP;
+    PERFORM finalize_itemized_schedule_a_tables(start_year, end_year, FALSE);
+    PERFORM finalize_itemized_schedule_b_tables(start_year, end_year, FALSE);
+END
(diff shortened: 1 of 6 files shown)


vrajmohan:
If the table exists, you get:
```
INFO:manager:Adding Schedule A and B cycles...
ERROR:manager:Failed to add partition cycles
Traceback (most recent call last):
  File "/Users/radhakrishnanvrajmohan/git/openFEC/.env/lib/python3.5/site-packages/sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "/Users/radhakrishnanvrajmohan/git/openFEC/.env/lib/python3.5/site-packages/sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
psycopg2.ProgrammingError: relation "ofec_sched_a_3003_3004" already exists
CONTEXT:  SQL statement "CREATE TABLE ofec_sched_a_3003_3004 (

                CHECK ( two_year_transaction_period in (3003, 3004) )

            ) INHERITS (ofec_sched_a_master)"
PL/pgSQL function add_partition_cycles(numeric,numeric) line 23 at EXECUTE


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 399, in add_itemized_partition_cycle
    count=amount
  File "/Users/radhakrishnanvrajmohan/git/openFEC/.env/lib/python3.5/site-packages/sqlalchemy/engine/base.py", line 914, in execute
    return meth(self, multiparams, params)
  File "/Users/radhakrishnanvrajmohan/git/openFEC/.env/lib/python3.5/site-packages/sqlalchemy/sql/elements.py", line 323, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/Users/radhakrishnanvrajmohan/git/openFEC/.env/lib/python3.5/site-packages/sqlalchemy/engine/base.py", line 1010, in _execute_clauseelement
    compiled_sql, distilled_params
```

data/functions/itemized.sql
```diff
@@ -16,6 +16,28 @@ END
 $$ LANGUAGE plpgsql;
 
 
+CREATE OR REPLACE FUNCTION add_partition_cycles(start_year NUMERIC, amount NUMERIC) RETURNS VOID AS $$
+DECLARE
+    end_year NUMERIC = start_year + (amount - 1) * 2;
+    master_table_name TEXT;
+    child_table_name TEXT;
+    schedule TEXT;
```
ccostino: One more small thing with this, @vrajmohan:  If the start year given is something like `2015` we want to make sure that it is handled appropriately, which means grabbing the cycle year based on the user input.  So in the case of `2015`, that should become `2016`, which is the cycle end year.

When I used `2015` locally, it created another child table called `ofec_sched_a_2014_2015`, which should not exist as that is not a valid cycle.

We have a [`get_cycle`](https://github.com/18F/openFEC/blob/develop/data/functions/utils.sql#L1) method in the SQL already, it should just be a matter of calling that with the `start_year` argument given and then using that returned value as the `end_year` and using `cycle - 1` as the start year for creating each child table.

Does this make sense?  Happy to hop on a quick chat if not!

PR #2543, opened 2017-07-18
Upgrade Celery to latest 4.0.x release

ccostino:
This changeset updates Celery to the latest current 4.0.x release.  We are doing this to keep our dependencies up to date and to help address some issues we have been having with our downloads recently.

We are following Celery's [4.0 upgrade documentation](http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#upgrading-from-celery-3-1) and making adjustments as necessary to get the new version working.

**TODO**

* [x] Fix serialization to not use Python's `pickle` module and instead allow for Celery 4.0's default of `JSON`

/cc @jmcarp - I won't ask for a formal review, but you're more than welcome to chime in if you'd like! :-)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ mandrill==1.0.57
 icalendar==3.9.1
 GitPython==1.0.1
 gunicorn==19.7.1
-gevent==1.2.1
+gevent==1.2.2
 webargs==0.18.0
 slacker==0.8.6
 ujson==1.33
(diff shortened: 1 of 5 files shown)


vrajmohan:
Looks good to me!

ccostino:
Thanks, @vrajmohan!  Do you mean just converting it to a string beforehand?  I think there'd be more to it than just that given the other query string processing that takes place a little further down the chain but if that'd simplify things I'd definitely like to explore that.

I'll sleep on it and perhaps we could catch up briefly tomorrow afternoon at some point to do a quick walk through to see?

ccostino:
@vrajmohan and I caught up about this some more and while it'd be a nice improvement, it's more work than appropriate for this PR.  We agreed it'd be best to come back to it at a later time, once this upgrade is done, so issue #2549 exists to capture this.

PR #2548, opened 2017-07-20
Fixes html_url property in filing end points

### LINKED ISSUE #2533: Add html_url value for /efile/filings/ and fix missing html_url for /filings
We've received a few comments recently about people looking for the links to the web-based view of filings. While we have a place to include these in the main filings tables, the `html_url` is currently `null` for many filings, though not all. (e.g. https://api.open.fec.gov/v1/efile/filings/?api_key=DEMO_KEY&form_type=F3). Also, it's completely absent from the `/efile/filings/` endpoint. 

(Note: It seems to work fine on the `/reports/` endpoints, which is why it shows up in the details panels. Those are populated by additional calls to that endpoint, whereas the table itself is just the filings endpoint).

**Completion criteria:**
- [ ] Fix missing `html_url` values on `/filings/`
- [ ] Add `html_url` property on `/efile/filings/`

rjayasekera:
Fixes #2533 

This change set fixes the html_url property in the filings end point and adds the property to the efile/filings end point.

**NOTE**

This requires running `update_schemas` in the database.

### DIFF
--- a/data/sql_updates/filings.sql
+++ b/data/sql_updates/filings.sql
@@ -74,10 +74,7 @@ with filings as (
             filing_history.form_tp
         ) as pdf_url,
         means_filed(begin_image_num) as means_filed,
-        case when means_filed(begin_image_num) = 'e-file' then
-          report_html_url(means_filed(begin_image_num), filing_history.cand_cmte_id::text, filing_history.file_num::text)
-          else null::text
-        end as html_url,
+        report_html_url(means_filed(begin_image_num), filing_history.cand_cmte_id::text, filing_history.file_num::text) as html_url,
         report_fec_url(begin_image_num::text, filing_history.file_num::integer) as fec_url,
         amendments.amendment_chain,
         --amendments.prev_file_num as previous_file_number,
@@ -147,10 +144,7 @@ rfai_filings as (
             'RFAI'::text
         ) as pdf_url,
         means_filed(begin_image_num) as means_filed,
-        case when means_filed(begin_image_num) = 'e-file' then
-          report_html_url(means_filed(begin_image_num), id::text, filing_history.file_num::text)
-          else null::text
-        end as html_url,
+        report_html_url(means_filed(begin_image_num), id::text, filing_history.file_num::text) as html_url,
         null::text as fec_url,
         null::numeric[] as amendment_chain,
         null::int as most_recent_file_number,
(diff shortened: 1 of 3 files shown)


ccostino:
@jontours or @LindsayYoung:  We exposed the `html_url` property in the `EFilingsSchema` but we weren't sure if it need to be exposed and added to the `BaseEfileSchema` - do either of you have a sense of whether or not that is needed?

noahmanger:
Just checking: what will the behavior be for paper filings? They'll still have `html_url: null` yes?

jontours:
In these instances of unwinding the Gordian Knot that is schemas.py it's best to map the right Schema object from the endpoint view.

jontours:
@noahmanger for paper it depends, if the file number is less than 1, then yes it will be null.

data/sql_updates/filings.sql
```diff
@@ -74,10 +74,7 @@ with filings as (
             filing_history.form_tp
         ) as pdf_url,
         means_filed(begin_image_num) as means_filed,
-        case when means_filed(begin_image_num) = 'e-file' then
-          report_html_url(means_filed(begin_image_num), filing_history.cand_cmte_id::text, filing_history.file_num::text)
-          else null::text
-        end as html_url,
+        report_html_url(means_filed(begin_image_num), filing_history.cand_cmte_id::text, filing_history.file_num::text) as html_url,
```
jontours: Thanks for cleaning up after my forgetfulness.

webservices/schemas.py
```diff
@@ -699,6 +699,7 @@ def remove_fec_url(self, obj):
     fields={
         'beginning_image_number': ma.fields.Str(),
         'ending_image_number': ma.fields.Str(),
+        'html_url': ma.fields.Str(),
```
jontours: There is a question about if this is the right schema, it is.   The other one that is similarly named are schemas that are used to build the efile endpoints, and shouldn't be referenced directly.  Schemas.py is a bit of a mess.

PaulClark2:
A follow-up on @noahmanger's questions and @jontours's response. In the data, the file numbers of paper filings can be negative. This means we do not yet have a .fec file for that report. In cases when the file number is negative is the return from the function still `html_url: null`?

jontours:
Yes @PaulClark2.

webservices/schemas.py
ccostino: Okay excellent, thanks!  So we shouldn't worry about the other schema at all then?

PR #2550, opened 2017-07-20
State filters

jontours:
First cut at state filters.  Kind of tricky this one, because sometimes filters are applied, and sometimes they aren't.


data/sql_updates/committee_history.sql
```diff
@@ -78,9 +78,9 @@ select distinct on (fec_yr.cmte_id, fec_yr.fec_election_yr)
     f1.cand_pty_tp as party_type,
     f1.cand_pty_tp_desc as party_type_full,
     f1.qual_dt as qualifying_date,
-    dates.first_file_date,
-    dates.last_file_date,
-    dates.last_f1_date,
+    cast(cast(dates.first_file_date as text) as date) as first_file_date,
+    cast(cast(dates.last_file_date as text) as date) as last_file_date,
+    cast(cast(dates.last_f1_date as text) as date) as last_f1_date,
```
ccostino: I'm curious about the double casting taking place here.  Are the column types such that it requires converting the values to text first before an actual date in order to enforce valid values, or something else?

webservices/args.py
```diff
@@ -207,6 +207,8 @@ def make_seek_args(field=fields.Int, description=None):
 candidate_list = {
     'q': fields.List(fields.Str, description=docs.CANDIDATE_NAME),
     'candidate_id': fields.List(IStr, description=docs.CANDIDATE_ID),
+    'min_first_file_date': fields.Date(description='Selects all candidates whose first filing was received by the FEC after this date'),
+    'max_first_file_date': fields.Date(description='Selects all candidates whose first filing was received by the FEC before this date'),
```
ccostino: Thanks for the clear documentation!

webservices/docs.py
```diff
@@ -668,6 +668,10 @@
         - N not yet a candidate\n\
         - P prior candidate\n\
 '
+LAST_F2_DATE = 'The day the FEC received the candidate\'s most recent Form 2'
+FIRST_CANDIDATE_FILE_DATE = 'The day the FEC received the candidate\'s first filing. \
+This is a F2 candidate registration.'
+LAST_CANDIDATE_FILE_DATE = 'The day the FEC received the candidate\'s most recent filing'
```
ccostino: Again, nice to see, thanks for making sure this is a part of this. :-)

webservices/resources/candidates.py
```diff
@@ -86,6 +93,8 @@ def build_query(self, **kwargs):
         if kwargs.get('election_year'):
             query = query.filter(models.Candidate.election_years.overlap(kwargs['election_year']))
 
+        print(query)
+
```
ccostino: Extra `print()` statement and blank line left over from debugging?

data/sql_updates/committee_history.sql
jontours: The columns as in are numeric, and I seem to remember having errors casting numeric straight to a date.  I can try casting directly in my console to see if it will work.

ccostino: We caught up about this and we should be fine here - this is because date conversions require strings to work with.

PR #2552, opened 2017-07-21
fixed boolean operator to make sure both legal and bulk downloads wil…

patphongs:
Needed to change boolean statement to `and` instead of `or`. This will ensure both legal and bulk-downloads will not be deleted from the bucket.

### DIFF
--- a/webservices/tasks/download.py
+++ b/webservices/tasks/download.py
@@ -192,6 +192,6 @@ def export_query(path, qs):
 @app.task
 def clear_bucket():
     for obj in task_utils.get_bucket().objects.all():
-        if not obj.key.startswith('legal') or not obj.key.startswith('bulk-downloads'):
+        if not obj.key.startswith('legal') and not obj.key.startswith('bulk-downloads'):
             obj.delete()
 
--- a/webservices/tasks/download.py
+++ b/webservices/tasks/download.py
@@ -191,7 +191,8 @@ def export_query(path, qs):
 
 @app.task
 def clear_bucket():
+    permanent_dir = ('legal', 'bulk-downloads',)
     for obj in task_utils.get_bucket().objects.all():
-        if not obj.key.startswith('legal') and not obj.key.startswith('bulk-downloads'):
+        if not obj.key.startswith(permanent_dir):
             obj.delete()
 


LindsayYoung:
```
@
```
app.task
def clear_bucket():
    for obj in task_utils.get_bucket().objects.all():
        if obj.key.startswith('legal') or obj.key.startswith('bulk-downloads'):
            pass
        else
            obj.delete()

LindsayYoung:
I think the original was correct, but it is really confusing. I think it might be clearer if we write it like this:
```
@app.task
def clear_bucket():
    for obj in task_utils.get_bucket().objects.all():
        if obj.key.startswith('legal') or obj.key.startswith('bulk-downloads'):
            pass
        else:
            obj.delete()
```

LindsayYoung:
Yes, it should be an and- but there is a lot of inverted logic in there, so I still think we might want to rewrite it to make it less confusing

webservices/tasks/download.py
```diff
@@ -192,6 +192,6 @@ def export_query(path, qs):
 @app.task
 def clear_bucket():
     for obj in task_utils.get_bucket().objects.all():
-        if not obj.key.startswith('legal') or not obj.key.startswith('bulk-downloads'):
+        if not obj.key.startswith('legal') and not obj.key.startswith('bulk-downloads'):
```
ccostino: There's a better way to go about this I think.  Python's [`str.startswith`](https://docs.python.org/3/library/stdtypes.html#str.startswith) method can also take in a `tuple` of strings to compare and if any of them match, it'll return `True`.  Example:

```python
matches = ('legal', 'bulk-downloads',)

if obj.key.startswith(matches):
    print('The object name starts with one of the matching strings.')
else:
    print('The object does not start with one of the matching strings.')
```

So, we could rewrite this by creating a `tuple` (`set`) of strings to match and then just letting the `startswith` method do the work for us:

```python
matches = ('legal', 'bulk-downloads',)

if not obj.key.startswith(matches):
    obj.delete()
```

That should be easier to read and extend in the future if need be. :-)

ccostino: Note that the `set` of prefixes to check for should be defined before the `for` loop, and please feel free to use a better/more descriptive name than `matches`!

LindsayYoung: Thanks @ccostino @vrajmohan and @patphongs !

vrajmohan:
Looks good to me. TIL - `startswith` takes tuples

Issue #2423, opened 2017-05-24
Database improvements for itemized schedule data


ccostino:
Related to issue #2373 

We should investigate a bit more on the database side of things to see if there's anything else we can do in or leverage from PostgreSQL to make our itemize schedule data more performant, especially in the case of being able to browse/filter/search records across multiple cycles.

There are a small handful of things we could look at that'll let us know if anything is possible:

* Query performance tuning for schedule A and B queries
* Testing SELECTs across multiple cycles (read: multiple child tables) and whether or not we need to limit how many
* Current index setup and if any others would help or hinder
* Other server settings we haven't tweaked yet but are exposed to us in RDS

I think these are all relatively lightweight things to focus on with what we have to see if we really are pushing the boundaries of what's possible in our current setup.

This work is partially related to the PR we have open to refactor the itemized schedule data that we partition (#2410) as well as issue #2394 (more directly to that one), but they're not dependent on one another.  I'm just noting that these things might be easier to accomplish with a simpler setup. :-)

## UPDATE ON 8/1/2017

We've made several improvements to the PostgreSQL server settings (some in response to recent production performance issues we've experienced), implemented a large chunk of refactoring (#2498) of the itemized schedule A and B setup, and added an initial set of indexes (#2530) to support working with itemized schedule data across cycles.

We've also started conducted some performance and usability tests (from the API) of this work and early results are promising for the pieces we know to be working, but less so for other parts (e.g., filtering by occupation - @LindsayYoung please correct me if I'm wrong).  We need to continue with our testing and tweak/add any index combinations that are either not working as expected or currently missing.

ccostino:
This work is happening as a part of #2446, it just breaks down the performance tuning pieces into more granular details.

ccostino:
@anthonygarvan is it safe to say that this has been happening as a part of your Elasticsearch research for #2373?

anthonygarvan:
Conclusion of [2373](https://github.com/18F/openFEC/issues/2373): performance should be acceptable across years if we have multicolumn indexes defined on ([query_field], [sort_field], sub_id), where query_field is any field we search by, and sort_field is either transaction date or transaction amount.

ccostino:
The next stage of this work is dependent on #2446 being finished first.

ccostino:
The initial refactoring has been completed as of #2498, paving the way for this work now.

ccostino:
[...]

PR #2558, opened 2017-07-24
[WIP] Remove two year cycle restriction on itemized schedule API endpoints

ccostino:
**WIP:  PLEASE DO NOT MERGE**

Part of #2423

This changeset removes the two year cycle restriction from the itemized schedule A and B API endpoints to enable searching and querying across multiple cycles.  Very early performance tests indicate that as long as the indexes are setup properly this is reasonably performant, even with large cross sections of data.

**NOTE**
* We still need to provide some type of restriction; simply asking for the whole data set results in a long-running query
* It might be possible to setup a restriction just based on min/max dates alone; that is worth exploring
* Alternatively, we could either allow multiple values for `two_year_transaction_period` or change that parameter to work on groupings of 4, 6, or 8 cycles

Again, this code only drops the restriction in the API for the moment and is only meant for testing and further development.  Work should be actively coordinated with the front end development in the web app to ensure consistency and proper UI/UX support.

/cc @LindsayYoung, @jontours, @anthonygarvan, @vrajmohan and anyone else interested in helping build this out from the API side:  please feel free to hop into this PR and carry it forward!

/cc @xtine and @noahmanger:  FYI to help coordinate with front end development.

### DIFF
--- a/tests/test_itemized.py
+++ b/tests/test_itemized.py
@@ -56,22 +56,22 @@ class TestItemized(ApiBaseTest):
         )
     #This is the only test that the years will have to be bumped when in a new cycle
     #maybe refactor to use some logic based on current year?
-    def test_two_year_transaction_period_default_supplied_automatically(self):
-        receipts = [
-            factories.ScheduleAFactory(
-                report_year=2016,
-                contribution_receipt_date=datetime.date(2016, 1, 1),
-                two_year_transaction_period=2016
-            ),
-            factories.ScheduleAFactory(
-                report_year=2018,
-                contribution_receipt_date=datetime.date(2018, 1, 1),
-                two_year_transaction_period=2018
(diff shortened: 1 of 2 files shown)


ccostino:
The tests were failing on one test in `test_itemized.py` due to the API returning all results given the lifted restrictions instead of just one.  I commented out the test for now in anticipation of modifying it based on any future work for implementing a new type of restriction.

LindsayYoung:
OK, I am picking this up. Thanks!

ccostino:
This PR is going to be open a little longer as it is needed for testing multi-cycle searches, so I am going to disconnect it from the older issue in favor of a more specific user story once we complete our testing.

LindsayYoung:
Work moved to #2632

PR #2562, opened 2017-07-26
Only return a candidate once

LindsayYoung:
So far, got the duplicates out of the elections list


jontours:
I got the senate results working right now.  I'll be in the office tomorrow to show you the work I did here and see if it makes sense @LindsayYoung.  Also need to get the tests passing.

LindsayYoung:
happy to talk it through when you get a chance

jontours:
Three of a the big bullet points are now taken care of.  I recommend a review and a merge at this point.  As the other bullet points are spread on two other endpoints.  So think it's better to review and merge and digestible chunks.  Let me know @LindsayYoung if you have any questions.  I tried adding you as a reviewer but I can't?  Maybe because you opened it?

webservices/common/models/elections.py
```diff
@@ -10,6 +10,8 @@ class ElectionResult(db.Model):
     cand_office = db.Column(db.String, primary_key=True, doc=docs.OFFICE)
     cand_office_st = db.Column(db.String, primary_key=True, doc=docs.STATE_GENERIC)
     cand_office_district = db.Column(db.String, primary_key=True, doc=docs.DISTRICT)
+    election_type = db.Column(db.String)
+    fec_election_yr = db.Column(db.Integer)
```
ccostino: Just a note here:  we don't need any doc strings for these as they're not exposed in the API?  If we do we can file another PR with them later, I see no reason to hold this up.

PR #2566, opened 2017-07-28
Feature/parse boolean operators

anthonygarvan:
Replaces previous parse logic for terms & phrases with the built in query_string, which supports boolean operations and proximity search.

The `q` parameter for all searches is not parsed as a `query_string` instead of a `match`. Documentation for `query_string` is [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html). Basically, it supports AND, OR, NOT, parenthesis, quotation marks for phrase match, proximity search, and wildcards. 

If the query is not parsable (eg., `president AND OR`), the query will return a json object with HTTP status code 400 and message "Could not parse query". This HTTP status will need to be handled by consumers of the API.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,6 @@ elasticsearch==2.4.0
 elasticsearch-dsl==2.1.0
 bandit==1.4.0
 git+https://github.com/18F/slate.git
-nltk==3.2.4
 
 # Marshalling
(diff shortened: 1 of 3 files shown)


vrajmohan:
The rest of the changes look good. Great job on discovering this feature from the documentation and eliminating many lines of code.

PR #2568, opened 2017-07-31
bigger pool

LindsayYoung:
Trying to set these directly


LindsayYoung:
@rohan this was the PR about getting a bigger pool for the API, the branch is behind but it shouldn't be too hard to pull in the circle changes

LindsayYoung:
I am not sure this solves the problems we have had with the pool over flow, I think we need to make sure that doesn't happen rather than make a bigger pool

PR #2570, opened 2017-07-31
Force using Trusty and PostgreSQL 9.6 in Travis CI

ccostino:
We are running into some issues with our tests failing in Travis and currently all signs point to something breaking down in the Travis build.  This changeset forces the build environment to use Trusty (which Travis is in the process of rolling out to all builds regardless), which in turn also supports using PostgreSQL 9.6 (which this project is running on).  We will continue making adjustments as necessary to see what is causing the tests to fail and what might be needed to get them to work.

This PR also pins all of our development dependencies to specific versions.  Please see issue #2573 for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
+dist: trusty
 language: python
 sudo: false
 cache: pip
 python:
   - "3.5"
 addons:
-  postgresql: "9.4"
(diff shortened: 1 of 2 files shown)


LindsayYoung:
Thanks!

PR #2581, opened 2017-08-02
FRQ-RFAI refactor - [WIP]

jontours:
This changes script to only use `FRQ` filings from `disclosure.f_rpt....` table.  Does *not* solve the issue with missing RFAI filings, at least not yet.


LindsayYoung:
@jontours is there something that needs to be done here or should we just close this?

jontours:
Guessing from Jeff's latest comment on a related issue that this is still a problem.  I'll catch up with this and try to provide an update on work needed.

LindsayYoung:
This is pre-migrations, so I am going to close this, we can use it as a reference later if we need it

PR #2586, opened 2017-08-04
Fixes precision issues by casting to a string

jontours:
I don't think this will break any uses, because most likely anybody actually paging through this data is using a script, and were hence most likely using a language like Python or Java, languages of which have no issues with a large integer.  But up to the most recent JavaScript library, the largest integer that can be referenced is 2^(54-1).

Front-end will need no changes for this, works as is and has been tested locally.


jontours:
closing in favor of #2587

Issue #2585, opened 2017-08-04
Nightly refresh process failing/not working correctly in the dev environment


ccostino:
We have been having some issues with the nightly processing in our development environment:

* Status emails haven't been going out
* Queries are getting locked up in the database
* The nightly refresh job has hung up and doesn't fully complete
* Celery sometimes doesn't get shut down gracefully and throws a small stack trace
* The nightly retry processing for schedule A and B data is still causing issues

I've been doing a lot of digging around this morning into the issues we've been having with the `dev` database and I'm reaching a couple of conclusions:  the extra nightly retry processing I had tried to put in place to catch records that we couldn't initially process with the schedule A and B queues appears to be causing far more problems than it is helping anything, and what happens if our nightly refresh ends up in a state where it's still running at the point it gets triggered again the next day (this should never be the case, it takes 1 - 2 hours at most to run on average).

Some extra context into things:

We started noticing these issues around July 29th.  I'm still not sure what ultimately started causing the deadlocks and row contention with the aggregate processing, but that's when the errors start getting reported (looking into the cloud.gov logs of the `dev` `celery-worker` app).  These ultimately culminate in the emails never getting sent out and a Mandrill (the email processing app we use) max retry error to be thrown, hence the lack of emails being sent out for `dev` from the nightly update bot we have setup.

Another thing to note:  even with the nightly processing itself shut off (we had done this as a hotfix recently), the triggers are still active, and I'm seeing errors in the RDS logs with duplicate keys trying to be inserted (even though we have code to remove existing records with matching keys first before inserting!).  Potentially due to these,  I've noticed some EOF and connection timeout errors too in the Celery worker logs.

Updating Celery in `dev` in prep for our next release seems like a good idea as there are a host of bug fixes in it, and I have caught several times when it throws a small stack trace when it shuts down, which may leave jobs in an unstable state.  `staging` and `prod` seem to be okay and have been updating fine each night, but they're still running on the 3.x series of Celery.

We also had the Redis connectivity issue early this week too, compounding a bit of this but not a direct cause (the initial errors happened a few days before).

Back to the problem at hand:

The main problem is the aggregate updates, those have gotten hung up in dev due to deadlocked processes when they attempt to run (and generally get stuck on the very first step, updating zip codes for schedule A).

Because the processes get locked, the main nightly refresh task doesn't quite "finish" despite having thrown some success messages - the underlying update aggregate queries are still active and locked in the DB.

[...]

PR #2589, opened 2017-08-04
Remove nightly retry processing of itemized schedule A and B data

ccostino:
Part of #2585

This changeset removes the nightly retry processing we had initially setup for the itemized schedule A and B data.  We have found this to be more cumbersome to get working and maintain than it is worth and it has resulted in several nightly updates failing due to deadlocked processes and record conflicts.

We will attempt to improve our processing and updating of itemized schedule A and B data through our continued refactoring that we had started previously to make the partitioning code more efficient and easier to maintain.

### DIFF
--- a/data/sql_setup/prepare_schedule_a.sql
+++ b/data/sql_setup/prepare_schedule_a.sql
@@ -1,12 +1,7 @@
--- Create table to hold sub_ids of records that fail during the nightly
--- processing so that they can be tried at a later time.
--- The "action" column denotes what should happen with the record:
---    insert, update, or delete
+-- Drop the old nightly refresh retry table and function if they still exist.
 drop table if exists ofec_sched_a_nightly_retries;
-create table ofec_sched_a_nightly_retries (
-    sub_id numeric(19,0) not null primary key,
-    action varchar(6) not null
-);
+drop function if exists retry_processing_schedule_a_records(start_year integer);
+
 
 -- Create queue tables to hold changes to Schedule A
 drop table if exists ofec_sched_a_queue_new;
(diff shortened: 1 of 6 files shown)


data/sql_setup/prepare_schedule_a.sql
```diff
@@ -1,12 +1,7 @@
--- Create table to hold sub_ids of records that fail during the nightly
--- processing so that they can be tried at a later time.
--- The "action" column denotes what should happen with the record:
---    insert, update, or delete
+-- Drop the old nightly refresh retry table and function if they still exist.
 drop table if exists ofec_sched_a_nightly_retries;
```
vrajmohan: Will this be removed in a subsequent PR?

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -2,15 +2,11 @@
 drop index if exists disclosure.nml_sched_b_link_id_idx;
 create index nml_sched_b_link_id_idx on disclosure.nml_sched_b (link_id);
 
--- Create table to hold sub_ids of records that fail during the nightly
--- processing so that they can be tried at a later time.
--- The "action" column denotes what should happen with the record:
---    insert, update, or delete
+
+-- Drop the old nightly refresh retry table and function if they still exist.
 drop table if exists ofec_sched_b_nightly_retries;
```
vrajmohan: ditto

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -101,13 +48,6 @@ begin
...
-        else
-            -- We weren't able to successfully retrieve a row from the view,
-            -- so keep track of this sub_id if we haven't already so we can
-            -- try processing it again each night until we're able to
-            -- successfully process it.
-            delete from ofec_sched_a_nightly_retries where sub_id = new.sub_id;
-            insert into ofec_sched_a_nightly_retries values (new.sub_id, 'insert');
         end if;
 
         return new;
```
vrajmohan: I was confused by this and had to RTFM to make sure I understood what is being returned. It may be better to use:
```
 RETURN NULL; -- result is ignored since this is an AFTER trigger
```
 as shown in an example on https://www.postgresql.org/docs/9.6/static/plpgsql-trigger.html#PLPGSQL-DML-TRIGGER

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -106,13 +54,6 @@ begin
...
-        else
-            -- We weren't able to successfully retrieve a row from the view,
-            -- so keep track of this sub_id if we haven't already so we can
-            -- try processing it again each night until we're able to
-            -- successfully process it.
-            delete from ofec_sched_b_nightly_retries where sub_id = new.sub_id;
-            insert into ofec_sched_b_nightly_retries values (new.sub_id, 'insert');
         end if;
 
         return new;
```
vrajmohan: ditto

data/sql_setup/prepare_schedule_a.sql
ccostino: Yes, I can file one after the next release goes out to make sure this is propagated through all environments.

data/sql_setup/prepare_schedule_b.sql
ccostino: Same note as above, but good catch!

data/sql_setup/prepare_schedule_a.sql
ccostino: Oh interesting - with this being an `AFTER` trigger the record returned ultimately doesn't matter because it's already been processed in the table that the trigger was fired off of.  I'm hesitant to make the change only because of all of the issues we've had here before with the incorrect return value being set in the past.

Would you be amenable to leaving this as is in this PR, but we can open a separate one to research more once we're certain the nightly update process itself is working fine again?

vrajmohan:
Looks great. Hurray for less code!

PR #2594, opened 2017-08-08
Add support for multi-column GIN indexes

ccostino:
Part of #2423

This changeset modifies our GIN indexes in the itemized schedule A and B data to be multi-column, similar to the other multi-column indexes we have setup.  The intent is to provide a boost in performance when searching/filtering on text/ts_vector fields across multiple cycles.

This requires the support of the `btree_gin` PostgreSQL extension, which is installed in Amazon RDS but not enabled by default.  It can be enabled by running the following:

```sql
CREATE EXTENSION btree_gin;
```
This will also require repartitioning to rebuild the indexes.

For more information about this, see the following bits of documentation:

* [PostgreSQL 9.6 extensions supported in Amazon RDS](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.FeatureSupport.Extensions.96x)
* [PostgreSQL `btree_gin` extension](https://www.postgresql.org/docs/9.6/static/btree-gin.html)
* [PostgreSQL Multicolumn documentation](https://www.postgresql.org/docs/9.6/static/indexes-multicolumn.html)

### DIFF
--- a/data/functions/itemized.sql
+++ b/data/functions/itemized.sql
@@ -94,8 +94,8 @@ BEGIN
         EXECUTE format('CREATE INDEX idx_%s_two_year_transaction_period_dt%s ON %I (two_year_transaction_period, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
 
         EXECUTE format('CREATE INDEX idx_%s_contributor_name_text_contb_receipt_dt%s ON %I USING GIN (contributor_name_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
-        EXECUTE format('CREATE INDEX idx_%s_contributor_employer_text_contb_receipt_dt%s ON %I USING GIN (contributor_name_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
-        EXECUTE format('CREATE INDEX idx_%s_contributor_occupation_text_contb_receipt_dt%s ON %I USING GIN (contributor_name_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_contributor_employer_text_contb_receipt_dt%s ON %I USING GIN (contributor_employer_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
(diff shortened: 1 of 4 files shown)


data/functions/itemized.sql
```diff
@@ -75,6 +75,10 @@ BEGIN
         END IF;
 
         -- Create indexes.
+        -- Note:  The multi-column GIN indexes require the btree_gin extension
+        --        https://www.postgresql.org/docs/current/static/btree-gin.html
+        --        This is installed and enabled in RDS by default.
```
vrajmohan: Huh, the PR summary says that it is **not** enabled.

data/functions/itemized.sql
```diff
@@ -88,6 +92,10 @@ BEGIN
         EXECUTE format('CREATE INDEX idx_%s_sub_id_line_num_dt%s ON %I (line_num, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
         EXECUTE format('CREATE INDEX idx_%s_two_year_transaction_period_dt%s ON %I (two_year_transaction_period, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
 
+        EXECUTE format('CREATE INDEX idx_%s_contributor_name_text%s ON %I USING GIN (contributor_name_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
+        EXECUTE format('CREATE INDEX idx_%s_contributor_employer_text%s ON %I USING GIN (contributor_name_text, contb_receipt_dt, sub_id)', child_table_root, index_name_suffix, child_table_name);
```
vrajmohan: Copy/paste error? The index columns are identical to the previous one,

data/functions/itemized.sql
ccostino: Argh, thanks for catching!

data/functions/itemized.sql
ccostino: I don't think they are, I had modified these to have `contb_receipt_dt` mixed with them, and the ones below now have `contb_receipt_amt` mixed in them (for schedule A).

ccostino: Oh but the names are the same, I'll fix that!

vrajmohan: The columns being indexed are the same?

PR #2602, opened 2017-08-11
API fixes for election view page

jontours:
Alabama issue resolved (although I'm seeing 24 candidates @PaulClark2, this can also be reflected by doing a select on ofec_candidate_history_mv for this special election).  Now debugging Colorado 6th district issues.

### DIFF
--- a/data/sql_updates/candidate_history.sql
+++ b/data/sql_updates/candidate_history.sql
@@ -46,6 +46,7 @@ select distinct on (fec_yr.cand_id, fec_yr.fec_election_yr)
     row_number() over () as idx,
     fec_yr.lst_updt_dt as load_date,
     fec_yr.fec_election_yr as two_year_period,
+    fec_yr.cand_election_yr as candidate_election_year,
     fec_yr.cand_id as candidate_id,
     fec_yr.cand_name as name,
     fec_yr.cand_state as address_state,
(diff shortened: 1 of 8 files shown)


jontours:
@PaulClark2 after doing some debugging facing some difficulties again in this case with special elections.  Namely the `cand_election_yr` for these special candidates is at max 2017, since that's the year it's held.  However, there isn't a good way to capture this.  I feel like a column such as `cand_election_yr_cycle` would be helpful?  In other words the cycle that the election occurred in?

LindsayYoung:
Did you try fec_election_yr?

jontours:
At first I considered that, as it would fix this specific case.  However that would would bring in invalid cases from regular senatorial cycles.  For instance, in this case, Shelby, who also has `fec_election_yr` of 2018.  You can see that here: 
```
 select candidate_id,name, two_year_period, active_through,election_years from ofec_candidate_history_mv where state ='AL' and office ='S' and two_year_period = 2018 and candidate_inactive = False;
```
These special elections are particularly nasty, as they fall on odd or years, or in this case do?  In fact they are so nasty, I'd argue they don't belong in a relational database that relies on relational algebra, or at least have them be in their own table?  I'm not sure, kind of at a loss.  Maybe it would makes sense to have a column for Senate class?  Since we know this is Jeff Sessions seat, that would be the best way to filter really.

jontours:
Actually, after a weekend to sleep on this, a programmtic solution I believe is possible.  I just needed some time away!

jontours:
Colorado and Alabama data looking good now (and probably many, many other examples).  Let me know if there are any questions about strategy used here.

PR #2606, opened 2017-08-15
Feature/eliminate queues

vrajmohan:
- Use the correct definition of `public.fec_fitem_sched_(a|b)_vw` and fix tests.
- Bypass queues and update itemized schedules directly from triggers
- Bypass queues and update aggregates directly from triggers


tests/test_integration.py
```diff
@@ -28,25 +29,34 @@ def make_factory():
...
         rpt_yr = 2016
+        amndt_ind = 'A'
 
-    return SchedAFactory, SchedBFactory
+    class FItemReceiptOrExp(SQLAlchemyModelFactory):
+        class Meta:
+            sqlalchemy_session = db.session
+            model = automap.classes.f_item_receipt_or_exp
+        v_sum_link_id = factory.Sequence(lambda n: n)
+        form_tp_cd = 'F3X'
```
ccostino: This is great to see, thank you!

tests/test_integration.py
```diff
@@ -392,18 +420,20 @@ def test_update_aggregate_create(self):
...
-        self._check_update_aggregate_existing('contbr_zip', 'zip', models.ScheduleAByZip)
-        self._check_update_aggregate_existing('contbr_st', 'state', models.ScheduleAByState)
-        self._check_update_aggregate_existing('contbr_employer', 'employer', models.ScheduleAByEmployer)
-        self._check_update_aggregate_existing('contbr_occupation', 'occupation', models.ScheduleAByOccupation)
+        faker = Faker()
+        self._check_update_aggregate_existing('contbr_zip', 'zip', models.ScheduleAByZip, faker.zipcode)
+        self._check_update_aggregate_existing('contbr_st', 'state', models.ScheduleAByState, faker.state_abbr)
+        self._check_update_aggregate_existing('contbr_employer', 'employer', models.ScheduleAByEmployer, faker.company)
+        self._check_update_aggregate_existing(
+            'contbr_occupation', 'occupation', models.ScheduleAByOccupation, faker.job)
```
ccostino: Ah, I was wondering what the `key_generator` was defined above. :-)

tests/test_integration.py
```diff
@@ -454,37 +488,61 @@ def get_existing():
...
+        )
+        db.session.commit()
+        manage.update_aggregates()
+        db.session.execute('refresh materialized view ofec_sched_a_aggregate_size_merged_mv')
+        self._clear_sched_a_queues()
         existing = get_existing()
         total = existing.total
         count = existing.count
-        self.SchedAFactory(
+        NEW_RECEIPT_AMOUNT = 538
```
ccostino: Out of curiosity, why pull this and `EXISTING_RECEIPT_AMOUNT` out as separate variables?  Ease of change and maintenance later on?  Totally fine, I just wanted to make sure I understand!

tests/test_integration.py
```diff
@@ -28,25 +29,34 @@ def make_factory():
...
-    class SchedBFactory(SQLAlchemyModelFactory):
+    class NmlSchedBFactory(SQLAlchemyModelFactory):
         class Meta:
             sqlalchemy_session = db.session
             model = automap.classes.nml_sched_b
         form_tp_cd = '11'
         disb_dt = datetime.datetime(2016, 1, 1)
         sub_id = factory.Sequence(lambda n: n)
         rpt_yr = 2016
+        amndt_ind = 'A'
```
vrajmohan: The relevant query fragments for the views are:
```
 WHERE sa.sub_id = fi.sub_id AND sa.amndt_ind::text <> 'D'::text AND sa.delete_ind IS NULL;
```
and
```
 WHERE sb.sub_id = fi.sub_id AND sb.amndt_ind::text <> 'D'::text AND sb.delete_ind IS NULL;
```

The value cannot be `D`. I chose `A` because I found rows with that value in the `dev` DB.

tests/test_integration.py
vrajmohan: I think it is easier when reading code to realize why a certain expression is being asserted when you have _intention revealing names_ as opposed to magi numbers. With that said, I haven't been consistent about changing all magic numbers to names :-(

ccostino: Ah, yes, I definitely agree, and no worries at all, we get what we can as we move along!

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -92,27 +323,18 @@ begin
         -- visit here:
         -- https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS
         if FOUND then
-            two_year_transaction_period = cast(get_cycle(view_row.rpt_yr) as smallint);
-
-            if two_year_transaction_period >= start_year then
+            if view_row.election_cycle >= start_year then
                 delete from ofec_sched_a_queue_old where sub_id = view_row.sub_id;
-                insert into ofec_sched_a_queue_old values (view_row.*, timestamp, two_year_transaction_period);
+                insert into ofec_sched_a_queue_old values (view_row.*, current_timestamp, view_row.election_cycle);
```
ccostino: Is this still needed if we're not going to use the queues anymore?

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -26,45 +26,279 @@ create index on ofec_sched_a_queue_old (two_year_transaction_period);
...
+                contributor_name_text = to_tsvector(concat($1.contbr_nm, '' '', $1.contbr_id)),
+                contributor_employer_text = to_tsvector($1.contbr_employer),
+                contributor_occupation_text = to_tsvector($1.contbr_occupation),
+                is_individual = is_individual($1.contb_receipt_amt, $1.receipt_tp, $1.line_num, $1.memo_cd, $1.memo_text),
+                clean_contbr_id = clean_repeated($1.contbr_id, $1.cmte_id),
+                two_year_transaction_period = get_cycle($1.rpt_yr),
+                line_number_label = expand_line_number($1.filing_form, $1.line_num)
+            WHERE
+                %I.sub_id = $1.sub_id',
+                child_table_name, child_table_name) USING view_row;
```
ccostino: Awesome to see all of this!

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -92,27 +323,18 @@ begin
...
-
-        if FOUND then
-            two_year_transaction_period = cast(get_cycle(view_row.rpt_yr) as smallint);
-
-            if two_year_transaction_period >= start_year then
-                delete from ofec_sched_a_queue_old where sub_id = view_row.sub_id;
-                insert into ofec_sched_a_queue_old values (view_row.*, timestamp, two_year_transaction_period);
-            end if;
-        end if;
-
```
ccostino: 🤔 I wonder if we even need the `UPDATE` portion in the delete trigger anymore... if we're going to process these sans-queues and take care of them in real time, then the `AFTER` trigger that is setup for `INSERT` and `UPDATE` should suffice since we have to wait until a record is (hopefully) available through the view, right?  In which case, we don't need to worry about tracking the changes ourselves anymore.

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -92,27 +323,18 @@ begin
...
         if FOUND then
-            two_year_transaction_period = cast(get_cycle(view_row.rpt_yr) as smallint);
-
-            if two_year_transaction_period >= start_year then
+            if view_row.election_cycle >= start_year then
                 delete from ofec_sched_a_queue_old where sub_id = view_row.sub_id;
-                insert into ofec_sched_a_queue_old values (view_row.*, timestamp, two_year_transaction_period);
+                insert into ofec_sched_a_queue_old values (view_row.*, current_timestamp, view_row.election_cycle);
+                IF tg_op = 'DELETE' THEN
+                    DELETE FROM ofec_sched_a_master WHERE sub_id = view_row.sub_id;
```
ccostino: This is what requires the trigger update(s) on the master table, correct?  Or would it also be easier to just target the proper child table directly in this case?

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -92,27 +323,18 @@ begin
...
         -- https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS
         if FOUND then
-            two_year_transaction_period = cast(get_cycle(view_row.rpt_yr) as smallint);
-
-            if two_year_transaction_period >= start_year then
+            if view_row.election_cycle >= start_year then
                 delete from ofec_sched_a_queue_old where sub_id = view_row.sub_id;
-                insert into ofec_sched_a_queue_old values (view_row.*, timestamp, two_year_transaction_period);
+                insert into ofec_sched_a_queue_old values (view_row.*, current_timestamp, view_row.election_cycle);
+                IF tg_op = 'DELETE' THEN
```
ccostino: I noticed this got shifted under finding the record first:  should we still check to see if this is a `DELETE` operation first before doing anything?

Then again, if we eliminate checking for `UPDATE`, this trigger could fire on just `DELETE` and therefore we don't need to check at all again (see note below). :-)

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -82,43 +314,30 @@ $$ language plpgsql;
...
+    -- up with a record full of NULL values.
+    -- "FOUND" is a PL/pgSQL boolean variable set to false initially in
+    -- any PL/pgSQL function and reset whenever certain statements are
+    -- run, e.g., a "SELECT INTO..." statement.  For more information,
+    -- visit here:
+    -- https://www.postgresql.org/docs/current/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS
+    if FOUND then
+        if view_row.election_cycle >= start_year then
+            delete from ofec_sched_b_queue_old where sub_id = view_row.sub_id;
+            insert into ofec_sched_b_queue_old values (view_row.*, current_timestamp, view_row.election_cycle);
```
ccostino: Same note as in schedule A, we shouldn't have to deal with the `old` queue tables anymore either, right?

data/sql_setup/prepare_schedule_a.sql
vrajmohan: I was keeping the queue creation in there until I fixed the code for the aggregates.

data/sql_setup/prepare_schedule_a.sql
vrajmohan: That was my thinking as well; this trigger will only fire for DELETEs eventually.

data/sql_setup/prepare_schedule_a.sql
vrajmohan: Good idea. Let me chew on it.

data/sql_setup/prepare_schedule_a.sql
ccostino: Oh right, sorry!

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +14,17 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
 group by cmte_id, cycle, employer
 ;
 
-alter table ofec_sched_a_aggregate_employer_tmp add column idx serial primary key;
+alter table ofec_sched_a_aggregate_employer_tmp add column idx serial;
```
ccostino: I'm going to use this script to capture most notes for all of the aggregate scripts, since they'll all be identical. :-)

Noting here to just double check on whether or not we need a unique identifier field for materialized views built off of these aggregate tables based on our conversation with @LindsayYoung from earlier!

Also, we do have models in the Python code that are built with all of these tables (found in [webservices/common/models/aggregates.py](https://github.com/18F/openFEC/blob/develop/webservices/common/models/aggregates.py) and those are all built off of the [`BaseModel`](https://github.com/18F/openFEC/blob/develop/webservices/common/models/base.py#L56), which references an `idx` field.  So if we are able to go down this route of not needing an arbitrary unique identifier field, we'd have to change some of the application code as well.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +14,17 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
...
+alter table ofec_sched_a_aggregate_employer_tmp add constraint uq_cmte_id_cycle_employer unique (cmte_id, cycle, employer);
```
ccostino: If we do need to keep a separate unique column, I take it this would have to be taken out?

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +14,17 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
...
-create index on ofec_sched_a_aggregate_employer_tmp (cmte_id, idx);
-create index on ofec_sched_a_aggregate_employer_tmp (cycle, idx);
-create index on ofec_sched_a_aggregate_employer_tmp (employer, idx);
-create index on ofec_sched_a_aggregate_employer_tmp (total, idx);
-create index on ofec_sched_a_aggregate_employer_tmp (count, idx);
+create index on ofec_sched_a_aggregate_employer_tmp (cmte_id);
+create index on ofec_sched_a_aggregate_employer_tmp (cycle);
+create index on ofec_sched_a_aggregate_employer_tmp (employer);
+create index on ofec_sched_a_aggregate_employer_tmp (total);
+create index on ofec_sched_a_aggregate_employer_tmp (count);
```
ccostino: Also noting that we may have to readjust these again based on the findings and the fact that the `idx` column is currently leveraged in the application code - given that we had to create those compound indexes for the itemized schedule data, would we have to maintain that here, or create different compound indexes?  Or would we be able to leverage PostgreSQL itself knowing what to combine, which [the documentation](https://www.postgresql.org/docs/9.6/static/indexes-multicolumn.html) says is usually more than sufficient?

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +14,17 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
...
-        select -1 as multiplier, cmte_id, rpt_yr, contbr_employer, contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, contbr_id
-        from ofec_sched_a_queue_old
-    ),
-    patch as (
-        select
-            cmte_id,
-            rpt_yr + rpt_yr % 2 as cycle,
-            contbr_employer as employer,
-            sum(contb_receipt_amt * multiplier) as total,
-            sum(multiplier) as count
```
ccostino: I just wanted to clarify, we're not losing these aggregate calculations based on the changes below, correct?  It doesn't appear that way but I just wanted to be sure!

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -3,84 +3,298 @@ drop table if exists ofec_sched_a_nightly_retries;
...
-alter table ofec_sched_a_queue_new add column timestamp timestamp;
-alter table ofec_sched_a_queue_old add column timestamp timestamp;
-alter table ofec_sched_a_queue_new add column two_year_transaction_period smallint;
-alter table ofec_sched_a_queue_old add column two_year_transaction_period smallint;
-create index on ofec_sched_a_queue_new (sub_id);
-create index on ofec_sched_a_queue_old (sub_id);
-create index on ofec_sched_a_queue_new (timestamp);
-create index on ofec_sched_a_queue_old (timestamp);
-create index on ofec_sched_a_queue_new (two_year_transaction_period);
-create index on ofec_sched_a_queue_old (two_year_transaction_period);
```
ccostino: *Rubs hands together with glee at seeing this.* :-)

Also, as with the aggregate methods, these remarks count for schedule B as well.

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -3,84 +3,298 @@ drop table if exists ofec_sched_a_nightly_retries;
...
-
--- Create trigger to maintain Schedule A queues for inserts and updates
+-- Create trigger to maintain Schedule A for inserts and updates
 -- These happen after a row is inserted/updated so that we can leverage pulling
 -- the new record information from the view itself, which contains the data in
 -- the structure that our tables expect it to be in.
-create or replace function ofec_sched_a_insert_update_queues() returns trigger as $$
+create or replace function ofec_sched_a_insert_update() returns trigger as $$
```
ccostino: Should we also keep in a `DROP FUNCTION` statement for the older function to make sure it gets removed?  Should that also be done at the time of the trigger recreation or just after?

*There's a note around those about this too.*

data/sql_setup/prepare_schedule_a.sql
ccostino: Restating how awesome it is to see this leveraged. :-)

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +14,17 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
...
+create index on ofec_sched_a_aggregate_employer_tmp (cmte_id);
+create index on ofec_sched_a_aggregate_employer_tmp (cycle);
+create index on ofec_sched_a_aggregate_employer_tmp (employer);
+create index on ofec_sched_a_aggregate_employer_tmp (total);
+create index on ofec_sched_a_aggregate_employer_tmp (count);
 create index on ofec_sched_a_aggregate_employer_tmp (cycle, cmte_id);
 
 -- Remove previous aggregate and rename new aggregate
 drop table if exists ofec_sched_a_aggregate_employer;
 alter table ofec_sched_a_aggregate_employer_tmp rename to ofec_sched_a_aggregate_employer;
```
ccostino: Do we also need a method to rename the indexes from their `*_tmp` names as well, like we did for schedule A and B?

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -3,84 +3,298 @@ drop table if exists ofec_sched_a_nightly_retries;
...
+                contributor_employer_text = to_tsvector($1.contbr_employer),
+                contributor_occupation_text = to_tsvector($1.contbr_occupation),
+                is_individual = is_individual($1.contb_receipt_amt, $1.receipt_tp, $1.line_num, $1.memo_cd, $1.memo_text),
+                clean_contbr_id = clean_repeated($1.contbr_id, $1.cmte_id),
+                two_year_transaction_period = get_cycle($1.rpt_yr),
+                line_number_label = expand_line_number($1.filing_form, $1.line_num)
+            WHERE
+                %I.sub_id = $1.sub_id',
+                child_table_name, child_table_name) USING view_row;
+            PERFORM increment_aggregates(view_row);
```
ccostino: Nice!

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -3,84 +3,298 @@ drop table if exists ofec_sched_a_nightly_retries;
...
 $$ language plpgsql;
 
 
--- Create trigger to maintain Schedule A queues deletes and updates
+-- Create trigger to maintain Schedule A deletes and updates
 -- These happen before a row is removed/updated so that we can leverage pulling
 -- the new record information from the view itself, which contains the data in
 -- the structure that our tables expect it to be in.
-create or replace function ofec_sched_a_delete_update_queues() returns trigger as $$
+create or replace function ofec_sched_a_delete_update() returns trigger as $$
```
ccostino: Same note as before regarding the older version of the function - we shouldn't let it hang out.

data/sql_setup/prepare_schedule_a.sql
ccostino: Looks like this is resolved now with this code being removed, right?

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -124,16 +328,100 @@ $$ language plpgsql;
...
 
 drop trigger if exists f_item_sched_a_after_trigger on disclosure.f_item_receipt_or_exp;
 create trigger f_item_sched_a_after_trigger after insert or update
-    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_a_insert_update_queues(:START_YEAR_AGGREGATE);
+    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_a_insert_update(:START_YEAR_AGGREGATE);
 
 drop trigger if exists f_item_sched_a_before_trigger on disclosure.f_item_receipt_or_exp;
 create trigger f_item_sched_a_before_trigger before delete or update
-    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_a_delete_update_queues(:START_YEAR_AGGREGATE);
+    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_a_delete_update(:START_YEAR_AGGREGATE);
```
ccostino: *As noted from above with the older functions.*

Should we have statements to drop the older functions at this point, where it might be safer to do so?  We'd need to have some `TODO` comments to remind us to take them out after this has been propagated across all environments, like with similar changes from before.

data/sql_setup/prepare_schedule_a.sql
```diff
@@ -124,16 +328,100 @@ $$ language plpgsql;
...
+            total = ofec_sched_a_aggregate_zip.total + excluded.total,
+            count = ofec_sched_a_aggregate_zip.count + excluded.count;
+        INSERT INTO ofec_sched_a_aggregate_size
+            (cmte_id, cycle, size, total, count)
+        VALUES
+            (view_row.cmte_id, view_row.election_cycle, contribution_size(view_row.contb_receipt_amt), view_row.contb_receipt_amt, 1)
+        ON CONFLICT ON CONSTRAINT uq_cmte_id_cycle_size DO UPDATE
+        SET
+            total = ofec_sched_a_aggregate_size.total + excluded.total,
+            count = ofec_sched_a_aggregate_size.count + excluded.count;
```
ccostino: These calculations are all equivalent to the older table construction replaced from above, correct?

Again, nicely done with the `ON CONFLICT` clause additions! :-)

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -7,118 +7,317 @@ create index nml_sched_b_link_id_idx on disclosure.nml_sched_b (link_id);
...
-create index on ofec_sched_b_queue_old (two_year_transaction_period);
-
-
--- Create trigger to maintain Schedule A queues for inserts and updates
+-- Create trigger to maintain Schedule A for inserts and updates
 -- These happen after a row is inserted/updated so that we can leverage pulling
 -- the new record information from the view itself, which contains the data in
 -- the structure that our tables expect it to be in.
-create or replace function ofec_sched_b_insert_update_queues() returns trigger as $$
+create or replace function ofec_sched_b_insert_update() returns trigger as $$
```
ccostino: Same remarks in this file as with schedule A - dropping functions, calculations, etc.

data/sql_setup/prepare_schedule_b.sql
```diff
@@ -130,16 +329,72 @@ $$ language plpgsql;
...
 create trigger f_item_sched_b_after_trigger after insert or update
-    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_b_insert_update_queues(:START_YEAR_AGGREGATE);
+    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_b_insert_update(:START_YEAR_AGGREGATE);
 
 drop trigger if exists f_item_sched_b_before_trigger on disclosure.f_item_receipt_or_exp;
 create trigger f_item_sched_b_before_trigger before delete or update
-    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_b_delete_update_queues(:START_YEAR_AGGREGATE);
+    on disclosure.f_item_receipt_or_exp for each row execute procedure ofec_sched_b_delete_update(:START_YEAR_AGGREGATE);
+
+CREATE OR REPLACE FUNCTION increment_aggregates(view_row fec_fitem_sched_b_vw) RETURNS VOID AS $$
```
ccostino: One addition remark:  This will be okay sharing the same name as the schedule A version?  I know the whole method signature is different, but I just want to confirm my understanding is correct.

Same remarks as above with checking on the calculations, too.

manage.py
```diff
@@ -12,7 +12,7 @@
 from flask_script import Server
 from flask_script import Manager
 
-from webservices import flow, partition
+from webservices import flow
```
ccostino: Wow, the last pieces...

webservices/tasks/refresh.py
```diff
@@ -17,7 +17,6 @@ def refresh():
     with mail.CaptureLogs(manage.logger, buffer):
         try:
             manage.update_functions()
-            manage.update_aggregates()
```
ccostino: This will be so nice to not have to do!

data/sql_setup/prepare_schedule_a.sql
vrajmohan: We need the `UPDATE` portion for the aggregates.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
vrajmohan: Even if we need to keep a separate unique column, this can still remain as you can have multiple unique constraints.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
vrajmohan: We shouldn't be, based on my understanding of the logic.

data/sql_setup/prepare_schedule_a.sql
vrajmohan: Good catch! We should have code to drop the old functions.

data/sql_setup/prepare_schedule_b.sql
vrajmohan: Good catch! I'll review the namespace and overloading features in `plpgsql` and report.

vrajmohan: It is permitted and they will not conflict - https://www.postgresql.org/docs/9.0/static/sql-createfunction.html. However, as it is subtle and we do not need overloading, I will rename the functions to be explicit.

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
vrajmohan: @LindsayYoung, can you point me to the code where the materialized views need the `idx` column?

vrajmohan: @ccostino, you're right, we would remove `BaseModel` as a parent for these models.

LindsayYoung: I was thinking of two constraints, but they don't apply in this case.

These are tables, I was thinking they were materialized views. To refresh materialized views concurrently there needs to be at least one unique key constraint. https://www.postgresql.org/docs/9.4/static/sql-refreshmaterializedview.html

The other consideration was that sqlalchemy needs a unique primary key or unique composites for all its models. We could make composites for the aggregates, but for most of the data I would not trust things to be unique.

So, this looks good. Thanks!

vrajmohan: Just to clarify, we are making composite unique keys with the columns that are being aggregated on.

ccostino: Awesome!  So it sounds like we can move forward with just the composite keys and update the models referencing the aggregate tables to use those instead (and not inherit from `BaseModel`)?

data/sql_setup/prepare_schedule_b.sql
ccostino: Thanks, that sounds like a good idea!

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
vrajmohan: Yes. In addition, we should explicitly name the indexes.

data/functions/utils.sql
```diff
@@ -50,3 +50,17 @@ BEGIN
...
+        SELECT indexname AS name
+        FROM pg_indexes
+        WHERE tablename = p_table_name;
+BEGIN
+
+    FOR index_name IN indexes_cursor LOOP
+        EXECUTE format('ALTER INDEX %1$I RENAME TO %2$I', index_name.name, regexp_replace(index_name.name, '_tmp', ''));
+    END LOOP;
+END
+$$ LANGUAGE plpgsql;
```
ccostino: Ah, nice to move this into a more appropriate place, thanks!

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
```diff
@@ -14,60 +17,18 @@ and is_individual(contb_receipt_amt, receipt_tp, line_num, memo_cd, memo_text, c
 group by cmte_id, cycle, employer
 ;
 
-alter table ofec_sched_a_aggregate_employer_tmp add column idx serial primary key;
+alter table ofec_sched_a_aggregate_employer_tmp add column idx serial;
+alter table ofec_sched_a_aggregate_employer_tmp add constraint uq_sched_a_aggregate_employer_tmp_cmte_id_cycle_employer unique (cmte_id, cycle, employer);
```
ccostino: One last question about this (and goes for all aggregate tables) - this constraint will be renamed with the indexes, right?  Or does that also have to be done separately to take care of the `_tmp` in it?

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
ccostino: Since the `idx` column is being kept in all cases here, we don't need to adjust the `BaseModel` inheritance, correct?

data/sql_incremental_aggregates/prepare_schedule_a_aggregate_employer.sql
vrajmohan: We don't _need_ to adjust the `BaseModel` inheritance as part of this PR. We should do this soon in another PR in order to remove superfluous code.

vrajmohan: The constraint will be renamed with the rest of the indexes. The successful run of the unit tests confirms it as the `ON CONFLICT ON` clause uses the constraint name without the `_tmp`.

PR #2616, opened 2017-08-22
Remove primary index, this is handled above

LindsayYoung:
This is running in dev for a and stage a and b now


LindsayYoung:
This is already covered in the last push in the release

PR #2620, opened 2017-08-23
[WIP] feature/ modularize index command 

LindsayYoung:
Working in postgres is not getting called correctly from management command.


lbeaufort:
Archiving this PR

Issue #2608, opened 2017-08-16
Create endpoints for Audit Search


jwchumley:
- [ ] use SQL from old audit search to create views/endpoints for new audit search

@pkfec @hcaofec @fec-jli

PaulClark2:
https://github.com/18F/openFEC/issues/2676 (edited)
 https://github.com/18F/openFEC/issues/2677 (edited)

PR #2626, opened 2017-08-25
[WIP] feature/ Create query endpoint for AuditSearch

rjayasekera:
Create endpoints for Audit Search #2608 
Adding tables to model for the pull down query


webservices/args.py
```diff
@@ -753,3 +753,10 @@ def _deserialize(self, value, attr, data):
         description=docs.COMMITTEE_TYPE_STATE_AGGREGATE_TOTALS
     ),
 }
+
+auditfinding = {
```
ccostino: PEP8-nit:  In Python, if you're dealing with something that is represented by two separate words, as in "Audit Finding," we would typically name this with an underscore, e.g., `audit_finding`.

webservices/common/models/auditsearch.py
```diff
@@ -0,0 +1,33 @@
+
+from .base import db, BaseModel
+from webservices import docs
+
+class AuditSearchMixin(object):
+    __table_args__ = {"schema": "auditsearch"}
+
+class AuditCase(db.Model, AuditSearchMixin):
```
ccostino: Just a note that we already went over [Python's MRO and order of mixins](https://stackoverflow.com/questions/10018757/how-does-the-order-of-mixins-affect-the-derived-class); will be fixed in a future commit to this PR!

webservices/common/models/auditsearch.py
```diff
@@ -0,0 +1,33 @@
...
+    finding_id = db.Column('finding_pk', db.String, index=True, primary_key=True, doc=docs.FINDING_ID)
+    finding = db.Column('finding', db.String, index=True, doc=docs.FINDING)
+    tier = db.Column('tier', db.Integer, index=True, doc=docs.TIER)
+
+class AuditFindingRel(db.Model, AuditSearchMixin):
+    __tablename__ = 'finding_rel'
+
+    relation_pk = db.Column('rel_pk', db.Integer, index=True, primary_key=True)
+    parent_finding_pk = db.Column('parent_finding_pk', db.Integer)
+    child_finding_pk = db.Column('child_finding_pk', db.Integer)
```
ccostino: We'll want to rename these object properties to be more consistent with the other classes.  So instead of `pk` at the end of these, we'd use `id` here.  The reference to the database column is what keeps the object/model properties connected to the corresponding database column, which you have defined explicitly here already. :-)

webservices/schemas.py
```diff
@@ -941,5 +941,13 @@ class TotalsCommitteeSchema(schemas['CommitteeHistorySchema']):
 register_schema(ScheduleAByStateRecipientTotalsSchema)
 register_schema(ScheduleAByStateRecipientTotalsPageSchema)
 
+AuditFindingSchema = make_schema(
+    models.AuditFinding,
+    options={'exclude': ('idx', 'finding_id', 'tier')},
```
ccostino: Two things here:

* `idx` is not a property that exists on the `AuditFinding` model, so it wouldn't be found here.
* By setting these fields in the `exclude` option, you will prevent them from displaying in the API results, which may not be what you want!  The called to `make_schema` with just the model will expose most things by default for you. :-)

If we remove the options that exclude these fields, this will make the fields appear in the API.

ccostino: Minor correction:  We probably do still want to exclude the ID, but we'll need to rename the field to match the model property.

tests/test_audit_search.py
```diff
@@ -0,0 +1,94 @@
...
+            ('issue', 'Coordinated Expenditures'),
+            ('election_cycle', 2002),
+            ('committee_id', 'C00161786'),
+            ('committee_name', 'COLORADO DEMOCRATIC PARTY'),
+            ('committee_designation', 'U'),
+            ('committee_type', 'Y'),
+            ('committee_description', 'Party'),
+            ('candidate_id', ''),
+            ('candidate_name', ''),
+            # ('audit_case_id', 1203),
```
LindsayYoung: I notice the these are still commented out, is this PR ready for review or is there still test work @pkfec and @rjayasekera ?

pkfec: @LindsayYoung 
Yes, i tested the existing filter fields, they worked as expected. Yes, this PR is ready for review.I am not sure, if we want to filter on other fields as well, which why i have commented those fields. If we decide not to have any filters, i can delete those lines.

PaulClark2: @pkfec I'm not sure I understand.  Issue (tiers), candidate name, committe ID, committee name and committee type should be filterable if we are trying to replicate the funcationality of the search system on classic.

rjayasekera: @PaulClark2 @Lindsay Yes, all those fields should be filterable.  And we can add them to API backend easily. We are awaiting to find out how the front end going to work, because records can be filter out and sort by front end logic also.  So it is matter of figuring out where the best place for the other filters.  Currently from API end point records can be filtered by finding, issue, election_cycle, cmte_id, cand_id.

lindsay: :wave: Just a note to double check you are pinging the right user. I've gotten 3 notifications on this thread, and I'm pretty sure that was unintentional.

pkfec:
My apologies, didn’t mean to send it to you. Please ignore.


Regards,
Priya

LindsayYoung:
Let's enable the filters. We want to have maximum flexibility for the web app so that we don't need API work if the front end specs change.

pkfec:
After the design session with Jennifer on tuesday, jun, rohan and myself discussed the following items that needs to be added to endpoint.
- add the election cycle date range 
- add additional filters
- rename the findings_id and issue_id to match the front name and be consistent.

tests/test_audit_search.py
ccostino: @lindsay we're very sorry about that, there was a bit of confusion on usernames!  Please feel free to remove yourself from any notifications on this; we'll try to do the same to clean things up.  Thanks!

ccostino: Sadly I'm afraid there's nothing we can do on our side, it's all on your end; as I said, please feel free to opt out of the review and notifications so you stop receiving things from us, apologies for the unexpected notifications!

LindsayYoung:
In Sublime to activate "remove trailing whitespace on save", click the menu items in Edit -> Whitespace .
To enable Whitespace at startup, modify your preference setting file.

(Doing the review right now, I can take care of some things directly, but I noticed the white space issues that can be dealt with automatically in Sublime so posting that here before I forget)

PR #2632, opened 2017-08-31
Feature/remove 2 yr restriction

LindsayYoung:
This should work 👍  

Old PR: https://github.com/fecgov/openFEC/pull/2558

### DIFF
--- a/webservices/common/util.py
+++ b/webservices/common/util.py
@@ -85,7 +85,7 @@ def year_defaults(**kwargs):
     # We plan on rolling this back the 6 year restriction in the future
     # but wanted to confirm performance on 6 years first.
     if min_date and max_date:
-        if (max_date - min_date).days > 2190:
+        if (max_date - min_date).days > 2192:
             raise ValidationError(
                 'Cannot search for more that a 6 year period. Adjust max_date and min_date or pick a two_year_period.',
                 status_code=422
--- a/webservices/common/util.py
+++ b/webservices/common/util.py
@@ -67,34 +67,27 @@ def year_defaults(**kwargs):
     min_date = kwargs.get('min_date')
     max_date = kwargs.get('max_date')
     cycle = kwargs.get('two_year_transaction_period')
-    # debug
-    print ('Function:')
-    print ([min_date, max_date, cycle])
+
     if min_date is None and max_date is None and cycle is None:
-        print(4)
         kwargs['two_year_transaction_period'] = CURRENT_CYCLE
         return kwargs
     if cycle is not None:
         return kwargs
     # this is for cases where only one date (min_date or max_date) is present
     if not min_date:
-        print(1)
-        begin_cycle = cycle - 1
+        begin_cycle = max_date.year - max_date.year % 2
         kwargs['min_date'] = datetime(begin_cycle, 1, 1)
         return kwargs
     if not max_date:
-        print(2)
+        cycle = min_date.year + min_date.year % 2
         kwargs['max_date'] = datetime(cycle, 12, 31)
         return kwargs
     # We plan on rolling this back the 6 year restriction in the future
     # but wanted to confirm performance on 6 years first.
     if min_date and max_date:
-        print(3)
         if (max_date - min_date).days > 2190:
             raise ValidationError(
                 'Cannot search for more that a 6 year period. Adjust max_date and min_date or pick a two_year_period.',
                 status_code=422
             )
-
-    print([min_date, max_date, cycle])
     return kwargs
--- a/webservices/common/util.py
+++ b/webservices/common/util.py
@@ -59,8 +59,8 @@ def year_defaults(**kwargs):
     defaults.
 
     Here are the conditions:
-        - if no date supplied default to the current 2-year period
         - if a cycle is selected, that will provide sufficient date filtering
+        - if no date supplied default to the current 2-year period
         - if there is a min and max date check to make sure there isn't more than a 6 year period
         - if cycle and only one min date or max-date, it adds the restriction with in the 2-year period
     """
@@ -68,10 +68,10 @@ def year_defaults(**kwargs):
     max_date = kwargs.get('max_date')
     cycle = kwargs.get('two_year_transaction_period')
 
-    if min_date is None and max_date is None and cycle is None:
-        kwargs['two_year_transaction_period'] = CURRENT_CYCLE
-        return kwargs
     if cycle is not None:
+        return kwargs
+    if min_date is None and max_date is None:
+        kwargs['two_year_transaction_period'] = CURRENT_CYCLE
         return kwargs
     # this is for cases where only one date (min_date or max_date) is present
     if not min_date:
--- a/webservices/common/util.py
+++ b/webservices/common/util.py
@@ -3,9 +3,11 @@ import os
 import flask
 import ujson
 
+from datetime import datetime
 
 dirname = os.path.dirname
 MAIN_DIRECTORY = dirname(dirname(dirname(__file__)))
+CURRENT_CYCLE = datetime.now().year + datetime.now().year % 2
 
 
 def get_full_path(*path):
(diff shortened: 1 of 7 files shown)


ccostino:
@LindsayYoung with this it seems it's safe to close and get rid of PR #2558, right?

LindsayYoung:
We need another round of testing after the improvements are pushed to prod

webservices/common/util.py
```diff
@@ -3,9 +3,11 @@
 import flask
 import ujson
 
+from datetime import datetime
 
 dirname = os.path.dirname
 MAIN_DIRECTORY = dirname(dirname(dirname(__file__)))
+CURRENT_CYCLE = datetime.now().year + datetime.now().year % 2
```
ccostino: I think we already have a `get_cycle` utility method defined somewhere - [in the config](https://github.com/18F/openFEC/blob/develop/webservices/config.py#L9-L21) (cycle start and cycle end, plus a `CURRENT_YEAR` constant).  Could/should those be leveraged instead, or is it more useful to have it handled this way now?  I just want to make sure we're not introducing duplicate logic. :-)

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+    max_date = kwargs.get('max_date')
+    cycle = kwargs.get('two_year_transaction_period')
+
+    if min_date is None and max_date is None and cycle is None:
+        kwargs['two_year_transaction_period'] = CURRENT_CYCLE
+        return kwargs
+    if cycle is not None:
+        return kwargs
+    # this is for cases where only one date (min_date or max_date) is present
+    if not min_date:
```
ccostino: This conditional may need to be more explicit, e.g., `if not min_date and max_date` to make sure we capture the case correctly.

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+        begin_cycle = max_date.year - max_date.year % 2
+        kwargs['min_date'] = datetime(begin_cycle, 1, 1)
+        return kwargs
+    if not max_date:
```
ccostino: Same note as above, except `if not max_date and min_date` in this case.

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+    have receipt date in the future, but we can keep it going now for
+    a smoother customer experience and try infer the most consistent
+    defaults.
+
+    Here are the conditions:
+        - if no date supplied default to the current 2-year period
+        - if a cycle is selected, that will provide sufficient date filtering
+        - if there is a min and max date check to make sure there isn't more than a 6 year period
+        - if cycle and only one min date or max-date, it adds the restriction with in the 2-year period
+    """
```
ccostino: Thank you for writing all of this!  Absolutely huge help! :-)

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+    min_date = kwargs.get('min_date')
+    max_date = kwargs.get('max_date')
+    cycle = kwargs.get('two_year_transaction_period')
+
+    if min_date is None and max_date is None and cycle is None:
+        kwargs['two_year_transaction_period'] = CURRENT_CYCLE
+        return kwargs
+    if cycle is not None:
+        return kwargs
```
ccostino: Even if the `cycle` is not `None`, are there any other checks we need to do here are all of the other values just ignored at this point?

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+        kwargs['min_date'] = datetime(begin_cycle, 1, 1)
+        return kwargs
+    if not max_date:
+        cycle = min_date.year + min_date.year % 2
+        kwargs['max_date'] = datetime(cycle, 12, 31)
+        return kwargs
+    # We plan on rolling this back the 6 year restriction in the future
+    # but wanted to confirm performance on 6 years first.
+    if min_date and max_date:
+        if (max_date - min_date).days > 2190:
```
ccostino: I was going to ask if it'd be possible do `(max_date - min_date).years` instead for clarity, but the resulting `datetime.timedelta` object doesn't have that property, d'oh!  Python, you've let me down! 😆

webservices/common/util.py
```diff
@@ -45,3 +47,47 @@ def get_class_by_tablename(tablename):
...
+            raise ValidationError(
+                'Cannot search for more that a 6 year period. Adjust max_date and min_date or pick a two_year_period.',
+                status_code=422
```
ccostino: Nice, thank you!

webservices/config.py
```diff
@@ -19,14 +19,15 @@ def get_cycle_end(year):
     return year if year % 2 == 0 else year + 1
 
 CURRENT_YEAR = datetime.datetime.now().year
+CURRENT_CYCLE = datetime.datetime.now().year + datetime.datetime.now().year % 2
```
ccostino: Same question as before - given the cycle methods defined above, do we need this?  And if so, should it be defined only here and not in the other spot too to cut down on duplication?

webservices/config.py
```diff
@@ -19,14 +19,15 @@ def get_cycle_end(year):
...
 
 SQL_CONFIG = {
     'START_YEAR': get_cycle_start(1980),
     'START_YEAR_AGGREGATE': get_cycle_start(2008),
     'END_YEAR_ITEMIZED': get_cycle_start(CURRENT_YEAR),
     'CYCLE_END_YEAR_ITEMIZED': get_cycle_end(CURRENT_YEAR),
     'PARTITION_START_YEAR': 1978,
-    'PARTITION_END_YEAR': 2018,
+    'PARTITION_END_YEAR': CURRENT_CYCLE,
```
ccostino: Ah I see what you're doing, :clap:!  That said, is there anything we need to account for on the DB side to make sure the app doesn't suddenly break on account of it referencing a partition child under the hood that doesn't exist yet?

In other words, let's say we're now in the 2020 cycle.  Right now, the DB won't automatically create this, we need to run the [`add_itemized_partition_cycle`](https://github.com/18F/openFEC/blob/develop/manage.py#L295-L321) management command before those tables exist.

I just want to make sure we're not creating a situation where the user would suddenly see an option that would throw an error!

webservices/common/util.py
ccostino: After chatting with some others about this, we might be able to leverage the [`dateutil.relativedelta`](https://dateutil.readthedocs.io/en/stable/relativedelta.html) object (from the `dateutil` library) to do this, as even going with the exact days difference ought to probably account for things like leap days and whatnot...  GitHub, you need a :shakes_fist: emoji. :-)

webservices/config.py
LindsayYoung: We should already have the next cycle tables built by January 1, 2020 or we would start getting insertion errors relatively quickly.  I would be fine making those tables now if you are concerned about it. 

We can also set the defaults on the web app separately from the API. In the future, we might want to hand code that value for a while to create more interesting presentation defaults, but I would rather have things work by default and over ride them later if people want something different. I feel like we had a lot of hand coded things to change this cycle that were not ideal and took us months to track down and change.

webservices/common/util.py
LindsayYoung: Let's just add 2 since that is as many leap years you can have in a 6 year period. As you know, we don't really have to be scientifically exact on this, it's just our guard rails while we try out our indexes.

webservices/common/util.py
LindsayYoung: Good catch!! Thanks

LindsayYoung: I fixed this by checking for cycle first and then checking for no min or max date. Thanks again

webservices/common/util.py
LindsayYoung: I think this is more straightforward since it is a simple formula, but maybe I am biased to my own preferences. I can change it Monday if you have a strong opinion about it.

lbeaufort:
@LindsayYoung we implemented this change with this new PR: https://github.com/fecgov/openFEC/pull/3674

LindsayYoung:
Congratulations FEC team! 🥇 

tears of joy

lbeaufort:
Thanks @LindsayYoung! Still figuring out the performance component, but we're moving forward!

ccostino:
Oh this is wonderful news @lbeaufort! Congratulations! :-)

PR #2638, opened 2017-09-05
WIP: Two styles for SQL unit tests

cmc333333:
Per conversation with @LindsayYoung @ccostino @vrajmohan, this includes an example of a very simple unit test using both a Python test harness and a pgTAP test harness.

Which do you all prefer?


LindsayYoung:
We would like to proceed with pgTAP tests. Knowing how to do SQL tests, and seeing a framework with them would be useful for the FEC outside this project, since they use a lot of SQL elsewhere. The sql tests are closer to the code we are testing, since we need a lot of views an tables that never exist as Python we would have to create a lot of them and it would be really easy to introduce a typing error.    

@ccostino or @vrajmohan anything else to add?

cmc333333:
Keep in mind that with both approaches, you'll need to define schemas that aren't spelled out right now -- these unit tests won't load the database dump. The difference between the approaches is in what interface is used to write the tests (and, by extension, which libraries can be used within those tests).

Including my thoughts from our previous conversation here, for reference:

> I think that web developers would generally prefer the Python system, but Vraj pointed out that FEC is more comfortable with straight SQL. The major pain point I see with the SQL approach is that it's not very flexible, but if that's FEC's world view, it might make sense. 

> I think the difference between the styles is in the test _runner_, since both are importing a single .sql file, exercising logic within the database, and validating the result.
> In one case (`tests/sql/functions/test_utils.py`), the database is accessed indirectly through sqlalchemy. The benefit here is that we have access to all of Python and its libs (libraries to mock models seems like a big win, for example) but we also gain indirection, needing to model concepts in Python. It certainly feels like overkill for such a simple test, but having access to that ecosystem might be important in the future. The second case (`tests/pgtap/functions/utils.sql`) is actually ran with `pgsql` (see `.travis.yml`), in the database environment. We lose a lot of Python's expressivity and libs, but gain straight access to the data like one might see in a database-centric UI.
> Personally, writing unit tests in SQL felt abnormal, but I'm much more comfortable in Python. If your developers are as comfortable if not more-so writing plpgsql than Python, I think the reverse would be true.

> you might like the unit tests in data-act, e.g. https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/tests/unit/dataactvalidator/test_fabs4_detached_award_financial_assistance_3.py#L15
> the idea here is that there's a specific validation in `fabs4_detached_award_financial_assistance_3.sql`. The test harness creates some mock data by using a factory (which allows them to specify only the fields that are relevant to the test), then runs that validation and verifies the results are what was expected.
> that approach is similar to `test_utils.py` -- Python's the test harness. It populates data through sqlalchemy models and does all of the asserts after the database function is called.
> an important difference is that the data-act tests don't have any data aside from the data populated within the test -- they're just testing a single unit of logic, in this case, a validation function.

> To write the unit tests in Python, you'll most likely want models for those tables so you can populate them with the test-specific data. To write the same tests in Postgres, you'd be doing `INSERT`s, so the model is still being described (if indirectly). Importantly, you don't want to load a big dump of data (as you might with an integration test) as you won't be able to split out what data is coming from your test and what data is coming from that blob.

LindsayYoung:
Yea, I understand that unit tests aren't going to use the test dataset. I think writing the schemas in SQL is less prone to error because that is what they already are, and we will have a lot of tables and views to recreate.

cmc333333:
:+1: I'll start work expanding the pgTAP tests.

LindsayYoung:
Thank you @cmc333333!

vrajmohan:
This is being followed up in a different issue as mentioned in https://github.com/fecgov/openFEC/issues/3272#issuecomment-425158863.

PR #2645, opened 2017-09-06
[MERGE WITH GIT FLOW] hotfix/inflated-totals

LindsayYoung:
…when there was a special and general election

H/T @jontours for helping me track this down!

### DIFF
--- a/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
+++ b/data/sql_incremental_aggregates/prepare_schedule_a_aggregate_size.sql
@@ -11,8 +11,8 @@ end
 $$ language plpgsql;
 
 -- Create initial aggregate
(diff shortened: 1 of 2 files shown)


ccostino:
@LindsayYoung should we fix this up and get it out the door, or abandon this PR at this point?  I think we did merge it into `develop` though to test, which means if we abandon we'll have to double check and revert in there before the next release or hotfix goes out.

LindsayYoung:
I am seeing this change incorporated here:

https://github.com/18F/openFEC/blob/8d7b892bca29d4da82e2da0bfdda0ee4b9ace109/data/migrations/V0026__more_public_schema_objects.sql#L70

We can make a new issue if there is an aspect that is not addressed we will have to re-write it as a migration.

PR #2647, opened 2017-09-07
[WIP] Add support for CircleCI service integration

ccostino:
**WIP:  PLEASE DO NOT MERGE**

This changeset begins the support for migrating our continuous integration over from Travis CI to CircleCI.  We need to move off of Travis CI per GSA requirements as it is no longer approve, and CircleCI is.  Once this fully working in CircleCI (meaning we can also deploy from CircleCI), we will shut off and remove the Travis CI integration.


ccostino:
@cmc333333, @adborden, @toolness, @LinuxBozo - would any of you be willing and able to take a few minutes to look over this CircleCI config quick to provide some feedback?  It's largely based off of @toolness's work with [Tock's CircleCI config](https://github.com/18F/tock/blob/master/.circleci/config.yml) and all I'm trying to do is mirror everything we have in our current [Travis CI config](https://github.com/18F/openFEC/blob/develop/.travis.yml) up to before deployment itself (so build the environment, install dependencies, run tests, and run code coverage).

For the most part it appears to be okay, except for the caching of dependencies (looks like a permissions issue based on the last build errors I'm seeing).  Now, we've had issues with that in the past with Travis and took it out, do you feel it's necessary here?  I'm seeing it in all of the examples and other project config files, so I opted to put it back in, but is it required?

Thank you all in advance for any help you may be able to provide!

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+          - fec-api-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "package.json" }}
+          # fallback to using the latest cache if no exact match is found
+          - fec-api-dependencies-
+
+      - run:
+          name: install dependencies
+          command: |
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install bandit codecov
```
LinuxBozo: Does it make sense to create a `requirements-ci.txt` file and include in follow line instead of this separate pip install?

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+            pip install -r requirements.txt -r requirements-dev.txt
+            npm install -g swagger-tools
```
LinuxBozo: You'll want to `sudo` here for the global swagger install. Pretty sure this is the cause of your build failure, not caching.

ccostino: Ah okay, thank you!

.circleci/config.yml
ccostino: 🤔 possibly, will look into that!

.circleci/config.yml
ccostino: @LinuxBozo is there anything required when using `sudo` for this, like in Travis, or safe and okay to just put in?

LinuxBozo: With the particular docker image you are using, from CircleCI, you just can use it out of the box. Otherwise, yes, you would need to make sure that your container supports using sudo, or run everything as root (usually a bad idea). 

BTW, here are basics and how Circle is doing it (and really same pattern can be used in any container):
 - makes sure sudo installed (https://github.com/circleci/circleci-images/blob/master/shared/images/Dockerfile-basic.template#L9..L13)
 - create a user https://github.com/circleci/circleci-images/blob/master/shared/images/Dockerfile-basic.template#L66..L67
 - give user sudo access https://github.com/circleci/circleci-images/blob/master/shared/images/Dockerfile-basic.template#L68..L69
 - drop privileges to that user to run the entrypoint/cmd
https://github.com/circleci/circleci-images/blob/master/shared/images/Dockerfile-basic.template#L74

ccostino: Awesome, thank you very much!

ccostino:
Okay interesting, that got us as far as the tests actually running, but they're erroring out due to being unable to connect to the database.  Do we have to explicitly set the connection string?  We didn't in Tock's config and I don't see that in CircleCI's Python example, but I have seen it elsewhere, so I'm just not sure.

ccostino:
I had a working theory that the connectivity does need to be specified as we do not have the FEC project setup within Docker (yet, at least).  Tock is, and I'm guessing CircleCI is leveraging that somehow with its setup.

In any case, I got as far as having the database server setup and the client tools configured, but still cannot get actual connectivity.  [The last build in CircleCI](https://circleci.com/gh/18F/openFEC/15) says it all:

```
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
Exited with code 2
```

[The documentation](https://circleci.com/docs/2.0/postgres-config/#postgresql-circleci-configuration-example) isn't much help here as it appears to be somewhat incomplete or at least hard to follow along as it doesn't give a concrete example of how all of the different setup pieces it covers work together (or what's actually required when).  I'm not sure what I'm missing or don't have configured right to make this work.

cmc333333:
@ccostino: CircleCI's running Python and Postgres on different containers, but they share the same port space. This means you won't be able to use the file-based socket, but you can use the network socket. Try running `psql` with `-h 0.0.0.0`. You'll also need to configure your connection string accordingly when you run the actual tests -- something like `postgres://postgres@0.0.0.0/cfdm_unit_test`

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+  build:
+    docker:
+      # CircleCI maintains a library of pre-built images documented at
+      # https://circleci.com/docs/2.0/circleci-images/
+      # use `-browsers` prefix for selenium tests, e.g. `<image_name>-browsers`
+
+      # Python
+      - image: circleci/python:3.5.3
+        environment:
+          - TZ=America/New_York
```
cmc333333: Small detail, but it'd be a bit cleaner if the environment were defined as a dict rather than a list:
```yml
environment:
    TZ: America/New_York
```

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+          command: dockerize -wait tcp://localhost:5432 -timeout 1m
+
+      - run:
+          name: Create the test database
+          # Commands listed here are from the CircleCI PostgreSQL config docs:
+          # https://circleci.com/docs/2.0/postgres-config/#postgresql-circleci-configuration-example
+          command: |
+            sudo apt-get install postgresql-client
+            echo ‘/usr/lib/postgresql/9.6/bin/:$PATH’ >> $BASH_ENV
+            psql -c "create database cfdm_unit_test;" -U postgres
```
cmc333333: Given that we've defined a database name in line 20, do we need to run this step? Is the `psql` cient used anywhere else?

.circleci/config.yml
```diff
@@ -0,0 +1,86 @@
...
+            py.test
+
+      - run:
+          name: Perform post-test checks
+          command: |
+            . .venv/bin/activate
+            cd ..
+            bandit -r openFEC
+            cd openFEC
+            codecov
```
cmc333333: Another small detail, but switching the directories around is a bit difficult to follow. Can bandit be ran on the current directory, like `bandit -r .`?

ccostino:
Thank you, @cmc333333, I'll give that a shot with the port/host next!  I had figured it was something to do with the separate containers but I wasn't sure what.

.circleci/config.yml
ccostino: Good call, I'll get that in the next change, thanks!

.circleci/config.yml
ccostino: We might not - I pulled this from our Travis setup, but if the database name defined above will take care of creating that for us, then no, we shouldn't need it!

.circleci/config.yml
ccostino: I'll give it a shot and see if it works, it'd be nice if it does to clean this up more!

ccostino:
@LindsayYoung looks like Bandit was the culprit there!

LindsayYoung:
@vrajmohan can you merge this when you get a chance? I just watched a successful test deploy form circle to dev

vrajmohan:
Nice work

PR #2648, opened 2017-09-08
Remove unused command

vrajmohan:


ccostino:
Ah good catch, @vrajmohan!  It looks like a test might need to be updated though, or at least investigated?  One of the Travis builds failed:  https://travis-ci.org/18F/openFEC/builds/273149717#L1341

ccostino:
Also, it looks like we should remove the `refresh_itemized` command from there and from `manage.py` as well (and its underlying methods) - that still got called during the nightly updates and failed:

```
Traceback (most recent call last):
  File "/home/vcap/app/webservices/tasks/refresh.py", line 20, in refresh
    manage.refresh_itemized()
AttributeError: module 'manage' has no attribute 'refresh_itemized'
```

That whole set of methods refers to the code that used to exist in the Python partition code, which is no longer there.

ccostino:
FYI that the Travis error seems to be breaking other PR tests now, too.

LindsayYoung:
Thank you both for getting this done!

PR #2655, opened 2017-09-13
feature/endpoint-for-totals-committe-type

pkfec:
…temporaliry commenting the code, since travis build is failing due to missing fields from the ofec_totals_parties_mv


tests/test_totals.py
```diff
@@ -34,7 +34,11 @@
     'last_debts_owed_by_committee': 42,
     'net_contributions': 123,
     'net_operating_expenditures': 321,
-    'last_debts_owed_to_committee': 42
+    'last_debts_owed_to_committee': 42,
+    'cmte_nm': None,
```
LindsayYoung: Using a string will create better tests since we expect their to be a string for these

data/sql_updates/candidate_fulltext.sql
```diff
@@ -15,9 +15,9 @@ with nicknames as (
         link.cmte_id = totals.committee_id and
         link.fec_election_yr = totals.cycle
     where
-        cmte_dsgn in ('P', 'A') and (
-            substr(cand_id, 1, 1) = cmte_tp or
-            cmte_tp not in ('P', 'S', 'H')
+        totals.cmte_dsgn in ('P', 'A') and (
```
LindsayYoung: I don't think we want to update to fulltext search. `ofec_candidate_fulltext_mv` is what powers the typeahead and we want to keep it as lean and fast as we can.

LindsayYoung:
Once you add the indexes you will want to add the fields to  the existing resources and args. 

Here are the examples from the RAD endpoint of the work that needs to be done for the filters. You can add these to the existing totals resources and args.

resources
Here are the resources  for RAD:
https://github.com/18F/openFEC/blob/ed51a68a7f3084efeb52e1c469b852a321daca32/webservices/resources/rad_analyst.py

args
look for the current args for the totals and add the new filters there
Here is the rad example:
https://github.com/18F/openFEC/blob/ed51a68a7f3084efeb52e1c469b852a321daca32/webservices/args.py#L524

data/sql_updates/candidate_fulltext.sql
pkfec: I had to make this change, to fix the "ambiguous column" error that i ran into. since "disclosure.cand_cmte_linkage" table and the mv "ofec_totals_combined_mv_tmp" are now referring to the same columns in candidate_fulltext.sql.

pkfec: i had a sql code review with @PaulClark2 and @fecjjeng on this last week , whether to fetch the columns for ofec_totals_combined_mv_tmp or from the cand_cmte_linkage. Paul suggested to fetch from the ofec_totals_combined_mv_tmp.
cc : @PaulClark2 & @fecjjeng

tests/test_totals.py
pkfec: done.

data/sql_updates/candidate_fulltext.sql
LindsayYoung: Thanks for the context. That makes sense.

pkfec:
@LindsayYoung Please review the code when you get a chance.

cc : @PaulClark2

LindsayYoung:
Thanks @pkfec!

PR #2656, opened 2017-09-13
Save the committe_id for use by downstream queries

vrajmohan:


vrajmohan:
Self merging PR as it is a pure fix of a test.

PR #2665, opened 2017-09-18
Restore default search by `_all` field

vrajmohan:
Restricting the search to only the `documents.text` field is too
restrictive, and moreover, does not work without a `nested` query.


LindsayYoung:
Thanks for catching this!

PR #2666, opened 2017-09-18
Remove copy-and-paste remnant

vrajmohan:


LindsayYoung:
Thanks for catching this!

PR #2668, opened 2017-09-20
Convert sched_a_by_size_merged_mv to a plain table

vrajmohan:
The materialized view was being used more as a table - it as being
rebuilt as opposed to be being refreshed. This commit converts it to a
real table, therby avoiding issues with dropping and rebuilding
_upstream_ tables; previously, these upstream tables could not be
dropped as they had materialized views depending on them.


data/converted_mvs/sched_a_by_size_merged.sql
```diff
@@ -0,0 +1,41 @@
...
+create unique index ofec_sched_a_aggregate_size_merged_tmp_idx on ofec_sched_a_aggregate_size_merged_tmp (idx);
+
+create index ofec_sched_a_aggregate_size_merged_tmp_cmte_idx on ofec_sched_a_aggregate_size_merged_tmp (cmte_id);
+create index ofec_sched_a_aggregate_size_merged_tmp_cycle on ofec_sched_a_aggregate_size_merged_tmp (cycle);
+create index ofec_sched_a_aggregate_size_merged_tmp_size on ofec_sched_a_aggregate_size_merged_tmp (size);
+create index ofec_sched_a_aggregate_size_merged_tmp_cycle_cmte_id on ofec_sched_a_aggregate_size_merged_tmp (cycle, cmte_id);
+
+drop table if exists ofec_sched_a_aggregate_size_merged;
+alter table ofec_sched_a_aggregate_size_merged_tmp rename to ofec_sched_a_aggregate_size_merged;
+select rename_indexes('ofec_sched_a_aggregate_size_merged');
```
ccostino: Would it make more sense to just have this be another script in the `data/sql_incremental_aggregates` folder instead, since this just another aggregate table for all intents and purposes?

manage.py
```diff
@@ -241,6 +241,7 @@ def update_schemas(processes=1):
         path = os.path.join('data', 'sql_updates', '{}.sql'.format(task))
         execute_sql_file(path)
     execute_sql_file('data/rename_temporary_views.sql')
+    execute_sql_file('data/converted_mvs/sched_a_by_size_merged.sql')
```
ccostino: See note above - or is it better due to the timing of when these scripts are run and still accounting for some dependencies?

data/converted_mvs/sched_a_by_size_merged.sql
vrajmohan: It is due to timing...it depends on `ofec_totals_combined_mv`

manage.py
vrajmohan: It is due to timing...it depends on `ofec_totals_combined_mv`

vrajmohan:
My intent is to unblock this quickly...all this should be refactored significantly when we use a migration tool...

manage.py
```diff
@@ -241,6 +241,8 @@ def update_schemas(processes=1):
...
+    execute_sql_file('data/converted_mvs/sched_a_by_state_recipient_totals.sql')
```
ccostino: Nice that these are in `update_schemas`, that makes it easier to ensure they're run appropriately!

PR #2670, opened 2017-09-21
Update Python and Node.js runtimes and README

ccostino:
This changeset updates the Python runtime to be `3.5.4` and makes some small adjustments to the README file to account for the recent switch from Travis CI to CircleCI.

This also makes adjustments for running with the latest version of Node.js, `8.6.0`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,4 +98,12 @@ jobs:
       - deploy:
           name: Deploy API
           command: |
+            export PATH=$HOME/bin:$PATH
+            export NVM_DIR="$HOME/.nvm"
+            [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+            [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+            . venv/bin/activate
+            nvm use default
+            npm run build
+            pip install -r requirements.txt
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -30,35 +30,42 @@ jobs:
...
-          name: Install Node.js
-          command: |
-            sudo apt-get update && curl -sL https://deb.nodesource.com/setup_6.x | sudo bash - && sudo apt-get install -y nodejs
+            sudo apt-get update -qq && sudo apt-get install -y build-essential postgresql-client
+            echo '/usr/lib/postgresql/9.6/bin/:$PATH' >> $BASH_ENV
 
       - restore_cache:
           keys:
-          - fec-api-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "package.json" }}
+          - fec-api-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "requirements-ci.txt" }}-{{ checksum "package.json" }}
```
LindsayYoung: I don't think we want to get rid of these, install all of those [here](https://github.com/18F/openFEC/pull/2670/files#diff-1d37e48f9ceff6d8030570cd36286a61R62) so we will only want to use the saved version if there are no changes in any of those files

LindsayYoung: Never mind!  This fixes that

LindsayYoung:
Thanks!

PR #2687, opened 2017-09-29
WIP Automated DB migration

vrajmohan:


data/refresh_materialized_views.sql
```diff
@@ -10,7 +10,7 @@ RETURNS INT AS $$
     FOR view IN SELECT matviewname FROM pg_matviews WHERE schemaname = schema_arg
     LOOP
       RAISE NOTICE 'Refreshing %.%', schema_arg, view.matviewname;
-      EXECUTE 'REFRESH MATERIALIZED VIEW CONCURRENTLY ' || schema_arg || '.' || view.matviewname;
```
LindsayYoung: Won't dropping concurrently cause downtime on refreshes?

vrajmohan: Thanks for catching it. The change was inadvertent - I was trying to populate the materialized views for the 1st time and `CONCURRENTLY` could not be used in that case.

manage.py
```diff
@@ -343,7 +213,51 @@ def refresh_materialized():
...
+        'large_aggregates': ['ofec_entity_chart_mv'],  # Anomaly
+        'candidate_fulltext': ['ofec_candidate_fulltext_mv'],
+        'totals_candidate_committee': ['ofec_totals_candidate_committees_mv']
+    }
+    graph = flow.get_graph()
+    for node in nx.topological_sort(graph):
+        for mv in materialized_view_names[node]:
+            logger.info('Refreshing %s', mv)
+            db.session.execute("REFRESH MATERIALIZED VIEW %s" % mv)
+
```
ccostino: Oh interesting!  You're leveraging the graph construction from `update_schemas` for the refresh mechanism here - do you think there'd be a way to leverage the graph as it's defined in `flow.py` directly though and have it run the `REFRESH MATERIALIZED VIEW` command instead of essentially copying the graph here?

This might throw a bit of a code smell, but what about `update_schemas` taking a flag instead that lets it know to do a full rebuild vs. just a refresh?

tests/common.py
```diff
@@ -132,11 +131,13 @@ def setUpClass(cls):
...
         cls.app_context.push()
         _reset_schema_for_integration()
-        try:
-            with open(os.devnull, 'w') as null:
-                subprocess.check_call(
-                    ['pg_restore', './data/subset.dump', '--dbname', TEST_CONN, '--no-acl', '--no-owner'],
-                    stdout=null,
-                )
-        except:
-            print('I am committing a try/ except crime')
```
ccostino: 😆 probably a good thing that this line is being eliminated!

manage.py
```diff
@@ -325,16 +204,7 @@ def update_all(processes=1):
...
-    update_functions(processes=processes)
-    load_districts()
-    load_pacronyms()
-    load_nicknames()
-    load_election_dates()
-    update_itemized('a')
-    update_itemized('b')
     update_itemized('e')
-    partition_itemized('a')
-    partition_itemized('b')
```
ccostino: Crazy to see all of these pieces being removed! :-)

manage.py
vrajmohan: Not sure what you mean by _copying the graph_? We are reusing the dependency relationships but need an additional `dict` to map filenames to view names as there is no consistent implicit mapping.

ccostino: Please disregard!  @vrajmohan and I went over this and it turns out that the dependency mapping is being used for this, the dictionary is separate for handling the view names vs. the file names.

ccostino:
@vrajmohan is it safe to close this PR since you have the new one up now?

vrajmohan:
I have kept this open only so that you or anyone else who commented on this PR needed to  review your comments until the new PR is merged.

vrajmohan:
Replaced by #2768 Feature/db migration

PR #2693, opened 2017-10-10
Match virtualenv name in CircleCI config to our ignore files

ccostino:
This changeset updates the CircleCI config to use the same name for the Python virtual environment it creates for testing and deployment as we have in the `.gitignore` and `.cfignore` files.  When the name does not match, the virtual environment directory and its contents are pushed up to the cloud.gov environment, which we do not want as it is extra cruft and could potentially cause environment conflicts/issues.


ccostino:
We're going to close this for now as something isn't right with the changes here.  We'll revisit this if it becomes a problem, but it was more of an issue in the CMS than the API; this was done for parity.

Issue #2683, opened 2017-09-29
Add MUR date to /legal/ API endpoint


PaulClark2:
Determine if  MUR date is available in elastic search/API
- [x] Ask Vraj if the data is already there
- [x] Add `open_date` and `close_date` to /legal/ - see Vraj's notes below 
- [x] Add to "description" like we do here: https://github.com/18F/openFEC/blob/develop/webservices/args.py#L163-L166
- [X] Add automated tests

vrajmohan:
The _indexed_ MUR data already has `open_date` and `close_date`: https://github.com/18F/openFEC/blob/develop/webservices/legal_docs/current_murs.py#L143. However, the API endpoint does not allow search by these fields. They need to be added in https://github.com/18F/openFEC/blob/develop/webservices/resources/legal.py#L155.

@AmyKort: Assuming that this task includes adding those fields to the search parameters?

AmyKort:
Yes! Thank you!

lbeaufort:
Thank you @vrajmohan for going over this with me! When it comes time, I'd like to take a shot at implementing this, if that's ok.

vrajmohan:
@lbeaufort, go for it! We are committing to add it in the current sprint, I will be happy to pair with you. Please take a look at https://github.com/18F/openFEC/blob/develop/webservices/resources/legal.py#L247-L261.

AmyKort:
Awesome! Thank you @vrajmohan !

lbeaufort:
Resolved by PR 18F/openFEC#2712. Thanks for all your help, @vrajmohan!

PR #2695, opened 2017-10-10
Feature/mur date filter

lbeaufort:
Related to #2683

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -181,9 +181,12 @@ query = {
     'mur_election_cycles': fields.Int(IStr, required=False, description='Filter MURs by election cycles'),
     'mur_document_category': fields.List(IStr, required=False,
         description='Filter MURs by category of associated documents'),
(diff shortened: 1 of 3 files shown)


webservices/args.py
```diff
@@ -181,6 +181,9 @@ def make_seek_args(field=fields.Int, description=None):
     'mur_election_cycles': fields.Int(IStr, required=False, description='Filter MURs by election cycles'),
     'mur_document_category': fields.List(IStr, required=False,
         description='Filter MURs by category of associated documents'),
+    'mur_open_date': fields.Date(required=False, description='Filter MURs by date opened'),
```
vrajmohan: Apologies for misleading you into thinking that these were only for documentation. They are actually parsed using https://webargs.readthedocs.io/en/latest/quickstart.html#validation.

Also, we need to define all 4 parameters with the right names: `mur_min_open_date`, `mur_max_open_date`, etc.

lbeaufort:
Closed in favor of https://github.com/18F/openFEC/pull/2712 per @vrajmohan

PR #2703, opened 2017-10-16
Added a slack function

LindsayYoung:
This slack function will slacks developers about the success or failure of the nightly updates. It should be easy to integrate for other things where we want more visibility .


LindsayYoung:
needs web hook to be added to env vars

lbeaufort:
I added the `SLACK_HOOK` credentials to `dev`, `stage`, and `prod`.

LindsayYoung:
Mandril is not shut off if we want to add some of that back

CONTRIBUTING.md
```diff
@@ -160,7 +160,7 @@ development tools:
 We currently have several monitoring tools in place:
 
 - New Relic, mainly for basic pings
-- A custom script powered by Mandrill to e-mail us nightly status reports in
+- A custom script powered by slacks us nightly status reports about
```
ccostino: Very nice to see the whitespace cleanup in here!

For this particular line, it looks like we're missing a word or two.  What about this to make it easier to understand?

`A custom script powered by Slack that posts a nightly status update about...`

manage.py
```diff
@@ -407,5 +408,13 @@ def load_efile_sheets():
         df.to_json(path_or_buf="data/" + table + ".json", orient='values')
 
 
+@manager.command
+def slack_message(message):
+    """ Sends a message to the bots channel. you can add this command to ping you when a task is done, etc.
+    run ./manage.py slack_message 'The message you want to post'
+    """
+    post_to_slack(message, '#bots')
```
ccostino: Oh awesome, this is a great idea!  Also, should we set the default channel as an environment variable instead?

requirements.txt
```diff
@@ -12,13 +12,11 @@ sqlalchemy-postgres-copy==0.3.0
 networkx==1.11
 newrelic==2.96.0.80
 SQLAlchemy==1.0.17
-mandrill==1.0.57
 icalendar==3.9.1
 GitPython==1.0.1
 gunicorn==19.7.1
 gevent==1.2.2
 webargs==0.18.0
-slacker==0.8.6
```
ccostino: Just confirming that we don't need this anymore as we're just posting directly over HTTPS to Slack now?

webservices/common/views.py
```diff
@@ -1,3 +1,4 @@
+import sys
```
ccostino: Looks like an extraneous `import` statement that's not used made it in here.

webservices/config.py
```diff
@@ -32,7 +32,6 @@ def get_cycle_end(year):
 REQUIRED_CREDS = (
     'SQLA_CONN',
     'FEC_SLACK_TOKEN',
-    'MANDRILL_API_KEY',
```
ccostino: Do we need to add the `SLACK_HOOK` environment variable here in addition to the `FEC_SLACK_TOKEN` that's already here, or no?

webservices/tasks/refresh.py
```diff
@@ -20,9 +22,10 @@ def refresh():
             manage.refresh_itemized()
             manage.refresh_materialized()
             download.clear_bucket()
+            slack_message = '*Success* nightly updates for {0} completed'.format(env.get_credential(NEW_RELIC_APP_NAME))
+            utils.post_to_slack(slack_message, '#bots')
```
ccostino: Just confirming that `NEW_RELIC_APP_NAME` is being used to identify the app here.  Also, as noted above, should we set a default environment variable for the channel name to post to?

webservices/tasks/refresh.py
```diff
@@ -20,9 +22,10 @@ def refresh():
...
         except Exception as error:
             manage.logger.exception(error)
-    try:
-        mail.send_mail(buffer)
-    except Exception as error:
-        logger.exception(error)
+            slack_message = '*ERROR* nightly update failed for {0}. Check logs.'.format(env.get_credential(NEW_RELIC_APP_NAME))
+            utils.post_to_slack(slack_message, '#bots')
```
ccostino: Same notes as above for the `NEW_RELIC_APP_NAME` and Slack channel name.

webservices/utils.py
```diff
@@ -335,3 +340,15 @@ def create_eregs_link(part, section):
     if section:
         url_part_section += '-' + section
     return '/regulations/{}/CURRENT'.format(url_part_section)
+
+def post_to_slack(message, channel):
+    response = requests.post(
+        env.get_credential('SLACK_HOOK'),
```
ccostino: Does the `SLACK_HOOK` environment variable need to be managed outside in the credential service?

webservices/utils.py
```diff
@@ -335,3 +340,15 @@ def create_eregs_link(part, section):
     if section:
         url_part_section += '-' + section
     return '/regulations/{}/CURRENT'.format(url_part_section)
+
+def post_to_slack(message, channel):
```
ccostino: As mentioned before, what if we set an environment variable for the default Slack channel to post to, but keep the parameter as you have to so we can override it.

webservices/utils.py
```diff
@@ -335,3 +340,15 @@ def create_eregs_link(part, section):
...
+    response = requests.post(
+        env.get_credential('SLACK_HOOK'),
+        data=json.dumps({
+            'text': message, 'channel': channel, 'link_names': 1,
+            'username': 'Ms. Robot', 'icon_emoji': ':robot_face:',
```
ccostino: ❤️ 😃

requirements.txt
LindsayYoung: yea it was too easy to justify importing a dependency

webservices/common/views.py
LindsayYoung: Nice catch!

webservices/tasks/refresh.py
LindsayYoung: I was a way to get text as to which environment it was running on without adding more variables

webservices/utils.py
LindsayYoung: Documenting here: We talked it over and it is easier to have a default that we don't have to configure across all environments, we could do an optional value you pass in the future.

webservices/utils.py
LindsayYoung: I think we need to add this to circle- I will do that now

LindsayYoung: done, now we can have things slack us in circle if we want it to

webservices/config.py
LindsayYoung: I thought about it but, I would not want not having the slack token to cause the build to fail. I assume we will notice that it is broken.

manage.py
ccostino: We opted not to after discussing this more - if we find that this'd be useful we can always do it in the future!

PR #2721, opened 2017-10-28
Set postgresql PATH correctly in CircleCI

vrajmohan:


ccostino:
A small note that we should do this for the CMS, too!

Issue #2706, opened 2017-10-20
Modify public.ofec_candidate_history_mv


PaulClark2:
See: https://github.com/18F/FEC/issues/4568

For Zink:  the inactive flag should be 2014 **f**, 2016 **f** and  2018 **t**

`public.ofec_candidate_history_mv` has 2014 **t**,, 2016 **t** and 2018 **f**

two_year_period | cand_election_yr |cand_id | name | candidate_inactive| active_through
-- | -- | -- | -- | -- | --
2014 | 2014 | H4MT01041 | ZINKE | t | 2018
2016 | 2016 | H4MT01041 | ZINKE | t | 2018
2018 | 2018 | H4MT01041 | ZINKE | f | 2018

Suggested fix: 
`LEFT JOIN disclosure.cand_inactive inactive ON fec_yr.cand_id::text = inactive.cand_id::text AND fec_yr.fec_election_yr < inactive.election_yr`

this shoud be  `...AND fec_yr.fec_election_yr >= inactive.election_yr`

Question: Does this solution handle cases were a candidate is active 2014, inactive in 2016 but active again in 2018?

ccostino:
@PaulClark2 this is a bit trickier than the proposed modification.  @hcaofec and I were looking into this more and determined that `>=` wouldn't be appropriate, but just a straight `=` might.  However, this did not fix the issue either and caused some other unintended side effects.

In looking at some more of the data, here are a couple examples of the current state:

```
select * from disclosure.cand_inactive where cand_id = 'H4MT01041';

  cand_id  | election_yr |          pg_date
-----------+-------------+----------------------------
 H4MT01041 |        2018 | 2017-03-05 09:23:56.845682
(1 row)


select candidate_id, candidate_election_year, name, candidate_status, candidate_inactive from ofec_candidate_history_mv where candidate_id = 'H4MT01041';

 candidate_id | candidate_election_year |     name      | candidate_status | candidate_inactive
--------------+-------------------------+---------------+------------------+--------------------
 H4MT01041    |                    2014 | ZINKE, RYAN K | C                | t
 H4MT01041    |                    2016 | ZINKE, RYAN K | C                | t
 H4MT01041    |                    2018 | ZINKE, RYAN K | C                | f
(3 rows)


select * from disclosure.cand_inactive where cand_id = 'H4MT01041';

  cand_id  | election_yr |          pg_date
-----------+-------------+----------------------------
[...]

PR #2727, opened 2017-11-01
modify sql script to correct value in column candidate_inactive

hcaofec:
sql script was modified to correct value in candidate_inactive column.
#2706

### DIFF
--- a/data/sql_updates/candidate_history.sql
+++ b/data/sql_updates/candidate_history.sql
@@ -78,7 +78,7 @@ left join elections using (cand_id)
 left join dates using (cand_id)
 left join disclosure.cand_inactive inactive on
     fec_yr.cand_id = inactive.cand_id and
-    fec_yr.fec_election_yr < inactive.election_yr
+    fec_yr.fec_election_yr = inactive.election_yr
 left join staging.ref_pty ref_party on fec_yr.cand_pty_affiliation = ref_party.pty_cd
 where max_cycle >= :START_YEAR and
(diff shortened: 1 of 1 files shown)


hcaofec:
@LindsayYoung @ccostino Please review

data/sql_updates/candidate_history.sql
```diff
@@ -78,7 +78,7 @@ left join elections using (cand_id)
 left join dates using (cand_id)
 left join disclosure.cand_inactive inactive on
     fec_yr.cand_id = inactive.cand_id and
-    fec_yr.fec_election_yr < inactive.election_yr
```
LindsayYoung: I think it is supposed to be < because we want to join with all years before the candidate was inactive.

hcaofec: Current logic is:      
" inactive.cand_id IS NOT NULL AS candidate_inactive 
  LEFT JOIN disclosure.cand_inactive inactive ON fec_yr.cand_id::text = inactive.cand_id::text AND fec_yr.fec_election_yr < inactive.election_yr  "
So when left joining disclosure.cand_inactive, it always returns an " inactive.cand_id " as long as the year is before the candidate was inactive.  And that puts "TRUE" to candidate_inactive column even though the candidate is actually active in that year.

Here is an example H4MT01041
@LindsayYoung

LindsayYoung: Thanks for finding and fixing this

PR #2737, opened 2017-11-04
Hotfix/restore queues

LindsayYoung:
Feature/restore queues #2729 was based on develop. I ported that here to have the same changes based on master


LindsayYoung:
sched B ran, sched A is running production now

PR #2738, opened 2017-11-06
Add AO status to legal AO endpoint

lbeaufort:
Along with https://github.com/18F/fec-cms/pull/1456, this resolves https://github.com/18F/fec-cms/issues/1390 

Originally intended to replace `is_pending` with `status` but that required a lot of front-end refactoring. Instead, I'm adding `status`.

-Verified that `ao.stage` data maps to AO status (0 = pending, 2 = withdrawn, 1 = final)
-add `status` display field and `ao_status` search parameter with values "Pending," "Withdrawn," and "Final."
-Slightly refactor SQL query - left join no longer needed, now need inner join to `aouser.ao` to pull in `stage`
-Verified searches still work in API where `ao_category=W` = this is based off document categories
-Tested AO search with `status` parameter - should be able to search for either
-Hard-coded withdrawn AO 2009-05 that has improperly been marked Final in internal database


webservices/legal_docs/advisory_opinions.py
```diff
@@ -14,24 +14,26 @@
...
-               FROM aouser.document
-               WHERE category IN ('Final Opinion', 'Withdrawal of Request')) AS finished
-        ON ao.ao_id = finished.finished
+        ao_parsed.ao_id,
+        ao_parsed.ao_no,
+        ao_parsed.name,
+        ao_parsed.summary,
+        ao_parsed.req_date,
+        ao_parsed.issue_date,
+        CASE ao.stage WHEN 0 THEN TRUE ELSE FALSE END AS is_pending,
```
vrajmohan: Would it be better to just retrieve `ao.stage` in the SQL query and do the `ao_status` and `ao_is_pending` evaluations in `python`?

vrajmohan: IMO, the logic would be simpler to understand in Python, and we could also keep this evaluation and the function `ao_status_to_stage` in the same Python file, so that they could be edited together (if there were future changes to be made, or future special cases to be handled like 2009-05).

lbeaufort: @vrajmohan works for me! Will do.

vrajmohan:
I made just one comment - it is a suggestion and I'll be happy to merge without it. Everything else looks great! Nice work, @lbeaufort!

webservices/legal_docs/advisory_opinions.py
```diff
@@ -14,24 +14,25 @@
...
-               FROM aouser.document
-               WHERE category IN ('Final Opinion', 'Withdrawal of Request')) AS finished
-        ON ao.ao_id = finished.finished
+        ao_parsed.ao_id,
+        ao_parsed.ao_no,
+        ao_parsed.name,
+        ao_parsed.summary,
+        ao_parsed.req_date,
+        ao_parsed.issue_date,
+        CASE ao.stage WHEN 2 THEN 'Withdrawn'
```
lbeaufort: @vrajmohan I removed the is_pending logic here, added the function on line 85, and added the is_pending logic on line 117.

vrajmohan: That's great! Sorry, I wasn't clear but ```CASE ao.stage WHEN 2 THEN 'Withdrawn'
            --Hard-code one withdrawn AO that has improperly been marked Final
            WHEN 1 THEN CASE ao.ao_no WHEN '2009-05' THEN 'Withdrawn' ELSE 'Final' END
            ELSE 'Pending' END AS stage``` would also be more clear in `python`.

webservices/legal_docs/advisory_opinions.py
```diff
@@ -83,6 +82,14 @@ def load_advisory_opinions(from_ao_no=None):
         ao_count += 1
     logger.info("%d advisory opinions loaded", ao_count)
 
+def ao_stage_to_pending(stage):
+    if stage == 'Pending':
+        is_pending = True
+    else:
+        is_pending = False
+
+    return is_pending
```
vrajmohan: In idiomatic `python`, the entire body of this function would be:
`return stage == 'Pending'`

webservices/legal_docs/advisory_opinions.py
vrajmohan: To elaborate, the function would be:
```
def ao_stage_to_status(ao_no, stage):
   AOS_WITH_INCORRECT_STAGE = ['2009-05']
   if stage == 2:
       return "Withdrawn"
   elif stage == 1:
       if ao_no in AOS_WITH_INCORRECT_STAGE:
           return "Withdrawn:
       else:
           return "Final"
  else:
      return "Pending"
```
This would make it easier to add more special cases.

lbeaufort:
@vrajmohan thanks for your suggestions! I tweaked it slightly to make `AOS_WITH_CORRECTED_STAGE` a dictionary lookup to make it easier to modify other AOs in the future.

PR #2740, opened 2017-11-08
Add handling for "C.F.R." in regulation citations and extra space in statute citations

lbeaufort:
Added REGEX handling for statutes formatted "52 U.S.C. § 30101" and "2 USC §437f" and regulations formatted "11 C.F.R. § 100.15"

Statute: `(?P<title>\d+)\s+U\.?S\.?C\.?\s+§*\s*(?P<section>\d+).*\.?`
Regulation: `(?P<title>\d+)\s+C\.?F\.?R\.?\s+§*\s*(?P<part>\d+)\.(?P<section>\d+)`

Added tests to confirm this worked

### DIFF
--- a/tests/test_advisory_opinions.py
+++ b/tests/test_advisory_opinions.py
@@ -38,6 +38,7 @@ def test_parse_ao_citations(text, ao_nos, expected):
     ("2 U.S.C. 441b, 441c, 441e", set([("2 U.S.C. 441b, 441c, 441e", 2, 441, 2, 441)])),
     (" 2 U.S.C. §437f", set([("2 U.S.C. §437f", 52, 30105, 2, 437)])),
     ("52 U.S.C. § 30101", set([("52 U.S.C. § 30101", 52, 30101, 52, 30101)])),
+    (" 2 USC §437f", set([("2 USC §437f", 52, 30105, 2, 437)])),
 ])
 def test_parse_statutory_citations(text, expected):
     assert parse_statutory_citations(text) == expected
--- a/tests/test_advisory_opinions.py
+++ b/tests/test_advisory_opinions.py
@@ -37,6 +37,7 @@ def test_parse_ao_citations(text, ao_nos, expected):
     ("52 U.S.C. 30116a", set([("52 U.S.C. 30116a", 52, 30116, 52, 30116)])),
     ("2 U.S.C. 441b, 441c, 441e", set([("2 U.S.C. 441b, 441c, 441e", 2, 441, 2, 441)])),
     (" 2 U.S.C. §437f", set([("2 U.S.C. §437f", 52, 30105, 2, 437)])),
+    ("52 U.S.C. § 30101", set([("52 U.S.C. § 30101", 52, 30101, 52, 30101)])),
 ])
 def test_parse_statutory_citations(text, expected):
     assert parse_statutory_citations(text) == expected
@@ -46,6 +47,7 @@ def test_parse_statutory_citations(text, expected):
     ("11 CFR §9034.4(b)(4)", set([(11, 9034, 4)])),
     ("11 CFR 300.60 and 11 CFR 300.62", set([(11, 300, 60), (11, 300, 62)])),  # TODO: Ranges
     ("11 CFR 300.60 through 300.65", set([(11, 300, 60)])),
+    ("11 C.F.R. § 100.15", set([(11, 100, 15)])),
 ])
 def test_parse_regulatory_citations(text, expected):
     assert parse_regulatory_citations(text) == expected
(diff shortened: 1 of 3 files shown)


webservices/legal_docs/advisory_opinions.py
```diff
@@ -58,18 +58,24 @@
     WHERE ao_id = %s
 """
 
-STATUTE_CITATION_REGEX = re.compile(r"(?P<title>\d+)\s+U.S.C.\s+§*(?P<section>\d+).*\.?")
-REGULATION_CITATION_REGEX = re.compile(r"(?P<title>\d+)\s+CFR\s+§*(?P<part>\d+)\.(?P<section>\d+)")
-AO_CITATION_REGEX = re.compile(r"\b(?P<year>\d{4,4})-(?P<serial_no>\d+)\b")
+STATUTE_CITATION_REGEX = re.compile(
+    r"(?P<title>\d+)\s+U.S.C.\s+§*\s*(?P<section>\d+).*\.?")
```
ccostino: I think you need to escape the `.` in the `U.S.C.` section, so that the regex looks like this:

`r"(?P<title>\d+)\s+U\.S\.C\.\s+§*\s*(?P<section>\d+).*\.?"`

You have this below in the citation regex, which is what you want!  What happens without escaping them is that you will still match a pattern that has different characters, e.g., `52 UxSxCx § 30101`.

Additionally, do you also need to be able to match `USC` without the `.` between each character?  If so, you'll need the extra `*` character after each `\.` like you have below in the citation regex.

tests/test_advisory_opinions.py
```diff
@@ -37,6 +37,7 @@ def test_parse_ao_citations(text, ao_nos, expected):
     ("52 U.S.C. 30116a", set([("52 U.S.C. 30116a", 52, 30116, 52, 30116)])),
     ("2 U.S.C. 441b, 441c, 441e", set([("2 U.S.C. 441b, 441c, 441e", 2, 441, 2, 441)])),
     (" 2 U.S.C. §437f", set([("2 U.S.C. §437f", 52, 30105, 2, 437)])),
+    ("52 U.S.C. § 30101", set([("52 U.S.C. § 30101", 52, 30101, 52, 30101)])),
```
ccostino: Excellent to see these test cases added! :-)

webservices/legal_docs/advisory_opinions.py
```diff
@@ -58,18 +58,24 @@
...
+
+REGULATION_CITATION_REGEX = re.compile(
+    r"(?P<title>\d+)\s+C\.*F\.*R\.*\s+§*\s*(?P<part>\d+)\.(?P<section>\d+)")
```
ccostino: Just noting that you have the `*` character here with `C.F.R.` and including the `*` modifies the preceding character and means "0 or more matches" - if this is intended, and looking at the test cases above it seems to be the case since you can have `CFR` or `C.F.R` match, then this is good!  I just wanted to make sure this was the desired outcome. :-)

webservices/resources/legal.py
```diff
@@ -227,7 +227,7 @@ def apply_ao_specific_query_params(query, **kwargs):
     citation_queries = []
     if kwargs.get('ao_regulatory_citation'):
         for citation in kwargs.get('ao_regulatory_citation'):
-            exact_match = re.match(r"(?P<title>\d+)\s+CFR\s+§*(?P<part>\d+)\.(?P<section>\d+)", citation)
+            exact_match = re.match(r"(?P<title>\d+)\s+C.*F.*R.*\s+§*(?P<part>\d+)\.(?P<section>\d+)", citation)
```
ccostino: Again, wouldn't this need to escape the `.` character so it matches it exactly?  `.` on its own means "match any one single character, regardless of what it is."

webservices/legal_docs/advisory_opinions.py
lbeaufort: Thanks, @ccostino! It seems like a good idea to me, but I didn't actually touch that part myself. @vrajmohan - do you see any problem with adding this to the U.S.C section? `U\.*S\.*C\.*`

webservices/legal_docs/advisory_opinions.py
lbeaufort: @ccostino yes, thank you! That's my understanding of the desired behavior.

webservices/resources/legal.py
lbeaufort: Yes, thanks! I forgot to update this section.

webservices/legal_docs/advisory_opinions.py
vrajmohan: Good catch, @ccostino! We may or may not have the ".", and if we have it we have only one of them, so it should be `U\.?S\.?C\.?`. Ditto for CFRs.

vrajmohan:
Nice work, @lbeaufort!

lbeaufort:
@ccostino and @vrajmohan thanks for your feedback! I made the changes and added a test, so this should be ready to go!

PR #2741, opened 2017-11-09
Update nvm and Node.js setup in CircleCI

ccostino:
This changeset updates the nvm and Node.js setup to use the latest versions of each.  Node.js has a new LTS version that we are now leveraging, so we want to make sure CircleCI is using it as well in our builds.  At the time of this PR, that is version 8.9.1 (Carbon).


ccostino:
Closing in favor of doing this against the current release branch.

PR #2757, opened 2017-11-15
increasing workers and resources, taing off debug, still working on b…

LindsayYoung:
adding more workers and instances

turning off debug

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -73,7 +73,7 @@ def sqla_conn_string():
 
 
 app = Flask(__name__)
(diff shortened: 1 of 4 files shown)


webservices/rest.py
```diff
@@ -73,7 +73,7 @@ def sqla_conn_string():
 
 
 app = Flask(__name__)
-app.debug = True
+# app.debug = True
```
afeld: You might want to have a way to enable this during development, but this works for now.

PR #2759, opened 2017-11-16
[MERGE WITH GIT FLOW] Improve logging setup and update New Relic and requests

ccostino:
**MERGE WITH GIT FLOW**

This changeset modifies the API logging configuration to ensure that as much info is possible is coming through into our logs to help provide visibility into what is happening at any given moment.  This includes a change to the Flask application configuration as well as the gunicorn startup command.

Furthermore, it also updates the newrelic and requests packages to ensure that we are using the latest versions and accounting for recent bugfixes and enhancements.

Information around this change was pulled from a few sources:

* https://github.com/benoitc/gunicorn/issues/379
* http://docs.gunicorn.org/en/latest/settings.html#loglevel
* http://flask.pocoo.org/docs/0.12/errorhandling/
* https://docs.cloudfoundry.org/devguide/deploy-apps/streaming-logs.html#types
* https://docs.python.org/3.5/library/logging.html

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,2 +1,2 @@
 python manage.py cf_startup
-gunicorn -k gevent -w 4 webservices.rest:app
+gunicorn --access-logfile - --error-logfile - --log-level info -k gevent -w 4 webservices.rest:app
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ Flask-SQLAlchemy==2.1
 python-dateutil==2.4.2
 sqlalchemy-postgres-copy==0.3.0
 networkx==1.11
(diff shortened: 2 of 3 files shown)


LindsayYoung:
Thanks for chiming in @sharms! We will add that as a follow up pr

ccostino:
@sharms Ah I see what mean, derive it from what we're always setting - is `$LOGLEVEL` something always set and available in the CF/cloud.gov environment?

ccostino:
Note though that for `gunicorn`, [the setting is `info`, not `INFO`](http://docs.gunicorn.org/en/latest/settings.html#loglevel).

PR #2763, opened 2017-11-20
Feature/clean up

LindsayYoung:
Removes outdated stuff in this repo.

We had an early version of the SSP and controls as code in out repo, but those are not the official ones, or up to date.

We used to use vagrant for local development, but we have not in years, so I am deleting that.

For review please double check I did not sweep anything we still use into this commit-
 the other stuff is really straight forward https://github.com/18F/openFEC/commit/246b3495a485b242643435ef0dadc67cea3aff04


webservices/resources/load.py
```diff
@@ -1,23 +0,0 @@
...
-        if data['api_key'] in write_authorized_tokens:
-            elasticsearch.helpers.bulk(es, (dict(_op_type='index', _source=doc, id=doc['doc_id']) for doc in data['docs']),
-              index='docs',
-              doc_type=data['doc_type'])
-
-            return {'success': True}
-        else:
-            msg = {'success': False, 'message':
-                "Your API token has not been authorized to write data to this application."}
-            return msg, 401
```
ccostino: @vrajmohan, we just want to double check that this is safe to remove.

vrajmohan: I didn't know that this existed. It should be OK to remove. What is SSP?

PaulClark2: SSP - system security plan

ccostino: Sounds like this is good to go, then; approving and merging!

PR #2768, opened 2017-11-22
Feature/db migration

vrajmohan:
I took the liberty of opening a fresh PR for DB migrations to have a cleaner history. There were several changes to deal with due to:
- Using a copy of the `prod` schema
- Merging changes from the "restoration of queues for processing itemized schedules"

I have left the https://github.com/18F/openFEC/pull/2687 PR  open in case you wanted to review your previous comments


ccostino:
@vrajmohan looks like we had a botched deploy in `dev` that wasn't cleaned up at the time of the last build.  It looks like this should just be a matter of rebuilding the job in CircleCI but wanted to double check!

.circleci/config.yml
```diff
@@ -33,6 +33,14 @@ jobs:
             sudo apt-get update -qq && sudo apt-get install -y build-essential postgresql-client
             echo 'export PATH=/usr/lib/postgresql/9.6/bin/:$PATH' >> $BASH_ENV
 
+      - run:
+          name: Install flyway
+          command: |
+            sudo apt-get update && sudo apt-get install -y default-jdk
+            mkdir flyway
+            curl -s https://repo.maven.apache.org/maven2/org/flywaydb/flyway-commandline/4.2.0/flyway-commandline-4.2.0-linux-x64.tar.gz | tar zxv -C flyway
```
ccostino: How often does `flyway` release updates?  It looks like roughly once a month for patch level updates, and every few months for minor and major releases.  I see that [version 5.0.0](https://flywaydb.org/documentation/releaseNotes) is in the works, which appears to include some updates around the minimum supported version of Java and Maven, JDBC driver updates, etc., all of which are probably important to stay on top of given security updates, vulnerability patches, etc.

I also ask because I'm wondering how we can best account for updates to this tool and make sure it's relatively easy/straightforward to maintain in our build pipeline.

.circleci/config.yml
```diff
@@ -71,6 +79,10 @@ jobs:
           name: Ensure database is available
           command: dockerize -wait tcp://localhost:5432 -timeout 1m
 
+      - run:
+          name: Create database roles
+          command: |
+              psql $SQLA_TEST_CONN -f data/create_roles.sql
```
ccostino: As noted in our overview conversation, we'll probably want to create an admin role specific for flyway for ATO/audit purposes.

/cc @LindsayYoung:  other than creating the role in the RDS environments and documenting that, is there anything else we need to do that I'm forgetting?

README.md
```diff
@@ -553,3 +554,20 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
...
+#### Installing `flyway`
+Download the version without JRE, e.g. `flyway-commandline-4.2.0.tar.gz` and expand the archive. Add `<target_directory>/flyway/flyway-4.2.0` to your `PATH`.
+
+#### How `flyway` works
+All database schema modification code is checked into version control in the directory `data/migrations` in the form of SQL files that follow a strict naming convention - `V<version_number>__<descriptive_name>.sql`. `flyway` also maintains a table in the target database called `schema_version` which tracks the migration versions that have already been applied.
+
+`flyway` supports the following commmands:
+- `info` compares the migration SQL files and the table `schema_version` and reports on migrations that have been applied and those that are pending.
+- `migrate` compares the migration SQL files and the table `schema_version` and runs those migrations that are pending.
+- `baseline` modifies the `schema_version` table to indicate that the database has already been migrated to a baseline version.
```
ccostino: It might be useful to link to the full documentation after this list, e.g.,:

> For more information and a complete list of commands, [please see the full documentation of the command line tool](https://flywaydb.org/documentation/commandline/)

It took me a few clicks around the site to find that page, it's not completely straightforward!

README.md
```diff
@@ -553,3 +554,20 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
...
+
+#### Deployment from CircleCI
+flyway is installed in `CircleCI`. During the deployment step, `CircleCI` invokes `flyway` to migrate the target database (depending on the target space). For this to work coirrectly, connection URLs for the target databases have to be stored as environment variables in CircleCI under the names `FEC_SQLA_CONN_DEV`, `FEC_SQLA_CONN_STAGE` and `FEC_SQLA_CONN_PROD`. The connection URL has to strictly adhere to the structure `postgresql://<username>:<password>@<hostname>:<port>/<database_name>`. Note that the database_name should be specified explicitly, unlike URLS for SQLAlchemy connections.
```
ccostino: Spelling error:  `coirrectly` => `correctly`.

Thanks for clearly explaining how this works!

data/create_roles.sql
```diff
@@ -0,0 +1,6 @@
+CREATE USER aomur_usr;
+CREATE USER fec;
+CREATE USER fec_read;
+CREATE USER openfec_read;
+CREATE USER real_efile;
+CREATE USER real_file;
```
ccostino: Just another note about the `flyway`-specific account to be added here too, if needed.

data/migrations/V9__disclosure_cand_valid_fec_yr.sql
```diff
@@ -0,0 +1,37 @@
+SET search_path = disclosure, pg_catalog;
```
ccostino: I noticed that this line is in most of the migrations; is `flyway` setting/adding this automatically when generating the migrations?  If so, I wonder why it doesn't just use the fully qualified name in the SQL itself just to be more explicit.  Is there an option that can control or change this?  No worries if not, just curious about it!

jdbc_utils.py
```diff
@@ -0,0 +1,10 @@
+import re
+def to_jdbc_url(dbi_url):
+    DB_URL_REGEX = re.compile(r'postgresql://(?P<username>[^:]*):?(?P<password>\S*)@(?P<host_port>\S*)$')
+    match = DB_URL_REGEX.match(dbi_url)
+    if match:
+        jdbc_url = 'jdbc:postgresql://{}?user={}'.format(
+            match.group('host_port'), match.group('username'))
+        if match.group('password'):
+            jdbc_url += '&password={}'.format(match.group('password'))
+        return jdbc_url
```
ccostino: Would it make more sense to just include this method in [`webservices/utils.py`](https://github.com/18F/openFEC/blob/develop/webservices/utils.py) instead of introducing a new file?  I'm not 100% against this, but given that we already have a `utils.py` file that seems like a natural place for this to live.

tasks.py
```diff
@@ -142,7 +143,9 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    ('dev', lambda _, branch: branch == 'develop'),
+#   DON'T MERGE - THIS IS FOR TESTING @TODO: Restore
+#    ('dev', lambda _, branch: branch == 'develop'),
```
ccostino: Flagging this in PR review too to make sure we don't forget about it!

tests/common.py
```diff
@@ -11,37 +12,46 @@
...
     rest.db.engine.execute('create schema public;')
 
+def _drop_schema():
+    rest.db.engine.execute('drop schema if exists public cascade;')
+    for schema in SCHEMAS:
+        rest.db.engine.execute('drop schema if exists %s cascade;' % schema)
+
+def _create_schema():
+    for schema in SCHEMAS:
+        rest.db.engine.execute('create schema %s;' % schema)
```
ccostino: Much cleaner, nice! :-)

tests/common.py
```diff
@@ -132,11 +146,9 @@ def setUpClass(cls):
...
         cls.app_context.push()
         _reset_schema_for_integration()
-        try:
-            with open(os.devnull, 'w') as null:
-                subprocess.check_call(
-                    ['pg_restore', './data/subset.dump', '--dbname', TEST_CONN, '--no-acl', '--no-owner'],
-                    stdout=null,
-                )
-        except:
-            print('I am committing a try/ except crime')
```
ccostino: 😆 Nice to see this get removed finally. :-)

tests/test_integration.py
```diff
@@ -527,10 +520,31 @@ def get_existing():
...
-            size=0,
-            cycle=2016,
-        ).first()
+        def get_existing():
+            return models.ScheduleABySize.query.filter_by(
+                size=0,
+                cycle=2016,
+            ).order_by(
+                models.ScheduleABySize.committee_id,
+            ).first()
```
ccostino: Out of curiosity, why the method definition inside of another method?  If I understand correctly, this block of code is defined as a method so that it executes and therefore does the database lookup at the time of the variable assignment down below so that we're guaranteed to retrieve the latest updates in the database at that point in the test, correct?  If so, is this just a matter of scoping?  I've always been a bit fuzzy on Python's flexibility in allowing this kind of design pattern.

tests/test_itemized.py
```diff
@@ -372,37 +372,6 @@ def test_image_number_range(self):
...
-            ScheduleA.contributor_id,
-            ScheduleA.committee_id,
-        )
-
-        rows = ScheduleA.query.all()
-        self.assertEqual(rows, individuals + earmarks)
-
-        rows = ScheduleA.query.filter(is_individual).all()
-        self.assertEqual(rows, individuals)
-
```
ccostino: By removing these from the `itemized` tests and moving them to the `integration` tests, does this help speed up the overall test execution and/or cut down on weird behaviors due to underlying test dependencies?  I don't think we had many, if at all, but was always wary of these things being spread out.

manage.py
```diff
@@ -373,25 +294,65 @@ def update_all(processes=1):
...
+        'totals_pac_party': ['ofec_totals_pacs_parties_mv', 'ofec_totals_pacs_mv', 'ofec_totals_parties_mv'],
+        'large_aggregates': ['ofec_entity_chart_mv'],
+        'candidate_fulltext': ['ofec_candidate_fulltext_mv'],
+        'totals_candidate_committee': ['ofec_totals_candidate_committees_mv']
+    }
+    graph = flow.get_graph()
+    for node in nx.topological_sort(graph):
+        for mv in materialized_view_names[node]:
+            logger.info('Refreshing %s', mv)
+            db.session.execute("REFRESH MATERIALIZED VIEW %s" % mv)
```
ccostino: One of the issues we ran into in the past was that due to some of the dependencies between the materialized views and tables (one of the Schedule A aggregates in particular), just calling `REFRESH MATERIALIZED VIEW` wasn't enough, hence the need to go through and recreate all of them and do the rename.

Will this work correctly given that they are being refreshed in the correct order?  Do we need to worry about the indexes at all and perform a reindex on them?

I definitely like that this is much simpler and leveraging the database itself more, but just want to be sure it'll work as we need it to!

webservices/tasks/refresh.py
```diff
@@ -19,7 +19,6 @@ def refresh():
             manage.update_functions()
```
ccostino: We need to remove this call since this method is no longer defined in `manage.py`; this is what's currently causing the nightly updates to fail in the `dev` environment. :-)

manage.py
```diff
@@ -373,25 +294,65 @@ def update_all(processes=1):
...
     load_pacronyms()
     load_nicknames()
     load_election_dates()
-    update_itemized('a')
-    update_itemized('b')
-    update_itemized('e')
-    partition_itemized('a')
-    partition_itemized('b')
-    rebuild_aggregates(processes=processes)
     update_schemas(processes=processes)
```
ccostino: `update_schemas` is removed above, so the call should be removed here as well, right?

In fact, what should be in place for the removed methods here if we need to do a complete refresh?  I don't see a means of rebuilding the API-managed relations and objects like we had previously; do we need to re-run the migrations or a subset of them, or is there another way we should do this?  Or did I miss something entirely?

webservices/tasks/refresh.py
```diff
@@ -19,7 +19,6 @@ def refresh():
...
             manage.update_aggregates()
             manage.refresh_itemized()
```
ccostino: This also needs to be removed by the looks of it.

webservices/tasks/refresh.py
```diff
@@ -19,7 +19,6 @@ def refresh():
...
-            manage.update_itemized('e')
             manage.update_schemas()
```
ccostino: And this removed as well, but should be replaced by `manage.refresh_materialized()`, right?

README.md
```diff
@@ -553,3 +554,20 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
 Database mirrors/replicas are supported by the API if the `SQLA_FOLLOWERS` is set to one or more valid connection strings.  By default, setting this environment variable will shift all `read` operations to any mirrors/replicas that are available (and randomly choose one to target per request if there are more than one).
 
 You can optionally choose to restrict traffic that goes to the mirrors/replicas to be the asynchronous tasks only by setting the `SQLA_RESTRICT_FOLLOWER_TRAFFIC_TO_TASKS` environment variable to something that will evaluate to `True` in Python (simply using `True` as the value is fine).  If you do this, you can also restrict which tasks are supported on the mirrors/replicas.  Supported tasks are configured by adding their fully qualified names to the `app.config['SQLALCHEMY_FOLLOWER_TASKS']` list in order to whitelist them.  By default, only the `download` task is enabled.
+
+### Database migration
+`flyway` is the tool used for database migration.
+
+#### Installing `flyway`
+Download the version without JRE, e.g. `flyway-commandline-4.2.0.tar.gz` and expand the archive. Add `<target_directory>/flyway/flyway-4.2.0` to your `PATH`.
```
ccostino: We should supply a link here, e.g., [https://flywaydb.org/getstarted/download`](https://flywaydb.org/getstarted/download) and specify that it's the command line tool you're looking for.

ccostino:
Another question, @vrajmohan:  When we're first setting up the project, should we still perform the `pg_restore` command or just create the database itself and run the migrations after?

ccostino:
Last question:  even though we are downloading the version of the tool without the JRE, we still need to install the JDK locally correct?  If so, what is the recommended way of doing that in our case?

.circleci/config.yml
vrajmohan: Major releases once or twice a year appears to be the frequency. I agree that we should keep on top of updates and security patches. As FEC starts to use `flyway` for more of their projects, it will be more important to do this. 

As we use `flyway` in its sweet spot (we don't rely on any edge case behavior), we should be able to easily upgrade to patch releases as well.

.circleci/config.yml
vrajmohan: Agreed.

README.md
vrajmohan: Done

jdbc_utils.py
vrajmohan: I thought of adding it to `webservices/utils.py` as well, but decided not to as this had nothing to do with webservices. I am open to suggestions.

data/migrations/V9__disclosure_cand_valid_fec_yr.sql
vrajmohan: The migrations were obtained by using `pg_dump` to obtain a schema-only SQL dump from `prod`. `pg_dump` sets `search_path` as opposed to schema-qualifying object names.

Also, this explains how we ended up with ~40 migrations. I started with a ~30,000 line dump file and broke it up into logical groups as well as making it easier to edit, by having smaller files!

tests/test_integration.py
vrajmohan: It is just a matter of scoping. It allows us to have multiple methods named `get_existing`, each specific to the test within which it is defined. e.g. in `test_update_aggregate_asize_existing` and `test_update_aggregate_size_existing_merged`.

tests/test_itemized.py
vrajmohan: Ideally all the tests in `test_itemized.py` should be moved into `test_integration`, or better still, into a directory like `tests/integration`. These tests use the test database, but bring their own schema by creating it from SQLAlchemy models. Instead they should be refactored to use the schema built using `flyway`. This was too much work to undertake.

I took the shortcut of only moving one test, `test_filter_individual_sched_a` as it uses the SQL function `is_individual`.

.circleci/config.yml
ccostino: Caught up with @LindsayYoung about this, we'll want to either leverage the existing API account we have created in RDS, or create a new one just for this.

data/migrations/V9__disclosure_cand_valid_fec_yr.sql
ccostino: Ah, okay, thank you for explaining! 🎉

jdbc_utils.py
ccostino: Okay, if you had also given it some thought and that was the reasoning, I'm 👍 on it. :-)

tests/test_integration.py
ccostino: Got it, thank you!

tests/test_itemized.py
ccostino: Oh good point, we've discussed this before.  Yes, that would be far too much work for in here, but a nice small start to help push it forward!

manage.py
vrajmohan: I think that this code ensures that the materialized views are refreshed in the correct order. We should test it on `dev` to make sure.

README.md
vrajmohan: Done

ccostino:
@vrajmohan, I think I forgot to ask this here or in one of our conversations as it's something @LindsayYoung had brought up in one of our joint reviews:  what happens in the case of the migration execution timing out or failing for some reason?  Is there a timeout setting for running any given migration?  What about a retry mechanism?  Do we have a way of ensuring that we won't end up in a broken state if the deploy has a hiccup?

I am assuming that transactions are a part of this and would play a role in some capacity, but I wanted to make sure that we took this into account fully, as well!

manage.py
vrajmohan: I have tested it on `dev` and it works correctly. I will be changing the `REFRESH MATERIALIZED VIEW` to add `CONCURRENTLY` for lesser interference with other `SELECT` queries.

vrajmohan:
Good question on migration failures and timeouts. Yes, it is indeed managed with transactions. See https://flywaydb.org/documentation/migration/ (specifically the 2nd bullet point in the section _Migration States_)

vrajmohan:
I have added documentation (and my opinion) on whether or not to download the version with a JRE).

ccostino:
A few more notes from testing locally and conversations @vrajmohan and I have had:

* Some adjustments might be needed to the documentation language around the installation - the non-JRE version is the recommended way for the reasons listed, but the JRE-included version is just fine too and works if you don't care for managing Java locally and/or run into installation issues. :-)
* Adding a command in the `tasks.py` file for `invoke` to use that can encapsulate creating the roles first, then running migrations, and finally loading the subset data will be a huge help in making this easy!
* Some documentation needs to be added that the location of the migrations is required for this to work (see the current `run_migrations` method in `tasks.py` of this PR).
* We discussed creating a data-only dump to be able to load in after the schema; this has the added benefit of serving as another validation test that the schema we end up with after the migrations is what we had prior to this change, because all of the test data should just fit right in as is. :-)

Thank you so much for all of the hard work and your patience and help with this, @vrajmohan!

vrajmohan:
I've addressed points 1 and 3 above, as well as instructions for creating roles. As far as creating a data-only dump and providing a command to load it all, I agree that it should be done. However, I suggest doing it as a separate (but immediately taken up) follow on story so that we can get this long-running work in as soon as possible.

ccostino:
@vrajmohan would creating the data-only dump also include creating the additional `invoke` task to help with that process?  If so then I agree, especially as other work coming in (e.g., audit search) will have to be synced up with the migrations.  Thanks!

README.md
```diff
@@ -94,21 +95,35 @@ To disable, run:
...
+
+#### Create a development database
+Before you can run this project locally, you'll need a development database.
+
+To create the development database, run:
+
+```
+createdb cfdm_test
+```
+
```
ccostino: Much clearer in this section, thank you!

README.md
```diff
@@ -553,3 +568,34 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
...
+#### Deployment from CircleCI
+flyway is installed in `CircleCI`. During the deployment step, `CircleCI` invokes `flyway` to migrate the target database (depending on the target space). For this to work correctly, connection URLs for the target databases have to be stored as environment variables in CircleCI under the names `FEC_SQLA_CONN_DEV`, `FEC_SQLA_CONN_STAGE` and `FEC_SQLA_CONN_PROD`. The connection URL has to strictly adhere to the structure `postgresql://<username>:<password>@<hostname>:<port>/<database_name>`. Note that the database_name should be specified explicitly, unlike URLs for SQLAlchemy connections.
+
+#### Running `flyway` manually
+You may need to run `flyway` manually in order to test migrations locally, or to troubleshoot migrations in production. There are 2 required parameters:
+- `-url` specifies the database URL. This is a JDBC URL of the form `jdbc:postgresql://<hostname>:<port>/<database>?user=<username>&password=<password>`.
+- `locations` specifies the directory where the migrations are stored. This is a value of the form `filesystem:<directory-path>`. In our case, if run from the project root, it would be `-locations=filesystem:data/migration`.
```
ccostino: 👏 this is a huge help, thank you!

vrajmohan:
Yes, it would. I have created #2788.

ccostino:
@vrajmohan there are two other things I noticed when trying to run the tests locally:

* If I just run the command `py.test` in the root project directory, I receive an error with the `slate` library being unable to import its own `utils` module.  I don't see any changes in the PR around the requirements so I'm not sure why that'd be the case (even after blowing away my virtualenv and recreating it to be sure).  Running `py.test tests/` works fine though.
* It looks like if you run the tests locally, you need to explicitly set the `SQLA_TEST_CONN` environment variable now.  The `to_jdbc_url` method is expecting to find the username and password as a part of the connection string when building the JDBC version, and if you don't have it explicitly set it fails to do this and ends up returning `None` for the URL.  It doesn't look like it'd be possible to leverage a connection string like `postgreql:///cfdm_unit_test` or `postgresql://:@/cfdm_unit_test` due to `flyway`'s requirements for a username and password, and specific connection formatting (e.g., `localhost:5432/cfdm_unit_test`, so we should probably update the instructions to include setting `SQLA_TEST_CONN` explicitly with your local connection info.

I'd like to know why just running `py.test` in the root project directory no longer works, but it appears to work just fine in CircleCI so it's likely just a local setup issue on my side.  The testing string documentation update should be the last thing needed here to wrap this up! :-)

README.md
```diff
@@ -564,3 +581,34 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
...
```
ccostino: `locations` is missing the dash at the start of this bullet, should be `-locations`.

README.md
```diff
@@ -564,3 +581,34 @@ Sorting fields include a compound index on on the filed to sort and a unique fie
...
+- `migrate` compares the migration SQL files and the table `schema_version` and runs those migrations that are pending.
+- `baseline` modifies the `schema_version` table to indicate that the database has already been migrated to a baseline version.
+
+For more information, see [Flyway documentation](https://flywaydb.org/documentation/).
+
+#### Running tests
+Tests that require the database will automatically run `flyway` migrations as long as `flyway` is in `PATH`.
+
+#### Deployment from CircleCI
+flyway is installed in `CircleCI`. During the deployment step, `CircleCI` invokes `flyway` to migrate the target database (depending on the target space). For this to work correctly, connection URLs for the target databases have to be stored as environment variables in CircleCI under the names `FEC_SQLA_CONN_DEV`, `FEC_SQLA_CONN_STAGE` and `FEC_SQLA_CONN_PROD`. The connection URL has to strictly adhere to the structure `postgresql://<username>:<password>@<hostname>:<port>/<database_name>`. Note that the database_name should be specified explicitly, unlike URLs for SQLAlchemy connections.
```
ccostino: We'll need to update these so they refer to the correct environment variable names:  `FEC_MIGRATOR_SQLA_CONN_DEV`, `FEC_MIGRATOR_SQLA_CONN_STAGE`, and `FEC_MIGRATOR_SQLA_CONN_PROD`.

PR #2781, opened 2017-11-30
Remove duplicate AO statute citations

lbeaufort:
Addresses https://github.com/18F/openFEC/issues/2762

For AO statute citations, prior citations and additional text were being added to the matches set [here](https://github.com/18F/openFEC/pull/2781/files#diff-324d3f17f824202b1e914e4bedbc79b5R271). These fields weren't being used and were creating duplicate statute citations for AO's.

AO 2012-36 [local link](http://127.0.0.1:8000/data/legal/advisory-opinions/2012-36/) now only showing the 2 citations. 

## Before
image

## After

- [x] Verify that we don't need AO statute citation former title/section anywhere else in the app
- [x] If we verify that we don't need them, remove statute text and former title/section from our `test_advisory_opinions.py` tests
- [x] Add tests for duplicate citations (regs and statutes)

Discovered bug with 2 USC mapping with letters, got most the way done with it, then it proved to be complicated, so rolled back and split up the work into a two issues - see https://github.com/18F/openFEC/issues/2803

### DIFF
--- a/tests/test_advisory_opinions.py
+++ b/tests/test_advisory_opinions.py
@@ -41,6 +41,7 @@ def test_parse_ao_citations(text, ao_nos, expected):
     (" 2 U.S.C. §437f", set([(52, 30105)])),
     ("52 U.S.C. § 30101", set([(52, 30101)])),
     (" 2 USC §437f", set([(52, 30105)])),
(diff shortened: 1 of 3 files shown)


vrajmohan:
Looks great!

PR #2782, opened 2017-11-30
Increase the threshold count to return the exact count for small queries

pkfec:


pkfec:
@LindsayYoung  Please review.

LindsayYoung:
did some quick testing and it looks good to me

PR #2783, opened 2017-12-01
Feature/revert calendar

LindsayYoung:
This will shift the source of the calendar endpoint to the he three fecapp.cal_event tables (known as the old calendar) This will give people a more consistent data entry experience and give FEC more control over the display of events.

- This can be a view now, so additions and changes should show up more quickly
- We lost the state filter because the data does not support it

The work is done but this requires filter changes on the front end so don't merge it until that is ready.


LindsayYoung:
Date filter works but some of the other fields are not working yet

fec-jli:
We have pull to local and run api 
"state" filters are no longer exist in /calendar-dates/ and  /calendar-dates/export/

LindsayYoung:
Got everything working- there is a change where we do need use calendar_category_id for category filtering, but it works pretty similarly 

Ready for you to look it over and merge Monday @fec-jli 

Thanks!

LindsayYoung:
@vrajmohan or @ccostino wanted to see if this needed to be refactored to work with the new migration work

ccostino:
@LindsayYoung yes, given the SQL changes.  We'll have to write a new migration to account for the changes at this point, right @vrajmohan?

vrajmohan:
Yes, it would need to be written as a migration.

LindsayYoung:
thanks for the flow catch- I thought I had gotten rid of the flow reference!

PR #2789, opened 2017-12-06
Build swagger-ui from within openFEC

xtine:
## Build Swagger from within openFEC
Addresses https://github.com/18F/openFEC/issues/2770.

I did some research and attempted to update swagger-ui to 3.x, however the differences were too great from our current swagger schema to warrant the time we would spend to get it looking right. Plus the layout of 3.x and 2x is almost completely different and we would need a bunch of extra customizations in addition to the accessibility tweaks we have for 2.x specifically.

Despite this, I did simplify the build process of our custom swagger-ui setup so it can be more maintainable in its current 2.x form:
- Move all relevant custom build files from https://github.com/18F/swagger-ui to a directory within this repo
- Edit and refactor Gulp tasks as defined in `gulpfile.js` to build swagger-ui into the `static/swagger-ui` directory to serve asset files
- Edit and simplify the main swagger template file at `webservices/swagger-ui.html` to contain references to new static file location
  - Moved JS scripts to bottom of template for faster loading
- Place all vendor JS files (jquery, handlebars, etc) in `static/swagger-ui/vendor`, which then can be easily updatable/maintainable in the future
- Edited the footer of API docs (to remove reference to 18F and out of date version info): screen shot 2017-12-05 at 9 49 21 pm
- Update README.md with information about swagger-ui file locations, customization and build instructions.

Once this PR is merged, then we can begin steps to deprecate https://github.com/18F/swagger-ui as a dependency.


README.md
```diff
@@ -7,7 +7,6 @@
 [![Code Climate](https://img.shields.io/codeclimate/github/18F/openFEC.svg)](https://codeclimate.com/github/18F/openFEC)
 [![Dependencies](https://img.shields.io/gemnasium/18F/openFEC.svg)](https://gemnasium.com/18F/openFEC)
 
-![Swagger validation badge](https://online.swagger.io/validator?url=https://api.open.fec.gov/swagger)
```
xtine: Currently our swagger spec file fails and doesn't display correctly, so I've removed it until we can figure out what's going on.

ccostino: This hasn't worked in the longest time, I've tried to fix it before and couldn't.  I wonder if this requires an update to the latest version to have it work?

README.md
```diff
@@ -70,7 +69,6 @@ Use `npm` to install JavaScript dependencies:
 
 ```
 nvm use --lts
-npm install -g swagger-tools
```
ccostino: AWESOME. :-)

README.md
```diff
@@ -209,6 +207,17 @@ redis-server
...
+We are using a customized 2.x version of swagger-ui to display our API developer documentation.
+
+### Template and Swagger spec file
+The base template for swagger-ui is located at `webservices/templates/swagger-ui.html`. The `{{ specs_url }}` template variable points to `http://localhost:5000/swagger` that is the swagger spec file for FEC specific model definitions and schema. Compiled and vendor assets are served from `static/swagger-ui/`.
+
+### Custom Swagger setup and build
+The swagger-ui package is within the `swagger-ui` directory. The `hbs` folder contains handlebars templates with 18F customizations, as do the files contained in the `js` and `less` folders. However, the `js/swagger-client.js` is the base v2.1.32 swager-ui file.
+
+All these files are then built and compiled via the `npm run build` command that runs Gulp tasks. Any modification should be done in the files in `swagger-ui` that will then be compiled and served in the `static/swagger-ui/` folder.
+
```
ccostino: Thank you for all of these details!

package.json
```diff
@@ -1,9 +1,19 @@
...
+    "gulp-less": "^3.3.2",
+    "gulp-order": "^1.1.1",
+    "gulp-rename": "^1.2.2",
+    "gulp-uglify": "^3.0.0",
+    "gulp-wrap": "^0.13.0",
+    "handlebars": "^4.0.11"
   },
   "scripts": {
-    "build": "mkdir -p static/swagger-ui && cp -R node_modules/swagger-ui/dist static/swagger-ui/"
+    "build": "gulp"
```
ccostino: Running `gulp` without any arguments will just run through the entire script, right?

ccostino: Yup, answered my own question. :-)

README.md
xtine: When I tried out 3.x it prominently showed a failing status badge on the API docs itself. I think your assumption might be correct, the spec file probably needs some update to have it pass on 3.x.

xtine:
@ccostino: thanks for noticing the missing images! just added them in.

PR #2790, opened 2017-12-06
Add to README - PR approval and getting most recent branches

lbeaufort:
Make sure PR is approved, then `git pull` on each of the three branches

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,9 +445,11 @@ We use git-flow for naming and versioning conventions. Both the API and web app 
     ```
 
 * [auto] `release/my-release` is deployed to `stage`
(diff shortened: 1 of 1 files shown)


xtine:
This is just a general comment and doesn't necessarily have to go into this PR, but I notice the openFEC README getting rather long. What would you think about splitting deployment documentation into a file like `DEPLOYING.md` and linking to it from README.md especially since it's only relevant to internal 18F/FEC folks?

ccostino:
@xtine definitely worth raising, thanks!  Some more changes will be coming into the file soon, so perhaps once those are in that'll be a good time to review the content again and split it apart, reorganize, etc.

PR #2792, opened 2017-12-07
check in ddl files for aggregate tables and fitem_sched tables transf…

fecjjeng:
…erred from FECP


ccostino:
As discussed with @fecjjeng, @hcaofec, @vrajmohan, and myself, we're closing this because the SQL is already captured in the baseline migrations that were created and the tables have existed in all environments for several months now with data flowing into them.

PR #2806, opened 2017-12-12
Add docs for flyway repair

vrajmohan:


README.md
```diff
@@ -602,6 +602,7 @@ All database schema modification code is checked into version control in the dir
 - `info` compares the migration SQL files and the table `flyway_schema_history` and reports on migrations that have been applied and those that are pending.
 - `migrate` compares the migration SQL files and the table `flyway_schema_history` and runs those migrations that are pending.
 - `baseline` modifies the `flyway_schema_history` table to indicate that the database has already been migrated to a baseline version.
+- `repair` repairs the `flyway_schema_history` table. Removes any failed migrations on databases.
```
ccostino: Thank you, @vrajmohan!  What are your thoughts on including this in the deployment process, perhaps as the step before running `migrate`?  Do you feel that is safe to do, or (as in the case of any migration errors themselves), better to let folks address the issue manually and directly?

vrajmohan: I would prefer to let folks address the issue manually. If this happens a lot and we are confident about the causes for it happening, we could add it as a step before `migrate`.

PR #2814, opened 2017-12-13
Reinstate nightly refreshes for schedule E data

### LINKED ISSUE #2813: Reinstate nightly updates for schedule E data
The recent work with the migrations did have one side effect:  we lost the nightly updates to schedule E.  That last bit of work around this had us removing all queue processing for schedule E and just [rebuilding the table each night](https://github.com/18F/openFEC/blob/master/webservices/tasks/refresh.py#L22) with the [table creation setup](https://github.com/18F/openFEC/blob/master/data/sql_setup/prepare_schedule_e.sql).

The `update_itemized` method was removed, though, and the table creation slightly modified with the migration work.  We'll have to devise a new way to make this refresh possible again.

ccostino:
Fixes #2813

This changeset adds support for schedule E nightly refreshes back into place.  The recent migration work accidentally removed the nightly refresh support for schedule E.

A huge h/t to @vrajmohan and @LindsayYoung for helping test and uncover this issue!

### DIFF
--- a/data/refresh/rebuild_schedule_e.sql
+++ b/data/refresh/rebuild_schedule_e.sql
@@ -1,0 +1,119 @@
+-- Create Schedule E table
+drop table if exists ofec_sched_e_tmp;
+create table ofec_sched_e_tmp as
+select
+    *,
+    cast(null as timestamp) as timestamp,
+    image_pdf_url(image_num) as pdf_url,
+    coalesce(rpt_tp, '') in ('24', '48') as is_notice,
+    to_tsvector(pye_nm) as payee_name_text,
+    now() as pg_date
+from fec_fitem_sched_e_vw;
+
+-- Add in records from the Schedule E notices view
+insert into ofec_sched_e_tmp
+select *,
+    cast(null as timestamp) as timestamp,
+    image_pdf_url(image_num) as pdf_url,
+    coalesce(rpt_tp, '') in ('24', '48') as is_notice,
+    to_tsvector(pye_nm) as payee_name_text,
+    now() as pg_date
+from fec_sched_e_notice_vw;
+
+update ofec_sched_e_tmp
+set exp_dt = coalesce(exp_dt, dissem_dt);
+
+-- Add in records for the Form 5 filings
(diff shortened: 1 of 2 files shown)


data/refresh/rebuild_schedule_e.sql
```diff
@@ -0,0 +1,119 @@
...
+insert into ofec_sched_e_tmp
+select *,
+    cast(null as timestamp) as timestamp,
+    image_pdf_url(image_num) as pdf_url,
+    coalesce(rpt_tp, '') in ('24', '48') as is_notice,
+    to_tsvector(pye_nm) as payee_name_text,
+    now() as pg_date
+from fec_sched_e_notice_vw;
+
+update ofec_sched_e_tmp
```
vrajmohan: Why is this a separate step? Why not deal with it in the 2 `INSERTS` above?

data/refresh/rebuild_schedule_e.sql
```diff
@@ -0,0 +1,119 @@
...
+
+-- Analyze the table
+analyze ofec_sched_e_tmp;
+
+-- Replace the existing table
+drop table if exists ofec_sched_e;
+alter table ofec_sched_e_tmp rename to ofec_sched_e;
+
+-- Rename indexes
+alter index ofec_sched_e_sub_id_pkey_tmp rename to ofec_sched_e_sub_id_pkey;
```
vrajmohan: All the `ALTER INDEXES` could be handled more succinctly with the function `rename_indexes`.

vrajmohan:
I have no context to review the logic in this code. All I can do is comment on general soundness of the SQL in a domain-ignorant way. 😄

vrajmohan:
Are there no tests for this code?

LindsayYoung:
I am going to merge this as-is with the follow up task here to simplify the update process here

https://github.com/18F/openFEC/issues/2825

ccostino:
@vrajmohan:  D'oh!  I thought the existing tests covered this case but perhaps not.  FWIW this is the same exact setup we had before prior to the DB migration, but @LindsayYoung and I were also talking more this morning about restructuring this as a materialized view.  We can be sure to get the appropriate tests in with that upcoming work.

PR #2700, opened 2017-10-13
Feature/refactor audit search endpoint

ccostino:

PR #2818, opened 2017-12-15
Feature/sample db

vrajmohan:


tasks.py
```diff
@@ -180,7 +181,10 @@ def deploy(ctx, space=None, branch=None, login=None, yes=False):
     # Target space
     ctx.run('cf target -o fec-beta-fec -s {0}'.format(space), echo=True)
 
-    run_migrations(ctx, space)
+    print("\nMigrating database...")
+    jdbc_url = to_jdbc_url(os.getenv('FEC_MIGRATOR_SQLA_CONN_{0}'.format(space.upper())))
+    run_migrations(ctx, jdbc_url)
+    print("Database migrated\n")
```
ccostino: Should these be `logger.info(...)` statements instead?  Or is this to account for output in CircleCI?  Would this be a case of using both?

tasks.py
```diff
@@ -216,8 +220,22 @@ def notify(ctx):
...
+    """
+    Load schema and data into the empty database pointed to by $SQLA_SAMPLE_DB_CONN
+    """
+
+    print("Loading schema...")
+    db_conn = os.getenv('SQLA_SAMPLE_DB_CONN')
+    jdbc_url = to_jdbc_url(db_conn)
+    run_migrations(ctx, jdbc_url)
+    print("Schema loaded")
+    print("Loading sample data...")
```
ccostino: Same note as above regarding the `print()` statements.

tasks.py
```diff
@@ -216,8 +220,22 @@ def notify(ctx):
...
+    subprocess.check_call(
+        ['psql', '-v', 'ON_ERROR_STOP=1', '-f', 'data/sample_db.sql', db_conn],
```
ccostino: Oh nice, will this just halt as soon as the first error is encountered?  Although it seems to halt anyway as soon as it hits something.

We also ran into a semi-related issue where the migrations took a long time in CircleCI, causing the cloud.gov token (used for the `cf` CLI tool) to timeout and error out of the rest of the deployment.  I have an idea on how to fix that:  wrap the call to check on the status of the app after migrating (the `cf app api` call) with a try/except and if the call fails, login again and then resume.

We can fix that in a separate PR though, but a broader question came up regarding how long a build can run for in CircleCI (there is a 10 minute timeout, at least by default, when there's no output).  Just wanted to flag it for your attention while it's in the back of my mind.

tasks.py
vrajmohan: This is a mere message to the user about stages pf completion. I felt there was no need for the kind of detail and configurability that `logging` provides. I wanted to reduce clutter.

tasks.py
vrajmohan: Yes, that is correct. Otherwise, if there are multiple errors during loading, you'll see it going through all of the same and it may be inconvenient to view earlier ones.

vrajmohan:
I was debating whether to add the schema qualifications in place or add new migrations. Since the triggers already work correctly in `prod`, I thought there was no point in "fixing" them. I can still be persuaded to make a separate migration.

tasks.py
ccostino: Ah okay - seeing as how this'll show up in our consoles locally or in the CircleCI output, which is all we're concerned about, that makes sense!

tasks.py
ccostino: Excellent, thanks!

ccostino:
If the triggers are working then that's totally fine, I see no reason to throw in another migration if it's not really needed.  Thanks, @vrajmohan!

README.md
```diff
@@ -102,12 +102,6 @@ To create the test database, run:
 createdb cfdm_unit_test
 ```
 
-Create the required roles in the test database by running:
-
-```
-psql <<db-connection-parameters>> -f data/create_roles.sql
-```
-
```
ccostino: Love that we can continue to simplify things! :-)

ccostino: Do we need to add documentation for the new `$SQLA_SAMPLE_DB_CONN` environment variable, though, so folks are aware of it and how to use it?

tasks.py
```diff
@@ -216,8 +220,22 @@ def notify(ctx):
...
-    print("\nMigrating database...")
-    jdbc_url = to_jdbc_url(os.getenv('FEC_MIGRATOR_SQLA_CONN_{0}'.format(space.upper())))
+@task
+def create_sample_db(ctx):
+    """
+    Load schema and data into the empty database pointed to by $SQLA_SAMPLE_DB_CONN
+    """
+
+    print("Loading schema...")
+    db_conn = os.getenv('SQLA_SAMPLE_DB_CONN')
```
ccostino: Is it required that we set this environment variable now?  Would it make sense to set a default?

README.md
ccostino: We also need to update the documentation on how to load the sample data.  Speaking of which...  if you're starting from scratch, it's a matter of creating the database, then running migrations, then loading the data, correct?  Could we setup an `invoke` task or a `manage.py` management command that will do all of this in one shot for us?

data/migrations/V0.1__create_roles.sql
```diff
@@ -0,0 +1,25 @@
+-- Create the roles needed in the DB if they do not already exist
+-- Unfortunately, Postgres lacks a "CREATE ROLE IF NOT EXISTS ..."
```
ccostino: Oh, also, in keeping with naming conventions going forward and helping with reading the migrations directory, could we rename this file to `V0000.1__create_roles.sql` please?  Some point soon we'll rename the others to have the leading zeros, this way it's easy to scan the directory and see things in order. :-)

tasks.py
```diff
@@ -216,8 +220,22 @@ def notify(ctx):
...
+    jdbc_url = to_jdbc_url(db_conn)
+    run_migrations(ctx, jdbc_url)
+    print("Schema loaded")
+    print("Loading sample data...")
+    subprocess.check_call(
+        ['psql', '-v', 'ON_ERROR_STOP=1', '-f', 'data/sample_db.sql', db_conn],
+    )
+    print("Sample data loaded")
+
+def run_migrations(ctx, jdbc_url):
```
ccostino: Looks like we're missing the `@task` decorator here if the intention is to be able to run this directly from the shell as an `invoke` task.

README.md
vrajmohan: The issue with creating a database in one command is that you now have to specify 2 DBs - one to log into, and the other that needs to be created. Or, we have to assume that the user has a default DB that is implicit and which we derive by removing the explicit DB parameter in the SQLA connection...

tasks.py
vrajmohan: I would rather go with the Zen of Python - "Explicit is better than implicit." :-)

tasks.py
vrajmohan: I was not intending for this to be run independently as an `invoke` task, but now that you mention it, it may be nice to have.

README.md
ccostino: Ah, I would leave it then as just a command for everything after the database (e.g., `cfdm_test`) has been created.

README.md
```diff
@@ -102,31 +102,30 @@ To create the test database, run:
...
+export SQLA_SAMPLE_DB_CONN=<psql:address-to-box>
 ```
 
-Ignore `user does not exist` error messages. Everything will still work!
+Load our sample data into the development database (`cfdm_test`) by running:
+
+```
+invoke create_sample_db
+```
+This will run `flyway` migrations on the empty database to create the schema, and then load sample data into this database from `data/sample_db.sql`.
```
ccostino: Thank you! 🎉 ❤️

README.md
```diff
@@ -102,31 +102,30 @@ To create the test database, run:
...
 createdb cfdm_test
 ```
 
-Load our sample data into the development database (`cfdm_test`) by running:
+Set the environment variable SQLA_SAMPLE_DB_CONN to point to this database, using:
 
 ```
-pg_restore --dbname cfdm_test --no-acl --no-owner data/subset.dump
-./manage.py refresh_materialized
+export SQLA_SAMPLE_DB_CONN=<psql:address-to-box>
```
ccostino: In this case, we should set it to `jdbc:postgresql://localhost:5432/cfdm_test` for working locally, correct?  If so it may help to just put this in there directly.  Also, doing this didn't work for me, but I suspect it's because I'm not including the username and password in it.

So instead, should it actually be `jdbc:postgresql://<username>:<password>@localhost:5432/cfdm_test`?

ccostino: D'oh still not working for me, even after doing that.  I keep seeing this message:

`ERROR: Invalid JDBC URL (should start with jdbc:) : None`

Which means the URL is not in the format that `to_jdbc_url` is expecting.  I thought we fixed this, though, so you didn't have to supply everything with just your local DB.

vrajmohan: I named it with a `SQLA_` prefix intending for it to be a `SQLAlchemy` URL, not a `jdbc` URL.

ccostino: Aha, that did it, thank you!

ccostino:
D'oh!  The merge of #2700 brought back `subset.dump`, so it just needs to be deleted again in here. :-)

PR #2823, opened 2017-12-18
[WIP] Use filename to create PDF instead of document_id for AOs and MURs

lbeaufort:
Use filename to create AO, MUR document pdfs instead of ID, ensuring we have permalinks. Add AO/MUR number to URL IE `legal/aos/2017-12/201712C_3.pdf`

We'll need to delete PDFs from S3 with old naming conventions.

- [ ] Manually deploy local branch to `dev` and reload all the AO's and MURs and test the new links - do the ones that have filenames with spaces work or do we have to modify them? Should we just go ahead and replace spaces with "_" ? "-"? I think dashes might be better, but the new naming convention has underscores in them so underscores might be more consistent:  "201712C_3.pdf"
 
- [ ] Reload all AOs, MURs with new naming convention in `prod`, `dev`, `stage`
- [ ] Delete all the old, unlinked PDFs from s3 in `prod`, `dev`, `stage`

### DIFF
--- a/webservices/legal_docs/advisory_opinions.py
+++ b/webservices/legal_docs/advisory_opinions.py
@@ -180,12 +180,27 @@ def get_documents(ao_id, bucket):
                 "text": row["ocrtext"],
                 "date": row["document_date"],
             }
+            # should we instead save the pdf_key in document['pdf_key']??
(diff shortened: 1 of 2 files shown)


lbeaufort:
Closed in favor of https://github.com/18F/openFEC/pull/2832

PR #2826, opened 2017-12-20
Renamed the migration files for better order in GitHub

LindsayYoung:
Renamed the migration files so that they will show up in order in GitHub, that will make it easier for us to look at the migrations.


vrajmohan:
Sorry, I beat you to it! :-)

PR #2833, opened 2017-12-22
Feature/refactor audit cmte audit vw

fec-jli:
Create a new V0045.1__public.auditcase_grant.sql
1) related issue#2827: https://github.com/18F/openFEC/issues/2827
Modify cmte_audit_vw to get cmte_desc from STAGING.REF_FILED_CMTE_TP
and add "GRANT" to all views

2)"GRANT" to all MVs


ccostino:
@fec-jli awesome work!  I just happened to poke my head in here and noticed the two migrations instead of one, though.  You'll want to add all of the changes in just the `V0045.1` migration because we've already run migrations up to `V0045`.  Anything added to a `V0044.1` would not get run on the next migration.

Just wanted to make sure I flagged that before you kept going further! :-)

fec-jli:
O,,Thank You reminder me. you are right.
I will put together into V0045.1.

data/migrations/V0045.1__public_auditsearch_grant.sql
```diff
@@ -0,0 +1,63 @@
...
+GRANT ALL ON TABLE finding_vw TO fec;
+GRANT SELECT ON TABLE finding_vw TO fec_read;
+GRANT SELECT ON TABLE finding_vw TO openfec_read;
+
+GRANT ALL ON TABLE finding_rel_vw TO fec;
+GRANT SELECT ON TABLE finding_rel_vw TO fec_read;
+GRANT SELECT ON TABLE finding_rel_vw TO openfec_read;
+
+
+SET search_path = public, pg_catalog;
```
ccostino: This line (and the space around it) is just extra from the other migration that was there, it's not needed again since it's at the top of the file. :-)

fec-jli: the top line is different schema: SET search_path = auditsearch, pg_catalog;
so I think we need this line ? right ? for public schema.
Thanks

ccostino: Oh good point, you're right, the operations are working against different schemas.  Please disregard! :-)

LindsayYoung:
Thanks this is great!

PR #2836, opened 2017-12-22
streaming the data to s3 for cached calls

pkfec:


LindsayYoung:
We paired on this and it is working 🥇 

thanks @pkfec and @ccostino

PR #2858, opened 2018-01-10
[WIP]Feature/switch to new aggregate tables

hcaofec:
switching APIs to use new aggregate tables ( dsc_ )
Migration scripts needs to be added too


data/migrations/V0047__alter_ofec_sched_a_aggregate_state_recipient_totals_mv.sql
```diff
@@ -0,0 +1,177 @@
...
+  ON public.ofec_sched_a_aggregate_state_recipient_totals_mv_TMP
+  USING btree
+  (total, idx);
+
+-- ------------------------------------------------------
+DROP MATERIALIZED VIEW IF EXISTS public.ofec_sched_a_aggregate_state_recipient_totals_mv;
+
+ALTER MATERIALIZED VIEW IF EXISTS public.ofec_sched_a_aggregate_state_recipient_totals_mv_TMP RENAME TO ofec_sched_a_aggregate_state_recipient_totals_mv;
+
+--SELECT public.rename_indexes ('ofec_sched_a_aggregate_state_recipient_totals_mv');
```
ccostino: Did this not work for you?  Are there any questions in using it?  I would think we'd still be able to leverage it here to avoid having to write all of the index names by hand below, but wanted to double check and see.

data/migrations/V0048__alter_ofec_sched_a_aggregate_size_merged_mv.sql
```diff
@@ -0,0 +1,113 @@
...
+  USING btree
+  (total, idx);
+
+
+-- ------------------------------------------------------
+DROP MATERIALIZED VIEW IF EXISTS public.ofec_sched_a_aggregate_size_merged_mv;
+
+ALTER MATERIALIZED VIEW IF EXISTS public.ofec_sched_a_aggregate_size_merged_mv_TMP RENAME TO ofec_sched_a_aggregate_size_merged_mv;
+
+--SELECT public.rename_indexes ('ofec_sched_a_aggregate_size_merged_mv');
```
ccostino: Same note as above regarding the use of this function.

webservices/resources/aggregates.py
```diff
@@ -28,7 +28,12 @@ def args(self):
 
     @property
     def sort_args(self):
-        return args.make_sort_args(validator=args.IndexValidator(self.model))
+        schema = None
+        if self.model and hasattr(self.model,'__table_args__') and 'schema' in self.model.__table_args__.keys():
+            schema = self.model.__table_args__.get('schema')
+        return args.make_sort_args(validator=args.IndexValidator(self.model, schema=schema))
```
ccostino: 👏 well done! :-)

PR #2864, opened 2018-01-17
Add districts missing from elections search

lbeaufort:
## Summary
Addresses https://github.com/18F/openFEC/issues/2844: Districts missing from election pages

Refactors the `/elections/search/` endpoint to use `dim_race_inf` as the base for all state/district information.

## Screenshots
**BEFORE: Lots of districts missing**
<kbd>
before
</kbd>

**AFTER: I verified every state**
<kbd>
after
</kbd>

**Remaining items:**
- [x] Update sort order
- [x] Fix counts
   - [x] Make count test
- [x] Manual testing
 - test all filters
 - make sure all districts are listed for each state in 2018
- [x] Make sure historical/future data looks the way we want it to - `dim_race_inf` goes way into the future, I think.


webservices/resources/elections.py
```diff
@@ -60,7 +60,7 @@ def get(self, **kwargs):
 
     def _get_elections(self, kwargs):
         """Get elections from ElectionsList model."""
-        query = db.session.query(ElectionsList).order_by(ElectionsList.district)
+        query = db.session.query(ElectionsList).order_by(ElectionsList.sort_order, ElectionsList.district)
```
LindsayYoung: I think we want to add this as the default sort order, so that people can over-ride it fi they want to. You should be able to add a default sort here https://github.com/18F/openFEC/pull/2864/commits/5a39eb3e398c9e8f081e8d0d898e5db370575b44#diff-1fb1f7e2bd8c60cb7128c0c501821eeaR55

PR #2885, opened 2018-01-25
feature/missing-filings-from-view

pkfec:


LindsayYoung:
You slacked me that this is ready but in the future you should also add a please review tag and edit the title to take out the [WIP]. That way it is clear that it is ready for review. :)

LindsayYoung:
Ok this looks like it has the new data we need. I think we might need to tweak the output so the web app can use it down stream:

If you look at the filings page that is consuming the API https://www.fec.gov/data/filings/?data_type=processed
screen shot 2018-01-26 at 12 16 41 pm

data/migrations/V0052__add_ofec_filings_all_mv.sql
```diff
@@ -0,0 +1,450 @@
...
+            filing_history.last_update_dt AS update_date,
+            report_pdf_url_or_null(filing_history.begin_image_num::text, filing_history.rpt_yr, com.committee_type::text, 'RFAI'::text) AS pdf_url,
+            means_filed(filing_history.begin_image_num::text) AS means_filed,
+            report_html_url(means_filed(filing_history.begin_image_num::text), filing_history.id::text, filing_history.file_num::text) AS html_url,
+            NULL::text AS fec_url,
+            NULL::numeric[] AS amendment_chain,
+            NULL::integer AS most_recent_file_number,
+            NULL::boolean AS is_amended,
+            true AS most_recent,
+            0 AS amendement_version,
```
lbeaufort: Should this be `amendment_version1`? @pkfec @LindsayYoung

LindsayYoung:
Adding my comment here for continuity, but I think what we need is a column in the view that will tell us if the record has a office associated with it or if there is a committee type of H, S or P. We can do this with a function that looks at both of these factors. We can then make that filterable on the API and point the website to use the new filter.

data/migrations/V0053__add_ofec_filings_all_mv.sql
```diff
@@ -0,0 +1,481 @@
...
+ALTER TABLE public.ofec_filings_all_mv
+  OWNER TO fec;
+GRANT ALL ON TABLE public.ofec_filings_all_mv TO fec;
+GRANT SELECT ON TABLE public.ofec_filings_all_mv TO fec_read;
+
+
+CREATE INDEX ofec_filings_all_mv_amendment_indicator_idx_idx1
+  ON public.ofec_filings_all_mv
+  USING btree
+  (amendment_indicator COLLATE pg_catalog."default", idx);
```
LindsayYoung: Why we are including the identifier in all the indexes? Maybe I am missing someting

data/migrations/V0053__add_ofec_filings_all_mv.sql
```diff
@@ -0,0 +1,481 @@
+
+CREATE OR REPLACE FUNCTION public.get_office_cmte_tp (p_office varchar, p_cmte_tp varchar)
+  RETURNS varchar AS
+$BODY$
+    begin
+
+        return coalesce(p_office, p_cmte_tp);
```
LindsayYoung: Let's only return this in the case that it is H, S or P, that will make the output easier to understand downstream.

LindsayYoung: I did a double check and we have seen in the past, the itemized resources performed better with the idx compound index because we always append the unique index for itemized resources:

https://github.com/18F/openFEC/blob/8e39a1f8e04915ade2e97898b96eda2ea686e906/webservices/common/views.py#L105

The finings endpoint inherits from the ApiResource so it doesn't need it:
 
https://github.com/18F/openFEC/blob/8e39a1f8e04915ade2e97898b96eda2ea686e906/webservices/resources/filings.py#L20

As you mentioned this existed before this PR, but let's simplify the indexes while we are in here. Thanks @fecjjeng and @pkfec!

data/migrations/V0053__add_ofec_filings_all_mv.sql
LindsayYoung: Did some research and in itemized sorts the unique key is always appended to the sort for consistent sorting in case of a tie. This only happens when the resource inherits from the ItemizedResource class:
https://github.com/18F/openFEC/blob/8e39a1f8e04915ade2e97898b96eda2ea686e906/webservices/common/views.py#L105

In this case, the filings resource inherits from the ApiBase and not Itemized, to the compound indexes won't help.
https://github.com/18F/openFEC/blob/8e39a1f8e04915ade2e97898b96eda2ea686e906/webservices/resources/filings.py#L20

This existed before this PR, but we can clean it up since we are making changes in here anyway. Thanks @fecjjeng and @pkfec

pkfec:
thank you @LindsayYoung

Issue #2871, opened 2018-01-19
Investigate issue with rename_indexes method in the DB


ccostino:
@hcaofec's and @fecjjeng's work with moving the aggregate tables in PR #2858 uncovered an issue with the `rename_indexes` method we built in the database.  This was built originally for the itemized schedule table construction and so it may not be as generic as we thought.  In any case, we should investigate this to make sure we can leverage it in the future.

We're not immediately able to reproduce this. We'll need to reconsider this the next time we move a table. Moving to icebox.

lbeaufort:
I'll test this on my local db and see if I can get some examples

lbeaufort:
I tested this locally and am unable to reproduce the issue.

PR #2886, opened 2018-01-25
[MERGE WITH GIT FLOW] Modify ofec_elections_list_mv to handle multiple incumbents

lbeaufort:
**Merge with git flow: Hotfix**

Fixes bad join in materialized view causing some districts to repeat: https://www.fec.gov/data/elections/?cycle=2018&state=KS&district=4

Add new SQL migration file to remove the DISTINCT on cand_id for incumbents - some races have more than one incumbent where there was a special. 

Districts that probably have this issue are: CA-34, GA-06, KS-04, MT-00, SC-05 and UT-03. All these districts had members that resigned  and now have a new incumbent.

1-25-2018 12-56-47 pm

### DIFF
--- a/data/migrations/V0051__alter_ofec_elections_list_mv.sql
+++ b/data/migrations/V0051__alter_ofec_elections_list_mv.sql
@@ -1,0 +1,72 @@
+SET search_path = public, pg_catalog;
+
+DROP MATERIALIZED VIEW IF EXISTS ofec_elections_list_mv_tmp;
+
+CREATE MATERIALIZED VIEW ofec_elections_list_mv_tmp AS
+
+WITH incumbents AS (
+--So we don't show duplicate elections
+    SELECT DISTINCT ON (
+            race_pk) --sometimes more than one incumbent
+        race_pk,
+        cand_id,
+        cand_name,
+        lst_updt_dt
+    FROM disclosure.cand_valid_fec_yr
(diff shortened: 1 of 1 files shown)


data/migrations/V0051__alter_ofec_elections_list_mv.sql
```diff
@@ -0,0 +1,72 @@
...
+CREATE INDEX ON ofec_elections_list_mv_tmp(incumbent_id);
+CREATE INDEX ON ofec_elections_list_mv_tmp(incumbent_name);
+
+--------------------------------------------------------
+
+DROP MATERIALIZED VIEW IF EXISTS ofec_elections_list_mv;
+
+ALTER MATERIALIZED VIEW IF EXISTS ofec_elections_list_mv_tmp RENAME TO ofec_elections_list_mv;
+
+SELECT rename_indexes('ofec_elections_list_mv');
```
ccostino: @lbeaufort please double check that the indexes do get renamed as expected. @fecjjeng reported that this wasn't always happening so we may have to either alter the index names manually or just create them as the last step. A few more details are in issue #2871 if you're curious. Thanks!

lbeaufort: @ccostino thanks! I double-checked on my local when I ran this, and they were getting moved over.

PR #2892, opened 2018-01-30
Change "/elections/list" endpoint zip search to use new table

lbeaufort:
## Summary
Currently, the elections search endpoint and other views use tables created from csv's to map congressional district to zip code (`ofec_fips_states` and `ofec_zips_districts`). This PR refactors the elections/search endpoint to use a new  `ref_zip_to_district` table instead of csv's in the materialized view `ofec_elections_list_mv` to make updating this data easier.

- Use the new `ref_zip_to_district` table created by https://github.com/18F/openFEC/pull/2877 in all the places the old tables `ofec_fips_states` and `ofec_zips_districts`are used
- Create duplicate (renamed) materialized view for `ofec_sched_a_aggregate_state_recipient_totals_mv` to use the new MV
- Update ElectionsListView resource to reference this new MV


data/migrations/V0052__copy_ofec_sched_a_aggregate_state_recipient.sql
```diff
@@ -0,0 +1,149 @@
+CREATE MATERIALIZED VIEW ofec_sched_a_aggregate_state_recipient_totals_2_mv AS
```
ccostino: Just out of curiosity, why the `2` in the name?  Are we still keeping the other materialized view?

Also, where is this materialized view being used?

webservices/resources/elections.py
```diff
@@ -85,31 +85,24 @@ def _get_elections(self, kwargs):
...
-            zips_districts,
-            fips_states,
-        ).join(
-            fips_states,
-            zips_districts.c['State'] == fips_states.c['FIPS State Numeric Code'],
-        ).filter(
-            zips_districts.c['ZCTA'].in_(kwargs['zip'])
+        districts = db.session.query(ZipsDistricts).filter(
+            cast(ZipsDistricts.zip_code, Integer).in_(kwargs['zip']),
+            ZipsDistricts.active == 'Y'
```
ccostino: Nice, using the new table definitely simplifies the code here!

webservices/common/models/elections.py
```diff
@@ -27,3 +27,14 @@ class ElectionsList(db.Model):
     cycle = db.Column(db.Integer)
     incumbent_id = db.Column(db.String, doc=docs.CANDIDATE_ID)
     incumbent_name = db.Column(db.String, doc=docs.CANDIDATE_NAME)
+
+
+class ZipsDistricts(db.Model):
+    __table_args__ = {'schema': 'staging'}
```
ccostino: The `staging` schema?  I may have missed this, so I just wanted to double check on it to make sure I know what it is. :-)

data/migrations/V0052__copy_ofec_sched_a_aggregate_state_recipient.sql
lbeaufort: @ccostino thanks for pointing this out! I misunderstood this before - I'm going to push a new commit with the standard "create TMP/rename TMP" migration.

webservices/common/models/elections.py
lbeaufort: @ccostino - it looks like that's where the table was put, with [this migration](https://github.com/18F/openFEC/blob/develop/data/migrations/V0050__add_ref_zip_to_district.sql). Is that ok?

LindsayYoung:
I'll merge when we get rid of the unused management commands

LindsayYoung:
never mind I just needed to refresh

webservices/common/models/elections.py
LindsayYoung: that was already added here: https://github.com/18F/openFEC/blob/83f59d5c2c1d4aa90aac036b276d4ca54340dbaa/data/migrations/V0050__add_ref_zip_to_district.sql

PR #2899, opened 2018-02-01
Fix elasticsearch reindex timeouts, revert global connection setting

lbeaufort:
Elasticsearch was timing out on [restore_from_staging_index](https://github.com/18F/openFEC/blob/develop/webservices/legal_docs/index_management.py#L461) and [move_archived_murs](https://github.com/18F/openFEC/blob/develop/webservices/legal_docs/index_management.py#L474) on `reindex` commands, which was messing with our ability to restore from a staging index. (Needed for management command `refresh_current_legal_docs_zero_downtime`)

This PR reverts the global elasticsearch connection settings, and passes the proper parameter to the elasticsearch `reindex` command to allow the resource-heavy task to finish.

[Elasticsearch 2.4 API docs on timeout](http://elasticsearch-py.readthedocs.io/en/2.4.0/api.html#timeout)
"Global timeout can be set when constructing the client (see Connection‘s `timeout` parameter) or on a per-request basis using `request_timeout` (float value in seconds) as part of any API call"

### DIFF
--- a/webservices/legal_docs/index_management.py
+++ b/webservices/legal_docs/index_management.py
@@ -458,7 +458,7 @@ def restore_from_staging_index():
         "index": "docs"
       }
     }
-    es.reindex(body=body, wait_for_completion=True, timeout='5m')
+    es.reindex(body=body, wait_for_completion=True, request_timeout=1500)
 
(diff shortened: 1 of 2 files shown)


vrajmohan:
Nice work figuring out the right parameters and where to set them!

PR #2901, opened 2018-02-02
Add candidate_pcc to /elections/ endpoint

lbeaufort:
## Summary

Addresses https://github.com/fecgov/fec-cms/issues/1910

This modifies the /elections/ endpoint to show `candidate_pcc` where a candidate has a linked committee has designation 'P' (Primary Campaign Committee.) Where two committees are marked 'P' in an election cycle (this really shouldn't happen) it takes the highest (aka most recent) committee ID. 

The plan is to refactor the /elections/ endpoint this sprint, but this *unblocks* the election profile redesign (cc: @patphongs @JonellaCulmer) 

**Note**: not all candidates have committees yet. Where there is no linked committee, the elections endpoint returns `null`.

This works across time where the data is there. Example, Graham:
http://127.0.0.1:5000/v1/elections/?office=senate&cycle=2014&state=sc&per_page=100
PCC is C00458828 in 2014
http://127.0.0.1:5000/v1/elections/?office=senate&cycle=2008&state=sc&per_page=100
PCC is C00364505 in 2008

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -161,6 +161,9 @@ class ElectionView(utils.Resource):
             totals_model.disbursements,
             totals_model.last_cash_on_hand_end_period.label('cash_on_hand_end_period'),
             totals_model.coverage_end_date,
+            sa.case(
+                [(CandidateCommitteeLink.committee_designation == 'P', CandidateCommitteeLink.committee_id)]  # noqa
+            ).label('candidate_pcc'),
         )
         pairs = join_candidate_totals(pairs, kwargs, totals_model)
         pairs = filter_candidate_totals(pairs, kwargs, totals_model)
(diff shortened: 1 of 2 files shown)


LindsayYoung:
Would we want to include authorized committees too?

lbeaufort:
@LindsayYoung we already output a list of all authorzied commitees in `committee_ids` already, so this just highlights which one is the PCC.

lbeaufort:
@LindsayYoung thanks for taking a look but this approach is on hold so I'm going to close the PR

vrajmohan:
Can you please rebase/squash these commits? The number of lines changed is lower than the number of commits :-)

lbeaufort:
@vrajmohan thanks for taking a look.
- Yes, I will squash these commits. You make a strong point :)
- Yes, the plan is to refactor this endpoint with #3076 but I wanted to get the `candidate_pcc` up first so I stop blocking the election summary pages redesign.

Issue #2791, opened 2017-12-06
Optimize Schedule B query


vrajmohan:
Based on the pgBadger report investigated in  https://github.com/18F/fec-cms/issues/1534, the schedule B query is executed often and takes an average of 8s to run:

Day | Hour | Count | Duration | Avg duration
------- | ----- | ------- | ------ | ------------
Dec 04 | 11 | 387 | 54m11s | 8s402ms
  | 12 | 387 | 54m18s | 8s419ms
  | 13 | 384 | 54m21s | 8s493ms
  | 14 | 383 | 54m30s | 8s538ms
  | 15 | 375 | 54m28s | 8s716ms
  | 16 | 376 | 54m12s | 8s650ms
  | 17 | 374 | 54m8s | 8s686ms
  | 18 | 323 | 49m15s | 9s150ms
  | 19 | 363 | 54m30s | 9s9ms
  | 20 | 351 | 53m15s | 9s105ms
  | 21 | 365 | 54m5s | 8s892ms
  | 22 | 184 | 27m27s | 8s955ms

An example actual query is:
```
SELECT
    ofec_sched_b_master.cmte_id AS ofec_sched_b_master_cmte_id,
    ofec_sched_b_master.rpt_yr AS ofec_sched_b_master_rpt_yr,
    ofec_sched_b_master.rpt_tp AS ofec_sched_b_master_rpt_tp,
    ofec_sched_b_master.image_num AS ofec_sched_b_master_image_num,
    ofec_sched_b_master.line_num AS ofec_sched_b_master_line_num,
    ofec_sched_b_master.tran_id AS ofec_sched_b_master_tran_id,
    ofec_sched_b_master.file_num AS ofec_sched_b_master_file_num,
    ofec_sched_b_master.entity_tp AS ofec_sched_b_master_entity_tp,
    ofec_sched_b_master.entity_tp_desc AS ofec_sched_b_master_entity_tp_desc,
    ofec_sched_b_master.recipient_cmte_id AS ofec_sched_b_master_recipient_cmte_id,
    ofec_sched_b_master.clean_recipient_cmte_id AS ofec_sched_b_master_clean_recipient_cmte_id,
    ofec_sched_b_master.recipient_nm AS ofec_sched_b_master_recipient_nm,
    ofec_sched_b_master.recipient_st1 AS ofec_sched_b_master_recipient_st1,
    ofec_sched_b_master.recipient_st2 AS ofec_sched_b_master_recipient_st2,
[...]

PR #2904, opened 2018-02-03
add_expression_based_index_for_sched_b_tables

fecjjeng:
add_expression_based_index_for_sched_b_tables is part of the attempt to address issue #2791


fecjjeng:
expreesion-based-indexes had been added to all the ofec_sched_b sub-tables according to the suggestions on issue #2791.  No index created on the ofec_sched_b_master table itself so this expression-based-index had not been added to the master table neither.

PR #2905, opened 2018-02-05
Add support for expression-based sorts in API

ccostino:
Addresses #2791

This changeset adds support for expression-based sorts in API queries, meaning that instead of sorting on a specific column we can choose to sort on an expression.  This is useful if an expression needs to be used to provide the desired value as opposed to what is stored in the column itself, especially if that expression is also used in the `WHERE` clause of the query.  Furthermore, this allows us to leverage expression-based indexes properly to ensure that the API queries are performant.

To create expression-based sorting options, you can follow the pattern developed for the `sort_expressions` hybrid property on the model in this PR, or followed a more [detailed set of instructions](https://github.com/18F/openFEC/wiki/Enabling-an-expression-based-sort) in the wiki.

### DIFF
--- a/webservices/resources/sched_b.py
+++ b/webservices/resources/sched_b.py
@@ -57,7 +57,7 @@ class ScheduleBView(ItemizedResource):
             args.make_sort_args(
                 default='-disbursement_date',
                 validator=args.OptionValidator(['disbursement_date', 'disbursement_amount']),
-            ),
+            )
         )
 
     def build_query(self, **kwargs):
@@ -77,7 +77,7 @@ class ScheduleBView(ItemizedResource):
 
 @doc(
     tags=['disbursements'],
-    description=docs.EFILING_TAG,
+    description=docs.EFILING_TAG
 )
 class ScheduleBEfileView(views.ApiResource):
     model = models.ScheduleBEfile
--- a/tests/test_itemized.py
+++ b/tests/test_itemized.py
@@ -343,6 +343,178 @@ class TestItemized(ApiBaseTest):
         )
         self.assertEqual(response.status_code, 422)
 
+    def test_pagination_with_sort_expression(self):
+        filings = [
+            factories.ScheduleBFactory()
+            for _ in range(30)
+        ]
+        page1 = self._results(api.url_for(ScheduleBView, **self.kwargs))
+        self.assertEqual(len(page1), 20)
+        self.assertEqual(
+            [int(each['sub_id']) for each in page1],
+            [each.sub_id for each in filings[:20]],
+        )
+        page2 = self._results(api.url_for(ScheduleBView, last_index=page1[-1]['sub_id'], **self.kwargs))
+        self.assertEqual(len(page2), 10)
+        self.assertEqual(
+            [int(each['sub_id']) for each in page2],
+            [each.sub_id for each in filings[20:]],
+        )
+
+    def test_pagination_with_null_sort_column_values_with_sort_expression(self):
+        filings = [
+            factories.ScheduleBFactory(disbursement_date=None)
+            for _ in range(5)
+        ]
+        filings = filings + [
+            factories.ScheduleBFactory(
+                disbursement_date=datetime.date(2016, 1, 1)
+            )
+            for _ in range(25)
+        ]
+        page1 = self._results(api.url_for(
+            ScheduleBView,
+            sort='disbursement_date',
+            **self.kwargs
+        ))
+        self.assertEqual(len(page1), 20)
+        self.assertEqual(
+            [int(each['sub_id']) for each in page1],
+            [each.sub_id for each in filings[5:25]],
+        )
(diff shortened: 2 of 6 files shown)


webservices/resources/sched_b.py
```diff
@@ -77,7 +77,7 @@ def build_query(self, **kwargs):
 
 @doc(
     tags=['disbursements'],
-    description=docs.EFILING_TAG,
+    description=docs.EFILING_TAG
```
lbeaufort: @ccostino - I'm never sure when a trailing comma is needed and when it isn't - is it a good guideline that a trailing comma is better for data structures and unnecessary for function calls?

ccostino: @lbeaufort that's the convention I've been following, mostly based on [PEP-8's recommendations for trailing commas](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)!  By and large that is the recommendation, though it does make the case for trailing commas in a function call if it is split across multiple lines.  I lean more toward consistency in that case, though, hence this particular change.

Speaking of consistency, [PEP-8 has a note about that too right at the start](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds). :-)

lbeaufort:
@ccostino and I are investigating the lack of pagination on the schedule B endpoint results (`schedules/schedule_b/`)

ccostino:
Pagination is definitely an issue - the paging metadata isn't coming through with this change, and it looks to be an unsupported piece in the underlying `marshmallow-pagination` library; the code is expecting a SQLAlchemy column and nothing else, so the SQLAlchemy expression causes an error to be thrown.  We're going to investigate a bit more to determine level of effort and what would be involved in fixing this and potentially shift gears to simply adding a new column that would always store the derived value of the `COALESCE` statement instead, index that, and adjust the query accordingly.

ccostino:
@lbeaufort, your testing and the pairing debug session we had yielded a host of things that I had missed in the initial implementation of this!  We were on the right track with the breakdown happening in `marshmallow-pagination`, but there were a couple of other points in the code that broke as well for similar reasons (mainly attempting to derive data types and values from the column and returned result(s) automatically).  Since an expression doesn't have a type associated with it, we have to supply that information with the expression.

To support this, I modified the property we add to the model by account for both the field type used for serialization in the pagination metadata returned with the results as well as the SQL data type of the value returned by the expression.  This solves the issue of not knowing what the data is and how it should be treated at the cost of some additional complexity under the hood.

Furthermore, there was the added issue of actually retrieving the value of the expression for the pagination construction.  This required a bit of SQLAlchemy black magic...  The pagination code that makes that initial call to `convert_value` in the underlying paginator object (which is what was bringing us into `marshmallow-pagination`) was operating on a result set row, meaning we had a record of data.  I figured that if I manipulated the query that was responsible for retrieving that record to adjust the `SELECT` portion of the statement to return the value of the sort expression we wanted and maintained the remaining `WHERE` and `ORDER BY` clauses, we'd get the value needed and could do the serialization ourselves.  Sure enough, this works, but again with an added cost of some additional complexity under the hood.

Lastly, tests!  Most of our itemized tests focus on just Schedule A because by and large A and B are identical.  With the sort expressions in place though, that's no longer the case.  I figured if we ran the handful of pagination tests that already exist against Schedule B in addition to Schedule A we'd be able to exercise this functionality and maintain or increase our test coverage.  This mostly works as intended (and helped me catch another break down point in the code!), but two of the new tests are failing.  I think this is a good place to stop, review some more in another pairing session, and try and solve there. :-)

Bonus:  I also updated the wiki page based on the new changes in this PR.

Thank you again for your help today!  It resulted in the small epiphany I had that lead to these latest changes and I wanted to try them out before losing them. :-)

ccostino:
One other thing to note - with pagination working again (despite the failing tests), we do still need to verify that the pagination is also *working properly* - in other words, records are still coming back ordered as they should be, not some other weird random order. :-)

lbeaufort:
@ccostino - it looks like `test_pagination_with_sort_expression` is possibly failing because we don't have the `ScheduleBFactory` creating dates, so it can't sort by `disbursement_date.` I'm still trying to come up with an elegant way to test that theory.  

I haven't gotten to `test_null_pagination_with_null_sort_column_values_ascending_with_sort_expression` yet but I will tomorrow!

ccostino:
Ah, nice catch @lbeaufort!  We should be able to test that by supplying them ourselves in the object construction in the test itself.  We can pair and/or I'll update the tests based on that and walk through it after!

ccostino:
Yup that is true, we can definitely review that. :-)

One thing I totally failed to keep in mind about this though is that in some cases we *want* empty `disbursement_date` values as we're testing that the expression statement (`COALESCE` in this case) is working properly with the sorting. :-)  So, it might be back to the drawing board for a cause, but still worth exploring a bit!

PR #2915, opened 2018-02-11
Feature/audit mv refactor and fix bug for multi columns sorting

fec-jli:
1)fix bug for multi columns sorting.
when 'sort' is list (['-cycle', 'committee_name', ] ), should setup multi=true to call utils.fetch_page() in views.py.

2)remove cmte_audit_vw and cand_audit_vw from 
ofec_audit_case_mv, 
ofec_committee_fulltext_audit_mv,
ofec_candidate_fulltext_audit_mv,

replace with sub query. so above 3 MVs no dependency.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,6 +10,7 @@ from webservices import args
 from webservices import rest
 from webservices import sorting
 from webservices.resources import candidate_aggregates
+from webservices.resources import audit
 from webservices.resources import elections
 from webservices.rest import db
 from webservices.tasks import utils
@@ -36,6 +37,32 @@ class TestSort(ApiBaseTest):
         ]
         query, columns = sorting.sort(models.Candidate.query, '-district', model=models.Candidate)
         self.assertEqual(query.all(), candidates[::-1])
+
+    def test_multi_column(self):
+        audit = [
+            factories.AuditCaseFactory(cycle=2012, committee_name='Boy', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Girl', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Zoo', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Abc', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='John', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+        ]
+        query, columns = sorting.multi_sort(models.AuditCase.query, ['cycle', 'committee_name', ], model=models.AuditCase)
+        self.assertEqual(query.all(), audit)
+
+    def test_multi_column_reverse_first_column(self):
+        audit = [
+            factories.AuditCaseFactory(cycle=2012, committee_name='Zoo', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Girl', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Abc', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='John', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Abc', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+        ]
+        query, columns = sorting.multi_sort(models.AuditCase.query, ['-cycle', 'committee_name', ], model=models.AuditCase)
+        self.assertEqual(query.all(), audit[::-1])
 
     def test_hide_null(self):
         candidates = [
--- a/webservices/common/views.py
+++ b/webservices/common/views.py
@@ -32,10 +32,14 @@ class ApiResource(utils.Resource):
     def get(self, *args, **kwargs):
         query = self.build_query(*args, **kwargs)
         count = counts.count_estimate(query, models.db.session, threshold=500000)
+        multi = False
+        if isinstance(kwargs['sort'], (list, tuple)):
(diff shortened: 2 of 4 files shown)


webservices/common/views.py
```diff
@@ -32,10 +32,14 @@ class ApiResource(utils.Resource):
...
         count = counts.count_estimate(query, models.db.session, threshold=500000)
+        multi = False
+        if isinstance(kwargs['sort'], (list, tuple)):
+            multi = True
+
         return utils.fetch_page(
             query, kwargs,
             count=count, model=self.model, join_columns=self.join_columns, aliases=self.aliases,
-            index_column=self.index_column, cap=self.cap,
+            index_column=self.index_column, cap=self.cap, multi=multi,
```
ccostino: @fec-jli this is a great catch in investigating and experimenting with what the issue was with the multi-sort not being supported when you were trying it out!  As I had mentioned when we caught up in PI planning, I know we've done this in a couple of other places and I had a chance to look into this some and now know why this appears to have not worked.

You're absolutely correct that you need to pass in the `multi` parameter with a value of `True`.  However, the fix shouldn't go here as this has ramifications across the whole API.  You have the right idea! :-)  We just need to adjust the where the change needs to go.

The changes in here should be reverted, and the fix (adjustment is more accurate) needs to go in the resource itself that you want to support mutli-sort in.  More details down below!

webservices/resources/audit.py
```diff
@@ -115,13 +115,16 @@ class AuditCaseView(ApiResource):
         ('candidate_id', model.candidate_id),
     ]
 
+    # @use_kwargs(
+    #     args.make_multi_sort_args(default=['-cycle', 'committee_name'])
+    # )
```
ccostino: This is safe to remove, it's not being used. :-)

webservices/resources/audit.py
```diff
@@ -115,13 +115,16 @@ class AuditCaseView(ApiResource):
...
     @property
     def args(self):
         return utils.extend(
             args.paging,
             args.AuditCase,
-            args.make_sort_args(
-                default='-cycle',
+            args.make_multi_sort_args(
+                default=['-cycle', 'committee_name', ]
```
ccostino: This is correct for adding the default parameters you want for the multi-sort support.  What's missing here is an override of the `get()` method in the base `ApiResource` object.  If your resource needs this support, then we need to create a resource-specific version of the `get()` method that makes the call to `utils.fetch_page()` with the `multi` parameter in it.

A couple of examples of this exist in the code where you can see how this is done:

* [`BaseFiling` resource](https://github.com/18F/openFEC/blob/develop/webservices/resources/filings.py#L66-L72)
* [`ReportsView` resource](https://github.com/18F/openFEC/blob/develop/webservices/resources/reports.py#L103-L119)

In both of these cases there is some other custom logic going, which I don't think is necessary in this case but am not 100% sure, it depends on the needs of the resource and requirements of the business logic.

In any case, an example to start with could be this:

```python
def get(self, **kwargs):
    query = self.build_query(**kwargs)
    count = counts.count_estimate(query, models.db.session, threshold=5000)
    return utils.fetch_page(query, kwargs, model=models.AuditCase, count=count, multi=True)
```

If you add this method in the `AuditCaseView` resource, this will get multi-sort working for you.  Again, I'm not sure if any other logic needs to go in place here, but you need at least these three lines as the `get()` method is what calls `build_query` and also provides a count estimate of results returned.  @LindsayYoung if you wouldn't mind taking a quick look to in case I missed something or misunderstood, I'd appreciate it!

Again, great job in finding what wasn't working!  We just need to adjust where the change is being made, but I'm glad you put this up in the PR so we could have a conversation around it in the review and certainly want to continue encouraging such exploration and experimentation as it yields valuable learning opportunities. :-)

If anything is unclear or if you have more questions please let me know!

fec-jli: Thanks @ccostino for your explanation.
for my understanding, ApiResource is a base class (in views.py). so 'get' function is base function to lunch request using 'get' method, which include some default codes and base feature . when a subclass extend ApiResource, if no more action wanted, don't need override 'get' function. like our AuditCaseView endpoint. AuditCaseView just use default 'get' call. so the 'get' function should support all base parameter to call as possible. for example, the 'sort' parameter, there are two datatype of call from args. one is args.make_sort_args which sort=string. another is args.make_multi_sort_args which sort is List datatype for multi columns sort. the 'get' function should handle the 'sort' to determine multi='false or =true'. most case sort=string,  AuditCaseView is the first one to use sort='List'.   so I put  code to check datatype first to determine multi value:  
 if isinstance(kwargs['sort'], (list, tuple)):
            multi = True
Is this reasonable? I have tested some other endpoint like candidates..using 'make_sort_args', they all work. 

for filings endpoint, it override 'get' function, generate some duplicate codes. not ease to maintain in the future.
for report endpoint, it extend 'utils.Resource', not ApiResource, it looks like rewrite get function which we can use in views.py. it forces multi=True in get function.

I am not sure if I understand correctly? Thanks again.

fec-jli: we don't need override 'get' function in all endpoint resource classes as long as it extend ApiResource  and use basic code of 'get' function in views.py. right now, 50% of our endpoints don't override 'get' function.

ccostino: Thanks, @fec-jli!  Yes, you're right that the `ApiResource` is the base (one of the base) API resource classes.

I took a second look at this after reading over your remarks and I think you're right, this does make more sense given what would need to be overridden and when.  I was following the established pattern we had from before but, but that doesn't mean it can't change for the better!

Furthermore, while type checking is generally a code smell in Python, this is a situation where it does make sense and you used `isinstance` to do this instead of `type`, which is [the preferred way of doing this when need be](https://docs.quantifiedcode.com/python-anti-patterns/readability/do_not_compare_types_use_isinstance.html). :-)

Defaulting `mutli` to `False` makes sense, so as far as the change in `ApiResource` goes, this does make sense as is!  Also, both the `BaseFiling` and `ReportView` resources do have some customization (the `BaseFiling` is very minor, it is the count threshold).

So in this case, yes, let's keep this all as is, but what's missing is a test or two to account for the API change.  We could have a specific test around the multi-sort in `tests/test_audit.py` for the audit search specific implementation, and another general test or two in `tests/test_utils.py` in the [`TestSort`](https://github.com/18F/openFEC/blob/develop/tests/test_utils.py#L22) object that exercises the multi-sort functionality.  Does this sound reasonable to you?

webservices/common/views.py
ccostino: Disregard this note!  Please see discussion below.

tests/test_utils.py
```diff
@@ -37,6 +38,32 @@ def test_single_column_reverse(self):
...
+            factories.AuditCaseFactory(cycle=2012, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Girl', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2012, committee_name='Abc', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Ted', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='John', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+            factories.AuditCaseFactory(cycle=2014, committee_name='Abc', primary_category_id=-1, sub_category_id=-2, audit_case_id=1000),
+        ]
+        query, columns = sorting.multi_sort(models.AuditCase.query, ['-cycle', 'committee_name', ], model=models.AuditCase)
+        self.assertEqual(query.all(), audit[::-1])
+
```
ccostino: Tests!  🎉 :-)

fec-jli: Thanks Carlo always :)

PR #2935, opened 2018-02-18
Fixed show sub_category results when select only primary_category_id.

fec-jli:
create V0059.1__public_audit_case_mv to modify mv to support 'allow users to get results by selecting only a primary category'.

issue: https://github.com/18F/openFEC/issues/2920

### DIFF
--- a/data/migrations/V0059.1__public_audit_case_mv.sql
+++ b/data/migrations/V0059.1__public_audit_case_mv.sql
@@ -62,7 +62,7 @@ audit_case_init AS (
     LEFT JOIN cand_info ON cand_info.cand_id::text = ac.cand_id::text 
     AND cand_info.fec_election_yr = ac.election_cycle
 ),
-audit_case_sebset AS (
+audit_case_subset AS (
     SELECT 
         audit_case_finding_info.primary_category_id,
         audit_case_finding_info.sub_category_id,
@@ -470,7 +470,7 @@ audit_case_all AS (
         audit_id,
         candidate_id,
         candidate_name
-    FROM audit_case_sebset
+    FROM audit_case_subset
     UNION ALL
     SELECT         
         10000+row_number() over () AS idx,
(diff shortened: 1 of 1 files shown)


data/migrations/V0059.1__public_audit_case_mv.sql
```diff
@@ -0,0 +1,752 @@
...
+        ac.audit_id::integer AS audit_id,
+        ac.cand_id AS candidate_id,
+        coalesce(cand_info.cand_name, 'N/A'::text) AS candidate_name
+    FROM auditsearch.audit_case ac
+    LEFT JOIN cmte_info ON cmte_info.cmte_id::text = ac.cmte_id::text 
+    AND cmte_info.fec_election_yr = ac.election_cycle
+    LEFT JOIN cand_info ON cand_info.cand_id::text = ac.cand_id::text 
+    AND cand_info.fec_election_yr = ac.election_cycle
+),
+audit_case_sebset AS (
```
rjayasekera: Is this audit_case_sebset the correct column name you want? (audit_case_subset)

fec-jli: audit_case_sebset is subquery name, not column name.

fec-jli:
Business logic notes:
1)when primary category= all, total results=1026 rows, means there are 1026(current total) audit case (distinct, in audit_case table)
2)when chose only primary category, return distinct audit case rows which include the primary category you chosen.
3)when chose sub category based on a primary category, one by one, the total audit case rows > the above rows
which means one audit case can include multiple sub category.
4)when chose some sub category, may return 0 rows. that means there is no any audit case based on this category
/sub-category pair yet.

fec-jli:
because the data structure is more complicate, this fix works fine for now. 
we may write a function in the future, that would be better. 
db team may help to write a function.

PR #2937, opened 2018-02-19
Add "exclude" capabilities to filters.py 

lbeaufort:
Add capability to search for results not matching a certain value

Update `filters.py` : `filter_match`, `filter_multi`, `filter_fulltext` to use `not_in` functionality to exclude results with a certain value when passed `parameter=-value`. I also check for when API is passed "key=A and key!=B" and handled that behavior by splitting out the arguments

- [x] Use list comprehension in `filter_multi`?
- [x] Add automated tests
- [x] Use [instance](https://github.com/18F/openFEC/pull/2937#discussion_r169325116)
- [x] update all filter functions: `filter_fulltext`
- [x] Add combo exclude/include test for `filter_multi`

___
Filters that didn't get updated with exclude capabilities:
`filter_range` exclude - possibly for the future, this hasn't been requested.
`filter_contributor_type`? `filter_election` - not needed

### DIFF
--- a/webservices/filters.py
+++ b/webservices/filters.py
@@ -5,17 +5,47 @@ from webservices import exceptions
 from webservices.common import models
 
 
+def is_exclude_arg(arg):
+    # Handle string and int excludes
+    return str(arg).startswith('-')
+
+
+def parse_exclude_arg(arg):
+    # Integers will come in as negative and strings will start with -
+    if type(arg) is int:
+        return abs(arg)
+    else:
+        return arg[1:]
+
+
 def filter_match(query, kwargs, fields):
     for key, column in fields:
         if kwargs.get(key) is not None:
-            query = query.filter(column == kwargs[key])
+            # Handle string and int excludes
+            if is_exclude_arg(kwargs[key]):
+                query = query.filter(column != parse_exclude_arg(kwargs[key]))
+            else:
+                query = query.filter(column == kwargs[key])
     return query
+
 
 def filter_multi(query, kwargs, fields):
     for key, column in fields:
         if kwargs.get(key):
-            query = query.filter(column.in_(kwargs[key]))
+            #loop through list and exclude or include
+            exclude_list = []
+            include_list = []
+            for value in kwargs[key]:
+                if is_exclude_arg(value):
+                    exclude_list.append(parse_exclude_arg(value))
+                else:
+                    include_list.append(value)
+            if len(exclude_list) > 0:
+                query = query.filter(column.notin_(exclude_list))
+            if len(include_list) > 0:
+                query = query.filter(column.in_(include_list))
     return query
+
 
 def filter_range(query, kwargs, fields):
     for (min_key, max_key), column in fields:
@@ -24,6 +54,7 @@ def filter_range(query, kwargs, fields):
         if kwargs.get(max_key) is not None:
             query = query.filter(column <= kwargs[max_key])
     return query
+
 
 def filter_fulltext(query, kwargs, fields):
     for key, column in fields:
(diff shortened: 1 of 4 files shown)


webservices/filters.py
```diff
@@ -5,18 +5,42 @@
 from webservices.common import models
 
 
+def is_exclude_arg(arg):
+    # Handle string and int excludes
+    return str(arg).startswith('-')
```
ccostino: I think this is an appropriate abstraction here - if we ever change how we want to handle URL parameters for filtering, particularly around exclusion, we just have to worry about this spot. :-)

webservices/filters.py
```diff
@@ -5,18 +5,42 @@
...
+
+
+def parse_exclude_arg(arg):
+    # Integers will come in as negative and strings will start with -
+    if type(arg) is int:
+        return abs(arg)
+    else:
+        return arg[1:]
```
ccostino: Nicely done, accounting for different parameter types!

webservices/filters.py
```diff
@@ -5,18 +5,42 @@
...
 
 
+def is_exclude_arg(arg):
+    # Handle string and int excludes
+    return str(arg).startswith('-')
+
+
+def parse_exclude_arg(arg):
+    # Integers will come in as negative and strings will start with -
+    if type(arg) is int:
```
ccostino: You might want to consider using [`isinstance`](https://docs.python.org/3/library/functions.html#isinstance) here instead (here's a [detailed explanation why on Stack Overflow](https://stackoverflow.com/a/1549854)), this way if `WebArgs` ever changes how parameters of an `int` type are handled by subclassing, we'll be covered. :-)

As an example, this line would become:

```python
if isinstance(arg, int):
```

webservices/filters.py
```diff
@@ -5,18 +5,42 @@
...
+
+
 def filter_match(query, kwargs, fields):
     for key, column in fields:
         if kwargs.get(key) is not None:
-            query = query.filter(column == kwargs[key])
+            if is_exclude_arg(kwargs[key]):
+                query = query.filter(column != parse_exclude_arg(kwargs[key]))
+            else:
+                query = query.filter(column == kwargs[key])
```
ccostino: Exactly what I had in mind, nice!

webservices/filters.py
```diff
@@ -5,18 +5,42 @@
...
     for key, column in fields:
         if kwargs.get(key):
-            query = query.filter(column.in_(kwargs[key]))
+            # handle combination exclude/include lists
+            exclude_list = [parse_exclude_arg(value) for value in kwargs[key] if is_exclude_arg(value)]
+            include_list = [value for value in kwargs[key] if not is_exclude_arg(value)]
+            if exclude_list:
+                query = query.filter(column.notin_(exclude_list))
+            if include_list:
+                query = query.filter(column.in_(include_list))
```
ccostino: Looks great so far!

webservices/filters.py
```diff
@@ -25,23 +49,35 @@ def filter_range(query, kwargs, fields):
...
     for key, column in fields:
         if kwargs.get(key):
-            filters = [
-                column.match(utils.parse_fulltext(value))
-                for value in kwargs[key]
-            ]
-            query = query.filter(sa.or_(*filters))
+            exclude_list = [parse_exclude_arg(value) for value in kwargs[key] if is_exclude_arg(value)]
+            include_list = [value for value in kwargs[key] if not is_exclude_arg(value)]
+            if exclude_list:
```
lbeaufort: @ccostino and @LindsayYoung - the `filter_fulltext` changes are the part I feel could use the most scrutiny.

ccostino: I am mulling this over in my head and this all looks right to me.  The spot I'm not 100% on (but would be easy to check) is that the query gets constructed correctly given the mix of `and` and `or`.

I think this is constructed correctly, where things that are excluded are all `and` within the query, and things to be included are all combined with `or` (in other words:  a value is not `x` and `y` and `z`, but is `a` or `b` or `c`).  I just want to be sure. :-)

lbeaufort: Thanks @ccostino! I agree - when I checked the queries I was a bit surprised not to see parentheses in there. I updated the issue to do lots more of manual testing, especially for combo include/exclude for fulltext filters. I've tested a fair amount so far but I think I might take an inventory of which endpoints use `match`, `multi` and `fulltext` to see if I can do some more extensive/comprehensive manual tests.

PR #2964, opened 2018-02-27
Use gradle to build flyway distZip app

vrajmohan:


data/flyway/build.gradle
```diff
@@ -0,0 +1,83 @@
...
+            }
+            from(createMigrationsDir) {
+                into "migrations"
+            }
+            from(createRunScript) {
+                into "bin"
+            }
+        }
+    }
+}
```
rjayasekera: Can we have a brief discussion of what this file is doing and a overall explanation of gradle in general.  Looks like we use gradle to package (zip) this application for distribution and unzip/copy in the target environment. Why point to mavenCentral in the repositories section?

data/flyway/manifest_headless_flyway.yml
```diff
@@ -0,0 +1,13 @@
...
+  buildpack: java_buildpack
+  health-check-type: process
+  no-route: true
+  command: /home/vcap/app/flyway/bin/run.sh && echo SUCCESS && sleep infinity
+  path: build/distributions/flyway.zip
+  env:
+      CLASSPATH: app/gradle/lib/flyway-commandline-5.0.2.jar:app/gradle/lib/flyway-core-5.0.2.jar
+      JAVA_HOME: /home/vcap/app/.java-buildpack/open_jdk_jre
+  services:
+      - fec-flyway-creds
```
rjayasekera: Looks like this application runs as soon as the deployment is done.  Is that true?  Then why sleep infinity after the run? Need to run once only right?

data/flyway/run.sh
```diff
@@ -0,0 +1,2 @@
+JDBC_URL=`echo $VCAP_SERVICES | jq -r '.["user-provided"][0]["credentials"]["JDBC_URL"]'`
+/home/vcap/app/flyway/bin/flyway -n -locations=filesystem:/home/vcap/app/flyway/migrations -url="${JDBC_URL}" info && echo SUCCESS
```
rjayasekera: Need to change "info" to "migrate" during the real run.

rjayasekera:
This is great.  Vraj, thanks for your help on this issue.

data/flyway/manifest_headless_flyway.yml
ccostino: @rjayasekera that's exactly why that instruction is in there, because we only want it to run once. :-)

When you deploy an application, it'll run as soon as the deployment finishes.  However, there are also health checks involved and in most cases an application will be something that is intended to run indefinitely, e.g., a web server/process for a web application.  If the process ends for some reason, Cloud Foundry will restart it automatically for us (or at least try to).

In this case, we're deploying an application that has a finite end to it and isn't supposed to be run again or indefinitely, so we want it to stop.  Hence the `sleep infinity` instruction, which will run as soon as the operation completes successfully.  This will ensure that the application runs successfully but won't run multiple times.

These instructions are outlined in detail in the [cloud.gov documentation](https://cloud.gov/docs/getting-started/one-off-tasks/#deploy-an-app-that-performs-a-task).  If anything we may want to make a couple of adjustments here to match what they have written (or cross check them with the latest Cloud Foundry documentation and go with what's best).

Does this answer your question?

data/flyway/run.sh
ccostino: Nice catch, @rjayasekera!

data/flyway/build.gradle
ccostino: @rjayasekera that's a good idea; do you feel we need to have that before this is merged or are you comfortable moving forward with this work and discussing after?

ccostino:
🎉 awesome!

PR #2979, opened 2018-03-05
changed transation to transaction in docs.py

johnnyporkchops:
Fix misspelling in   /schedules/schedule_b/{sub_id}/ two_year_transaction_period description.

It looks like in [this PR](small definition tweak #1723) some changes were made and a word was misspelled, so this PR fixes that.


lbeaufort:
Nice catch, @johnnyporkchops!

PR #2990, opened 2018-03-06
[WIP]Feature/remove extra coalesce

hcaofec:


lbeaufort:
@hcaofec @fec-jli instead of removing it here, would it be possible to figure out why there's an extra one being added, and not add it there? 
cc: @ccostino because he worked on the expression-based sort work

hcaofec:
@lbeaufort  Thanks for reviewing it.  You are right that we should consider the case-sensitive situation.

In schedule B query case, by the expression-based sort, left_index already has coalesce function.  without checking,   **_sa.func.coalesce(left_index, comparator_** put another coalesce to it. That's how the double coalesce  happens.

[link for left_index](https://github.com/18F/openFEC/blob/develop/webservices/utils.py#L105) 
[link for sort_column](https://github.com/18F/openFEC/blob/develop/webservices/utils.py#L217)
[link for sort_column](https://github.com/18F/openFEC/blob/develop/webservices/sorting.py#L104)

@ccostino

lbeaufort:
@hcaofec thanks for the tip on testing this locally. I talked with @LindsayYoung who thinks that case sensitivity won't matter here. While the performance improvement might be negligible, I do think this is easier to read, if you agree: 
```
 if 'coalesce' not in str(left_index):
    left_index = sa.func.coalesce(left_index, comparator)
```

fec-jli:
how about, if don't call sort_expression, then don't need check left_index  ?

`             if not self.sort_column[3]:
                 comparator = self.max_column_map.get(
                     str(left_index.property.columns[0].type).lower()
                 )
   				left_index = sa.func.coalesce(left_index, comparator)            
             else:
                 comparator = self.max_column_map.get(self.sort_column[5])       
            	 if str(left_index).lower().find('coalesce') == -1:
                    left_index = sa.func.coalesce(left_index, comparator)`

LindsayYoung:
Great explanation on the issue in your comment @hcaofec

Of the proposed changes I think the easiest to read version of this change is:
```
 if 'coalesce' not in str(left_index):
    left_index = sa.func.coalesce(left_index, comparator)
```
Though, the current one works as well. @hcaofec do you have a preference?

Small thing- I took off the [WIP] we really just want that naming for PRs that are not ready for review yet.

webservices/utils.py
```diff
@@ -113,7 +113,11 @@ def _fetch(self, last_index, sort_index=None, limit=None, eager=True):
             else:
                 comparator = self.max_column_map.get(self.sort_column[5])
 
-            left_index = sa.func.coalesce(left_index, comparator)
+            
+            if 'coalesce' not in str(left_index):
+                left_index = sa.func.coalesce(left_index, comparator)
```
ccostino: Nicely done, @hcaofec! :-)

PR #3002, opened 2018-03-12
Fix Legal search endpoint 500 error for unexpected parameters

fec-jli:
## Summary (required)
For our API design, we can follow the “gracious" way to handle request parameters.

Scenario1) if miss mandatory parameter:
	a) if have default. set to default.
	b) if don’t have default, abort right away, don’t need execute query

Scenario2) if miss need parameter:
	set to default or None

Scenario3)For extra parameters or bad value parameters:
	ignore those parameters quietly

this ticket only fix /legal/search/ endpoint with nonsense parameters: type.
we will set default type=all

- Addresses # [_Issue number_]
[https://github.com/18F/openFEC/issues/2934](https://github.com/18F/openFEC/issues/2934)

in legal.py:
if doc_types is not in one of these 4 doc types :'statutes', 'regulations', 'advisory_opinions', 'murs'
then reset type = all (= four default search types)

and add more except handle.

—[1] Reproduce error and fix:
1)Check out “fix_500_error” branch on local.
set SQLA_CONN to dev db

2) ./manage.py runserver

3) on localhost, correct link: type=all, no error
http://127.0.0.1:5000/v1/legal/search/?type=all&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY

on dev api, correct link: type=all, no error
https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=all&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY

4) on dev api, error link: type=error_type, 
https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=error_type&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY
(because type is not in one of these('statutes', 'regulations', 'advisory_opinions', 'murs'), will get 500 error.

on localhost fixed, reset type=all, so no error, get some results.
http://127.0.0.1:5000/v1/legal/search/?type=error_type&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY

-[2]Test with extra parameters:
1)on dev api, as long as type value is correct and q is correct , will ignore extra parameters, get some results
https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=all&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY&nonsenseword=nonoe%26%26+type+%25SystemRoot%25%5C%5Cwin.ini+%26%26+&hits_returned=20&from_hit=0

type is not correct, will get 500 error:
https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=error_type&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY&nonsenseword=nonoe%26%26+type+%25SystemRoot%25%5C%5Cwin.ini+%26%26+&hits_returned=20&from_hit=0

2)on localhost. type is not correct, no error, get some results based on type=all
http://127.0.0.1:5000/v1/legal/search/?type=error_type&q=public&hits_returned=3&from_hit=0&api_key=DEMO_KEY&nonsenseword=nonoe%26%26+type+%25SystemRoot%25%5C%5Cwin.ini+%26%26+&hits_returned=20&from_hit=0

-[3] if q with nonsenseword parameter, will get 500 error on dev 
https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=regulations&q=+public+%26%26+type+%25SystemRoot%25%5C%5Cwin.ini+%26%26+&hits_returned=20&from_hit=0&api_key=DEMO_KEY

on localhost. will catch 400 “Elasticsearch failed to execute query”
http://127.0.0.1:5000/v1/legal/search/?type=regulations&q=+public+%26%26+type+%25SystemRoot%25%5C%5Cwin.ini+%26%26+&hits_returned=20&from_hit=0&api_key=DEMO_KEY

http://127.0.0.1:5000/v1/legal/search/?q=president%20AND%20OR&type=advisory_opinions

Legal Doc Search

### DIFF
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -85,18 +85,29 @@ class UniversalSearch(utils.Resource):
         else:
             doc_types = [kwargs.get('type')]
 
+            # if doc_types is not in one of these 4 doc types :'statutes', 'regulations', 'advisory_opinions', 'murs'
+            # then reset type = all (= four default search types)
+            if doc_types[0] not in ('statutes', 'regulations', 'advisory_opinions', 'murs'):
+                doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
+
         hits_returned = min([200, hits_returned])
 
         results = {}
         total_count = 0
 
         for type_ in doc_types:
-            query = query_builders.get(type_)(q, type_, from_hit, hits_returned, **kwargs)
             try:
+                query = query_builders.get(type_)(q, type_, from_hit, hits_returned, **kwargs)
                 formatted_hits, count = execute_query(query)
+            except TypeError as te:
+                logger.info(te.args)
+                raise ApiError("Not a valid search type", 400)
             except RequestError as e:
                 logger.info(e.args)
-                raise ApiError("Could not parse query", 400)
+                raise ApiError("Elasticsearch failed to execute query", 400)
+            except:
+                logger.info("Unexpected Error")
+                raise ApiError("Unexpected Error", 400)
             results[type_] = formatted_hits
             results['total_%s' % type_] = count
             total_count += count
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -26,6 +26,7 @@ INNER_HITS = {
     }
 }
 
+ALL_DOCUMENT_TYPES = ['statutes', 'regulations', 'advisory_opinions', 'murs']
 
 class GetLegalCitation(utils.Resource):
     @property
@@ -81,14 +82,14 @@ class UniversalSearch(utils.Resource):
         }
 
         if kwargs.get('type', 'all') == 'all':
-            doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
+            doc_types = ALL_DOCUMENT_TYPES
         else:
             doc_types = [kwargs.get('type')]
 
-            # if doc_types is not in one of these 4 doc types :'statutes', 'regulations', 'advisory_opinions', 'murs'
-            # then reset type = all (= four default search types)
-            if doc_types[0] not in ('statutes', 'regulations', 'advisory_opinions', 'murs'):
-                doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
+            # if doc_types is not in one of ALL_DOCUMENT_TYPES
+            # then reset type = all (= ALL_DOCUMENT_TYPES)
+            if doc_types[0] not in ALL_DOCUMENT_TYPES:
+                doc_types = ALL_DOCUMENT_TYPES
 
         hits_returned = min([200, hits_returned])
 


webservices/resources/legal.py
```diff
@@ -85,18 +85,29 @@ def get(self, q='', from_hit=0, hits_returned=20, **kwargs):
         else:
             doc_types = [kwargs.get('type')]
 
+            # if doc_types is not in one of these 4 doc types :'statutes', 'regulations', 'advisory_opinions', 'murs'
+            # then reset type = all (= four default search types)
+            if doc_types[0] not in ('statutes', 'regulations', 'advisory_opinions', 'murs'):
+                doc_types = ['statutes', 'regulations', 'advisory_opinions', 'murs']
```
ccostino: This is a great start in accounting for this and resolves the issue! :-)  I think there are a couple of opportunities for improvement, too:

* I noticed that you're checking for the presence of the value provided in a `tuple` instead of a `list` (you have a collection of values wrapped in parentheses instead of square brackets).  This isn't inherently bad, especially in this case as you're just checking to see if the value exists from a known set, but it might lead someone to infer that the value you're checking (`doc_types[0]`) is also in a `tuple`, which it is not (it's a `list` as defined above on line 86).
* This collection of values are being repeated on three lines that I see, so this might be an opportunity for a bit of refactoring to clean this up a bit and prevent duplication (also known as DRY - Don't Repeat Yourself).

In terms of refactoring, I'd approach it like this:

* Create a new constant, defined just under the `INNER_HITS` constant you see at lines 18 - 27 in the same file), and call it `ALL_DOCUMENT_TYPES` or something to that effect, and set it like this:  `ALL_DOCUMENT_TYPES = ['statutes', 'regulations', 'advisory_opinions', 'murs',]`
* Change line 84 to be this instead:  `doc_types = ALL_DOCUMENT_TYPES`
* Change line 90 to this:  `if doc_types[0] not in ALL_DOCUMENT_TYPES`
* Change line 91 to this:  `doc_types = ALL_DOCUMENT_TYPES`

Does this make sense?  By doing this, you're only defining the document types in one place so if they ever change in the future, it's just the constant that has to be modified and all of the rest of the logic leverages the constant.  Let me know if you have any questions or other ideas!

webservices/resources/legal.py
```diff
@@ -85,18 +85,29 @@ def get(self, q='', from_hit=0, hits_returned=20, **kwargs):
...
+            except TypeError as te:
+                logger.info(te.args)
+                raise ApiError("Not a valid search type", 400)
             except RequestError as e:
                 logger.info(e.args)
-                raise ApiError("Could not parse query", 400)
+                raise ApiError("Elasticsearch failed to execute query", 400)
+            except:
+                logger.info("Unexpected Error")
+                raise ApiError("Unexpected Error", 400)
```
ccostino: Nicely done with the improvements to the error handling!  It's always good having more specific errors. :-)

webservices/resources/legal.py
fec-jli: great suggestion. every clean and clear. will do.
Thanks.

Issue #2854, opened 2018-01-04
Create itemized tables (sched_b) and transfer data from fecp to postgresql database in cloud


fecjjeng:
This is derived from Issue #2686: Golden Gate data loading time.

fecjjeng:
This issue is derived from issue #2686 and is related to issue #2724

fecjjeng:
Currently, coding of the PL/SQL package and Java programs are done. Initial load and daily updates process to the cloud databases of the itemized tables are done. The tables are in separated set of tables in a different schema (disclosure) so not to impact any current APIs accessing the aggregate tables in public schema.

LindsayYoung:
We will want to modify the API code- mostly model references so that it references the new tables. We will want to adjust the tests so that they use the schema  and we will want to add a migration that deletes the old tables.

fecjjeng:
Comparison between table structure of disclosure.fec_fitem_sched_b (new) and public.ofec_sched_b_master (old), following is the result
<img width="1269" alt="sched_b_striucture_comparison" src="https://user-images.githubusercontent.com/24943130/37847042-df55b98a-2ea5-11e8-8527-16dbb6af9724.png">
public.ofec_sched_b_master (old) table has two more columns: timestamp and two_year_transaction_period.  
Data in column timestamp is used in queue tables processing, which will not be needed in the data population process for disclosure.fec_fitem_sched_b tables
Data in column two_year_transaction_period is exactly the same as in column election_cycle in each and every row, so it is not needed

hcaofec:
need to check if column two_year_transaction_period is used by endpoints or materialized views or views before dropping it.
@fecjjeng @ccostino

ccostino:
Yes, we've caught up about this some in Slack but need to have a full discussion around it.  The `two_year_transaction_period` is used by quite a few of the endpoints as it is the column that powers the partitions of the schedule A and B data for now (the `CHECK CONSTRAINT` is based off of it; the endpoints expect it as it's a required field, too.

There is an opportunity to consolidate and cleanup here, so I think we can find a path forward, we'll just have to go over everything and plan accordingly. :-)

fecjjeng:
The FECP-driven tables use election_cycle to power the partitions of schedule A ad B  (the CHECK CONSTRAINT is based on the election_cycle column).  I think there is several possible route we can take.  We will discuss to see what is the best way to proceed.

fecjjeng:
[...]

PR #3005, opened 2018-03-12
[WIP] Add more robust itemized schedule nightly processing

ccostino:
**WIP:  PLEASE DO NOT MERGE**

This changeset introduces more resilience to the nightly processing of the itemized schedule data by checking for a specific error condition and attempting to fix it before trying to process the data again.  We find that at times we run into a situation where an itemized record will be present in the new queue table as well as the master table, which indicates the need for an update, but the record is absent from the old queue table.  This change rectifies that by retrieving the values from the master table and creating a copy in the old queue, then retrying the processing again.  If the processing succeeds or fails in a fatal way (another type of error entirely or an error in trying to copy the record into the old queue), the loop is broken from the processing so the rest of the nightly refresh can continue.

TODO:
* [ ] Review the problem and proposed solution
* [ ] Update tests as necessary

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -207,12 +207,18 @@ def refresh_itemized_a():
     """Used to refresh the itemized Schedule A data."""
 
     logger.info('Updating Schedule A...')
-    message = partition.SchedAGroup.process_queues()
 
-    if message[0] == 0:
-        logger.info(message[1])
-    else:
-        logger.error(message[1])
+    while True:
+        message = partition.SchedAGroup.process_queues()
+
+        if message[0] == 0:
+            logger.info(message[1])
+            break
+        elif message[0] == 1:
+            logger.warning(message[1])
+        elif message[0] >= 2:
+            logger.error(message[1])
+            break
 
     logger.info('Finished updating Schedule A.')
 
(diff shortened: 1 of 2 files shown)


ccostino:
Another thought I had about this is we could create another short process before the main queue processing that looks for these kind of records in this situation and clean them all up prior to doing the full processing.  In other words, find all records in the master tables that also have matching records in the new queue tables but not the old queue tables, then either copy them to the old queue tables or just remove them outright from the master tables (remember, this processing happens for both schedule A and B, and has been an issue for both).  Once that is done, then we can do the processing as we had before.

I think that might be a little bit cleaner than having nested `try/except` blocks, but the work so far is at least a start based on the idea I had after chatting with @rjayasekera a bit yesterday.

@rjayasekera, @hcaofec, and @fecjjeng, what do you all think of this so far?  We should review over video chat/in person, but I wanted to make sure I captured all of my thoughts and a proof of concept before I lost them! :-)

fecjjeng:
We should accelerate the process of switching to use partitioned tables trasferred from fecp as well.  I had created the first issue (#2854) for the first step of this process.

ccostino:
Oh, you're right, thanks for the reminder @fecjjeng!  I wonder then if that work should take place instead of this then?

ccostino:
I'm definitely in agreement!  Recapping what we discussed:

* Stop work going down this path
* Repartition schedule B in `dev`; do this again if we need to in the interim before we're able to complete the switch
* Start the work to switch to @fecjjeng's partitioned tables instead, which won't require the queue tables and nightly refresh processing

Thanks again, everyone!

Issue #2646, opened 2017-09-07
Improve performance when automatically loading MURs


vrajmohan:
The code that refreshes MURs works by discovering the oldest document to be reloaded and then reloads every document after that. Unlike AOs, it is not only recent MURs that can be updated. Therefore this code is suboptimal for MURs. 

The code needs to be refactored to only reload the actual documents changed.

This issue has been created by pulling out the remaining work from #2445

 - Consider updating the view
- [ ] As part of this work, make a migration to filter for `case_type=MUR` for `cases_with_parsed_case_serial_numbers`. See https://github.com/18F/openFEC/blob/develop/data/migrations/V0021__fecmur_schema_objects.sql#L97
- [ ] Update team & docs with `load_current_murs` new flag is `-m` -  `./manage.py load_current_murs -m '2804R'`

lbeaufort:
Once this work is done, we could explore having celery worker check for modified AOs and MURs more often - maybe every 5 min

AmyKort:
Merged, so closing

PR #3010, opened 2018-03-15
Modify MUR reload to only load specific MURs, check every 5 minutes

lbeaufort:
Questions:
- [x] ~Should `cases_with_parsed_case_serial_numbers` have an index on it? It'll have 2,000 rows.~ Underlying table `"case"` has `    "case_pkey" PRIMARY KEY, btree (case_id)`" - regular views inherit the indexes of the underlying tables.
- [x] ~Is there a better way than checking for all AOs/MURs changed in the last day? It looks like we reload them all day the day they’re modified. If I change that to every 5 minutes, will that be a big performance hit? Maybe we could look for ones modified in the past X hours?~ The plan is to check for any modified AOs or MURs in the past 8 hours instead of 24. By going from every 15m to every 5m we increase the load by 3 times. So, going from 24h -> 8h would even it up.

## Summary

- Addresses https://github.com/fecgov/openFEC/issues/2646

_Change MUR reload to only load the specified MUR(s) and not continue updating every subsequent MUR. Unlike AOs, MURs are not published sequentially._

_Run the celery task to check for modified MURs/AOs every 5 minutes._

- Run the migration file - `flyway migrate`
- celery -
Set up and run celery [using our wiki](https://github.com/fecgov/openFEC/wiki/Set-up-Redis-Celery-on-local-and-test-the-downloads)
If no MURs have been modified on `dev`, change the pg_date of some MURs:

**1) connect to dev DB**
**2) Update pg_date:**
```
BEGIN
UPDATE fecmur.cases_with_parsed_case_serial_numbers
SET pg_date = NOW()
WHERE case_no IN ('7000', '7004)
COMMIT
```

- manage.py -
Make sure elasticsearch is running and you have the latest credentials for `dev` s3 (see [legal resources instructions](https://docs.google.com/document/d/1e_sj6xD8hJiyCp20LjK8jITJlCBJHsLzfo6rPnvJBaw/edit) for more info
`./manage.py load_current_murs -m '2804R'` to load one or 
`./manage.py load_current_murs` to load all

- Automatic reload of modified MURs with celery task
- Manual reload with manage.py task


webservices/tasks/legal_docs.py
```diff
@@ -17,13 +17,11 @@
     LIMIT 1;
 """
 
-RECENTLY_MODIFIED_STARTING_MUR = """
+RECENTLY_MODIFIED_MURS = """
     SELECT case_no, pg_date
     FROM fecmur.cases_with_parsed_case_serial_numbers
     WHERE pg_date >= NOW() - '1 day'::INTERVAL
```
lbeaufort: I added the `case_type = MUR filter` to the VIEW with a new migration file ^^

data/migrations/V0067__filter_mur_view.sql
```diff
@@ -0,0 +1,32 @@
...
+PR #3010
+Filter existing view cases_with_parsed_case_serial_numbers
+    by case_type = 'MUR'
+*/
+
+SET search_path = fecmur, pg_catalog;
+
+DROP VIEW IF EXISTS cases_with_parsed_case_serial_numbers_tmp;
+
+CREATE VIEW cases_with_parsed_case_serial_numbers_tmp AS
```
vrajmohan: We don't need to create a view with an `_tmp` suffix and then rename it at the end. PostgreSQL provides `CREATE OR REPLACE VIEW...` to do this in one shot - https://www.postgresql.org/docs/10/static/sql-createview.html

webservices/legal_docs/current_murs.py
```diff
@@ -19,10 +19,19 @@
...
     ORDER BY case_serial
 """
 
+SINGLE_MUR = """
+    SELECT DISTINCT
+        case_id,
+        case_no,
+        name
+    FROM fecmur.cases_with_parsed_case_serial_numbers
+    WHERE case_type = 'MUR'
```
vrajmohan: This clause can be removed now that the underlying view has it.

PR #3003, opened 2018-03-12
[WIP]Fix schedule A zip search

hcaofec:
This PR will refactor the zip code search for SA. No matter what zip code a user enters, it will always find returns transactions  that match the first 5-digit of the zip code

[#2980](https://github.com/fecgov/openFEC/issues/2980#issuecomment-371876195)

- new type of filter was added : filter_multi_start_with
- this new filter is used by SA
- test case was modofied: test_filter_sa_zip

## How to test the changes locally
-run API locally and test the searching feature

Issue #2980, opened 2018-03-05
Fix zipcode search - limit match to left(5) of parameter value.


lbeaufort:
See related cms issue: https://github.com/fecgov/fec-cms/issues/1391

When a 9-digit zip is passed to the `/schedules/schedule_a/` endpoint, we should truncate it to 5 characters. Not all filers report 9-digit zips, so searching by zip+4 returns unreliable data. In addition, currently searching by 5-digit zips is excluding records that have 9-digit zips. Also, we should add some automated tests to test_itemized to confirm this behavior. 

- [ ] `schedule_a.py` - match the  inputted zip value to left 5 characters to the left(5) of `contributor_zip` column to  in the `ScheduleA` model - do this in the resource with `if kwargs.get('contributor_zip'):...`
- [ ] `test_itemized.py` - add some automated tests for zip+4 searches

Should all schedules work this way? only the ones that search by zip. Maybe next look at `ScheduleAByZip` - looks like it doesn't return results when you pass +4.
https://api.open.fec.gov/v1/schedules/schedule_a/by_zip/?api_key=DEMO_KEY&zip=200162409&page=1&per_page=20

**Example**

PaulClark2:
For context: 

The  ZIP code data in the itemized transactions of scheudles is inconsistent. Some filers provide 5-digit ZIP codes and some filers provide 9-digit ZIP codes. Currently, if a user is searching for ZIP code 20009-2415, the app's behavior varies depending if the user searches for 20009 or for 20009-2415. 

The underlying data might be `200092415` or `20009`. We do not standardize ZIP codes to make them all ZIP+4. When a user enters a 5-digit ZIP code, we do not return any transactions with 9-digit ZIP codes even if first 5-digits match. If a user enters a 9-digit ZIP code, we do not return any transactions with 5-digit ZIP codes even if first 5-digits match. To make the results returned to users predictable, , we plan to match left(5) of the requested ZIP code to left(5) of the ZIP code in the database.

hcaofec:
My thoughts:

first, make sure any non numeric character is removed from zip argument ( by endpoint );

when user enter a 5-digit zip ( or 6, 7, 8 by error ), the endpoint should return all transactions with first 5-digit match no matter how many digits of the zip code they have in database  ( could be 5 ,6,7,8, 9, or more in table  );

[...]

Issue #2962, opened 2018-02-27
Investigate why 422 (missing parameters) errors are coming through as 500 (Internal Server) errors


lbeaufort:
We've designed the API to throw 422 (missing parameters) errors when required parameters are missing. However, it looks like the production environment is throwing 500 (internal server) errors when we're trying to throw 422 errors. This is problematic in that it makes user errors look like API bugs and time-consuming to diagnose.

Locally, this throws a 422 for missing district but throws a 500 on the server: 

500 error:
https://api.open.fec.gov/v1/elections/?API_KEY=DEMO_KEY

422 error:
http://localhost:5000/v1/elections/

```
{
message: {
cycle: [
"Missing data for required field."
],
office: [
"Missing data for required field."
]
},
status: 422
}
```

We should make sure our develop/stage/production environment return the proper server errors.

lbeaufort:
Update: it looks like production servers might be throwing 500 when we're trying to throw 422 errors. Locally, this throws a 422 for missing district but throws a 500 on the server. 

see `check_election_arguments`: https://github.com/18F/openFEC/blob/develop/webservices/utils.py

ccostino:
Furthermore, it appears that `office` is required for this endpoint but it is not showing up as required in the API (Swagger) docs.  Compounding that is the fact that `district` is a conditionally required parameter if `office` is present, so there are a few things that need to be adjusted and double checked to make sure this is wired up correctly.

ccostino:
It seems like there are two things happening here:

* `ApiErrror` exceptions aren't being surfaced properly, which could be impacting the API in a broad manner
* This particular endpoint (`/elections`) is not properly handling its required and conditional arguments

lbeaufort:
@PaulClark2 @jwchumley - This error was reported by an API user - it definitely causes usability problems for data users.

lbeaufort:
@ccostino @LindsayYoung : I think the issue and a workaround are described here, but I don't really follow the technical implementation of workaround. Do you mind taking a look?

Issue: https://github.com/flask-restful/flask-restful/issues/280
Workaround: https://github.com/flask-restful/flask-restful/issues/280#issuecomment-280648790

lbeaufort:
@fec-jli it looks like it's swallowing the 400 error you're expecting here too:
From https://github.com/fecgov/openFEC/pull/3002
-[3] if q with nonsenseword parameter, will get 500 error on prod, after release will catch 400 exception.
[...]

PR #3012, opened 2018-03-19
Feature/add filter multi start with

hcaofec:
## Summary (required)

- Addresses [#2980](https://github.com/fecgov/openFEC/issues/2980)

This PR will refactor the zip code searching for Schedule A. No matter what zip code a user enters, it will always return transactions that match the first 5-digit of the zip code

- new type of filter was added : filter_multi_start_with
- this new filter is used by schedule A endpoint
- three test cases are created for this fix
- indexed are added to improve performance
- exact 5-digit zip code limitation are added
- functions are added to handle exclude/include list

- run API locally and test the searching feature

**old PR of this ticket**
[#3003](https://github.com/fecgov/openFEC/pull/3003)


tests/test_itemized.py
```diff
@@ -106,15 +107,29 @@ def test_filter(self):
...
+            factories.ScheduleAFactory(contributor_zip=9678912),
+            factories.ScheduleAFactory(contributor_zip=967891234)
         ]
 
         results = self._results(api.url_for(ScheduleAView, contributor_zip=96789, **self.kwargs))
-        self.assertEqual(len(results), 1)
-        self.assertEqual(results[0]['contributor_zip'], '96789')
+        self.assertEqual(len(results), 3)
+        
+        results = self._results(api.url_for(ScheduleAView, contributor_zip=9678912, **self.kwargs))
```
vrajmohan: Even though this is what we discussed, it just feels wrong to me as far as the API is concerned. We should do one of:
1. Accept only (exactly) 5-digit zip codes for searches and return rows that have those 5 digits as a prefix.
2. Accept 5-digit zip codes for searches and return rows that have those 5 digits as a prefix. If 9 digits are given, do an exact match on those 9 digits. 

As the extra 4 digits in ZIP+4 is a mailing route that can change, don't accept 6, 7 or 8 digits.

Pinging @ccostino, @lbeaufort, @fecjjeng, @rjayasekera, @fec-jli .

webservices/filters.py
```diff
@@ -71,6 +71,25 @@ def filter_fulltext(query, kwargs, fields):
                 query = query.filter(sa.or_(*filters))
     return query
 
+def filter_multi_start_with(query, kwargs, fields):
+    for key, column in fields:
+        if kwargs.get(key):
+            exclude_list = [parse_exclude_arg(value) for value in kwargs[key] if is_exclude_arg(value)]
```
vrajmohan: These 2 lines occur so often. Perhaps they can be extracted into a function?

tests/test_itemized.py
lbeaufort: @vrajmohan - Option 1 would be my preference. Do you envision throwing validation if an API user supplies a zip over 5 digits? We do have some weirdness with 400 validation errors showing up as 500 errors, so that might be misleading.

For option 2, here's why I wouldn't recommend it:
**Data is inconsistent**
1 | Laura | 999 E St | 20463
2 | Laura | 999 E St | 204630923

**User searching for 204630923 would expect it to return:**
1 | Laura | 999 E St | 204630923
2 | Laura | 999 E St | 204630923

**..But because of inconsistent data, it only returns:**
2  | Laura  | 999 E St | 204630923

**and is missing**
1 | Laura | 999 E St | 20463
**...even though that address is in that area**

vrajmohan: I also like Option 1. Yes, I would throw a validation error for any zip other than 5 digits.

lbeaufort: Sounds good to me. Here's the issue for the 400 errors - looks like it's tagged for this 5.3. https://github.com/fecgov/openFEC/issues/2962

webservices/filters.py
hcaofec: Good point!  We should create a separate ticket as enhancement for this issue.

tests/test_itemized.py
hcaofec: This is more like product feature to me. Maybe we should get stakeholders to weigh in on this issue. Would there be case when user enters other than 5-digit zip?  
If user enters 123, the API will return sets starting with 123
if user enter 123456789, the API will return sets starting with 12345

maybe validation for minimum digit entered for zip instead?

@PaulClark2 @jwchumley

fec-jli:
It is better to do validation at front-end, save api call.
because we use html 5, so it is easy to do this, something like this: 
<Label>ZIP Code</Label><input type="text" pattern="[0-9]{5}" title="Five digit zip code" />

business rule can decide min-digit limitation.

hcaofec:
Is there an example of openFEC endpoint performing validations against user-entered data now?  What is the rule of what error message to use when a user enters an invalid data?   Should it be included in this PR  or a separate ticket?
@lbeaufort @fec-jli

fec-jli:
@hcaofec For I know, maybe wrong, no validation error message from backend, only some exception messages in log files.
 on front-end, there are some messages for user.

lbeaufort:
@hcaofec @fec-jli check out `ApiError`: https://github.com/fecgov/openFEC/search?utf8=%E2%9C%93&q=ApiError&type=

webservices/filters.py
vrajmohan: I suggest that as the work required is so small, we just do it rather than create a separate ticket, groom it, assign it in a sprint, etc.

webservices/filters.py
```diff
@@ -31,10 +40,10 @@ def filter_match(query, kwargs, fields):
 
 def filter_multi(query, kwargs, fields):
     for key, column in fields:
-        if kwargs.get(key):
-            # handle combination exclude/include lists
-            exclude_list = [parse_exclude_arg(value) for value in kwargs[key] if is_exclude_arg(value)]
-            include_list = [value for value in kwargs[key] if not is_exclude_arg(value)]
+        # if kwargs.get(key):
```
lbeaufort: I think we want to still check for `if kwargs.get(key):` here. If not, we need to fix the indentation. Also, we need two new lines between functions here.

webservices/filters.py
```diff
@@ -17,6 +17,15 @@ def parse_exclude_arg(arg):
     else:
         return arg[1:]
 
+def build_exclude_list(value_list):
```
lbeaufort: Rather than calling this all the time, even without a value, 
I think we should add back in `if kwargs.get(key):` before parsing the values. Also, it might be slightly more effective to do one function like this:
```
def split_list_args(value_list):
    exclude_list = [parse_exclude_arg(value) for value in value_list if is_exclude_arg(value)]
    include_list = [value for value in value_list if not is_exclude_arg(value)]
    return exclude_list, include_list
```
and then
``exclude_list, include_list = split_list_args(kwargs.get(key))``

webservices/filters.py
```diff
@@ -54,9 +63,8 @@ def filter_range(query, kwargs, fields):
 
 def filter_fulltext(query, kwargs, fields):
     for key, column in fields:
-        if kwargs.get(key):
-            exclude_list = [parse_exclude_arg(value) for value in kwargs[key] if is_exclude_arg(value)]
-            include_list = [value for value in kwargs[key] if not is_exclude_arg(value)]
+            exclude_list = build_exclude_list(kwargs.get(key))
```
lbeaufort: Fix indentation if not adding back in the `if` statement

webservices/filters.py
```diff
@@ -71,6 +79,23 @@ def filter_fulltext(query, kwargs, fields):
                 query = query.filter(sa.or_(*filters))
     return query
 
+def filter_multi_start_with(query, kwargs, fields):
+    for key, column in fields:
+            exclude_list = build_exclude_list(kwargs.get(key))
```
lbeaufort: see above - indentation

webservices/resources/sched_a.py
```diff
@@ -76,6 +80,16 @@ def args(self):
     def build_query(self, **kwargs):
         query = super().build_query(**kwargs)
         query = filters.filter_contributor_type(query, self.model.entity_type, kwargs)
+
+        if kwargs.get('contributor_zip'):
+            for value in kwargs['contributor_zip']:
+                if re.search('^\d{5}$',value) is None and re.search('^[-]\d{5}$',value) is None:
```
vrajmohan: Just a suggestion - `'^-?\d{5}$'` will achieve the same result with a single expression.

lbeaufort:
Really great work, @hcaofec! Thank you!

Issue #2884, opened 2018-01-25
Update cloud.gov manifest setup in API


ccostino:
[Manifest inheritance](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#minimize-duplication) has been deprecated in Cloud Foundry in favor of another type of setup using something called a YAML anchor.

Instead of inheriting manifest files from each other, you declare a defaults section in your manifest file that other sections then reference.  Unfortunately what this means for us is increased duplication since we have separate manifests for each space (and in the case of the API, separate manifests for each application to boot), but we want to make sure this is accounted for before the manifest inheritance is just shut off completely on us (this change happened about a year ago).

To update the API we'll have to do the following:

* [ ] Copy everything in the base manifest files to their respective application files for each space
* [ ] Make sure each manifest file for each application in each space still works as expected
* [ ] change host -> routes

LindsayYoung:
Good catch! Thank you for making this ticket

AmyKort:
Merged, so closing.

PR #3015, opened 2018-03-19
Update manifests - inheritance and host are depreciated

lbeaufort:
## Summary (required)

- Addresses #2884 

[Manifest inheritance](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#minimize-duplication) and [the host attribute](http://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#route-attribute) have been deprecated in Cloud Foundry.

To update the API we'll have to do the following:

* [x] Copy everything in the base manifest files to their respective application files for each space
* [x] tested a deploy to `dev` with a temporary change to the branch used in the deploy task: https://circleci.com/gh/fecgov/openFEC/56
* [X] change host -> routes
* [x] Remove unused `manifest_feature.yml`


manifests/manifest_celery_beat_prod.yml
```diff
@@ -1,7 +1,22 @@
...
+  PRODUCTION: True
+  WEB_CONCURRENCY: 4
+services:
+  - fec-api-search
+  - fec-redis
+  - fec-creds-prod
+  - fec-s3-prod
 applications:
   - name: celery-beat
     instances: 1
```
ccostino: Interesting, we have one instance in `prod` but 2 in `dev`?  We may want to take another look at this quick.

@LindsayYoung, should we swap these values or just reduce `dev`?  This is the `celery-beat` app, not the worker.

manifests/manifest_celery_worker_dev.yml
```diff
@@ -1,7 +1,20 @@
...
+  NEW_RELIC_LOG: stdout
+  WEB_CONCURRENCY: 4
+services:
+  - fec-api-search
+  - fec-redis
+  - fec-creds-dev
+  - fec-s3-dev
 applications:
   - name: celery-worker
     instances: 2
```
ccostino: See note in the `prod` manifest file for this, it seems off to me.

ccostino: I caught up with @LindsayYoung and confirmed my suspicions, we want just one instance of this because it is the piece that handles the scheduled tasks.  With multiple instances we don't want to run the risk of tasks firing off more than once!

If you wouldn't mind adjusting this back to `1`, @lbeaufort, then this PR will be ready to go. :-)

manifests/manifest_celery_beat_prod.yml
ccostino: Disregard!  I was mixing up this manifest for the `celery-beat` app with the `celery-worker` app!

manifests/manifest_celery_worker_dev.yml
ccostino: Okay, total FAIL on my part, I'm sorry!  I mixed this manifest up with the `celery-beat` manifest above.  We're actually okay here, and it may make sense to keep 2 instances just so that we continue being able to test the setup of multiple instances, especially if we make any changes to the setup of Celery.

PR #3018, opened 2018-03-21
Prevent flask-restful from taking over all errors

vrajmohan:
Fixes fecgov/openFEC#2962

Thanks to @lbeaufort for discovering flask-restful/flask-restful#578.

It was difficult to investigate this in local environments as they run
in debug mode where all exceptions are propagated. 

## Summary (required)

- Addresses #2962 

- Turn off debug mode by running
```
python manage.py runserver --no-debug
```
- Access the API endpoint with an invalid query, e.g., `curl -v http://localhost:5000/v1/elections/`

You should get a clean 422 response with a proper JSON message:
```
{
  "message": "Required parameter \"office\" not found.",
  "status": 422
}
```

-  This would primarily benefit public API users by providing better error responses

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -101,6 +101,8 @@ app.config['SQLALCHEMY_FOLLOWERS'] = [
     for follower in env.get_credential('SQLA_FOLLOWERS', '').split(',')
     if follower.strip()
 ]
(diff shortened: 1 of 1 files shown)


fec-jli:
wow, you are so great.
I setup no debug mode on local, get same error code as on dev server.

PR #3026, opened 2018-03-23
Help debug and improve the cached calls for the API

ccostino:
This changeset attempts to resolve a couple of issues we noticed with the cached calls:

* URL paths being cached that should not be
* Downloads being cached when they are already managed and cached separately

Furthermore, it adds a bit more error handling and comments to help describe what is happening and how.


ccostino:
Thank you @pkfec and @vrajmohan for the help with this and fixing the test, I appreciate it! :-)

ccostino:
There is one remaining issue with this:  in the event of an error being thrown the cached response is retrieved from S3 if it can be found, but if that is successful then it also attempts to cache the cached response again:

```
INFO:rest.py:An API error occurred with the status code of 500.
INFO:rest.py:Attempting to retrieving the cached request from S3...
INFO:rest.py:Successfully retrieved cached request from S3.
INFO:rest.py:Attempting to cache request at: <URL here>
```

ccostino:
It turns out this is a non-issue; while it's sub-optimal, we're also going to be expiring/deleting the cache on a regular basis and as soon as there's another successful response the stale/old one will be overwritten.

Thank you, @LindsayYoung, for talking through this last bit with me!

Issue #3041, opened 2018-04-02
Incorrect total refunds on `/candidate/{candidate_id}/totals` endpoint


lbeaufort:
Thanks for finding this @jwchumley!

The "contribution_refunds" value in the `/candidate/{candidate_id}/totals` endpoint is wrong. 

Example on front end: https://www.fec.gov/data/candidate/H2MD05155/?cycle=2016
Example API call: https://api.open.fec.gov/v1/candidate/H2MD05155/totals/?page=1&cycle=2016&sort=-cycle&api_key=DEMO_KEY&per_page=20

`refunded_individual_contributions: 3990`
`refunded_political_party_committee_contributions: 0`
`refunded_other_political_committee_contributions: 2500`
[total] `contribution_refunds` should be `6490`, but it's 
[total] `contribution_refunds: 2500`

lbeaufort:
PR is merged: https://github.com/fecgov/openFEC/pull/3042

PR #3042, opened 2018-04-02
Fix contribution_refunds total

lbeaufort:
## Summary (required)

- Addresses https://github.com/fecgov/openFEC/issues/3041

_Changes materialized view `ofec_totals_candidate_committees_mv` to display the proper contribution refunds total._

[Original calculation:](https://github.com/fecgov/openFEC/blob/develop/data/migrations/V0026__more_public_schema_objects.sql#L5307) `sum((p.pol_pty_cmte_contb + p.oth_cmte_ref)) AS contribution_refunds`
[New calculation: ](https://github.com/fecgov/openFEC/pull/3042/files#diff-f16707a1e2de3aafc7a39eee794c2e97R149) ` sum(p.ttl_contb_ref) AS contribution_refunds`

- Test the migration on your local db using Flyway (`flyway migrate`)
- Check out this branch and change the model (`webservices/common/models/totals.py`) to point to `ofec_totals_candidate_committees_mv_tmp` on line 136 and 162 (I created this temp view on the dev DB until the SQL migration is run on `dev`)
- Set `SQLA_CONN` to the `dev` db and run this branch locally
- Look at  http://localhost:5000/v1/candidate/H2MD05155/totals/?page=1&cycle=2016&sort=-cycle&per_page=20. Per the issue https://github.com/fecgov/openFEC/issues/3041 the `contribution_refunds` should be 6490

-  Candidate pages: https://www.fec.gov/data/candidate/H2MD05155/?cycle=2016
- /candidate/{candidate_id}/totals endpoint: https://api.open.fec.gov/v1/candidate/H2MD05155/totals/?page=1&cycle=2016&sort=-cycle&api_key=DEMO_KEY&per_page=20

### DIFF
--- a/data/migrations/V0071__update_ofec_totals_candidate_committees_mv.sql
+++ b/data/migrations/V0071__update_ofec_totals_candidate_committees_mv.sql
@@ -336,8 +336,7 @@ UNION ALL
     election_totals_with_ending_aggregates.last_report_year,
     election_totals_with_ending_aggregates.last_net_operating_expenditures,
     election_totals_with_ending_aggregates.last_net_contributions
-   FROM election_totals_with_ending_aggregates
-  WITH NO DATA;
+   FROM election_totals_with_ending_aggregates;
 
 ----Permissions------
 
--- a/data/migrations/V0071__update_ofec_totals_candidate_committees_mv.sql
+++ b/data/migrations/V0071__update_ofec_totals_candidate_committees_mv.sql
@@ -1,0 +1,371 @@
+--
+-- Name: ofec_totals_candidate_committees_mv; Type: MATERIALIZED VIEW; Schema: public; Owner: fec
+--
+
+SET search_path = public, pg_catalog;
+
+DROP MATERIALIZED VIEW IF EXISTS ofec_totals_candidate_committees_mv_tmp;
+
+CREATE MATERIALIZED VIEW ofec_totals_candidate_committees_mv_tmp AS
+ WITH last_cycle AS (
+         SELECT DISTINCT ON (v_sum.cmte_id, link.fec_election_yr) v_sum.cmte_id,
+            v_sum.rpt_yr AS report_year,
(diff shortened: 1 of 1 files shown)


fec-jli:
with DB team help, we found that some candidates which contribution_refunds=null, but refunded_individual_contributions or
refunded_political_party_committee_contributions or
refunded_other_political_committee_contributions
has value.

so question:  Can we use contribution_refunds column as contribution total refund directly? 
or use formula to add all three above ?

ex:
select * from public.ofec_totals_candidate_committees_mv_tmp 
where contribution_refunds is NULL
and candidate_id='S2AZ00299'
and cycle=2012

@PaulClark2 @jwchumley

lbeaufort:
@fec-jli thanks for taking a look! It looks like that particular example reflects what was filed: http://docquery.fec.gov/pdf/852/12021103852/12021103852.pdf

It makes the most sense to me to pull directly from the data, rather than calculating a field.

fec-jli:
Thank for your link: [http://docquery.fec.gov/pdf/852/12021103852/12021103852.pdf](http://docquery.fec.gov/pdf/852/12021103852/12021103852.pdf)
you can see in 20d "total contribution refunds" field is empty, so this field is optional for user.
in 20c has value: $215. 

if you use 20d directly, will return null. is that correct?

lbeaufort:
@fec-jli totals aren't required, but it does look like we're using the `ttl_...` fields for other subtotals, which is reflective of the underlying data.

fec-jli:
I am not sure how our data to be loaded into tables. 
but for this candidate, the total-refund will return $0, because our api return $0 too. which is not correct? should return $215?right?

PaulClark2:
@fec-jli @lbeaufort we should use what's in the underlying data. We should not calculate a line item total if the committee can report it. So, for the example above we'd display "None" for total contribution refunds because that's what the committee reported.  

image

fec-jli:
got it. just report what ever committee reported. Thanks Paul.

Issue #2803, opened 2017-12-11
Fix statute mapping where 2 USC contains a letter


lbeaufort:
Mapping is incorrect for 2 USC to 52 USC where 2 USC contained a letter. For example, "2 USC 441b" is being parsed as "2 USC 441" which is causing lookup mistakes. 1980-01 is an example.

- [x] Fix bug where REGEX isn't accounting for characters (IE "2 USC 441b" is being parsed as "2 USC 441") which is causing lookup mistakes. 1980-01 is an example:
>>**Note** Changed statute section from `int` to `str` as a result 
- [x] Fix elasticsearch: `./manage.py load_advisory_opinions -f '1980-01'` (needed to reinitialize legal docs because we made a schema change) `./manage.py initialize_current_legal_docs`
- [ ] Add tests for handling these characters properly in `reclassify_archived_mur_statutory_citation`
**Steps**
Write for the current wrong behavior (test 1)
Pass on develop
Fix for correct behavior (test 2)
Fail on develop
git stash changes, checkout branch
Write for the wrong behavior (test 1)
Fail on branch
Write for correct behavior (test 2)
Pass on branch
Check in

Test 1 (incorrect)
437f == 30105
441b == 441

Test 2 (correct)
437f == 52 U.S.C. §30108 
441b == 52 U.S.C. §30118
441a-1== 30117

![image](https://user-images.githubusercontent.com/31420082/33460746-973f5fb2-d5fd-11e7-84fd-4e7d5118961d.png)

add to python file `logger.setLevel(logging.DEBUG)`
logging.debug
python debugger (ipdb, stack tracing)
![image](https://user-images.githubusercontent.com/31420082/33460691-52711934-d5fd-11e7-9b69-a618ef1c2967.png)
- [ ] figure out how to restore archived murs from a backup without reloading them from scratch (thanks @vrajmohan)
- [ ] re-initialize and re-load legal docs on production: `./manage.py refresh_legal_docs_zero_downtime`

PR #3044, opened 2018-04-04
Fix AO statute mapping where section contains a letter

lbeaufort:
## Summary (required)

- Addresses https://github.com/fecgov/openFEC/issues/2803

_Fix bug where REGEX isn't accounting for characters (IE "2 USC 441b" is being parsed as "2 USC 441") which is causing lookup mistakes. 1980-01 is an example. Add tests for handling these characters properly_

### Before
~2 U.S.C. §437f == 52 U.S.C. §30105~
~2 U.S.C. §441b == 2 U.S.C. §441~

### After
2 U.S.C. §437f == 52 U.S.C. §30108
2 U.S.C. §441b == 52 U.S.C. §30118
2 U.S.C. §441a-1 == 52 U.S.C. §30117

- Get Elasticsearch up and running
- get the latest S3 credentials from `dev`
- run `./manage.py initialize_current_legal_docs` which recreates the docs index and loads current AO's. Quit the process when it gets to MURs - not needed.
- Check http://localhost:5000/v1/legal/search/?type=advisory_opinions&ao_no=1980-01&ao_category=F and the statutory citations should be 52 30108 and 52 30118 (vs. [dev](https://fec-dev-api.app.cloud.gov/v1/legal/search/?type=advisory_opinions&ao_no=1980-01&ao_category=F) which still shows 2 441)

### DIFF
--- a/tests/test_advisory_opinions.py
+++ b/tests/test_advisory_opinions.py
@@ -35,13 +35,14 @@ def test_parse_ao_citations(text, ao_nos, expected):
 
 
 @pytest.mark.parametrize("text,expected", [
-    ("2 U.S.C. 432h", set([(52, 30102)])),
-    ("52 U.S.C. 30116a", set([(52, 30116)])),
-    ("2 U.S.C. 441b, 441c, 441e", set([(2, 441)])),
-    (" 2 U.S.C. §437f", set([(52, 30105)])),
-    ("52 U.S.C. § 30101", set([(52, 30101)])),
-    (" 2 USC §437f **test with no . in USC**", set([(52, 30105)])),
-    ("52 U.S.C. § 30101 **newline needed to ensure two entries**,\n 52 USC. § 30101 other words", set([(52, 30101)])),
+    ("2 U.S.C. 432", set([(52, '30102')])),
+    ("52 U.S.C. 30116(a", set([(52, '30116')])),
+    ("2 U.S.C. 441b, 441c, 441e", set([(52, '30118')])),
+    ("2 U.S.C. 441a-1", set([(52, '30117')])),
+    (" 2 U.S.C. §437f", set([(52, '30108')])),
+    ("52 U.S.C. § 30101", set([(52, '30101')])),
+    (" 2 USC §437f **test with no . in USC**", set([(52, '30108')])),
+    ("52 U.S.C. § 30101 **newline needed to ensure two entries**,\n 52 USC. § 30101 other words", set([(52, '30101')])),
 ])
 def test_parse_statutory_citations(text, expected):
     assert parse_statutory_citations(text) == expected
@@ -254,7 +255,7 @@ class TestLoadAdvisoryOpinions(BaseTestCase):
 
         actual_ao = next(get_advisory_opinions(None))
 
-        assert actual_ao["statutory_citations"] == [{'title': 52, 'section': 30101}]
+        assert actual_ao["statutory_citations"] == [{'title': 52, 'section': '30101'}]
 
     @patch("webservices.legal_docs.advisory_opinions.get_bucket")
     @patch("webservices.legal_docs.advisory_opinions.get_elasticsearch_connection")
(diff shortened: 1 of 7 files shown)


webservices/legal_docs/advisory_opinions.py
```diff
@@ -60,7 +67,7 @@
 """
 
 STATUTE_CITATION_REGEX = re.compile(
-    r"(?P<title>\d+)\s+U\.?S\.?C\.?\s+§*\s*(?P<section>\d+).*\.?")
+    r"(?P<title>\d+)\s+U\.?S\.?C\.?\s+§*\s*(?P<section>\d+\d+[-0-9a-z]*).*\.?")
```
vrajmohan: This has the undesired consequence of requiring that section numbers be at least _2_ digits followed by 0 or more digit-or-lowercase-alphabet-or-dash.

e.g. try parsing `52 U.S.C. 3`.

tests/test_advisory_opinions.py
```diff
@@ -35,13 +35,14 @@ def test_parse_ao_citations(text, ao_nos, expected):
...
 @pytest.mark.parametrize("text,expected", [
-    ("2 U.S.C. 432h", set([(52, 30102)])),
-    ("52 U.S.C. 30116a", set([(52, 30116)])),
-    ("2 U.S.C. 441b, 441c, 441e", set([(2, 441)])),
-    (" 2 U.S.C. §437f", set([(52, 30105)])),
-    ("52 U.S.C. § 30101", set([(52, 30101)])),
-    (" 2 USC §437f **test with no . in USC**", set([(52, 30105)])),
-    ("52 U.S.C. § 30101 **newline needed to ensure two entries**,\n 52 USC. § 30101 other words", set([(52, 30101)])),
+    ("2 U.S.C. 432", set([(52, '30102')])),
+    ("52 U.S.C. 30116(a", set([(52, '30116')])),
```
vrajmohan: Now that we allow alphabets and hyphens in the section, we should add a test like:
```
("52 U.S.C. 30116a", set([(52, '30116a')])),
```

lbeaufort: Thank you! I had added a test for 2 U.S.C. §441a-1 == 52 U.S.C. §30117 but I somehow lost it when re-working the (very stale) branch. I'll add it back.

lbeaufort: Oh nevermind, I see it now. I'll add a "52 U.S.C. 30116a === 52 U.S.C. 30116a" test too.

webservices/legal_docs/advisory_opinions.py
lbeaufort: Thank you, I've actually spent some time trying to understand REGEX better and I have a long way to go. It looks like the 2nd `\d+` is the problem, so it should be `(?P<title>\d+)\s+U\.?S\.?C\.?\s+§*\s*(?P<section>\d+[-0-9a-z]*).*\.?`. Do you agree?

vrajmohan: `(?P<section>\d+[-0-9a-z]*)` matches "one or more digits that may _optionally_ be followed by 0 or more digit-or-lowercase-alphabet-or-dash and would also match "123a-4".

`(?P<section>\d+[-a-z]*)` matches "one or more digits that may _optionally_ be followed by 0 or more lowercase-alphabet-or-dash and would match "123a" and "123a-b" but not "123a-4".

Not sure which one is better, but my hunch is the latter.

lbeaufort: Thanks @vrajmohan! I think we want to catch  2 USC 441a-1 so I'll test the first option.  http://uscode.house.gov/editorialreclassification/t52/Reclassifications_Title_52.html

lbeaufort:
@vrajmohan and @pkfec I've addressed Vraj's earlier comments and this is ready for your review. Thanks!

PR #3048, opened 2018-04-06
update ofec_sched_e refresh script to specify data type for column sc…

fecjjeng:
…hedule_type

## Summary (required)

- Addresses # 3046

This problem is caused by one of the column in table ofec_sched_e has data type 'unknown', so when two cmte_ids are input and SQLAlchemy try to do the union, it throws an error and can not proceed.   The script that re-create the table ofec_sched_e had been updated to specify the column type.

- Script data/refresh/rebuild_schedule_e.sql can be run locally in the local database (the local database need to be up-to-date to have all the depending views).  After this sql script had been run, 
The API schedules/schedule_e should not throw error.

-  schedules/schedule_e

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()


data/refresh/rebuild_schedule_e.sql
```diff
@@ -1,31 +1,437 @@
...
+  image_num varchar(18),
+  file_num numeric(7,0),
+  link_id numeric(19,0),
+  orig_sub_id numeric(19,0),
+  sub_id numeric(19,0) NOT NULL,
+  filing_form varchar(8),
+  rpt_tp varchar(3),
+  rpt_yr numeric(4,0),
+  election_cycle numeric,
+  "timestamp" timestamp without time zone,
```
vrajmohan: Can we get rid of this column? We only seem to be storing `NULL` in it.

data/refresh/rebuild_schedule_e.sql
```diff
@@ -1,31 +1,437 @@
...
+ cycle,
+ cast(null as timestamp) as timestamp,
+ image_pdf_url(image_num) as pdf_url,
+ coalesce(rpt_tp, '') in ('24', '48') as is_notice,
+ to_tsvector(pye_nm) as payee_name_text,
+ now() as pg_date
 from fec_sched_e_notice_vw;
 
-update ofec_sched_e_tmp
-set exp_dt = coalesce(exp_dt, dissem_dt);
```
vrajmohan: Is this not needed any more? I could not see it being handled in the `INSERT`.

vrajmohan: I see now that the `UPDATE` was called in 2 places. Can we take care of it in the `INSERT` itself?

fecjjeng: Yes, in the original script the update had been called twice so I remove one of them.  
I agree, it is better handle it in the insert statement, especially F57 does not have dissem_dt and hence not need to calculate.
I will update the script accordingly.

data/refresh/rebuild_schedule_e.sql
fecjjeng: The timestamp column was added by 18F, I assume probably was planned to be used by the queue process.  My understanding is that now sched_e is not using the queue process so it is always null.  Is this correct?  If that is the case, I will drop this column.  Please let me know.

I had also checked the itemized.py in models, it seems scheduleE model does not refer to this column.

vrajmohan: Unfortunately it is also a poor name for a column, making it difficult to search in code.

fecjjeng: I agree.  Also usually I will try not to name a column with a reserved word...

Alternately, I can skip this column in the insert statement for now.  At least it will save us from doing the cast(null as timestamp) conversion in vain.  Thoughts?

data/refresh/rebuild_schedule_e.sql
fecjjeng: This calculation had been moved to insert statement from sched_e.  
Timestamp column in ofec_sched_e had been dropped.
Queue process related objects (tables, triggers, and functions) for ofec_sched_e had been cleaned out from database since ofec_sched_e is not populated by queue process.

PR #3068, opened 2018-04-17
Update Swagger UI to the latest release

ccostino:
**PLEASE DO NOT MERGE** until this testing is complete: https://github.com/fecgov/openfec/issues/3170

This changeset finishes our migration away from the old custom forked version of Swagger UI that we had been using before and puts the newest version available in its place out of the box.  There are a few things that we need to adjust and we may want to incorporate some of the customizations we had before, so this will be a work in progress for a bit of time.

Furthermore, I was not sure how best to manage the static assets.  For now I have just included the whole `static` directory as is, but perhaps we just want to keep the directory itself and not its contents, and have Swagger UI built with every deploy.  This would mean that a developer would have to do this locally too as a part of project setup, though.  I wanted to gather input from the team first before settling on a solution to this.

The high-level changes include the following:

* Removed existing `./static/` and `./swagger-ui/` directories
* Removed `gulpfile.js`
* Removed all dependencies from `package.json` and replaced them with just `swagger-ui-dist`
* Rewrote all of the `npm` build scripts to account for the `swagger-ui-dist` package
* Adjusted the API error handling slightly to log the actual error happening
* Replaced most of the HTML in the `./webservices/templates/swagger-ui.html` with the HTML from the new version of Swagger UI
* Updated `flask-apispec` to the latest release


apburnes:
The repo is now setup for possible future customization using the swagger-ui with react. `npm run build` compiles the swagger-ui source and styles consumed by the `/developers` page.

ccostino:
Or, should we work the customizations into this PR if they'll be relatively straightforward to put in place?  My main concern there is the PR getting too unwieldy and staying open for too long. :-)

apburnes:
@patphongs should we merge this directly into a `release` branch to let users test in staging or to `develop`?

ccostino:
Tying conversations together from Slack and stand-ups:  we're going to keep this targeting `develop`, and merge it in after the next release branch is cut this morning. :-)

pkfec:
@apburnes 
I tried to test the swagger updated and here is what i found:

1. Very first thing i noticed that the new swagger  openFEC API documentation requires API KEY/VALUE. Usually this is not the case for LOCAL or DEV environments. API_KEY is only required in PROD env.

- Please find the screenshot below:
  screen shot 2018-05-29 at 3 10 09 pm


2.  After i entered the default DEMO_KEY value and click on **Authorize** on above screen and choose an endpoint with  **required** fields, API did return some results.

- Please find the screenshot below:
  screen shot 2018-05-29 at 10 28 56 pm

  However, in the above mentioned endpoint, i change the _**two_year_transaction_period**_ 
 required parameter value to 2016, and search for results.  API did not return any results and _**two_year_transaction_period**_ parameter value is defaulted to 2018 , which did not return any results either.

3.  I tried _**/rad-analyst/**_ endpoint (without any **required**  fields), the API returned some results. However, i tried to search rad analyst by **_name_** "John" , no results were returned.

- Please find the screenshot below:
  screen shot 2018-05-29 at 10 57 30 pm


cc @patphongs  @rjayasekera

pkfec:
@apburnes 

I  have tested your latest fix by querying few endpoints on my local swagger API. API seem to return results. Below are the endpoints that i tested. Swagger is throwing some weird error on audit-case endpoint, i have attached the screen shots for your reference below. Please take a look:

**filer resource**
-  /state-election-office/
- /rad-analyst/

**disbursement**
-  /schedules/schedule_b/

**audit** 
when i clicked on /audit-case/  endpoint swagger through some error
screen shot 2018-05-30 at 11 22 35 pm

cc @jwchumley

apburnes:
@pkfec when trying out th `audit-case/` route with the production api spec using the [swagger demo site](http://petstore.swagger.io/), the console is throwing the same error but the results look as expected. This is being thrown by a component inside the `swagger-ui` package and we should probably discuss more as a group.

fec-jli:
@apburnes  I have fixed /audit-case/ endpoint  type error. in branch feature/fix_audit_type_error.
Can you review that? if it is ok Can you merge into feature/update-swaggerui branch? Thanks

apburnes:
Closed #3171

Issue #3045, opened 2018-04-04
Switching API to the fecp-driven sched_b tables


fecjjeng:
This is to deploy the work completed in derived from #2854 

We are switching API to access fecp-driven sched_b tables.

make API code changes to use fecp-driven sched_b tables
modify view definition scripts that utilize these tables
evaluate and create appropriate indexes on the fecp-driven sched_b tables

fecjjeng:
fec_fitem_sched_b tables prior to election cycle 2002 do not have as many data and also are less queried.  Howefer, as suggested by Carlo, for consistency, and also for possible future feature of multiple-cycle queries, indexes still added for election cycle 1976 to 2002

patphongs:
The dependency https://github.com/fecgov/openfec/issues/3098 will be worked on in sprint 5.6. @ccostino @fecjjeng has confirmed that we will be leaving the tables and related processes in place for now as a safety net and still deploying these changes.

PR #3086, opened 2018-04-25
Feature/switch schedule b

fecjjeng:
## Summary (required)
Issus #3045 switch to use fecp-driven sched_b tables

After the switch, the schedule_b queries will be pointed to fecp-driven fec_fitem_sched_b tables.  When run the application locally, try api for schedule_b 
http://localhost:5000/developers/#!/disbursements/get_schedules_schedule_b
Make sure the API is still working. 

After the switch, the schedule_b queries will be pointed to fecp-driven fec_fitem_sched_b tables.  When run the application locally, try api for schedule_b.  The will impact schedule_b related API.  However, the changes should be transparent to the users.
-  

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

### DIFF
--- a/data/migrations/V0078__add_indexes_on_fec_fitem_sched_b_tables.sql
+++ b/data/migrations/V0078__add_indexes_on_fec_fitem_sched_b_tables.sql
@@ -1,0 +1,1381 @@
+-- Issus #3045 is to switch to use fecp-driven sched_b tables
+-- creating all these indexes during release time will take LONG time and also cause performance problems .
+-- Since tables are not being used before switch, indexes had been built during non-peak hour.  
+-- However, migration file is still submitted to keep the track of the integrity of the migration process.
+-- suppress a particular error reporting (error: 42P07 creating indexes if they already exist without error out)
+-- Using create index if not exists will also generate a LOT of warning message.
+
+-- fec_fitem_sched_b tables prior to election cycle 2002 do not have as many data and also are less queried.  No need for indexes except primary key
+-- but for consistency, and also for possible future feature of multiple-cycle queries, indexes still added for election cycle 1976 to 2002
+-- -------------------------------
+-- -------------------------------
+--   fec_fitem_sched_b_1975_1976
+-- -------------------------------
+-- -------------------------------
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_clean_recipient_cmte_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (clean_recipient_cmte_id COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_recipient_st ON disclosure.fec_fitem_sched_b_1975_1976  USING btree (recipient_st COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_recipient_city ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (recipient_city COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_cmte_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (cmte_id COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_disb_amt ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (disb_amt);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_disbursement_description_text ON disclosure.fec_fitem_sched_b_1975_1976 USING gin (disbursement_description_text);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_image_num ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (image_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_recipient_name_tex ON disclosure.fec_fitem_sched_b_1975_1976 USING gin (recipient_name_text);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_rpt_yr ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (rpt_yr);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_line_num ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (line_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_disb_dt ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (disb_dt);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_coalesce_disb_dt ON disclosure.fec_fitem_sched_b_1975_1976 USING btree ((COALESCE(disb_dt, '9999-12-31'::date::timestamp without time zone)));
+ANALYZE disclosure.fec_fitem_sched_b_1975_1976;
+-- -------------------------------
+-- -------------------------------
+--   fec_fitem_sched_b_1977_1978
+-- -------------------------------
+-- -------------------------------
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_clean_recipient_cmte_id ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (clean_recipient_cmte_id COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_recipient_st ON disclosure.fec_fitem_sched_b_1977_1978  USING btree (recipient_st COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_recipient_city ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (recipient_city COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_cmte_id ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (cmte_id COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_disb_amt ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (disb_amt);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_disbursement_description_text ON disclosure.fec_fitem_sched_b_1977_1978 USING gin (disbursement_description_text);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_image_num ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (image_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_recipient_name_tex ON disclosure.fec_fitem_sched_b_1977_1978 USING gin (recipient_name_text);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_rpt_yr ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (rpt_yr);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_line_num ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (line_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_disb_dt ON disclosure.fec_fitem_sched_b_1977_1978 USING btree (disb_dt);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1977_1978_coalesce_disb_dt ON disclosure.fec_fitem_sched_b_1977_1978 USING btree ((COALESCE(disb_dt, '9999-12-31'::date::timestamp without time zone)));
+ANALYZE disclosure.fec_fitem_sched_b_1977_1978;
+-- -------------------------------
+-- -------------------------------
+--   fec_fitem_sched_b_1979_1980
+-- -------------------------------
+-- -------------------------------
+CREATE INDEX IF NOT EXISTS idx_sched_b_1979_1980_clean_recipient_cmte_id ON disclosure.fec_fitem_sched_b_1979_1980 USING btree (clean_recipient_cmte_id COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_1979_1980_recipient_st ON disclosure.fec_fitem_sched_b_1979_1980  USING btree (recipient_st COLLATE pg_catalog."default");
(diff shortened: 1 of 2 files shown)


data/migrations/V0078__add_indexes_on_fec_fitem_sched_b_tables.sql
```diff
@@ -0,0 +1,1381 @@
...
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_image_num ON disclosure.fec_fitem_sched_b_2001_2002 USING btree (image_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_recipient_name_text ON disclosure.fec_fitem_sched_b_2001_2002 USING gin (recipient_name_text);
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_rpt_yr ON disclosure.fec_fitem_sched_b_2001_2002 USING btree (rpt_yr);
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_line_num ON disclosure.fec_fitem_sched_b_2001_2002 USING btree (line_num COLLATE pg_catalog."default");
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_disb_dt ON disclosure.fec_fitem_sched_b_2001_2002 USING btree (disb_dt);
+CREATE INDEX IF NOT EXISTS idx_sched_b_2001_2002_coalesce_disb_dt ON disclosure.fec_fitem_sched_b_2001_2002 USING btree ((COALESCE(disb_dt, '9999-12-31'::date::timestamp without time zone)));
+ANALYZE disclosure.fec_fitem_sched_b_2001_2002;
+
+
+
```
ccostino: I'm curious about the change in style of how the indexes are created once you hit 2003; was there a reason for not just using the simpler `CREATE INDEX` statements as above?

Another thing about the index creation - these don't quite mirror what we have on the current `ofec_sched_b_xxxxx` tables, where the indexes are multicolumn to efficiently support the paging, filtering, and sorting we have on the site.  I know the PostgreSQL documentation says that PostgreSQL is usually able to combine indexes effectively on its own, but our research into this has shown that's not been the case with the site and we've needed to create the indexes with each combination of columns.  I'm happy to review this further with you if it'd be helpful to walk through and demonstrate what this means if anything is unclear!

I call this out because I'm concerned that we'll lose some of the performance gains we've made, especially if we're still looking to enable multi-cycle filtering/searching in the future.

fecjjeng: The change in style of how the indexes are created after year 2003 (Also stated in the migration file itself) is as followed:
Creating all these indexes during release time will take LONG time and also cause performance problems, especially for the bigger partition tables.  Since tables are not being used before switch, indexes had been built during non-peak hour.  However, migration file is still submitted to keep the track of the integrity of the migration process.  Suppress a particular error reporting (error: 42P07 creating indexes if they already exist without error out).   Using create index if not exists will also generate a LOT of warning message.  Indexes for the smaller tables will be created normally using the migration file itself.

fecjjeng: For the sample DB, this is another issue Joseph Kuan is currently working on.  He is planning on using a local postgresql database, which we had kept all the base tables data current to server as the sample DB instead of using the sample DB script to load small set of sample data.

fecjjeng: Helen and Jean had conducted analysis on the multiple column indexes vs combined single column indexes queries.  Postgresql DB had combine indexes effectively on its own with acceptable performance.  So we decided to use the single column indexes and let the DB engine handle the combination.  However, I did not consider the pagination/sorting enough during the analysis.  I am in the process of re-evaluating the indexes now.

ccostino: D'oh, I completely missed the note at the top of the migration file regarding the index creation, I am sorry about that, @fecjjeng!  Thanks for bringing it to my attention. :-)

I'm curious to hear more about what's happening with the sample DB, but if that work is happening outside of this we can follow up where it's more appropriate instead of focusing on it here.

If you need any help with the index analysis please let me know!

ccostino: @fecjjeng yes, the `COALESCE`-based indexes are definitely needed against the `disb_dt`, at least as the API endpoint is structured right now.  I can see not wanting to include the older style, which I think is okay for now, but please be mindful if the endpoint changes in the future.  In other words, if the query construction changes and `disb_dt` ends up being used directly again, the regular style indexes will be needed again.

For the third one, this is because we allow sorting by disbursement amount (labeled as `disb_amt`) in addition to the disbursement date, so we need a set of indexes that make use of it as well.  `last_disbursement_amount` is in reference to an [additional argument and field](https://github.com/fecgov/openFEC/blob/develop/webservices/args.py#L511) added to the API to help with the sorting and pagination.  It is used to keep track of the last amount of the previous page when paging through results sorted by disbursement amount (and there's a corresponding one for `last_disbursement_date` as well).  This is because of the way the paging is ultimately handled when the query is constructed.

Most of the paging code itself happens with the [`SeekCoalescePaginator`](https://github.com/fecgov/openFEC/blob/develop/webservices/utils.py#L83) in this case, which is accessed by the [`ItemizedResource`](https://github.com/fecgov/openFEC/blob/develop/webservices/common/views.py#L56) for resources that are specific to the itemized schedule, as in this case.

All of this is to say that yes, a set of indexes is needed with `disb_amt`. :-)

Did I answer your questions?  Please let me know if anything is still unclear!

ccostino: Also, don't forget about the documentation around how the expression based indexes are supported, that can help explain things too!  [Enabling an expression based sort](https://github.com/fecgov/openFEC/wiki/Enabling-an-expression-based-sort)

fecjjeng: Yes, I remembered we were working on the function-based indexes for disb_dt.
I realized that my first test does not show disb_amt in the query since I was calling the API in a mixed way.  Now when I select disbursement_amount in the sort field and run the next page with last_disbusement_amount, the query does include the disb_amt in the where AND order by field.  Then I noticed a new question, the following is the where/order by clause I got:
WHERE disclosure.fec_fitem_sched_b.two_year_transaction_period = %(two_year_transaction_period_1)s AND disclosure.fec_fitem_sched_b.cmte_id IN (%(cmte_id_1)s) AND (coalesce(disclosure.fec_fitem_sched_b.disb_amt, NULL), disclosure.fec_fitem_sched_b.sub_id) < (%(param_1)s, %(param_2)s) ORDER BY disclosure.fec_fitem_sched_b.disb_amt DESC, disclosure.fec_fitem_sched_b.sub_id DESC
LIMIT %(param_3)s

The where clause has coalesce for disb_amt but the order by clause does not.  I remember last time when these two clause does not match, the performance is not optimized for disb_dt.  Should we made the indexes to use the coalesce disb_amt and change the models/itemized.py to add the coalesce disb_amt to make it consistene?

ccostino: Oh good question, @fecjjeng!  We should dig into that further to see why a `COALESCE` statement is being generated for use with disbursement amount, especially just to default to `NULL` anyway, that doesn't make sense and causes the very issue we were trying to avoid with using `COALESCE` for the disbursement date.

From the app code side, we're only defining the disbursement date for use with a `COALESCE` in the `ORDER BY` clause ([as seen here](https://github.com/fecgov/openFEC/blob/develop/webservices/common/models/itemized.py#L343-L355)), so this is with the SQL construction side.  Like I said, we should investigate why because that doesn't seem right or useful, but there might be something I'm also missing/forgetting.

fecjjeng: @ccostino I agree.  I will create a ticket to investigate this issue.  For the indexes, I will create set #1 and Set #3 for now.

ccostino: That sounds great, thanks @fecjjeng! :-)

hcaofec: Thanks for all the research @fecjjeng @ccostino 
From Jean's post above, it looks like the COALESCE with disbursement_amount was created when pagination happens.  We definitely should investigate it.

fecjjeng:
Laura's V0078 had been merged.  So this one can be merged IF it pass the review.

data/migrations/V0079__add_indexes_on_fec_fitem_sched_b_tables.sql
```diff
@@ -0,0 +1,2424 @@
...
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_cmte_id_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (cmte_id COLLATE pg_catalog."default", disb_amt, sub_id);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_disb_desc_text_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING gin (disbursement_description_text, disb_amt, sub_id);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_image_num_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (image_num COLLATE pg_catalog."default", disb_amt, sub_id);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_rcpt_name_text_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING gin (recipient_name_text, disb_amt, sub_id);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_rpt_yr_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (rpt_yr, disb_amt, sub_id);
+CREATE INDEX IF NOT EXISTS idx_sched_b_1975_1976_line_num_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_1975_1976 USING btree (line_num COLLATE pg_catalog."default", disb_amt, sub_id);
```
ccostino: In checking this over and cross-referencing the existing indexes I noticed that we may be missing one more combination, that of the `COALESCE` `disb_dt` value paired first with the `disb_amt` and `sub_id` following.  Example:

```sql
CREATE INDEX IF NOT EXISTS idx_sched_b_xxxx_yyyy_colsc_disb_dt_disb_amt_sub_id ON disclosure.fec_fitem_sched_b_xxxx_yyyy USING btree ((COALESCE(disb_dt, '9999-12-31'::date::timestamp without time zone)), disb_amt, sub_id);
```

We have it going in the other direction with `disb_amt` coming first, but not this way, and I wonder if that's still contributing to some performance issues now that I think about it.

ccostino: In looking at this a bit more, we can filter and sort by both an amount and a date, so I think we do need this; I'm not sure why we didn't have it before, must've been an oversight!

fecjjeng: I had created three different indexes:
idx_sched_b_2017_2018_disb_amt_sub_id (disb_amt, sub_id);
idx_sched_b_2017_2018_colsc_disb_dt_disb_amt_sub_id ((COALESCE(disb_dt, ''9999-12-31''::date::timestamp without time zone)), disb_amt, sub_id);
idx_sched_b_2017_2018_disb_dt_amt_sub_id(disb_dt, disb_amt, sub_id);
  
I then run the following query (base on the sql from API):
explain analyze
select cmte_id, two_year_transaction_period, disb_dt
FROM disclosure.fec_fitem_sched_b 
WHERE two_year_transaction_period = 2018 
AND disb_dt >= '2018-02-01'::date
ORDER BY disb_amt DESC, sub_id DESC
LIMIT 20;

Limit  (cost=0.58..280.74 rows=20 width=40) (actual time=0.028..0.262 rows=20 loops=1)
  ->  Merge Append  (cost=0.58..48570525.81 rows=3467401 width=40) (actual time=0.027..0.251 rows=20 loops=1)
        Sort Key: fec_fitem_sched_b.disb_amt DESC, fec_fitem_sched_b.sub_id DESC
        ->  Sort  (cost=0.01..0.02 rows=1 width=94) (actual time=0.008..0.008 rows=0 loops=1)
              Sort Key: fec_fitem_sched_b.disb_amt DESC, fec_fitem_sched_b.sub_id DESC
              Sort Method: quicksort  Memory: 25kB
              ->  Seq Scan on fec_fitem_sched_b  (cost=0.00..0.00 rows=1 width=94) (actual time=0.003..0.003 rows=0 loops=1)
                    Filter: ((disb_dt >= '2018-02-01'::date) AND (two_year_transaction_period = '2018'::numeric))
        ->  Index Scan Backward using idx_sched_b_2017_2018_disb_amt_sub_id on fec_fitem_sched_b_2017_2018  (cost=0.56..48527183.27 rows=3467400 width=40) (actual time=0.017..0.235 rows=20 loops=1)
              Filter: ((disb_dt >= '2018-02-01'::date) AND (two_year_transaction_period = '2018'::numeric))
              Rows Removed by Filter: 160
Planning time: 2.922 ms
Execution time: 0.301 ms

idx_sched_b_2017_2018_disb_amt_sub_id (disb_amt, sub_id) is the index the DB optimizer used.  Based on this result, I will create this index to all the tables and re-submit the migration file again.

fecjjeng: updated migration to add extra indexes had been pushed to the branch.

ccostino: Wait, are we sure that's the SQL from the API?  Sorry, now I have one more question... 😆 I promise this will be the last one though. :-)

I want to verify the query being generated by the API to ensure we're basing the index creation on the right information.  The reason why I ask this is because the test queries we've generated in the past _do_ have `COALESCE` statements in them, plus a host of other things. Your investigation into what the query planner is great and definitely useful, I just want to make sure we're basing the findings on the right information, and I'm sorry I didn't provide a clearer example of what I was going for.  We can work through this tomorrow though and get it ironed out quick to wrap this up. :-)

Thanks so much for working through this with me, @fecjjeng!

fecjjeng:
This is the query output from API (inputs are min_date: 2018-02-01, two_year_transaction_period: 2018, sort_by_field: -disbursement_amt):
WHERE disclosure.fec_fitem_sched_b.two_year_transaction_period = %(two_year_transaction_period_1)s AND disclosure.fec_fitem_sched_b.disb_dt >= %(disb_dt_1)s ORDER BY disclosure.fec_fitem_sched_b.disb_amt DESC, disclosure.fec_fitem_sched_b.sub_id DESC
LIMIT %(param_1)s
2018-05-03 20:55:11,912 INFO sqlalchemy.engine.base.Engine {'rpt_yr_1': 2, 'rpt_yr_2': 2, 'two_year_transaction_period_1': 2018, 'disb_dt_1': datetime.date(2018, 2, 1), 'param_1': 20}

ccostino:
Okay, after a bit more investigation, we've verified we're looking at the same queries and seeing the correct performance plans!  Thanks, @hcaofec and @fecjjeng!

PR #3084, opened 2018-04-25
Feature/upgrade elasticsearch

fec-jli:
upgrade elasticsearch 24 —> 56

PR #3106, opened 2018-05-07
Revert ES56 Services 

pkfec:
Elastic Search Service was upgraded from 24 to 56 in this PR https://github.com/fecgov/openFEC/pull/3084

While testing this new ES56 service on **dev**, we encountered Gateway and ReadTime outs.
We are still following up with cloud.gov team to resolve the timeout issues.

Until then we are rolling back the ES56 services and will be using the existing ES24 service on cloud.


lbeaufort:
@pkfec can you please update the PR description to explain why we're backing out the changes?

Issue #3103, opened 2018-05-04
Restore Paper filings missing from API /reports/ and /filings/


AmyKort:
At least 591 paper filings available from the bulk downloads seem to be missing from the API's /reports/ endpoint: when we fetch a /reports/ endpoint for a committee, the response does not include any records with these filing ids in the file_number field. 366 of those filings are accessible through the /filings/ endpoint. It seems that the filings which are available through /filings/ but not /reports/ show up on the website, but the summary fields for those are shown as $0 regardless of the filing's actual content.

The filing ids for the 366 filings available through /filings/ but not /reports/ are in the document below:
missing_filings.txt

To resolve this issue, we need to:
- [x] create new view that includes F3X plus F3 submitted by PAC and party committee
- [ ] API work to use the new view
- [ ] test

@hcaofec @pkfec @jwchumley 

Continues work at https://github.com/fecgov/openFEC/issues/3038

hcaofec:
There are two different reasons for causing this issue. 
Besides the reason identified in #3038 , another cause was identified while working on this ticket.  It's from the logic in one of the functions:
 _resolve_committee_type

This function does not have logic for rpt_yr, only for cycle.  When rpt_yr is passed to API, API will use the latest committee type ( type in first row returned order by cycle desc ), then api uses different tables based on committee type.  This logic will not work correctly when committee type changed over different cycles.  Here is an example:


FILE_NUM | CMTE_ID | FORM_TP | RPT_TP | RPT_YR | BEGIN_IMAGE_NUM 
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
1884270 | C00258525 | F3 | Q2 | 2013 | 13031090700 
897290 | C00258525 | F3 | Q3 | 2013|13031130211

 Type for C00258525:  
2014    H;  2016  U.
when rpt_yr= 2013, api sets type = U ,  uses pac_party table for F3 and will not return any report;
Instead when cycle=2014, api sets type = H ,  uses house_senate table for F3,  and 1884270 and 897290 will be returned.

So logic to handle rpt_yr will be added to fix this.

@jwchumley @AmyKort

PR #3139, opened 2018-05-16
create new materialized view to include form_3 submitted by pac and party

hcaofec:
## Summary 

Addresses [#3103](https://github.com/fecgov/openFEC/issues/3103) 

-V0081 was added to data/migration.  

 - If local database has enough sample data, the materialized view should have this row:
FILE_NUM | CMTE_ID | FORM_TP | RPT_TP | RPT_YR | BEGIN_IMAGE_NUM | 
1206344 | C00156554 | F3 | YE | 2017 | 201801240300187403

- otherwise, run a select statement against the query part only


vrajmohan:
Is this going to replace `public.ofec_reports_pacs_parties_mv`? If not, can it be defined using `public.ofec_reports_pacs_parties_mv`?

vrajmohan:
Also, if it is not going to replace `ofec_reports_pacs_parties_mv`, can it be a regular view defined using `ofec_reports_pacs_parties_mv`?

hcaofec:
Eventually, the api endpoint that is using ofec_reports_pacs_parties_mv now will switch to this new materialized view.   If ofec_reports_pacs_parties_mv is not used by any other api, it will be dropped.

The logic from ofec_reports_pacs_parties_mv was copied here instead of using the view itself is to reduce the dependency between materialized views.

hcaofec:
I guess this view will replace public.ofec_reports_pacs_parties_mv

hcaofec:
Thanks.  I will modify the script

Issue #3014, opened 2018-03-19
Add swagger test for when "cached calls" is turned on


lbeaufort:
Something like: 
```
    def test_swagger_with_cache(self):

        os.environ['CACHE_ALL_REQUESTS'] = True
        try:
            utils.validate_swagger(spec)
        except exceptions.SwaggerError as error:
            self.fail(str(error))
```
Note: This is passing when we expect it to fail - need to figure out why.

AmyKort:
ice boxing until Swagger tests are complete.

pkfec:
**Issue before**:
To enable cache calls on local, we have to the set the **export CACHE_ALL_REQUEST=True**. By default the variable value is set to False. In the past, we were caching all api endpoints besides non api url's **(ex: http://127.0.0.1:5000/)** which led the swagger (OpenFEC API Documentation) failures.

Now that we have refactored the **Cache Calls** logic, which saves only the api endpoints, the swagger loads fine now.

ccostino:
@pkfec, you're right that we refactored the cached calls logic to only cache the API endpoints themselves and this fixes the problem that we had with the existing test.  However, refactoring that test a bit to provide this added check of the documentation loading with cached calls being set to `True` to turn it on helps ensure that if we ever experience a regression in the code, it'll be caught.

Furthermore, we should be adding a few more tests around the cached calls themselves to test that the caching is working correctly.  What I mean by this is given a handful of URLs, some API endpoints and some not, a test or set of tests could verify that the API endpoint calls are cached and others are not, and we could test that none of the URLs are cached if the cached calls functionality is disabled.  There could be one or two other things to test as well, this is just a start.

We went ahead without adding these tests before because of how long the feature had been in development and as a team we decided not to delay things further.  The other decision that was made in response to this was that tests would be written in the future to account for the changes and new feature, and that's what this issue encompasses.

Does this make sense?  If there are any questions or things still aren't clear please ask!

pkfec:
Swagger Test:
Setting the cache-calls env variable to True or False, has no impact on Swagger API documentation.  Swagger Test is only looking if the the variable is EXISTS or NOT, but not necessary checking the value of the variable. There is no way that the swagger API is going to break, with the way it is right now.
Thats my concern. 

Additional Tests:
I am with you on writing additional tests to check if caching happens for API endpoints and non api endpoints. This should be tracked in a different issue ,i guess. But i am happy to write these test anytime.

pkfec:
[...]

PR #3140, opened 2018-05-16
Feature/add swagger test cache calls

pkfec:
## Summary (required)
Added two swagger tests to verify that the openfec api swagger documentation loads fine when cache_all_requests environment variable is set to either True Or false
- Addresses # [_Issue number_]
https://github.com/fecgov/openFEC/issues/3014

- 1. set the cache_all_requests env variable to true.
- 2. run the server
- 3. launch the api (http://localhost:5000))
- 4. set the cache_all_requests env varibale to false.
- 4. repeat steps (ii) and (iii) 

OpenFEC Swagger API Documentation
http://localhost:5000/ should load the api documentation without any errors.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -4,11 +4,8 @@ import unittest
 import sqlalchemy as sa
 import factory
 import manage
-
-from webservices.env import env
 from sqlalchemy.ext.automap import automap_base
 from sqlalchemy.orm.session import make_transient
-from faker import Faker
 from factory.alchemy import SQLAlchemyModelFactory
 from apispec import utils, exceptions
 from tests import common, factories
@@ -75,7 +72,7 @@ class TestSwagger(unittest.TestCase):
     def test_swagger_without_cache(self):
         # Test if Swagger API documentation loads
         # when cache all requests is disabled
-        os.environ['CACHE_ALL_REQUESTS'] = 'False'
+        os.environ['CACHE_ALL_REQUESTS'] is False
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
@@ -84,7 +81,7 @@ class TestSwagger(unittest.TestCase):
     def test_swagger_with_cache(self):
         # Test if Swagger API documentation loads
         # when cache all request is enabled
-        os.environ['CACHE_ALL_REQUESTS'] = 'True'
+        os.environ['CACHE_ALL_REQUESTS'] is True
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -69,19 +69,16 @@ TOTALS_MODELS = [
 
 class TestSwagger(unittest.TestCase):
 
-    def test_swagger_without_cache(self):
-        # Test if Swagger API documentation loads
-        # when cache all requests is disabled
-        os.environ['CACHE_ALL_REQUESTS'] is False
+    def test_swagger(self):
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
             self.fail(str(error))
 
     def test_swagger_with_cache(self):
-        # Test if Swagger API documentation loads
-        # when cache all request is enabled
-        os.environ['CACHE_ALL_REQUESTS'] is True
+        # Test if Swagger API documentation loads OK
+        # when cache all requests is enabled
+        os.environ['CACHE_ALL_REQUESTS'] = 'True'
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1,23 +1,21 @@
+import os
 import datetime
 import unittest
-
 import sqlalchemy as sa
+import factory
+import manage
+
+from webservices.env import env
 from sqlalchemy.ext.automap import automap_base
 from sqlalchemy.orm.session import make_transient
-
 from faker import Faker
-import factory
 from factory.alchemy import SQLAlchemyModelFactory
-
 from apispec import utils, exceptions
-
-import manage
 from tests import common, factories
 from webservices.rest import db
 from webservices.spec import spec
 from webservices.common import models
 from webservices.common.models import ScheduleA
-
 
 def make_factory():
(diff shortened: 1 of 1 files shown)


pkfec:
@ccostino @lbeaufort Please review this PR.

tests/test_integration.py
```diff
@@ -1,24 +1,22 @@
...
 from sqlalchemy.ext.automap import automap_base
 from sqlalchemy.orm.session import make_transient
-
 from faker import Faker
-import factory
 from factory.alchemy import SQLAlchemyModelFactory
-
 from apispec import utils, exceptions
-
-import manage
```
ccostino: Out of curiosity, was all of this cleanup intentional here?  If so that's fine, I just want to make sure!

tests/test_integration.py
```diff
@@ -74,12 +72,23 @@ class Meta:
 
 class TestSwagger(unittest.TestCase):
 
-    def test_swagger_valid(self):
+    def test_swagger_without_cache(self):
+        # Test if Swagger API documentation loads
+        # when cache all requests is disabled
+        os.environ['CACHE_ALL_REQUESTS'] = 'False'
```
ccostino: For these values, you don't need the quotes - we want the actual `True` and `False` value, not the string representation.

Furthermore, `'False'` will actually evaluate to `True` because it is not an empty string.  You can see more information about this in the [Python documentation](https://docs.python.org/3/library/stdtypes.html#truth-value-testing).

tests/test_integration.py
```diff
@@ -74,12 +72,23 @@ class Meta:
...
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
             self.fail(str(error))
 
+    def test_swagger_with_cache(self):
+        # Test if Swagger API documentation loads
+        # when cache all request is enabled
+        os.environ['CACHE_ALL_REQUESTS'] = 'True'
```
ccostino: Same note as above regarding the setting of this value.

tests/test_integration.py
pkfec: **os.environ['CACHE_ALL_REQUESTS'] = False**
The above expression is evaluating a string against a boolean value. I ran into a **TypeError** during testing. So i changed the  boolean value to a string. so that we are comparing two strings

Or i could have the statement like this may be:
**os.environ['CACHE_ALL_REQUESTS'] is False**

tests/test_integration.py
pkfec: Thought of cleaning up the unused imports.

pkfec:
@ccostino 
I have updated the tests little bit. Please review .

tests/test_integration.py
```diff
@@ -74,12 +69,23 @@ class Meta:
 
 class TestSwagger(unittest.TestCase):
 
-    def test_swagger_valid(self):
+    def test_swagger_without_cache(self):
+        # Test if Swagger API documentation loads
+        # when cache all requests is disabled
+        os.environ['CACHE_ALL_REQUESTS'] is False
```
ccostino: As noted in a previous comment, we need to keep the value as a string.  We also want to set the environment variable here, not perform an equality check, so the statement should read as follows:

```python
os.environ['CACHE_ALL_REQUESTS'] = 'False'
```

tests/test_integration.py
```diff
@@ -74,12 +69,23 @@ class Meta:
...
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
             self.fail(str(error))
 
+    def test_swagger_with_cache(self):
+        # Test if Swagger API documentation loads
+        # when cache all request is enabled
+        os.environ['CACHE_ALL_REQUESTS'] is True
```
ccostino: Same note as above, only for setting it to `'True'`:

```python
os.environ['CACHE_ALL_REQUESTS'] = 'True'
```

tests/test_integration.py
pkfec: I think this is what i had before. I am happy to revert the chagnes :)

https://github.com/fecgov/openFEC/pull/3140/files/d189ac0bb68faef0b253d36cf5df73bf96e347c6#diff-7eb47fc18d4a6196f0940784484a5bf3R78

pkfec: First i have to set the env variable and then  evaluate
 os.environ['CACHE_ALL_REQUESTS'] = 'False'?

ccostino: It is what you had before, and this statement is also what is setting the variable; that's all that needs to happen in the test as the code being tested handles the rest. :-)

Though, in thinking about this a bit more, this is where this may need to shift a bit; you and @lbeaufort both pointed out that this is just testing the validation of the generated API documentation itself, not actually going to the documentation URL and loading it up.

I'm sorry for the confusion, it's in part just due to lack of time to be able to focus.  In any case, these tests ought to be re-written to load the Swagger documentation itself by attempting to visit the documentation URL and seeing if a `200` response is given or something to that effect.  Given these proposed changes to the tests, we should keep the one test as we had it originally to validate the Swagger UI itself.

@lbeaufort does this make sense to you, based on what we discussed?

lbeaufort: @ccostino - that is what I was thinking too. I'm sorry I haven't spoken up sooner, I've been swamped.

pkfec:
@ccostino 

- Add the original swagger test that validates the API specs.
- Remove the swagger test when cache calls is disabled.
- Update the swagger test when cache calls is enabled.  set and evaluate the **CACHE_ALL_REQUESTS**  env variable as a string.

Please review.

tests/test_integration.py
```diff
@@ -74,12 +69,20 @@ class Meta:
...
+    def test_swagger(self):
         try:
             utils.validate_swagger(spec)
         except exceptions.SwaggerError as error:
             self.fail(str(error))
 
+    def test_swagger_with_cache(self):
+        # Test if Swagger API documentation loads OK
+        # when cache all requests is enabled
+        os.environ['CACHE_ALL_REQUESTS'] = 'True'
```
ccostino: This line still needs to be removed from here as we've identified it's not appropriate in this test; we need it in a different test instead.

lbeaufort:
Because we decided as a team to remove cached calls in #3235 - this test is no longer needed.

PR #3153, opened 2018-05-23
Feature/upgrade sqlalchemy

vrajmohan:


webservices/common/models/base.py
```diff
@@ -58,4 +58,4 @@ def create_session(self, options):
 
 class BaseModel(db.Model):
     __abstract__ = True
-    idx = db.Column(db.Integer, primary_key=True)
+    idx = db.Column(db.Integer, primary_key=True, autoincrement=True)
```
ccostino: Based on the note I saw in your commit message, we were relying on this parameter being set automatically before with the `1.0.x` version?

webservices/resources/elections.py
```diff
@@ -176,6 +176,7 @@ def _get_pairs(self, totals_model, kwargs):
     def _get_latest(self, pairs):
         latest = db.session.query(
             pairs.c.cash_on_hand_end_period,
+            pairs.c.candidate_id,
```
ccostino: I saw your note on this too, thank you for indicating why this is now needed!  Just to verify, this change doesn't actually change the query construction or performance?

webservices/utils.py
```diff
@@ -139,6 +138,7 @@ def _get_index_values(self, result):
         """Get index values from last result, to be used in seeking to the next
         page. Optionally include sort values, if any.
         """
+        from webservices.common.models import db
```
ccostino: Was this change pulled in from elsewhere or a part of this changeset explicitly?  I ask because of the logging adjustments that were looked at yesterday, too.

webservices/common/models/base.py
vrajmohan: Correct!

webservices/utils.py
ccostino: Ah I see, thank you!

webservices/resources/elections.py
ccostino: Yes, you're right!  We were relying on the `distinct` clause to add the column into the `SELECT` portion of the query automatically, as noted in the piece of the migration guide you linked to.  Thanks!

ccostino:
@patphongs, would you like to give this a look before we move ahead with merging it?

tests/test_integration.py
```diff
@@ -460,6 +462,7 @@ def test_unverified_filers_excluded_in_committees(self):
 
         self.assertEqual(committee_history_count, committee_history_verified_count)
 
+    @pytest.mark.filterwarnings("ignore:Skipped unsupported reflection")
```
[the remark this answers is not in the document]
vrajmohan: This commit would have fixed it - https://github.com/fecgov/openFEC/pull/3153/commits/671638ab33b1a590ad50eaa1a3f32d8c2c97578a. If you refresh your branch, you should see it fixed.

FYI, this issue also existed with the old SQLAlchemy.

PR #3164, opened 2018-05-25
add form_tp to report mv

hcaofec:
## Summary (required)

- Addresses #3103  [_Issue number_]

- add form_tp column to change hard coded the value of form type in model file -

- If local database has enough sample data, the materialized view should have this row:
FILE_NUM | CMTE_ID | FORM_TP | RPT_TP | RPT_YR | BEGIN_IMAGE_NUM |
1206344 | C00156554 | Form 3 | YE | 2017 | 201801240300187403

- otherwise, run a select statement against the query part only


fecjjeng:
Sample database has rows for 'Form 3',  I had pull the branch to local machine and run the following query:
select form_tp, count(*)
from public.ofec_report_pac_party_all_mv
group by form_tp;

"Form 3X";1532
"Form 3";8


Also run the query part only against DEV DB with 
where file_number = 1206344

The resulted row has form_tp = 'Form 3'

fecjjeng:
Have you add this new view to manage.py and flow.py file to include it in the daily refresh.

fecjjeng:
downloaded branch to local machine and tested.  MV refreshed with expected number of rows in cfdm_test (with sample data).

PR #3203, opened 2018-06-11
[V0093] create operations log endpoint

pkfec:
## Summary (required)

Create new endpoint for operations log and write python tests for this endpoint.

- Resolves # https://github.com/fecgov/fec-cms/issues/2008

Write up new model, resource and args to support this new endpoint.

- checkout the branch  **feature/create-operations-log-endpoint**
- export SQLA_CONN to point to **dev** database
- run the server (./manage.py runserver)
- In a browser, open the API documentation
- go to **operations-log ** endpoint which is under **/filings/** section 

- enter a **candidate_committe_id** (a required field) and search for results.

-  A new API endpoint is created.


pkfec:
@hcaofec @fec-jli @rjayasekera @lbeaufort  
Please review this PR when you get a chance.

pkfec:
@jwchumley i have moved the operations-log endpoint to **filings** section

fec-jli:
Do we need make multi search by **candidate_committee_id** and **report_year** ?

pkfec:
i am not sure, if we want to search by multiple committee id's and report years. It is unlike other endpoints where we want to find the report details for a given committee, with different transaction status. we even do not know how the front-end design is going to be and i would suggest we leave this as is now and revisit later.
cc @jwchumley

pkfec:
@jwchumley reviewed this endpoint on my local env today. He suggested to change the candidate_committee_id column to be multi column search.  And add few searchable columns to this endpoint :
- ending_image_number
- form_type
- rpt_type

pkfec:
@fec-jli @hcaofec  : Please review.

cc @jwchumley

tests/test_operation_logs.py
```diff
@@ -0,0 +1,23 @@
+
+from tests import factories
+from tests.common import ApiBaseTest
+from webservices.rest import api
+from webservices.resources.operations_log import OperationsLogView
+
+class TestOperationsLog(ApiBaseTest):
```
lbeaufort: @pkfec great work on this! My only thought is - since we're filtering for only `status_num=1` in the API, I think we should add an automated test to make sure other `status_num=1` values are not returned. Thoughts?

webservices/common/models/operations_log.py
```diff
@@ -0,0 +1,27 @@
...
+    amendment_indicator = db.Column('amndt_ind', db.String,
+            doc="Type of the report.N(new), A(amended) or T(cancel)")
+    report_type = db.Column('rpt_tp', db.String,
+            doc="Monthly, quarterly or other period covered reports")
+    summary_data_complete_date = db.Column('pass_1_entry_dt', db.DateTime,
+            doc="Date when the report is loaded in the database")
+    summary_data_verification_date = db.Column('pass_1_verified_dt', db.DateTime,
+            doc="Same day or a day after the report is loaded in the database")
+    transaction_data_complete_date = db.Column('pass_3_entry_done_dt', db.Date,
+            doc="Date when the report is processed completely")
```
lbeaufort: We're missing newlines for some of these files

webservices/resources/operations_log.py
```diff
@@ -0,0 +1,48 @@
...
+logger = logging.getLogger(__name__)
+
+
+@doc(tags=['filings'], description=docs.OPERATIONS_LOG)
+class OperationsLogView(ApiResource):
+    model = models.OperationsLog
+    schema = schemas.OperationsLogSchema
+    page_schema = schemas.OperationsLogPageSchema
+
+    filter_multi_fields = [
```
lbeaufort: Should we add indexes to the `operations_log` table to correspond with these filters?

webservices/common/models/operations_log.py
pkfec: Done

webservices/resources/operations_log.py
pkfec: This is not the final endpoint, but yes i need to create one or two indexes on this endpoint. 
The endpoint keep changing based off @jwchumley feedback that i get after testing in my local env.

tests/test_operation_logs.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertEqual(len(results), 0)
+
+    def test_search_cand_cmte_id(self):
+
+        response = self.app.get(api.url_for(OperationsLogView, candidate_committee_id='01', report_year=2012))
+        self.assertEquals(response.status_code, 200)
+
+    def test_search_cand_cmte_id_with_status_verified(self):
+
+        results = self._results(api.url_for(OperationsLogView, status_num=1))
```
lbeaufort: @pkfec thanks for adding another test! Because we're testing that `status_num=1` maybe we could `assert` that all the results are `status_num=1`  rather than there being 4 results, in case we add more test data in the future. Something like this, but I'd test it out to be sure :)
```
for result in results:
    self.assert(result.get('status_num') == 1 )
```

pkfec: results object only yields the actual count of the records found in test data.  From response object we can decode the data.  I couldnt find a way to compare the response object against the status_num=1.

webservices/resources/operations_log.py
```diff
@@ -0,0 +1,53 @@
...
+            args.make_multi_sort_args(
+                default=default_sort,
+            ),
+        )
+
+    def build_query(self, *args, **kwargs):
+        try:
+            query = super().build_query(*args, **kwargs)
+            # get the records whose status is verified
+            results = query.filter_by(status_num=1)
```
vrajmohan: build_query returns a `query`. It would be misleading to call it `results`.

webservices/resources/operations_log.py
```diff
@@ -0,0 +1,53 @@
...
+        except:
+            raise ApiError("Unexpected Server Error", 500)
```
vrajmohan: This hides the root cause of the exception and makes it difficult to diagnose the underlying issue.

pkfec: are you suggesting to take off the entire try/except block from here, since we have a master exception handler in rest.py?

vrajmohan: Either that, or if we choose to catch the exception here, we should at least log what it was and not suppress it. When we used this pattern for the coverage dates endpoint, this handler made it difficult to debug.

webservices/resources/operations_log.py
pkfec: changed it to query.

webservices/resources/operations_log.py
pkfec: The master exception handler is handing the exceptions, and giving meaningful error message. I have removed the try/except block from this file.

vrajmohan:
Not a must have, but I would recommend judicious use of `git cherry-pick` and `git rebase` to simplify the commit history.

pkfec:
@vrajmohan  Please review.

tests/test_operation_logs.py
vrajmohan: This test is giving the misleading impression that I could query the endpoint by `status_num`, ie, `api.url_for(OperationsLogView, status_num=0)`, `api.url_for(OperationsLogView, status_num=1)`, etc. Is that correct? Aren't we fixing the query to use `status_num == 1` in https://github.com/fecgov/openFEC/pull/3203/files#diff-362d4bedb02a4abf10fc6a6868a7b021R39?

pkfec:
@vrajmohan do not merge this PR yet. I have one additional change, i.e. add this new MV to the  daily refresh manage task.

Will ping you once i add the changes.

pkfec:
@vrajmohan operations_log MV is now added to the manage task that does the daily refresh.
Please review now.

data/migrations/V0087__operations_log_mv.sql
```diff
@@ -0,0 +1,51 @@
...
+    o.pass_3_coding_dt,
+    o.pass_3_entry_done_dt,
+    o.receipt_dt,
+    o.beginning_coverage_dt,
+    o.ending_coverage_dt,
+    o.file_num,
+    o.prev_file_num
+   FROM staging.operations_log o
+     LEFT JOIN disclosure.f_rpt_or_form_sub f ON o.sub_id = f.sub_id
+  WHERE (o.form_tp::text <> ALL (ARRAY['RFAI'::character varying::text, 'ICI'::character varying::text, 'F1Z'::character varying::text, 'F2Z'::character varying::text])) AND o.beg_image_num IS NOT NULL AND (o.status_num = ANY (ARRAY[0::numeric, 1::numeric]))
```
lbeaufort: It might be more readable this way. Do we want to bring in `status_num` `0`?
```
WHERE o.form_tp NOT IN ('RFAI', 'ICI', 'F1Z', 'F2Z') 
  AND o.beg_image_num IS NOT NULL 
  AND o.status_num IN (0, 1)
```

data/migrations/V0087__operations_log_mv.sql
```diff
@@ -0,0 +1,51 @@
...
+WITH DATA;
+
+ALTER TABLE public.ofec_operations_log_mv
+  OWNER TO fec;
+GRANT ALL ON TABLE public.ofec_operations_log_mv TO fec;
+GRANT SELECT ON TABLE public.ofec_operations_log_mv TO fec_read;
+
+CREATE INDEX ofec_operations_log_mv_status_num_idx1
```
lbeaufort: It looks like we need a `UNIQUE INDEX` in order to refresh materialized. When I ran `invoke create_sample_db` I got an error message: 

```
cannot refresh materialized view "public.ofec_operations_log_mv" concurrently
HINT:  Create a unique index with no WHERE clause on one or more columns of the materialized view.
 [SQL: 'REFRESH MATERIALIZED VIEW CONCURRENTLY ofec_operations_log_mv']
```

data/migrations/V0087__operations_log_mv.sql
lbeaufort: 👍 Looks like we're bringing in the `0`'s with this migration.

data/migrations/V0087__operations_log_mv.sql
pkfec: not sure if we need a  unique index to refresh MV. I will look into this.

data/migrations/V0087__operations_log_mv.sql
hcaofec: status =0 are records that not verified yet

data/migrations/V0087__operations_log_mv.sql
lbeaufort: We do. Happy to help if you get stuck!

pkfec: Since we switched to MV, i see that a unique index is required. I have one unique index in place now. Thanks for flagging this.

data/migrations/V0087__operations_log_mv.sql
pkfec: The original requirement is to pull records with      status_num =1 records only. 
@jwchumley is looking at the SQL.

pkfec:
@vrajmohan Please review

data/migrations/V0087__operations_log_mv.sql
vrajmohan: Can you please address https://github.com/fecgov/openFEC/pull/3203/files#r199188265?

data/migrations/V0087__operations_log_mv.sql
```diff
@@ -0,0 +1,48 @@
...
+
+SET search_path = public;
+
+CREATE MATERIALIZED VIEW public.ofec_operations_log_mv AS 
+ SELECT o.sub_id,
+    o.status_num,
+    o.cand_cmte_id,
+    o.beg_image_num,
+    o.end_image_num,
+        CASE
```
vrajmohan: These 4 lines can be replaced with `COALESCE(f.form_tp, regexp_replace(o.form_tp, 'A|N|T', ''))`

data/migrations/V0087__operations_log_mv.sql
```diff
@@ -0,0 +1,48 @@
+-- Materialized View: public.ofec_operations_log_mv
+
+DROP MATERIALIZED VIEW IF EXISTS public.ofec_operations_log_mv;
+
+SET search_path = public;
+
+CREATE MATERIALIZED VIEW public.ofec_operations_log_mv AS
```
vrajmohan: With the concern about too many materialized views, have we confirmed that one is actually needed?

pkfec: The request came from @jwchumley, since it was not easy to get the form_types from the existing operations_log view.

tests/factories.py
```diff
@@ -427,3 +427,8 @@ class Meta:
         model = models.StateElectionOfficeInfo
     state = 'VA'
     office_type = 'STATE CAMPAIGN FINANCE'
+
+class OperationsLogFactory(BaseFactory):
+    class Meta:
+        model = models.OperationsLog
+    candidate_committee_id = '01'
```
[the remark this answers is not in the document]
pkfec: Not needed unless its a primary key. I will delete line#434.

data/migrations/V0087__operations_log_mv.sql
vrajmohan: I meant: "Would a _plain_ view suffice instead of a _materialized_ view"?

pkfec: @jwchumley @ChingKuan Could either one of you respond to this?

tests/test_operation_logs.py
```diff
@@ -0,0 +1,33 @@
...
+from tests import factories
+from tests.common import ApiBaseTest
+from webservices.rest import api
+from webservices.resources.operations_log import OperationsLogView
+
+class TestOperationsLog(ApiBaseTest):
+
+    def setUp(self):
+        super().setUp()
+        factories.OperationsLogFactory(candidate_committee_id='00', report_year=2000, status_num=1),
```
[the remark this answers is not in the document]
pkfec: Not needed. I will delete this.

data/migrations/V0087__operations_log_mv.sql
jwchumley: I'd prefer to include the zeros but that would mean we also need to be able to filter by status.

data/migrations/V0087__operations_log_mv.sql
ChingKuan: Either one is good, only less than 200 msec difference in performance.  View is easy to maintain and no disk spaces request.  I've created a view (public.ofec_operations_log_vw) , you can try it.
--
select * from public.ofec_operations_log_vw where cand_cmte_id = 'C00329920'; -- execution time 198.6 mesc. 
--
select * from public.ofec_operations_log_mv where cand_cmte_id = 'C00329920'; -- execution time 0.38 mesc.

tests/test_operation_logs.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_search_cand_cmte_id(self):
+
+        response = self.app.get(api.url_for(OperationsLogView, candidate_committee_id='01', report_year=2012))
+        self.assertEquals(response.status_code, 200)
+
+    def test_unverified_reports(self):
+
+        results = self._results(api.url_for(OperationsLogView, candidate_committee_id='01', status_num=0))
+        self.assertEqual(len(results), 1)
```
lbeaufort: Since we're testing for the `status_num` value, I think we can take out the less meaningful (and less maintainable) test for exactly 1 result. If later we added more test data, it could break the test.

webservices/docs.py
```diff
@@ -1194,3 +1194,37 @@ def add_ytd(var):
 LINK_TO_REPORT = '''
 URL for retrieving the PDF document
 '''
+
+SUB_ID = '''
+A unique identifier of the transactional report.
+'''
+
+STATUS_NUM = '''
```
lbeaufort: If we only allow users to see `status_num` 0 and 1, we shouldn't let them choose 4 and 9 in the dropdown.

pkfec: removed the 4 and 9 text from the docs.

webservices/args.py
```diff
@@ -778,3 +778,13 @@ def make_seek_args(field=fields.Int, description=None):
...
 }
+
+operations_log = {
+    'candidate_committee_id': fields.List(IStr, description=docs.CAND_CMTE_ID),
+    'report_type': fields.List(IStr, description=docs.REPORT_TYPE),
+    'beginning_image_number': fields.List(fields.Str, description=docs.BEGINNING_IMAGE_NUMBER),
+    'report_year': fields.List(fields.Int, description=docs.REPORT_YEAR),
+    'form_type': fields.List(IStr, description=docs.FORM_TYPE),
+    'amendment_indicator': fields.List(IStr, description=docs.AMENDMENT_INDICATOR),
+    'status_num': fields.List(IStr, description=docs.STATUS_NUM),
```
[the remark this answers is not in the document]
pkfec: can pick either one value not both.

pkfec:
@lbeaufort 
I have updated the **status_num** filter to a multi-select option on the api. 
Can you review the PR?

PR #3216, opened 2018-06-18
Feature/periodic refresh ao

### LINKED ISSUE #2856: Set up periodic refresh of Advisory Opinions
We need to refresh Advisory Opinions so that older AO's can see when newer AO's reference them. I would suggest creating a daily celery task in the evening to see if any AO's have been published that day. We might want to reload all advisory opinions once a week in case that one day got missed.

`./manage.py load_advisory_opinions`

- [x] Create a celery task to check at the end of the day if an AO has been published that day, and reload all AO's
- [x] Create a weekly task to reload all AO's - maybe weekends?
- [x] Make sure this doesn't interfere with the existing refresh task - maybe do them on different time windows.

rjayasekera:
## Summary (required)
Set up a schedule to refresh all AOs daily if there are any changes to AOs for the past 24hr period.  But on Sundays reload of all AOs whether there are any changes to AOs or not.

- Resolves #2856 

# Include a summary of proposed changes.

- Modify celery-worker schedule in __init.py__
- Create a  new schedule task in legal_docs.py

Run redis-server, elasticsearch, celery-beat, and celery-worker locally by modifying the schedule as needed.  Setup SQLA_CONN to dev database.

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()


webservices/tasks/__init__.py
```diff
@@ -12,19 +13,25 @@
     schedule = {
         'refresh': {
             'task': 'webservices.tasks.refresh.refresh',
-            'schedule': crontab(minute=0, hour=9),
+            'schedule': crontab(minute=0, hour=9),
```
vrajmohan: Should we fix this as well not that we have  realized that the times are all UTC?

webservices/tasks/legal_docs.py
```diff
@@ -9,6 +12,14 @@
 
 logger = logging.getLogger(__name__)
 
+DAILY_MODIFIED_STARTING_AO = """
```
vrajmohan: Can't we reuse the `RECENTLY_MODIFIED_STARTING_AO` query for this purpose as well?

webservices/tasks/legal_docs.py
```diff
@@ -30,6 +41,28 @@ def refresh():
         refresh_aos(conn)
         refresh_murs(conn)
 
+@app.task(once={'graceful': True}, base=QueueOnce)
+def reload_all_aos():
+    """
+    Reload all AOs daily (except Sunday) if there were any new or modified AOs found for the past 24 hour period
+    But on Sundays reload all AOs whether there were changes to AOs or not
+    """
+    if (datetime.date.today().strftime("%A") == 'Sunday'):
```
vrajmohan: Instead of burying it in this `if` statement, would it be better to have 2 separate schedules with 2 separate tasks?

webservices/tasks/__init__.py
```diff
@@ -12,19 +13,25 @@
     schedule = {
         'refresh': {
             'task': 'webservices.tasks.refresh.refresh',
-            'schedule': crontab(minute=0, hour=9),
+            'schedule': crontab(minute=0, hour=9), 
+        },
+        
+        'reload_all_aos': {
+            'task': 'webservices.tasks.legal_docs.reload_all_aos',
+            'schedule': crontab(minute=15, hour=1),
```
vrajmohan: What is this schedule?

webservices/tasks/__init__.py
rjayasekera: I think this is fine.  We want to refresh data 5am EST (9am UTC)

webservices/tasks/__init__.py
rjayasekera: this is when we plan to run reload of all AOs task daily.

webservices/tasks/legal_docs.py
rjayasekera: I thought 8 hrs going back is not sufficient.

vrajmohan: That makes sense

webservices/tasks/__init__.py
```diff
@@ -10,21 +11,37 @@
 schedule = {}
 if env.app.get('space_name', 'unknown-space').lower() != 'feature':
     schedule = {
+        # Refresh all Materialized views at 9am UTC daily
```
vrajmohan: Comments are usually good to have, but they suffer from the fact that they are often left unmodified when code changes. For this reason, it is often better to rename methods instead of adding comments. E.g., `webservices.tasks.refresh.refresh` should become `webservices.tasks.refresh.refresh_materialized_views`.

Instead of explaining that the schedule is for 9am UTC daily, it may be better to do the following:
1. Add a single comment highlighting that all schedules are UTC
2. Say _why_ 9am UTC is the best time for refreshing materialized views.

webservices/tasks/__init__.py
```diff
@@ -10,21 +11,37 @@
...
+        
+        # If any changes in AOs during the day all AOs get reloaded at 1am UTC
+        'reload_all_aos': {
+            'task': 'webservices.tasks.legal_docs.reload_all_aos',
+            'schedule': crontab(minute=0, hour=1, day_of_week='mon,tue,wed,thu,fri,sat'),
+        },
+        
+        # All AOs get reloaded every Sunday at 1am UTC whether any changes to AOs or not
+        'reload_all_aos_wkly': {
+            'task': 'webservices.tasks.legal_docs.reload_all_aos_wkly',
```
vrajmohan: The method name should say what it does `reload_all_aos`. The schedule should not leak into the method name. We would not want to change the method name if we decided to run this every morning at 5AM UTC.

webservices/tasks/legal_docs.py
```diff
@@ -30,6 +43,35 @@ def refresh():
...
+    Reload all AOs daily (determine by schedule) if there were any new or modified AOs found for the past 24 hour period
+    """
+    with db.engine.connect() as conn:
+        row = conn.execute(DAILY_MODIFIED_STARTING_AO).first()
+        if row:
+            logger.info("AO found %s modified at %s", row["ao_no"], row["pg_date"])
+            logger.info("Daily (%s) reload of all AOs starting", datetime.date.today().strftime("%A"))
+            load_advisory_opinions()
+            logger.info("Daily (%s) reload of all AOs completed", datetime.date.today().strftime("%A"))
+            slack_message = 'Daily reload of all AOs completed in {0} space'.format(env.get_credential('NEW_RELIC_APP_NAME'))
```
vrajmohan: It was confusing to me that we are using 'NEW_RELIC_APP_NAME'. It may be better to have a utility method like `get_app_name()`. Inside that method, we could add a comment saying that we leverage 'NEW_RELIC_APP_NAME'.

vrajmohan:
Looks great!

PR #3225, opened 2018-06-22
Remove 'won' from /elections/

### LINKED ISSUE #3076: Elections 'won' Field Frequently Incorrect
I am using the openFEC API for a data science project for my college coursework, and I've noticed recently that the `/elections/` endpoint often returns a `results` list wherein the `won` field is returned as `false` for the victor of the election.  I have observed this issue for the 2016, 2008, 2000, 1996, 1992, 1988, and 1982 election cycles.

Attached is some example data collected using Powershell from the 2016, 2012, 2008, and 2004 election cycles.  True negatives are circled in yellow, true positives are circled in green, and false negatives are circled in red.  I have encountered no false positives.
fec api election endpoint error annotated

I have additionally tested this issue using `curl` instead of PowerShell's `Invoke-WebRequest` to ensure that the issue was not a result of binding issues, and I further tested the issue using demo functionality of the [openFEC Developers](https://api.open.fec.gov/developers/#!/financial/get_elections) page.  In all cases, the issue persists.

I am unsure if this issue is the result of an issue with the database, or if there is an issue in the [_get_records method](https://github.com/fecgov/openFEC/blob/88593e0bb4f052b61621c2a791bd34ed6598bc9a/webservices/resources/elections.py#L214) for ElectionsView.  I am not an FEC employee, so I may be a bit limited in what I can do in the former case.  Regardless, I am happy to help solve this problem any way I can!

lbeaufort:
## Summary

- Resolves #3076 

_Remove 'won' from /elections/ until we find a better data source for won/lost results._

- Run this branch locally and test an elections query and make sure `won` is showing up http://localhost:5000/v1/elections/?api_key=DEMO_KEY&cycle=2018&office=house&state=CO&stateFull=Colorado&district=07&per_page=0&sort_hide_null=true

-  https://api.open.fec.gov/v1/elections - The `won` field doesn't show up on the front end - this is an API-only change

### DIFF
--- a/data/migrations/V0092__drop_ofec_election_result_mv.sql
+++ b/data/migrations/V0092__drop_ofec_election_result_mv.sql
@@ -1,0 +1,10 @@
+/*
+Resolves #3706
+
+Drop ofec_election_result_mv - removing elections/won until we can determine a better data source
+
+*/
+
+SET search_path = public;
+
+DROP MATERIALIZED VIEW IF EXISTS ofec_election_result_mv
--- a/manage.py
+++ b/manage.py
@@ -275,7 +275,6 @@ def refresh_materialized(concurrent=True):
         'committee_history': ['ofec_committee_history_mv'],
         'communication_cost': ['ofec_communication_cost_mv'],
         'communication_cost_by_candidate': ['ofec_communication_cost_aggregate_candidate_mv'],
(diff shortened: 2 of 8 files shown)


JonellaCulmer:
Flagging, @pkfec This merge is dependent upon your change, #3203.

Issue #3221, opened 2018-06-20
Elections endpoint is including joint fundraisers and inflating candidate receipts and disbursements


jwchumley:
https://www.fec.gov/data/elections/senate/VA/2018/

There are several issues with the SQL in the elections endpoint:

1. it is including joint fundraisers so any candidate that has one or more joint fundraiser has inflated figures (because the joint fundraising committee transfers money to the candidate committee so the money is already there)
  
- [x] add AND ofec_cand_cmte_linkage_mv.cmte_dsgn in ('A','P') to the anon_4 where clause

2. There are two tables that we use to make sure candidates are not on the list:
   disclosure.cand_inactive (for candidates that withdraw from an election) which is already included in one of the underlying views and:
   disclosure.unverified_cand_cmte (for candidates or committees that have been removed for administrative reasons) which is not accounted for

- [ ] Add unverified candidate statement to the where clause of the **anon_4** sub select:  
AND ofec_cand_cmte_linkage_mv.cand_id not in (select cand_cmte_id from disclosure.unverified_cand_cmte)

fec-jli:
This(elections) endpoint is created based on ** ofec_totals_house_senate_mv**
then leftjoin a couple of queries using SQLAlchemy orm function.
Here is the final query, Could you look at it? Thanks:
`SELECT DISTINCT anon_1.candidate_id AS anon_1_candidate_id, anon_1.candidate_election_year AS anon_1_candidate_election_year, 
anon_1.candidate_name AS anon_1_candidate_name, 
anon_1.party_full AS anon_1_party_full, 
anon_1.incumbent_challenge_full AS anon_1_incumbent_challenge_full, 
anon_1.office AS anon_1_office, 
anon_1.total_receipts AS anon_1_total_receipts, 
anon_1.total_disbursements AS anon_1_total_disbursements, 
anon_1.cash_on_hand_end_period AS anon_1_cash_on_hand_end_period, anon_1.committee_ids AS anon_1_committee_ids, 
anon_1.coverage_end_date AS anon_1_coverage_end_date, 
anon_1.candidate_pcc_id AS anon_1_candidate_pcc_id, 
anon_2.candidate_id AS anon_2_candidate_id, 
anon_2.cash_on_hand_end_period AS anon_2_cash_on_hand_end_period, 
[...]

PR #3255, opened 2018-06-29
Remove the joint fundraisers inflating money from totals

fec-jli:
## Summary 
remove joint fundraiser inflated figures from totals

- Resolves  The first Item of Issue [3221] 
[https://github.com/fecgov/openFEC/issues/3221](https://github.com/fecgov/openFEC/issues/3221)

1)checkout openFEC feature branch:  feature/joint_fundraisers_inflating_mone
2)run py.test and run server,
3)compare Total receipts, Total disbursements, Cash on hand 
from prod and local. will see difference as screenshot below.
[http://127.0.0.1:8000/data/elections/senate/VA/2018/](http://127.0.0.1:8000/data/elections/senate/VA/2018/)
[https://www.fec.gov/data/elections/senate/VA/2018/](https://www.fec.gov/data/elections/senate/VA/2018/)

on production:
screen shot 2018-06-27 at 12 13 07 am

on local:
screen shot 2018-06-27 at 12 12 37 am

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -286,6 +286,7 @@ def join_candidate_totals(query, kwargs, totals_model):
         sa.and_(
             CandidateCommitteeLink.committee_id == totals_model.committee_id,
             CandidateCommitteeLink.fec_election_year == totals_model.cycle,
+            CandidateCommitteeLink.committee_designation.in_(['P', 'A']),
         )
     )
 
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -273,8 +273,16 @@ election_durations = {
     'house': 2,
 }
 
-
 def join_candidate_totals(query, kwargs, totals_model):
+    """
+    Joint Fundraising Committees raise money for multiple
+    candidate committees and transfer that money to those committees.
+    By limiting the committee designations to A and P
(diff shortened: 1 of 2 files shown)


webservices/resources/elections.py
```diff
@@ -286,6 +286,7 @@ def join_candidate_totals(query, kwargs, totals_model):
         sa.and_(
             CandidateCommitteeLink.committee_id == totals_model.committee_id,
             CandidateCommitteeLink.fec_election_year == totals_model.cycle,
+            CandidateCommitteeLink.committee_designation.in_(['P', 'A']),
```
lbeaufort: Thanks for making this change! Like @vrajmohan mentioned in the earlier PR, it would be good to add a comment as to why this is needed, as well as add automated tests. The test should add other committee designation data and assert that they are not picked up in the totals.

fec-jli: Thanks for review and good suggestion.
I will add some comment for the reason to add this condition.
for automated tests, I don't have idea how to do that.
here, committee_designation is hard code in['P','A'], committee_designation is not a filter, not in schema for output. no idea how to add other committee designation create the instance of object?

vrajmohan: I can pair with you to help with adding a test.

Issue #3235, opened 2018-06-25
Remove S3 caching


vrajmohan:
The current implementation of S3 allows the use of the application only if the following conditions are met:
App server is up but for some reason, DB/Elasticsearch connections are down
The request is being repeated with the same params since the cache was cleared at 2am UTC.

_Note:_ 

1. _This cache has failed us when we failed to do the Elasticsearch upgrade in prod._
2. _The S3 caching of responses was the reason that NewRelic was not alerting on dev when we removed an underlying table_

In addition, the logging done by this code is incorrect and does not indicate when the cache is actually being used.

**Completion criteria:** 
- [ ] Flip the feature flag to disable S3 caching
- [ ] Deploy as hotfix to lessen logging records
- [ ] Delete caching code

pkfec:
I would like to work on disabling the cache requests and remove the celery task that uploads the cached requests to S3 bucket in this sprint.

I need an extra day to test the changes throughly, but i can wrap up the work before this sprint ends.

PR #3257, opened 2018-07-02
Feature/disable s3 cache calls

pkfec:
## Summary (required)

To disable caching the requests into S3.

- Resolves https://github.com/fecgov/openFEC/issues/3235

_Include a summary of proposed changes.

- Remove the `CACHE_ALL_REQUEST= True` env variable from manifest_*.yml files

1. manifest_api_dev.yml
2. manifest_api_prod.yml
3. manifest_api_stage.yml
4. manifest_celery_beat_prod.yml
5. manifest_celery_beat_stage.yml
6. manifest_celery_worker_prod.yml
7. manifest_celery_worker_stage.yml

- Remove the cache_calls  python code from rest.py
- Remove the celery task that upload the requests to S3.
- Remove the  cache_calls python tests 

- checkout the `feature/disable-s3-cache-calls` branch.
- export the `SQLA_CONN `to point to `Dev` DB 
- start the server `./manage.py runserver`
- check the local server log to make sure the requests are not cached.
- check the s3 bucket in `dev` space to make sure that requests are not cached.

-  S3 bucket.


pkfec:
@vrajmohan 
 Since i remove the exception handling section from rest.py the below test was throwing APIError. I tweaked the test a little bit i am not sure if we really need to raise exceptions in python tests.
`def test_download_forbidden(self):` 
Please review this PR.

webservices/rest.py
```diff
@@ -223,65 +162,21 @@ def add_caching_headers(response):
...
-        else:
-            app.logger.error(
-                'An error occured while retrieving the cached file from S3.'
-            )
-            raise exceptions.ApiError(
-                'The requested URL could not be found.'.format(request.url),
-                status_code=http.client.NOT_FOUND
-            )
-    else:
-        raise exceptions.ApiError(
```
vrajmohan: We should still raise this exception if we cannot return a response. We may need to revisit exception handling so that we are in control of what we return to the client.

tests/test_downloads.py
```diff
@@ -143,7 +143,9 @@ def test_download_cached(self, export, get_cached):
         assert not export.delay.called
 
     def test_download_forbidden(self):
-        with pytest.raises(ApiError):
+        # should the tests throw exceptions?
```
vrajmohan: The test is asserting that a particular exception is raised. Nothing wrong with that.

tests/test_downloads.py
```diff
@@ -143,9 +143,7 @@ def test_download_cached(self, export, get_cached):
         assert not export.delay.called
 
     def test_download_forbidden(self):
-        # should the tests throw exceptions?
-        # with pytest.raises(ApiError):
-        with self.assertRaises(Exception):
```
vrajmohan: This is again a candidate for `git rebase -i`. Otherwise, we are confusing future maintainers by modifying and reverting code within the same PR.

webservices/rest.py
```diff
@@ -165,15 +165,16 @@ def add_caching_headers(response):
 @app.errorhandler(Exception)
 def handle_exception(exception):
     wrapped = ResponseException(str(exception), ErrorCode.INTERNAL_ERROR, type(exception))
-
+    requested_url = request.url
```
vrajmohan: Do we need this variable?

pkfec: Thought of logging the URL so we know which one failed.

vrajmohan: I am not sure I follow, it is not being used in the following logging call. It is used in one place, which is:
```
raise exceptions.ApiError('Could not process the request %s'.format(requested_url),
            status_code=http.client.NOT_FOUND)
```
but that could use `request.url` directly, right?

pkfec: either way is fine i think.

vrajmohan: I disagree. Sometimes, it makes sense to create a variable to hold a field from an object, when the name adds more clarity, especially when the value is being used in multiple places. In this case, we are adding code with no payoff.

pkfec:
@vrajmohan  Can you review for one last time?

pkfec:
@vrajmohan Please review.

Issue #3054, opened 2018-04-10
Enable zipcode search for foreign zipcodes 


lbeaufort:
As a follow-up to the changes here https://github.com/fecgov/openFEC/pull/3012/files we should enable zipcode search for foreign zipcodes.

From an API user: 

>We saw that the change was made to the API to support 9 digit zips within the 5 digit zip search - that is really great, and very helpful, thank you for making this happen!

>One thing we did notice is that foreign postal codes no longer seem to be supported, and when we call the API with one of those postal codes (say for London) we get an error. I'm not quite sure on the best way to handle that but just wanted to flag it for you. 

London zipcode format: `SW1A 1AA`
Canada zipcode: `A1A 1A1`

Example : https://api.open.fec.gov/v1/schedules/schedule_a/?sort=contribution_receipt_date&contributor_zip=SW1A%201AA&two_year_transaction_period=2018&api_key=DEMO_KEY&per_page=20

- [x] Enable zipcode search for zipcodes greater than 5 digits
- [x] Match on the first 5 characters
- [x] Test performance of /schedules/schedule_a/

cc: @PaulClark2 @jwchumley

vrajmohan:
This may have been introduced by https://github.com/fecgov/openFEC/issues/2980

hcaofec:
in the process of getting consensus on business rules with PMs.

PR #3264, opened 2018-07-05
[WIP]Enable zipcode search for foreign zipcodes, and remove input of 5-digits restricition

hcaofec:
## Summary (required)

- Resolves [#3054](https://github.com/fecgov/openFEC/issues/3054)
1) Enable zipcode search for zipcodes greater than 5 digits.
2) Enable search for foreign zipcode that has letters, space, and -,  and to exclude certain special character.  
3) The results still match on the first 5 characters.
 4) refactor test case accordingly.

1) run this branch locally, using DEV database
2) test this endpoint locally :  /schedules/schedule_a/
    
    two_year_transaction_period=2016
    contributor_zip: using the value in **zip code** column
    
    

zip code | counts before change | counts after change |  
---- | -- | -- | --
10050 | 26 | 26 |  
100502814 | error code = 400 | 26 |  
10050-2814|error_code = 400 | 26
L2A2W4 | error code = 400 | 3 |  L2A2W4 ( 3 )
L5K 2B3 | error code = 400 | 3 | L5K 2G1;   L5K 2G1;    L5K 2B3

The counts  concur with the results by running query against database directly.

### DIFF
--- a/tests/test_itemized.py
+++ b/tests/test_itemized.py
@@ -111,24 +111,27 @@ class TestItemized(ApiBaseTest):
         [
             factories.ScheduleAFactory(contributor_zip=96789),
             factories.ScheduleAFactory(contributor_zip=9678912),
-            factories.ScheduleAFactory(contributor_zip=967891234)
+            factories.ScheduleAFactory(contributor_zip=967891234),
+            factories.ScheduleAFactory(contributor_zip='M4C 1M7')
         ]
-
-        results = self._results(api.url_for(ScheduleAView, contributor_zip=96789, **self.kwargs))
+        results = self._results(api.url_for(ScheduleAView, contributor_zip=967893405, **self.kwargs))
         self.assertEqual(len(results), 3)
     
-    def test_invalid_sched_a_zip(self): 
-        response = self.app.get(api.url_for(ScheduleAView,contributor_zip=9678912,cycle=2018))
-        self.assertEqual(response.status_code,400)
+        results = self._results(api.url_for(ScheduleAView, contributor_zip='M4C 1M55', **self.kwargs))
+        self.assertEqual(len(results), 1)
 
-        response = self.app.get(api.url_for(ScheduleAView,contributor_zip='9678-',cycle=2018))
+        contributor_zips = ['M4C 1M5555',96789]
+        results = self._results(api.url_for(ScheduleAView, contributor_zip=contributor_zips,**self.kwargs))
+        self.assertEqual(len(results), 4)
+
+    def test_invalid_sched_a_zip(self): 
+        response = self.app.get(api.url_for(ScheduleAView,contributor_zip='96%',cycle=2018))
         self.assertEqual(response.status_code,400)
 
     def test_filter_multi_start_with(self):
         [
             factories.ScheduleAFactory(contributor_zip=1296789)
         ]
-
         results = self._results(api.url_for(ScheduleAView, contributor_zip=96789, **self.kwargs))
         self.assertEqual(len(results), 0)
 
--- a/tests/test_itemized.py
+++ b/tests/test_itemized.py
@@ -1,4 +1,3 @@
-
 import datetime
 
 import sqlalchemy as sa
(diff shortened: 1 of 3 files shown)


tests/test_itemized.py
```diff
@@ -111,24 +111,27 @@ def test_filter_sched_a_zip(self):
...
+        results = self._results(api.url_for(ScheduleAView, contributor_zip=967893405, **self.kwargs))
         self.assertEqual(len(results), 3)
     
-    def test_invalid_sched_a_zip(self): 
-        response = self.app.get(api.url_for(ScheduleAView,contributor_zip=9678912,cycle=2018))
-        self.assertEqual(response.status_code,400)
+        results = self._results(api.url_for(ScheduleAView, contributor_zip='M4C 1M55', **self.kwargs))
+        self.assertEqual(len(results), 1)
+
+        contributor_zips = ['M4C 1M5555',96789]
```
vrajmohan: PEP8 violation

tests/test_itemized.py
```diff
@@ -111,24 +111,27 @@ def test_filter_sched_a_zip(self):
...
-
-        results = self._results(api.url_for(ScheduleAView, contributor_zip=96789, **self.kwargs))
+        results = self._results(api.url_for(ScheduleAView, contributor_zip=967893405, **self.kwargs))
         self.assertEqual(len(results), 3)
     
-    def test_invalid_sched_a_zip(self): 
-        response = self.app.get(api.url_for(ScheduleAView,contributor_zip=9678912,cycle=2018))
-        self.assertEqual(response.status_code,400)
+        results = self._results(api.url_for(ScheduleAView, contributor_zip='M4C 1M55', **self.kwargs))
+        self.assertEqual(len(results), 1)
```
vrajmohan: Are we sure this is intended? 'M4C 1M55' will match 'M4C 1M7'.

webservices/resources/sched_a.py
```diff
@@ -80,16 +80,18 @@ def args(self):
     def build_query(self, **kwargs):
         query = super().build_query(**kwargs)
         query = filters.filter_contributor_type(query, self.model.entity_type, kwargs)
-
+        zip_list=[]
         if kwargs.get('contributor_zip'):
             for value in kwargs['contributor_zip']:
-                if re.search('^-?\d{5}$',value) is None:
+                if re.search('[^a-zA-Z0-9-\s]+',value):
```
vrajmohan: We don't need the trailing plus.

hcaofec:
are we using the same matching on first 5 digits/characters rule for foreign zipcodes?  
for example, if searing for 'L5K 2B3',  the below three rows will return with satisfying condition,
zipcode like 'L5K 2%'

 L5K 2G1;   L5K 2G1;    L5K 2B3

@PaulClark2 @jwchumley @lbeaufort

PaulClark2:
@hcaofec I think matching on the first five characters is an okay place to start. I do have one question. What happens if the postal code is less than five characters? For example, Iceland uses 3-digit postal codes.

vrajmohan:
My $0.02: If we detect the zipcode to be 9-digit US (with or without a - between 5th and 6th), use the 1st 5 digits for a match. In all other cases, perform an exact match.

hcaofec:
If API searches by 10050, then zipcode=100502814 in database will not be returned. This is listed as one of the issues to be fixed in original ticket.

@vrajmohan

vrajmohan:
I was thinking that when we store the postal code we would create an expression based index (which uses the above logic) and then query it using the same expression.

tests/test_itemized.py
hcaofec: yes for now since we would rather be over-inclusive than under-inclusive

lbeaufort:
@hcaofec I'm sorry I ran out of time to finish reviewing this but will look on Monday. I'm having an issue with testing some London zipcodes, and want to find more test examples in the data. 

For this example, it finds no results but shows `count: 8412, pages: 281`

http://127.0.0.1:5000/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=true&data_type=processed&two_year_transaction_period=2018&contributor_zip=WC2N+5DU&contributor_zip=SW1A+1AA&sort=-contribution_receipt_date&per_page=30

screen shot 2018-07-20 at 3 47 54 pm

hcaofec:
looking into it. @lbeaufort

hcaofec:
One solution is using actual count of the query like [this](https://github.com/fecgov/openFEC/blob/develop/webservices/common/counts.py#L20)

But in current api, since there is no minimum requirement of the length of search string, nor there is maximum restriction on how many zipcodes can be searched, the result set could be large if user searches by 1 digit or 5+ zipcodes at once, which will be very slow to get count.

zipcode | actual count
-- | --
22 | 884748
20170,22101,221 | 144415
20170,22101|35501 

In order to get accurate count in time, we may want to reduce the size of results by endpoint by adding this:
1) set minimum length of zipcode = 3 ( this is the same rule used in classic website ). For example, 2, 23 are not accepted. 123, 1234, 12345, 123465 are ok
2) disable multiple zipcode search. Only 1 zipcode per api call.

Another possible solution is refactor the query to have planner return more accurate estimate.

JonellaCulmer:
@hcaofec Are we good to go on this with the review from @vrajmohan, or does @lbeaufort still need to review before we merge this PR?

lbeaufort:
I've confirmed the count issue is an existing one, and shouldn't hold up the foreign zip code work. I'm going to keep testing today. Example: https://api.open.fec.gov/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=true&data_type=processed&two_year_transaction_period=2018&contributor_zip=88888&contributor_zip=77777&sort=-contribution_receipt_date&per_page=30

PR #3268, opened 2018-07-06
[DO NOT MERGE] Add pgTap test for amendment chain view

vrajmohan:
Install pgTap using https://pgtap.org/documentation.html.

Run the tests using `pg_prove  --verbose -d fectest tests/sql/efiling_amendment_chain.sql`


lbeaufort:
We've decided to stick with `pytest` for testing SQL / MV logic. Example: https://github.com/fecgov/openFEC/blob/develop/tests/integration/test_amendment_chain.py#L14

PR #3290, opened 2018-07-18
feature/add-es-timeout

pkfec:
Add retries on ElasticSearch connection when timeout happens.


lbeaufort:
@pkfec let's test this on `dev` before we merge it.

pkfec:
Not need. Closing it.

PR #3291, opened 2018-07-18
[Close without merge] Support Schedule A filtering by committee_type

vrajmohan:


fec-jli:
Thanks, This is really helpful for us to learn how to use contains_eager to filter fields inside sub-object.
Do we need create a new filter function in filters.py,  maybe like this:
 `def filter_fields_in_sub_object(query, model, kwargs, fields)`

in case we need more filters from sub-object.

vrajmohan:
Yes, that may be a useful function to create. I would caution against too many of those - explicit is better than implicit.

Right now, I am still experimenting.

webservices/resources/sched_a.py
```diff
@@ -56,10 +56,6 @@ def amount_column(self):
     filter_multi_start_with_fields = [
         ('contributor_zip', models.ScheduleA.contributor_zip),
     ]
-    query_options = [
-        sa.orm.joinedload(models.ScheduleA.committee),
-        sa.orm.joinedload(models.ScheduleA.contributor),
-    ]
```
rjayasekera: I am not sure why you remove `sa.orm.joinedload(models.ScheduleA.contributor)` .

rjayasekera:
@vrajmohan Thank you.  This is very helpful to understand aliased, eager query joined.  And also now I understand why you always say make 'logical commits' so the other person can understand the program changes better.

rjayasekera:
Query by committee_type performance is very good.  At some point we need to make this a multi select field I suppose.

webservices/resources/sched_a.py
rjayasekera: I think that is because now the relationship happens in the sched_a resource file
`committee_committee_history_alias = aliased(models.CommitteeHistory)
    contributor_committee_history_alias = aliased(models.CommitteeHistory)`

vrajmohan: That is correct.

vrajmohan:
An index on `ofec_committee_history_mv.committee_type` may also help.

rjayasekera:
Close without merge this pull request.  Created a new pull request.

PR #3292, opened 2018-07-19
[WIP]Refactor elections endpoint to handle candidate finance totals export function

fec-jli:
## Summary
1)In order to enable 'export' button on these two sections in election summery page, we need refactor elections endpoint to extend ApiResouce and handle download function.
These two sections use same endpoint: `resource.ElectionView`
a)Candidate financial totals section:
[https://www.fec.gov/data/elections/senate/VA/2018/?tab=election-data](https://www.fec.gov/data/elections/senate/VA/2018/?tab=election-data)

b)Candidate information section:
[https://www.fec.gov/data/elections/senate/VA/2018/?tab=about-election](https://www.fec.gov/data/elections/senate/VA/2018/?tab=about-election)

- Resolves Issue# [https://github.com/fecgov/openFEC/issues/3254](https://github.com/fecgov/openFEC/issues/3254)

## Include a summary of proposed changes
fec-cms need to be changed to enable the button and make event trigger work.

## How to test the changes locally
1)setup openFEC locally
check out openFEC feature branch: feature/refactor_elections_endpoint

set some variables:
export SQLA_CONN=dev
export access_key_id=
export secret_access_key=
export bucket=
export region=

2) setup fec-cms locally
checkout fec-cms feature branch:  feature/enable_cft_export_button
and point to your local api:
export FEC_WEB_API_KEY_PUBLIC=xxxxxxx
export FEC_WEB_API_KEY=xxxxxxx
export FEC_CMS_ENVIRONMENT=LOCAL
export FEC_API_URL=http://localhost:5000
export DATABASE_URL=postgresql://:@/cfdm_cms_test

3)start redis-server
redis-server

4)Setup Celery Worker and start
export SQLA_CONN=xxxxxxx
export access_key_id=
export secret_access_key=
export bucket=
export region=
celery worker --app webservices.tasks --loglevel INFO

5)test:
example1 url: [http://127.0.0.1:8000/data/elections/senate/VA/2018/?tab=election-data](http://127.0.0.1:8000/data/elections/senate/VA/2018/?tab=election-data)
Click "Export" on Candidate financial totals section, download the file.
screen shot 2018-07-19 at 10 36 00 am

example2 url  [http://127.0.0.1:8000/data/elections/senate/VA/2018/?tab=about-election](http://127.0.0.1:8000/data/elections/senate/VA/2018/?tab=about-election)
Click "Export" on Candidate information section, download the file.
screen shot 2018-07-19 at 10 36 20 am

PR #3303, opened 2018-07-24
Fix elections endpoint to handle per_page=0

fec-jli:
Summary:
when fec-cms pass per_page=0 (means: get all results, don't do pagination),  
the elections endpoint not handle this case, 
so set cap=0 to avoid doing pagination.

related PR: [https://github.com/fecgov/openFEC/pull/3292](https://github.com/fecgov/openFEC/pull/3292)

How to test locally:
1) checkout branch : feature/fix_elections_per_page
2)DB--dev
3)loach cms point to local api
4)test elections summary pages to make sure all data sets result correctly.

### DIFF
--- a/webservices/resources/elections.py
+++ b/webservices/resources/elections.py
@@ -7,6 +7,7 @@ from webservices import docs
 from webservices import utils
 from webservices import filters
 from webservices import schemas
+from webservices.common import counts
 from webservices.utils import use_kwargs
 from webservices.common import models
 from webservices.common.views import ApiResource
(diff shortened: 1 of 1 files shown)


webservices/resources/elections.py
```diff
@@ -128,6 +129,19 @@ def args(self):
             ),
         )
 
+    def get(self, *args, **kwargs):
+        query = self.build_query(*args, **kwargs)
+        count = counts.count_estimate(query, models.db.session, threshold=500000)
```
pkfec: Any reason why the threshold value is set to high value?

Issue #3270, opened 2018-07-09
Intermittent page not found errors from AO search


rjayasekera:
This issue originally reported by NewRelic alert system.  NewRelic is pinging the URL (https://www.fec.gov/data/legal/advisory-opinions/2018-08/) every 30 min from different part of the country.  Occasionally  `AO 2018-08`  is not returned therefore we are presented with `Page not found` message screen.  We can reproduce this error if you access the above URL right after you received the alert email.  If you refresh the page couple of times page will show up.

![image](https://user-images.githubusercontent.com/20324717/42458079-25cb68ae-8367-11e8-9e03-3e52359d24c1.png)

![image](https://user-images.githubusercontent.com/20324717/42458119-3fcb4d82-8367-11e8-8b7b-ec1f54e0f466.png)

This error also can be found in Kibana

July 8th 2018, 21:24:30.059   - 404 GET /data/legal/advisory-opinions/2018-08/ (10253.221 ms)
July 8th 2018, 21:24:30.056  -  404 GET /data/legal/advisory-opinions/2018-08/ (10239.801 ms)
July 8th 2018, 20:42:33.537 - 404 GET /data/legal/advisory-opinions/2018-08/ (10259.113 ms)
July 7th 2018, 23:25:03.407  -  404 GET /data/legal/advisory-opinions/2018-01/ (10234.212 ms)
July 7th 2018, 20:03:54.124  -  404 GET /data/legal/advisory-opinions/2018-01/ (10255.1 ms)

vrajmohan:
I have been able to isolate the issue to the API layer and reproduce it on `dev` and `stage` as well with:
```
curl -v "https://fec-dev-api.app.cloud.gov/v1/legal/docs/advisory_opinions/2016-08"
curl -v "https://api-stage.open.fec.gov/v1/legal/docs/advisory_opinions/2016-08?API_KEY=DEMO_KEY"

```

vrajmohan:
With the removal of S3 caching from `dev` and `stage`, we have better visibility into the root cause of the error.  The API fails with:
```
[...]

PR #3304, opened 2018-07-25
feature/test-es-connection-timeouts

pkfec:
Deploying a hot fix to address: https://github.com/fecgov/openFEC/issues/3270

1. ElasticSearch connection is set to retry on timeout. timeouts are set as below:
    - Elasticsearch(url, timeout=5, max_retries=5, retry_on_timeout=True)

2. Added a new locust task for advisory_opinions. 
    - This task is specific to AO's. We can run locust tests on advisory_opinions from now on. 

3. Renamed existing locust MUR task.
    - Made this task specific to MUR.

### DIFF
--- a/locustfile.py
+++ b/locustfile.py
@@ -194,11 +194,17 @@ class Tasks(locust.TaskSet):
         self.client.get('legal/search', name='legal_search', params=params)
 
     @locust.task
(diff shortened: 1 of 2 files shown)


pkfec:
@fec-jli @lbeaufort  Can you please review?

PR #3311, opened 2018-07-27
Fixed dead SQL links in API docs

### LINKED ISSUE #3296: Dead link in Documentation needs to be fixed
Our documentation at https://api.open.fec.gov/developers/#/receipts/get_schedules_schedule_a has a dead link to a SQL function for defining individual contributions (https://github.com/fecgov/openFEC/blob/develop/data/functions/individual.sql)..

Got more clarification from @lbeaufort about where these `/data/function` sql links should go now that we moved over to the migrations method. See below:

- Fix dead links:
- [x] Individual contributions should go to https://github.com/fecgov/openFEC/blob/master/data/migrations/V0003__disclosure_schema_functions.sql#L192-L218
- [x] Schedule B by purpose should go to https://github.com/fecgov/openFEC/blob/master/data/migrations/V0003__disclosure_schema_functions.sql#L34-L77

patphongs:
## Summary

- Resolves #3296

_There were 2 dead links to sql functions in the API documentation. Fixed link for individual contributions sql function and purpose by disbursement sql function._

- Look at webservices/docs.py to make sure dead links from individual contributions sql function and purpose by disbursement sql now go to correct URL.

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -296,6 +296,7 @@ or `/totals` endpoints.
 When comparing the totals from reports to line items. the totals will not match unless you
 only look at items where `"is_individual":true` since the same transaction is in the data
 multiple ways to explain the way it may move though different committees as an earmark.
+See the `is_individual` sql function within the migrations for more details.
 
 For the Schedule A aggregates, such as by_occupation and by_state, include only unique individual
 contributions. See below for full methodology.
@@ -339,12 +340,11 @@ Line number with description
 
 Of those transactions,[under $200, and having "earmark" in the memo text OR transactions having the codes 11A, 12, 17, 17A, or 18], we then want to exclude earmarks.
 
-This is [the sql function](https://github.com/fecgov/openFEC/blob/develop/data/functions/individual.sql) that defines individual contributions:
 '''
 
 SCHEDULE_A = SCHEDULE_A_TAG + '''
-The data is divided in two-year periods, called `two_year_transaction_period`, which
-is derived from the `contribution_receipt_date`. If no value is supplied, the results
+All receipt data is divided in two-year periods, called `two_year_transaction_period`, which
+is derived from the `report_year` submitted of the corresponding form. If no value is supplied, the results
 will default to the most recent two-year period that is named after the ending,
 even-numbered year.
 
@@ -388,7 +388,7 @@ reported as part of forms F3, F3X and F3P.
 
 SCHEDULE_B = SCHEDULE_B_TAG + '''
 The data is divided in two-year periods, called `two_year_transaction_period`, which
-is derived from the `disbursement_date`. If no value is supplied, the results will
+is derived from the `report_year` submitted of the corresponding form. If no value is supplied, the results will
 default to the most recent two-year period that is named after the ending,
 even-numbered year.
 
@@ -421,8 +421,8 @@ large result sets are approximate; you will want to page through the records unt
 '''
 
 SCHEDULE_B_BY_PURPOSE = '''
-Schedule B receipts aggregated by disbursement purpose category. To avoid double counting, memoed items are not included.
-Purpose is a combination of transaction codes, category codes and disbursement description.  See [the sql function](https://github.com/fecgov/openFEC/blob/7d2c058706f1b385b2cc18d75eb3ad0a1fba9d52/data/functions/purpose.sql)
+Schedule B disbursements aggregated by disbursement purpose category. To avoid double counting, memoed items are not included.
+Purpose is a combination of transaction codes, category codes and disbursement description. See the `disbursement_purpose` sql function within the migrations for more details.
 '''
 
 SCHEDULE_C_TAG = '''


patphongs:
> I am not sure if it is reasonable to link to the migration files since the flyway migration is a history line that keeps changing. We are not saving a fix final copy of the database code like other objects such as Python code. For example, V0003 (actually anything prior to V0041) is a "snapshot" that Varaj took as the baseline. The particular function you were mentioning (public.ofec_sched_b_update_aggregate_purpose) had been dropped in a later version of migration file. So maybe a general description instead of links to the actual DB code is a more generic solution.

@lbeaufort what do you think of @fecjjeng request to leave it as a general description instead of adding a link?

lbeaufort:
@patphongs I think a description would probably be more useful. We could also include the name of the function so people could look for it.

patphongs:
@lbeaufort @fecjjeng I made some slight modifications to this PR to remove the sql function link and added the function names to the documentations in an attempt to make this more generic. If you could please review and let me know if you have any updates.

lbeaufort:
@fecjjeng did you say that the `is_individual` and `disbursement_purpose` logic is on the Oracle side now? So if it changes, there isn't anywhere we can point people to the logic?

fecjjeng:
@lbeaufort is_individual for populating schedule_a and disbursement_purpose for populating schedule_b are still in postgresql side, and they are still being used to populate the corresponding columns in these tables.  It is the public.ofec_sched_b_update_aggregate_purpose function that populate the aggregate table that had been dropped.

fecjjeng:
Replace the dead link to the sql file with a general description for  is_individual and disbursement_purpose is correct.  It does fix the dead link problem while providing the readers the function name if they would like to read.  Good job.

While reading the doc.py, I have a question:  
SCHEDULE_A = SCHEDULE_A_TAG + '''
Individual contributions data is divided in two-year periods, called `two_year_transaction_period`, which
is derived from the `contribution_receipt_date`. If no value is supplied, the results
will default to the most recent two-year period that is named after the ending,
even-numbered year. 
My understanding is that the two_year_transaction_period is from the report_year submitted of the corresponding form, not the `contribution_receipt_date`.
Same case in SCHEDULE_B = SCHEDULE_B_TAG + '''

webservices/docs.py
```diff
@@ -339,14 +339,13 @@
...
-This is [the sql function](https://github.com/fecgov/openFEC/blob/develop/data/functions/individual.sql) that defines individual contributions:
 '''
 
 SCHEDULE_A = SCHEDULE_A_TAG + '''
-The data is divided in two-year periods, called `two_year_transaction_period`, which
+Individual contributions data is divided in two-year periods, called `two_year_transaction_period`, which
 is derived from the `contribution_receipt_date`. If no value is supplied, the results
 will default to the most recent two-year period that is named after the ending,
-even-numbered year.
+even-numbered year. See the `is_individual` sql function within the migrations for more details.
```
lbeaufort: I think this would make more sense on line 299.

patphongs:
@lbeaufort I have moved the `is_individual` reference to the /schedules/schedule_a doc string as requested. Could you please review again and merge if you think it looks good?

lbeaufort:
@patphongs thanks! Can you consolidate your commits to one commit please?

webservices/docs.py
```diff
@@ -339,11 +340,10 @@
 
 Of those transactions,[under $200, and having "earmark" in the memo text OR transactions having the codes 11A, 12, 17, 17A, or 18], we then want to exclude earmarks.
 
-This is [the sql function](https://github.com/fecgov/openFEC/blob/develop/data/functions/individual.sql) that defines individual contributions:
 '''
 
 SCHEDULE_A = SCHEDULE_A_TAG + '''
-The data is divided in two-year periods, called `two_year_transaction_period`, which
+Individual contributions data is divided in two-year periods, called `two_year_transaction_period`, which
```
lbeaufort: @patphongs sorry I missed this before - Schedule A isn't just for individual contributions - it's all "receipts"

webservices/docs.py
```diff
@@ -422,7 +422,7 @@
 
 SCHEDULE_B_BY_PURPOSE = '''
 Schedule B receipts aggregated by disbursement purpose category. To avoid double counting, memoed items are not included.
```
lbeaufort: This should say "Schedule B disbursements" or "Schedule B expenditures" or "Schedule B spending data" but they're not receipts.

fecjjeng:
@patphongs  As mentioned in my comment above, in line 347
Individual contributions data is divided in two-year periods, called two_year_transaction_period, which
is derived from the **contribution_receipt_date**. If no value is supplied, the results
will default to the most recent two-year period that is named after the ending,
even-numbered year.
My understanding is that the two_year_transaction_period is from the report_year submitted of the corresponding form, not the contribution_receipt_date.
This is under
SCHEDULE_A = SCHEDULE_A_TAG + '''
Same case in SCHEDULE_B = SCHEDULE_B_TAG + ''' as well

Since this ticket is to address the dead-link problem and the solution does fix the problem, I have no problem we merge the PR.  Do you want to open a separate ticket for the source of contribution_receipt_date?

PR #3322, opened 2018-08-01
Add endpoint-specific caching

### LINKED ISSUE #3237: Set caching headers in the API
Do a single endpoint this sprint as a test.

We removed the caching headers set by the API entirely when trying to ensure that refreshed legal documents show up promptly. Consequently, we are not leveraging any caching by api.data.gov, and API users do not benefit from any caching. As discussed in https://docs.google.com/document/d/1vUDTrnWKpbUDmUv90pEIINV79MSlkg03qoBoXpzl9Wc/edit?usp=sharing, we should reinstate caching - the cache duration may be different for legal (no cache?), efiling (10m?) and the rest (1h?). 

In addition, we should also fix https://github.com/fecgov/openFEC/blob/develop/webservices/rest.py#L157 as it gives the false impression that we are caching.

- [x] Reinstate caching - the cache duration may be different for legal (no cache?), efiling (10m?) and the rest (1h?). 
- [x] Remove `if app.config['MAX_CACHE_AGE'] is not None:`[here](https://github.com/fecgov/openFEC/blob/develop/webservices/rest.py#L157)

**Cache duration**
efiling: none
legal: 5 minutes
calendar: 5 minutes per @AmyKort 
other: 1 hour

lbeaufort:
## Summary (required)

- Resolves #3237 

_Add endpoint-specific caching: Default is 1 hour, legal and calendar are 5 minutes, efiling is none._

Run this branch locally and look at the response headers using `CURL`:
- default: `curl -I http://localhost:5000/` 1 hour = 3600s
- legal: `curl -I "http://localhost:5000/v1/legal/search/?description=excessive&ao_is_pending=True&type=advisory_opinions&api_key=DEMO_KEY"` 5 min = 300s
- calendar: `curl -I http://localhost:5000/v1/calendar-dates/` 5 min = 300s
- efile: `curl -I http://localhost:5000/v1/schedules/schedule_a/efile/?api_key=DEMO_KEY` = 0

-  API caching

## Related PRs

Remove S3 caching #3235

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -152,14 +152,31 @@ def limit_remote_addr():
                 abort(403)
 
 
+def get_cache_duration(url):
+
+    EFILING_CACHE = 0
+    LEGAL_CACHE = 60 * 5
+    CALENDAR_CACHE = 60 * 5
+    DEFAULT_CACHE = 60 * 60
+
+    if '/efile/' in url:
+        return EFILING_CACHE
+    elif '/calendar-dates/' in url:
+        return CALENDAR_CACHE
+    elif '/legal/' in url:
+        return LEGAL_CACHE
+
+    return DEFAULT_CACHE
+
+
 @app.after_request
 def add_caching_headers(response):
-    if app.config['MAX_CACHE_AGE'] is not None:
-        response.headers.add(
-            'Cache-Control',
-            'public, max-age={}'.format(app.config['MAX_CACHE_AGE'])
-        )
 
+    cache_duration = get_cache_duration(request.path)
+    response.headers.add(
+        'Cache-Control',
+        'public, max-age={}'.format(cache_duration)
+    )
     return response
 
 


webservices/rest.py
```diff
@@ -152,14 +152,29 @@ def limit_remote_addr():
                 abort(403)
 
 
+def get_cache_duration(url):
+
+    EFILING_CACHE = None
+    LEGAL_CACHE = 60 * 5
+    DEFAULT_CACHE = 60 * 60
+
+    if 'efile' in url:
```
vrajmohan: It may be better to use '/efile/' to ensure that we don't have false positives.

webservices/rest.py
```diff
@@ -152,14 +152,29 @@ def limit_remote_addr():
...
+        return EFILING_CACHE
+    if 'legal' in url:
```
vrajmohan: `elif` would be better here.

fec-jli:
It looks great.
one question: Can you explain the means for those [1]..[2]..[3]  and [1]-,[3]+...? Thanks
screen shot 2018-08-02 at 8 31 02 pm

lbeaufort:
Thanks @fec-jli! I think `CURL` is interpreting things after `&` as new arguments - it needs quotes to avoid the [1][2][3]:
`curl -I "http://localhost:5000/v1/legal/search/?description=excessive&ao_is_pending=True&type=advisory_opinions&api_key=DEMO_KEY"` I'll also update the instructions.

vrajmohan:
Minor correction. It is not `curl` that is interpreting the things after `&` as new arguments. It is the shell. `curl` only sees the argument before the `&` when it is not enclosed in quotes. The shell thinks that you want to run the `curl` command in the background (because of the "trailing" `&`) and then assumes that what follows is _another_ command to run.

This can be seen by running
```
echo foo&bar=3
echo "foo&bar=3"
```

PR #3349, opened 2018-08-29
[Close without merge]Add new spender committee type filter in ScheduleBView.

fec-jli:
## Summary (required)
1)due to `spender committee type` field is not in fec_fitem_sched_b table. we have to create sub-object using spender_committee_id to get committee history data from ofec_committee_history_mv. 

2)The filter function of endpoint works well. but have some performance issues for a few committee_type.
     a) For those committee types which not exist in fec_fitem_sched_b-b table, result never return back.
    short-term solution: create hardcode array `FILER_COMMITTEE_TYPES` to constrain those committee 
     types.

     on front-end, add dynamic dropdown list for committee_type , the list will be changed dynamically when `Transaction Time Period` chose different option.
screen shot 2018-08-29 at 3 54 11 pm

3)for cycle=2018 and committee_type=W, the performance is still not good. need think about later.

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -214,8 +214,14 @@ candidate_detail = {
 candidate_list = {
     'q': fields.List(fields.Str, description=docs.CANDIDATE_NAME),
     'candidate_id': fields.List(IStr, description=docs.CANDIDATE_ID),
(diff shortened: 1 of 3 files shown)


fec-jli:
We like to keep the code in this PR and branch for future reference. for how to handle sub-object using eagle load. so close this one now, and open another one to Add new spender committee type filter in ScheduleBView.

PR #3368, opened 2018-09-08
Feature/refactor ao mur test schema

vrajmohan:


tests/common.py
```diff
@@ -107,7 +107,16 @@ def _results(self, qry):
         response = self._response(qry)
         return response['results']
 
-
 def assert_dicts_subset(first, second):
     expected = {key: first.get(key) for key in second}
     assert expected == second
+
+def get_test_jdbc_url():
```
lbeaufort: `tests/common.py` is a better place for this function, thanks!

PR #3387, opened 2018-09-17
Add Swagger docs for legal API

### LINKED ISSUE #3377: Add /legal/ to API developer docs
So that API users can access our legal resources API, developers can test the API more easily, and front-end developers can design the front-end for new legal features more easily, we should add the /legal/ endpoint to the API developer documentation here: https://api.open.fec.gov/developers/

We've run into issues with `flask-apispec` because legal resources don't have a `model` file (they're stored in `elasticsearch`, not `postgres`)

Old [WIP] PR: https://github.com/fecgov/openFEC/pull/3074

vrajmohan:
Resolves #3377

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -1226,3 +1226,9 @@ The Operations log contains details of each report loaded into the database. It 
 used as status check to determine when all of the data processes, from initial entry through
 review are complete.
 '''
+LEGAL = '''
+Explore relevant statutes, regulations and Commission actions.
+'''
+LEGAL_SEARCH = '''
+Search legal documents by type, or across all document types using keywords, parameter values and ranges.
+'''
--- a/webservices/legal_docs/responses.py
+++ b/webservices/legal_docs/responses.py
@@ -1,0 +1,709 @@
+import copy
+
+MUR_ADR_COMMON_SCHEMA = {
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "open_date": {
+                "type": "string",
+                "format": "date"
+            },
+            "close_date": {
+                "type": "string",
+                "format": "date"
+            },
+            "no": {
+                "type": "string"
+            },
+            "name": {
+                "type": "string"
+            },
+            "election_cycles": {
+                "type": "integer"
+            },
+            "documents": {
+                "type": "array",
+                "items": {
+                    "type": "object",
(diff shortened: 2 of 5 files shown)


webservices/docs.py
```diff
@@ -1225,4 +1225,10 @@ def add_ytd(var):
 The Operations log contains details of each report loaded into the database. It is primarily
 used as status check to determine when all of the data processes, from initial entry through
 review are complete.
-'''
\ No newline at end of file
+'''
+LEGAL = '''
+Explore relevant statutes, regulations, Commission actions and court cases.
```
lbeaufort: Can users see court cases in /legal/?

webservices/legal_docs/responses.py
```diff
@@ -0,0 +1,709 @@
...
+                            "  is prohibited by law.  6. Pitts sought  to conceal the  <em>embezzlement</em> by under",
+                            "  execute  his various  <em>embezzlements</em>. That  month, Mr.  Pitts also  transferred",
+                            " that  was used to carry out his  <em>embezzlement</em> scheme.  Pitts also  did not  ",
+                            " the  <em>embezzlement</em> by under-reporting the  transfer of payroll  deductions from"
+                        ],
+                        "document_highlights": {}
+                    }
+                ],
+                "total_murs": 2,
+                "admin_fines": [],
```
lbeaufort: Can we take out these empty `admin_fines` and `adrs` lists, now that the examples have been fleshed out above?

webservices/docs.py
vrajmohan: I cribbed it from 
screen shot 2018-09-18 at 4 11 02 pm

Should I drop court cases?

webservices/legal_docs/responses.py
vrajmohan: Did you mean take them out, or provide examples of them _if_ the query did return ADRs and Admin fines?

webservices/docs.py
lbeaufort: It looks like court cases aren't in the legal search, so I think so - otherwise people might expect them to be there.

lbeaufort: Looks like https://www.fec.gov/legal-resources/court-cases/ is a hard-coded or wagtail page, and the "browse" option takes them to classic.fec.gov

PR #3392, opened 2018-09-19
Fix "most recent" logic for Forms 1, 2, and 1M

### LINKED ISSUE #3096: /filings/ endpoint marking more than one `most_recent` Form 1 (Statement of Organization)
- [ ] Add a testing ticket

So that committees and the public can see the most recent version of the Form 1 (Statement of Organization), we need to update the logic to handle when a committee files multiple Form 1's as new and only show the newest as "Current Version" and the others as "Past Version."

@patphongs discovered this issue.

Example: (https://www.fec.gov/data/committee/C00639013/?tab=filings)

8-31-2017 3-35-01 pm

Background: Using FECFile and other 3rd party vendors, committees can amend their Statement of Organization (Form 1) by filing it as new. Then, an overnight process updates it to show as amended. As a result, there is no amendment number associated with the file that was submitted as new and now displays as amended.

Potential fix: Add to /filings/ "amendment chain" logic to handle multiple F1Ns for the same committee.  If the amendment chain logic fails (>1 current version for the same filing type - for statements of org. ) then use the most recently filed one as the current version and mark the others as past version.

- [X] F2 also has the same issue? Find an example to confirm
API example
https://fec-dev-api.app.cloud.gov/v1/filings/?api_key=DEMO_KEY&sort_hide_null=false&sort=-receipt_date&per_page=10&page=1&candidate_id=H8CA05035&form_type=F99&form_type=RFAI&request_type=-1&request_type=-2&request_type=-3&request_type=-4&request_type=-5&request_type=-6&request_type=-7&request_type=-8&request_type=-9
Other front-end example:
https://fec-dev-proxy.app.cloud.gov/data/candidate/H8WI01024/?tab=filings
~- [ ] Run the PGtap amendment chain test (#3268)~ PGTap test is for efiling amendment_chain
- [X] Write a new python test that fails for current behavior
- [X] modify F1 (and F2?) function logic for amendments -- create special case for committees that file multiple  Form 1 "New"s.

Existing amendment chain view is complicated and not a good place to fix the logic for F1's and F2's. We should investigate where else we can fix the logic upstream.

lbeaufort:
## Summary (required)

- Resolves #3096 

**There should be only one `most_recent` F1 per committee and one F2 for candidates**

- Create a view for every materialized view that depends on `ofec_filings_amendments_all_vw`
- Modify amendment_chain logic for Forms 1, 1M, and 2
- Ensure single current version for F1, F1M and F2
- Add integration test for filing amendments, tests for amended F1s and F3s

committee filings page: '/committee/<committee_id>/filings/' > `filings.FilingsView` > `BaseFilings`
candidate filings page: '/filings/' > `filings.FilingsList` > `BaseFilings`

`BaseFilings` >  `ofec_filings_all_mv` > `ofec_filings_amendments_all_mv` > `disclosure.f_rpt_or_form_sub`

- Drop/Create `cfdm_test`/`invoke create_sample_db`
- `unset SQLA_CONN` and run your local api
- `unset FEC_API_URL` and run your local cms

Testing links: 
API
http://localhost:5000/v1/filings/?per_page=100
http://localhost:5000/v1/filings/?most_recent=true&per_page=100&form_type=F1
http://localhost:5000/v1/committee/C00559492/filings/?per_page=100
http://localhost:5000/v1/filings/?per_page=100&form_type=F2 (note that there isn't much F2 test data)
CMS
http://localhost:8000/data/committee/C00559492/?tab=filings&cycle=2016
http://localhost:8000/data/committee/C00559492/?tab=filings&cycle=2014

- Candidate filings tab
- Committee filings tab
- Processed filings in data tables


tests/integration/test_amendment_chain.py
```diff
@@ -0,0 +1,294 @@
...
+            'report_type': 'Q2',
+            'form_type': 'F3'
+        }
+        form_3_q2_amend_1 = {
+            'committee_id': 'C006',
+            'report_year': 2018,
+            'amendment_indicator': 'A',
+            'receipt_date': datetime.date(2018, 8, 15),
+            'file_number': 20002,
+            'amendment_chain': [20001, 20002],
```
vrajmohan: As we are mimicking the `efiling_amendment_chain_vw` logic, this field will have the entire longest chain, not the _longest chain so far_. This is different from the current behavior.

lbeaufort: @vrajmohan I didn't find that to be the case - check out `1170005` here: https://fec-dev-api.app.cloud.gov/v1/committee/C00330894/filings/?api_key=5yyI90SU3Xb73TVlv4wrEhQxYcCwMWCywQiGdYbJ&sort_hide_null=false&sort=-coverage_end_date&sort=report_type_full&sort=-beginning_image_number&per_page=100&page=1&form_type=F3&form_type=F3X&form_type=F3P&form_type=F3L&form_type=F4&form_type=F5&form_type=F7&form_type=F13&form_type=RFAI&report_type=-24&report_type=-48&request_type=-1&request_type=-5&request_type=-6&request_type=-9&cycle=2018

lbeaufort: Sorry if I misunderstood - do you mean, we should change the current behavior to show the longest chain so far?

lbeaufort: Looks like the front-end uses `amendment_chain` to generate the amendment number (which would also explain why it's wrong so often) https://github.com/fecgov/fec-cms/blob/develop/fec/fec/static/js/modules/helpers.js#L395-L418

vrajmohan: Actually there is a bigger question. When we force the most recent file number to be the highest file number for F1, F1M and F2, should we (do we need to) "fix" the `amendment_chain` as well? If yes, how?

lbeaufort: I think that we'd want to take the file ID's in descending order, and stop at the current file ID, if that's not too hard to implement.

lbeaufort: After discussing it with @PaulClark2, the plan forward is to make the F1/2/1M amendment chain logic match the existing F3 behavior for /filings/ so we don't have to change the front-end logic or crack open the F3 logic. I'll put in an issue to reconcile the /efile/filings amendment chain logic (goes all the way through the most recent filing) and the /filings/ amendment chain logic (stops at the filing you're looking at).

data/migrations/V0108__fix_processed_amendment_chain.sql
```diff
@@ -0,0 +1,3247 @@
+CREATE OR REPLACE VIEW ofec_processed_filing_vw AS
```
vrajmohan: It may be better to make this view specific for all-but-forms-1-and-2 (is there a better name) and create a similar view for forms-1-and-2 and use the logic from lines 76-95. The materialized view can then be a simple union of these 2 views, making a clear separation between complex computations and collation.

lbeaufort: @vrajmohan that seems like a good plan. I'll be in training today through Thursday and would be able to make the change by Friday at the latest, but I don't mind if you prefer to get the change in faster.

lbeaufort:
👍 @vrajmohan this looks good to merge to me. I messaged @pkfec about migration numbers.

PR #3401, opened 2018-09-24
Use cache key prefix to fix repo cache build errors in Stage

patphongs:
## Summary

Fixes cached build errors in openFEC repo before deploy tonight.


patphongs:
re-did with cherry picking develop changes

PR #3403, opened 2018-09-24
Update Flask and Flask-related packages.

qqss88:
## Summary (required)

- Resolves # [_Issue number_]

_Include a summary of proposed changes._

## How to test the changes locally

-

## Impacted areas of the application
List general components of the application that this PR will affect:

those 4 packages have been updated to the newest version.
Flask==1.02
Flask-Cors==3.0.6
Flask-Script==2.0.6
Flask-RESTful==0.3.6

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

PR #3419, opened 2018-10-04
Disables strict slashes in API, add 404 error handler

patphongs:
## Summary

- Related #3403 

_The latest upgrade of flask was causing requests from the CMS to 500 in the API due to lack of a trailing slash from the CMS request call. To fix this, this PR disables `strict_slashes` so that both types of requests will work._

Solution from: https://stackoverflow.com/questions/33241050/trailing-slash-triggers-404-in-flask-path-rule

- Run local API and make sure that both of these endpoint sill render (notice one has a trailing slash after history and the other does not):
http://localhost:5000/v1/candidate/H8GA06195/history?api_key=DEMO_KEY
http://localhost:5000/v1/candidate/H8GA06195/history/?api_key=DEMO_KEY
- Run local CMS environment and try to render a candidate or committee page based on these changes:
http://localhost:8000/data/candidate/H8GA06195/
http://localhost:8000/data/committee/C00657353/

-  Flask slash settings

## Related PRs

qqss88:feat-3371 | #3403

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -64,6 +64,7 @@ def initialize_newrelic():
 initialize_newrelic()
 
 app = Flask(__name__)
+app.url_map.strict_slashes = False
 
 
 def sqla_conn_string():
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -192,7 +192,12 @@ def handle_exception(exception):
         )
     )
     raise exceptions.ApiError('Could not process the request',
-            status_code=http.client.NOT_FOUND)
+        status_code=http.client.NOT_FOUND)
+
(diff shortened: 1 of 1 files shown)


lbeaufort:
@patphongs this looks great! As we discussed, in testing I found that bad routes and favicon.ico requests were throwing exceptions and 500 errors rather than 404's. I pushed a commit that adds a custom 404 error handler per http://flask.pocoo.org/docs/0.12/patterns/errorpages/#error-handlers

patphongs:
@lbeaufort I have reviewed your 404 error handler updates and they work well. Thanks!

PR #3465, opened 2018-10-31
Add handling for include/exclude AO citations

### LINKED ISSUE #3461: AOs: Short-term fix for AO missed/false positive citations
What we're after:
So that we can quickly update AO citations that are missed or accidentally included by the AO loading process, we should have a list-like thing with "extra" citations that aren't picked up by the REGEX due to bad OCR text or unusual formats.

- [x] Design python data structure(s) `lookup` to hold this info - maybe AO #, citation, date added, added by, exclude/include ....anything else relevant. (Separate include/exclude lookup?)
- [x] Modify the AO reload management task to also check this `lookup` for citations missed excluded ( 1 - remove false positives 2 - add missed)

See https://github.com/fecgov/openFEC/issues/3393 as an example

See also, past research on false positives: 
https://docs.google.com/spreadsheets/d/1-nf-oAPtw3LisZInuDHcQ7JdpyEQXRISZ4kY1ONrFcE/edit?usp=drive_web&ouid=114351900669501389307

https://docs.google.com/spreadsheets/d/161cnTtYzXzXW2iNxZPIX9rYhko2ZV2deHYFNDP-R49o/edit#gid=0

For a more elegant long-term solution, see #3302

lbeaufort:
## Summary (required)

- Resolves #3461

Add handling for manually including/excluding AO citations (AO, statute, regulation)

- Uncomment `logger.setLevel(logging.DEBUG)` in `advisory_opinions.py`, get local elasticsearch set up, and `./manage.py load_advisory_opinions -f '1999-40'` and `2017-03` (you can cancel after each one loads). 

- AO '2017-03' shouldn't cite: ['2010-11', '2011-12', '2015-16']
- AO '1999-40' should cite (11, 110, 3)

-  Advisory opinion AO, statute, regulatory citations

### DIFF
--- a/tests/test_ao_citations.py
+++ b/tests/test_ao_citations.py
@@ -117,3 +117,5 @@ def test_validate_regulation_citation(title, part, expected):
 ])
 def test_parse_regulatory_citations(text, expected):
     assert parse_regulatory_citations(text) == expected
+
+# Test exclude/include
--- a/tests/test_ao_citations.py
+++ b/tests/test_ao_citations.py
@@ -123,13 +123,12 @@ def test_parse_regulatory_citations(text, expected):
 class TestManualCitations(unittest.TestCase):
     def test_exclude_citations(self):
         original_ao_citations = set(['2011-12', '2017-03', '2014-18', '2010-11', '2004-41', '2007-13', '1996-26', '2005-17', '2017-01', '2016-02', '2012-21', '2012-23', '2014-11', '2002-15', '2006-12', '2015-16', '2014-21'])
-        expected_citations = set(['2017-03', '2014-18', '2004-41', '2007-13', '1996-26', '2005-17', '2017-01', '2016-02', '2012-21', '2012-23', '2014-11', '2002-15', '2006-12', '2014-21'])
-
         fixed_citations = fix_citations('2017-03', 'ao', original_ao_citations)
-        assert fixed_citations == expected_citations
+        assert '2010-11' not in fixed_citations
+        assert '2011-12' not in fixed_citations
+        assert '2015-16' not in fixed_citations
 
     def test_include_citations(self):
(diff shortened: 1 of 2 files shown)


tests/test_ao_citations.py
```diff
@@ -117,3 +118,18 @@ def test_validate_regulation_citation(title, part, expected):
 ])
 def test_parse_regulatory_citations(text, expected):
     assert parse_regulatory_citations(text) == expected
+
+
+class TestManualCitations(unittest.TestCase):
+    def test_exclude_citations(self):
+        original_ao_citations = set(['2011-12', '2017-03', '2014-18', '2010-11', '2004-41', '2007-13', '1996-26', '2005-17', '2017-01', '2016-02', '2012-21', '2012-23', '2014-11', '2002-15', '2006-12', '2015-16', '2014-21'])
```
vrajmohan: The only comment I would make is that the test data is much shorter so that it is easy to tell at a glance the difference between the original and expected citations.

tests/test_ao_citations.py
```diff
@@ -117,3 +118,18 @@ def test_validate_regulation_citation(title, part, expected):
...
+        expected_citations = set(['2017-03', '2014-18', '2004-41', '2007-13', '1996-26', '2005-17', '2017-01', '2016-02', '2012-21', '2012-23', '2014-11', '2002-15', '2006-12', '2014-21'])
+
+        fixed_citations = fix_citations('2017-03', 'ao', original_ao_citations)
+        assert fixed_citations == expected_citations
+
+    def test_include_citations(self):
+        original_reg_citations = set([(11, 110, 4), (11, 110, 1), (11, 102, 9), (11, 102, 6), (11, 114, 5), (11, 100, 8), (11, 114, 1), (11, 100, 5), (11, 103, 3), (11, 104, 14), (11, 100, 6), (11, 102, 8), (11, 114, 7)])
```
vrajmohan: ditto

AmyKort:
Thank you @vrajmohan !

PaulClark2:
Thanks, @vrajmohan

tests/test_ao_citations.py
lbeaufort: Thanks so much for taking a look! I had it that way before and couldn’t make up my mind - I agree that it was much easier to read. I will change it back :)

Issue #3390, opened 2018-09-19
Example 1: Disbursements missing from /schedule_b/by_recipient/


lbeaufort:
Some disbursements are missing from /schedule_b/by_recipient/.

From an API user:

>There are a few cases where the vendor names I’m searching with yield no results through the API but yield results through the manual search on the FEC website.

Example 1: 
First, considering the vendor DRL Publications

The following request, which is trying to get a list of every committee that spent money on DRL Publications yields no results:
https://api.open.fec.gov/v1/schedules/schedule_b/by_recipient/?api_key=DEMO_KEY&cycle=2016&per_page=20&recipient_name=DRL%20PUBLICATIONS&page=1

However, the same search on /schedule_b/ and the front end returns one result:
https://api.open.fec.gov/v1/schedules/schedule_b/?api_key=DEMO_KEY&sort_hide_null=true&data_type=processed&recipient_name=drl+publications&two_year_transaction_period=2016&min_date=01%2F01%2F2015&max_date=12%2F31%2F2016&sort=-disbursement_date&per_page=30

https://www.fec.gov/data/disbursements/?two_year_transaction_period=2016&data_type=processed&recipient_name=drl+publications&min_date=01%2F01%2F2015&max_date=12%2F31%2F2016

Example 2 has a different cause: https://github.com/fecgov/openFEC/issues/3429

- [x] Make necessary python changes after DB work is done: https://github.com/fecgov/openFEC/issues/3431
- [ ] Notify user of resolution (original email was sent to apiinfo@fec.gov)

lbeaufort:
The logic preventing this from appearing seems to be in `disclosure.dsc_sched_b_aggregate_recipient` and I'm looking into getting that logic (It's currently on the Oracle side). 

I'm guessing at least for example 1, we may be excluding memos. Example 2 is on line 29 - I want to see what other filters are on the aggregate table.

lbeaufort:
One possible solution @fecjjeng and I discussed is showing $0 totals for recipients that have only memoed transactions.

Example:
$100 Itemized:
Jean $50
Laura $50
Paul $50 X Memo

$100 By recipient aggregate:
Name, Amount, "Memo/Description"
Jean $50
Laura $50
Paul $0 Memo



logic for `disclosure.dsc_sched_b_aggregate_recipient`:
```
select cmte_id, rpt_yr + mod(rpt_yr, 2) as cycle,
[...]

PR #3481, opened 2018-11-07
Switch to new schedule B aggrgeate tables with memo totals

hcaofec:
## Summary (required)

- Resolves # [_3390_](https://github.com/fecgov/openFEC/issues/3390)

_Currently some disbursements are missing from schedule B aggregate endpoints.  Most of those disbursements are memoed items.  To fix this, totals and counts of memoed items are added to three aggregate tables_

-Test this link and API will return 1 result ( using DEV database ): 
http://127.0.0.1:5000/v1/schedules/schedule_b/by_recipient/?cycle=2016&per_page=20&sort_null_only=false&sort_hide_null=false&page=1&recipient_name=DRL%20Publications

current API returns 0:
https://api.open.fec.gov/v1/schedules/schedule_b/by_recipient/?api_key=DEMO_KEY&per_page=20&page=1&sort_null_only=false&cycle=2016&recipient_name=DRL%20Publications&sort_hide_null=false

Current disbursement aggregate pages only display non-memo total.  With endpoint returns memo-total now, we may need to think whether want to display memo total or not?
https://www.fec.gov/data/committee/C00575795/?tab=spending&cycle=2016

### DIFF
--- a/tests/test_aggregates.py
+++ b/tests/test_aggregates.py
@@ -13,6 +13,7 @@ from webservices.resources.aggregates import (
     ScheduleBByPurposeView,
     ScheduleBByRecipientView,
     ScheduleEByCandidateView,
+    ScheduleBByRecipientIDView,
 )
 from webservices.resources.candidate_aggregates import (
     ScheduleABySizeCandidateView,
@@ -111,6 +112,8 @@ class TestCommitteeAggregates(ApiBaseTest):
             'cycle': 2012,
             'total': aggregate.total,
             'count': aggregate.count,
+            'memo_total': aggregate.memo_total,
+            'memo_count': aggregate.memo_count,
         }
         self.assertEqual(results[0], expected)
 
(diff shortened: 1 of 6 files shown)


webservices/schemas.py
```diff
@@ -584,7 +584,7 @@ class CandidateHistoryTotalSchema(schemas['CandidateHistorySchema'],
         'recipient_name': ma.fields.Str(),
     },
     options={
-        'exclude': ('committee', 'recipient')
+        'exclude': ('idx', 'committee', 'recipient')
     },
 )
```
hcaofec: Will excluding `idx` from api returns break someone’s app or script using the current endpoint?
@lbeaufort @fec-jli

webservices/common/models/aggregates.py
```diff
@@ -45,15 +45,24 @@ class ScheduleAByOccupation(BaseAggregate):
     occupation = db.Column(db.String, primary_key=True, doc=docs.OCCUPATION)
 
 
-class ScheduleBByRecipient(BaseAggregate):
+class BaseDisbursementAggregate(BaseAggregate):
```
lbeaufort: This is a helpful abstraction, good call!

tests/test_aggregates.py
```diff
@@ -137,9 +140,42 @@ def test_disbursement_recipient(self):
...
+    def test_disbursement_recipient_id_total(self):
+        committee = factories.CommitteeHistoryFactory(cycle=2012)
+
+        aggregate = factories.ScheduleBByRecipientIDFactory(
+            committee_id=committee.committee_id,
+            cycle=committee.cycle,
+            recipient_id='C00507368',
+            total=4000,
+            count=2,
+            memo_total=0,
```
lbeaufort: it's great to add a new test here - would factory values other than 0 be more likely to catch a mistake?

PR #3484, opened 2018-11-08
fix dev branch requirements.txt file.

qqss88:
## Summary (required)
fix the dev branch requirements.txt file
- Resolves # [_Issue number_]
local testing issue
git+https://github.com/fecgov/apispec.git@dev#egg=apispec -->
git+https://github.com/fecgov/apispec.git@dev

-

-  

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/fecgov/apispec.git@dev#egg=apispec
+git+https://github.com/fecgov/apispec.git@dev
 prance[osv]>=0.11 # use apispec[validation] once we no longer fork apispec
 cfenv==0.5.2
 invoke==0.15.0


lbeaufort:
@qqss88 I'm having trouble figuring out what the original issue was, and how this PR addresses it. How can I replicate the error and confirm that this fixes it? 
`#egg=`... is supported by `pip` - https://pip.pypa.io/en/stable/reference/pip_install/#git

Any context you can give on this issue would be helpful, thanks!

qqss88:
@lbeaufort Hi Laura: I think the issue here is that I am not sure the egg name is correct - even it is correct, you need a '-e' option when you run pip to install it. People usually do not use that option when doing 'pip install -r requirements.txt'. That is why John run into that 'module not found' issue. I think it is much easier to remove it from requirements.txt file when we still use pip. 

we have more customized packages like this: we can add a egg name for each and make sure all of them work. but we need to test it and we need people to be aware that they need to use the '-e' option along with it.

Issue #3467, opened 2018-10-31
Add searchability to filings/operations-log endpoint


jwchumley:
Members of the Press need to be able to easily determine if a report has completed processing so they will know if the transactions are available. We need to make the following existing data elements searchable. Modifying this endpoint: https://api.open.fec.gov/developers/#/filings/get_operations_log_

- [x] status_num 
- [x] amendment_indicator
- [x] receipt_date
- [x] coverage_end_date
- [x] transaction_data_complete_date

**Completion criterion**
- [x] adding these elements to the API call returns the correct results.
- [x] Add automated tests

PR #3493, opened 2018-11-19
Add searchability to filings/operations-log endpoint

hcaofec:
## Summary (required)

- Resolves [#3467](https://github.com/fecgov/openFEC/issues/3467)

- More data fields are added to the endpoint to expand its searchability _

Test this link and compare the result to sql query result
http://127.0.0.1:5000/v1/operations-log/?sort_hide_null=false&min_receipt_date=05%2F01%2F2018&sort_null_only=false&sort_nulls_last=false&sort=-report_year&per_page=20&page=1


tests/test_operation_logs.py
```diff
@@ -34,4 +35,18 @@ def test_verified_reports(self):
         results = self._results(api.url_for(OperationsLogView, candidate_committee_id='02', status_num=1))
         self.assertEqual(len(results), 1)
         for result in results:
-            self.assertTrue(1, result['status_num'])
\ No newline at end of file
+            self.assertTrue(1, result['status_num'])
+
+    def test_receipt_date(self):
```
pkfec: Well done @hcaofec! Could you add another test case to test the data range on receipt_date column? A simple test to make sure that the date range is picked up correctly.Here is an example to test the data range filter:
https://github.com/fecgov/openFEC/blob/develop/tests/test_filings.py#L35

hcaofec: test case has been added.
@fec-jli @pkfec

fec-jli:
good work.
Can you add test case for 'overage_end_date' and 'transaction_data_complete_date'?
same as test case for 'receipt_date'. Thanks

tests/test_operation_logs.py
pkfec: @hcaofec Looks good. I will merge this PR.

PR #3505, opened 2018-11-23
Remove mur-specific query fields - use 'case'

### LINKED ISSUE #3504: Remove the `mur_` filters from the API
As part of the API work here: fecgov/openFEC#3365, we added generic `case_` filters that work for MURs, ADRs, and AFs. We have refactored the front end search to use the generic `case_` filters so we should remove the `mur_` filters from the API.

- Consider adding filters without prefixes so we can streamline further -prefixes make sense for `_no` and `_name` but they're confusing to keep straight for respondents and af-specific fields.

lbeaufort:
## Summary (required)

- Resolves #3504 
- Remove mur-specific query fields - use `case_`
- Fix admin fine filter typo

- Run ES locally, do a mur query: http://localhost:5000/v1/legal/search/?case_no=7400

-  MUR search

### DIFF
--- a/tests/test_legal.py
+++ b/tests/test_legal.py
@@ -211,14 +211,14 @@ class SearchTest(unittest.TestCase):
                                      doc_type=mock.ANY)
 
     @patch.object(es, 'search')
-    def test_query_dsl_with_mur_filters(self, es_search):
+    def test_query_dsl_with_case_filters(self, es_search):
         response = self.app.get('/v1/legal/search/', query_string={
                                 'q': 'embezzle',
                                 'type': 'murs',
-                                'mur_min_open_date': '2012-01-01',
-                                'mur_max_open_date': '2013-12-31',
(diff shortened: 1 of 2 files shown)


lbeaufort:
@fec-jli thanks for taking a look! 

>if I run old field name like: http://localhost:5000/v1/legal/search/?mur_no=7400
I got all 3 murs return. Is this ok? we control field name from cms side?
Yes, because we're removing `mur_no` as a parameter, the filter won't work, and will return all results

>2)if I run DSL to check mur data on local (new field name)
**curl localhost:9200/docs/murs/case_7400 | python -m json.tool**
no result.
I have to run this and get result (old field name):
**curl localhost:9200/docs/murs/mur_7400 | python -m json.tool**

This query uses the `doc_id`, which will start with the case type (in your example, mur_7400). doc_id needs to be unique, and some cases have the same number (something like AF 100, ADR 100).

PR #3625, opened 2019-03-13
Add application-level IP blocking

lbeaufort:
## Summary (required)

- Temporary workaround while we work on https://github.com/fecgov/fec-cms/issues/2735

Add application-level IP blocking. To block an IP, add to the `BLOCKED_IPS` list on line 37 and do a hotfix. This should be a temporary solution until we can reach out to the user, or determine whether activity is malicious

- This is really only testable with a manual deploy - add your IP address to `BLOCKED_IPS`, manually deploy, and then try/fail to query API directly and through the CMS.

-  API/CMS IP access

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -133,9 +133,9 @@ def handle_error(error):
 
 
 # api.data.gov
-trusted_proxies = ('54.208.160.112', '54.208.160.151')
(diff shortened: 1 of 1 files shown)


pkfec:
@lbeaufort Changes look good.  I will test the fix tonight from home.

lbeaufort:
This was implemented with a hotfix: https://github.com/fecgov/openFEC/pull/3628

PR #3695, opened 2019-04-12
Limit access to /downloads/ to one API key

### LINKED ISSUE #3651: Implement download endpoint limits
In order to rate limit the amount of download requests per IP, we need to restrict the downloads endpoint to only 1 download key. Then we will be able to apply rate limits to that key as desired. This will protect us from a single IP requesting too many downloads all within a short timeframe.

Rate limit downloads

**Completion criteria:**

- [X] Check IPs for internal network traffic to see if rate limits will interfere with internal downloads
For wired and wireless public IP's for FEC HQ (see #security-practices channel for the IPs) and the most download requests we got was 20/hour. One anomaly was load testing on 4/1 - those can been discarded. After finding this, I think we should target **60 requests/hour/IP**, which is triple our busiest internal day in the past 90 days (logs: https://logs.fr.cloud.gov/goto/433b8ffe039c6dcbb885bb3b235ea74a
- [x] Create new download key, one per environment. CMS should only use this key for downloads. CMS `dev` and `feature` should share a key
- [x] Rate limit the new key. Each ping to check task status is considered a request. We may want to start with allowing 100 requests per hour per IP. Calculations are as follows: 100 requests / 4 average pings per download request = ~25 CSV download files can be requested per hour per IP
- [x] Create API restrictions on the `/download/` endpoint to only whitelist requests from new download keys
- [X] Save `env vars` in `stage` and `prod` environments for `DOWNLOAD_WHITELIST_API_KEY_ID` (save key ID and not key name).  (`dev` doesn't use the API umbrella)
- [x] Create a follow-up front-end ticket for error handling and update the download key in the CMS. On the front end, add error handling messages for exceeding the request limit. Ex: "You have exceeded the maximum allowed download requests this hour". https://github.com/fecgov/fec-cms/issues/2842
- [X] Make a ticket to unblock particular IPs from downloads (included in the PR as `env var`)
- [X] Turn on downloads protection with `RESTRICT_DOWNLOADS=True` in `stage` (We will do `prod` after the deploy)

lbeaufort:
## Summary (required)

Resolves #3651 

- Limit access to /downloads/ to one API key, only when `RESTRICT_DOWNLOADS` is turned on
- Turn `BLOCKED_IPS` into an `env var`

- This is hard to test locally because it needs the API umbrella in place. There's currently a manual deploy on `stage` (`invoke deploy --space stage --skip-migrations`)

-  Downloads

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -135,7 +135,7 @@ def handle_error(error):
 
 # api.data.gov
 TRUSTED_PROXIES = ('54.208.160.112', '54.208.160.151')
-BLOCKED_IPS = env.get_credential('BLOCKED_IPS', [])
+BLOCKED_IPS = env.get_credential('BLOCKED_IPS', '')
 FEC_API_WHITELIST_IPS = env.get_credential('FEC_API_WHITELIST_IPS', False)
 # Search this key_id in the API umbrella admin interface to look up the API KEY
 DOWNLOAD_WHITELIST_API_KEY_ID = env.get_credential('DOWNLOAD_WHITELIST_API_KEY_ID')
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -135,6 +135,7 @@ def handle_error(error):
 
 # api.data.gov
 TRUSTED_PROXIES = ('54.208.160.112', '54.208.160.151')
+# Save blocked IPs as a long string, ex. "1.1.1.1, 2.2.2.2, 3.3.3.3"
 BLOCKED_IPS = env.get_credential('BLOCKED_IPS', '')
 FEC_API_WHITELIST_IPS = env.get_credential('FEC_API_WHITELIST_IPS', False)
 # Search this key_id in the API umbrella admin interface to look up the API KEY
@@ -143,8 +144,11 @@ RESTRICT_DOWNLOADS = env.get_credential('RESTRICT_DOWNLOADS')
 
 @app.before_request
 def limit_remote_addr():
-    """If `FEC_API_WHITELIST_IPS` is set, reject all requests that are not
-    routed through the API Umbrella.
+    """
+    If `FEC_API_WHITELIST_IPS` is set:
+    - Reject all requests that are not routed through the API Umbrella
+    - Block any flagged IPs
+    - If we're restricting downloads, only allow requests from whitelisted key
     """
     falses = (False, 'False', 'false', 'f')
     if FEC_API_WHITELIST_IPS not in falses:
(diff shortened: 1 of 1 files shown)


webservices/rest.py
```diff
@@ -135,8 +135,11 @@ def handle_error(error):
 
 # api.data.gov
 TRUSTED_PROXIES = ('54.208.160.112', '54.208.160.151')
-BLOCKED_IPS = ('95.216.186.66', '109.252.57.42')
+BLOCKED_IPS = env.get_credential('BLOCKED_IPS', '')
```
lbeaufort: Add comment for how to add to these

webservices/rest.py
```diff
@@ -135,8 +135,11 @@ def handle_error(error):
...
 FEC_API_WHITELIST_IPS = env.get_credential('FEC_API_WHITELIST_IPS', False)
+# Search this key_id in the API umbrella admin interface to look up the API KEY
+DOWNLOAD_WHITELIST_API_KEY_ID = env.get_credential('DOWNLOAD_WHITELIST_API_KEY_ID')
+RESTRICT_DOWNLOADS = env.get_credential('RESTRICT_DOWNLOADS')
 
 @app.before_request
 def limit_remote_addr():
```
lbeaufort: Update docstring

webservices/rest.py
```diff
@@ -154,6 +161,11 @@ def limit_remote_addr():
                 abort(403)
             if source_ip in BLOCKED_IPS:
                 abort(403)
+            if RESTRICT_DOWNLOADS not in falses and '/download/' in request.url:
+                # 'X-Api-User-Id' header is passed through by the API umbrella
+                request_api_key_id = request.headers.get('X-Api-User-Id')
+                if request_api_key_id != DOWNLOAD_WHITELIST_API_KEY_ID:
+                    abort(403)
```
qqss88: Looks great. One minor issue is that I am not 100% sure 'RESTRICT_DOWNLOADS' should be within 'FEC_API_WHITELIST_IPS' conditional block, or we should keep them in parallel logically? I guess we always set FEC_API_WHITELIST_IPS to true. 

Great thought and smart implementation on handling this download hacking issue:-)

lbeaufort: Thank you @qqss88! We can only protect downloads when the API umbrella is in place, which corresponds with `FEC_API_WHITELIST_IPS`. I think a better name might be `REQUIRE_API_UMBRELLA` or something like that. Thoughts?

qqss88: sure - that clear things up. And yes, that name make sense better:-)

lbeaufort: For now, this is holding up the release so I'll put in a new issue for the rename, since we need the `env var` to match.

PR #3732, opened 2019-04-29
fix parse_fulltext match

### LINKED ISSUE #3716: fix parse_fulltext()
**Problem**
When model fields are specified as `fulltext` fields (e.g., `RadAnalystView`'s `filter_fulltext_fields`) there appears to be behavior leading to **unexpected API returns** (notable when the database column is of type `varchar` and when the searched database value is in the form of single string having no whitespace characters but having special characters, e.g., a full email address or a company name such as "company.com".

**Background**
The `filter_fulltext` method calls the `parse_fulltext` method, which takes an user-specified input field and replaces any `\W` characters with `' '` (one whitespace character) and returns a formatted query as an argument to `to_tsquery()`.

For example, if the user input is `user@fec.gov` and the corresponding searched values are in `table.email` (thing they are searching for) then the resulting query will be of the form:

```sql
select <columns> from table
where table.email @@ to_tsquery('user: * & fec:* & gov:*');
```

And the query will return no rows. But the query

```sql
select <columns> from table
where table.email @@ to_tsquery('user@fec.gov');
```
will return the desired rows.

**Suggested steps**
- [x] Investigate the behavior when column types are specified as `TSVECTOR`
- [ ] Investigate other tables/views where this behavior may give unexpected results (e.g., aggregate by employer)
- [ ] consider adding to automated test coverage

**Proximate code**
```python
def filter_fulltext(query, kwargs, fields):
    for key, column in fields:
        if kwargs.get(key):
            exclude_list = build_exclude_list(kwargs.get(key))
            include_list = build_include_list(kwargs.get(key))
            if exclude_list:
                filters = [
                    sa.not_(column.match(utils.parse_fulltext(value)))
                    for value in exclude_list
                ]
                query = query.filter(sa.and_(*filters))
            if include_list:
                filters = [
                    column.match(utils.parse_fulltext(value))
                    for value in include_list
                ]
                query = query.filter(sa.or_(*filters))
    return query
```

and subroutine `parse_fulltext()`:
```python
def parse_fulltext(text):
    return ' & '.join([
        part + ':*'
        for part in re.sub(r'\W', ' ', text).split()
    ])
```

jason-upchurch:
## Summary (required)

Added 
```python                
for value in include_list:
filters.append(column.match(value))
```
as well as similar code for `exclude_list` in `filters.py` to make sure literal field inputs were considered by `fulltext_filter`. Testing is ongoing.

- Resolves #3716 

Modify `parse_fulltext` definition to return search string with terms at most numbering what the user entered and includes only non-special characters.

- [ ] `pytest`
- [ ] `./manage.py runserver` and verify key endpoint fields return expected results

- Anywhere fulltext filter fields are used
- See issue #3716 for full listing of application areas using fulltext fields for filtering

## Related PRs

branch | PR
------ | ------
feature/3716-parse_fulltext-bug | [link](https://github.com/fecgov/openFEC/issues/3716)

### DIFF
--- a/webservices/filters.py
+++ b/webservices/filters.py
@@ -72,16 +72,16 @@ def filter_fulltext(query, kwargs, fields):
                     sa.not_(column.match(utils.parse_fulltext(value)))
                     for value in exclude_list
                 ]
-                for value in exclude_list:
-                    filters.append(column.match(value))
+ #               for value in exclude_list:
+ #                   filters.append(column.match(value))
                 query = query.filter(sa.and_(*filters))
             if include_list:
                 filters = [
                     column.match(utils.parse_fulltext(value))
                     for value in include_list
                 ]
-                for value in include_list:
-                    filters.append(column.match(value))
+#                for value in include_list:
+#                    filters.append(column.match(value))
                 query = query.filter(sa.or_(*filters))
     return query
 
--- a/webservices/filters.py
+++ b/webservices/filters.py
@@ -72,16 +72,12 @@ def filter_fulltext(query, kwargs, fields):
                     sa.not_(column.match(utils.parse_fulltext(value)))
                     for value in exclude_list
                 ]
- #               for value in exclude_list:
- #                   filters.append(column.match(value))
                 query = query.filter(sa.and_(*filters))
             if include_list:
                 filters = [
                     column.match(utils.parse_fulltext(value))
                     for value in include_list
                 ]
-#                for value in include_list:
-#                    filters.append(column.match(value))
                 query = query.filter(sa.or_(*filters))
     return query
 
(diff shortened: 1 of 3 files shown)


webservices/filters.py
```diff
@@ -72,12 +72,16 @@ def filter_fulltext(query, kwargs, fields):
                     sa.not_(column.match(utils.parse_fulltext(value)))
                     for value in exclude_list
                 ]
+ #               for value in exclude_list:
```
lbeaufort: Please take out these commented lines

jason-upchurch: thank you! done.

jason-upchurch:
Thank you!

PR #3830, opened 2019-06-20
Remove credentials from migration url

lbeaufort:
## Summary (required)

Resolves https://github.com/fecgov/fec-accounts/issues/182

- Remove username/password from migration url, pass as username/password parameters to prevent displaying password in URL
- Strip out username from postgres error messages, preserve Flyway-specific error messages because they don't print username and we need them for debugging
- Fail build when deploy task fails (it was failing silently with `return` and that made it look like the build passed when it really didn't. See http://docs.pyinvoke.org/en/1.2/api/exceptions.html#invoke.exceptions.Exit for info on the built-in "clean" exit without stack trace)

**Successful migration**
-  Set `FEC_MIGRATOR_SQLA_CONN_DEV`
- Target `dev` space
- Make sure you don't have any new migration files (you can check out this branch)
- Run `invoke deploy --space dev --migrate-database`
- Migrations should "run" even though there are new ones

**Unsuccessful migration**
- Change the `FEC_MIGRATOR_URL_DEV` to something that doesn't exist (you can test bad DB name, bad username, bad password)
- Note the error message (talk to @lbeaufort if you want to know more about how to test this properly)

- DB migrations

## Related PRs

Original implementation of Circle automated migrations: https://github.com/fecgov/openFEC/pull/2768

### DIFF
--- a/jdbc_utils.py
+++ b/jdbc_utils.py
@@ -1,11 +1,8 @@
 import re
 def to_jdbc_url(dbi_url):
-    DB_URL_REGEX = re.compile(r'postgresql://(?P<username>[^:]*):?(?P<password>\S*)@(?P<host_port>\S*)$')
+    DB_URL_REGEX = re.compile(r'(?P<host_port>\S*)$')
     match = DB_URL_REGEX.match(dbi_url)
     if match:
-        jdbc_url = 'jdbc:postgresql://{}?user={}'.format(
-            match.group('host_port'), match.group('username'))
-        if match.group('password'):
-            jdbc_url += '&password={}'.format(match.group('password'))
+        jdbc_url = 'jdbc:postgresql://{}/fec'.format(match.group('host_port'))
         return jdbc_url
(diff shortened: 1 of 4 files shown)


fec-jli:
1) for local test:
export  FEC_MIGRATOR_SQLA_CONN_DEV= postgresql://xxxxxx:xxxxxxx@fec-govcloud-dev.xxxxxxxxx.us-gov-west-1.rds.amazonaws.com:5432/fec
`invoke deploy --space dev —migrate-database `

2)for circleci:
a) set FEC_MIGRATOR_SQLA_CONN_DEV with wrong password
wrong_psw


2)set FEC_MIGRATOR_SQLA_CONN_DEV with wrong port number
wrong_port_number

jason-upchurch:
For reference, list of `postgres@9.6` errors:
https://www.postgresql.org/docs/9.6/errcodes-appendix.html

PR #3811, opened 2019-06-07
Redefine function definitions for TSVECTOR columns

jason-upchurch:
## Summary (required)

- Resolves #3798 

redefine `fec_fitem_sched_b_insert()` such that `to_tsvector` splits on (and removes) non-word characters when generating `recipient_name_text` column.

## How to test the changes locally

- [ ] check out this branch
- [ ] run pytest
- [ ] Test migrations with drop/create cfdm_test and `invoke create_sample_db`
- [ ] query `cfdm_test` for tables/columns identified in https://github.com/fecgov/openFEC/issues/3798#issuecomment-501311701 and verify no special characters in the `<col_nam_prefix>_text` columns.
- [ ] for any `varchar` column that has a corresponding `tsvector` column, searching the `varchar` column in the form `select * from <table> where <column> @@ to_tsquery('part_1:* & part_2:* & ... & part_n: *');` should return that `varchar` column data. 
- [ ] Additionally, if the `varchar` column data is parsed to split on and remove any nonword characters, i.e., contain only [a-zA-Z0-9] where spaces and special characters split the string into words, that parsing should match the contents of the `tsvector` column (locally `tsvector` entries may have stems of full words rather than the full word itself.

## Impacted areas of the application
List general components of the application that this PR will affect:

-  all places where 'disclosure.fec_fitem_sched_b.recipient_name_text` is in use

PR #3864, opened 2019-07-12
Handle diacritical marks in data

### LINKED ISSUE #3852: account for accents in alphanumeric input for tsvector
## Summary ##
Filers may submit text that contains accents. Currently, non-English alphabet and non-numeric characters are stripped from filed data to construct the `tsvector` column that is searched. Accents need to be passed and not stripped.

## Completion criteria ##
- [x] install postgres extension `unaccent` for all environments and schema
    - [x]  `CREATE EXTENSION unaccent;` (related issue #3862)
- [x] make SQL function to generate the input argument to `to_tsvector`, i.e., return a string to be used as input. 
    - [x] When the `unaccent`ed string is the same as the input string, treat as:
        `regexp_replace(<string>, '[^a-zA-Z0-9]', ' ', 'g'))`
    - [x] When the `unaccent`ed string is different from the input string, treat as:
        `concat(regexp_replace(unaccent(<string>), '[^a-zA-Z0-9]', ' ', 'g'), ' ', 
regexp_replace(<string>, '[^[:alnum:]]', ' ', 'g'));`
- [x] make new migration file (reference `V0142_`)
- [x] account for diacritical marks in `utils.parse_fulltext()`
- [x] update `requirements.txt` with necessary libraries
- [ ] add diacritical marks test for `pytest`

jason-upchurch:
## Summary (required)
Allows users to search for data containing diacritical marks as well as unaccented versions.

- Resolves #3852

## Proposed changes
* add migration `V0151__` and `V0152` to define function to generate input argument to `to_tsvector` calls and use the function in migrations, respectively
* update `parse_fulltext` in `utils.py` to account for possible input containing diacritical marks
* add test cases to `/tests/integration/test_tsvector_generation.py`

- [ ] check out this branch (feature/3852-handle-diacritical-marks-in-data)
- [ ] run `pytest`
- [ ] try `swagger ui` and input accents on known existing data into fields known to use `tsvector`, e.g., `contributor_employer` on `schedule_a` or `recipient_name` on `schedule_b`. Those known data should still be returned.

-  `api`, front-end datatables

## Related PRs

branch | PR
------ | ------
feature/other_pr | [3811](https://github.com/fecgov/openFEC/pull/3811)


fecjjeng:
@jason-upchurch 
I am reviewing the migration files in this PR.  
In trigger functions
disclosure.fec_fitem_sched_b_insert() 
disclosure.fec_fitem_sched_b_insert() 
The clean_recipient_cmte_id/clean_contbr_id, that are part of the recipient_name_text/contributor_name_text, are actually cmte_id, not actually text fields.  So there is no need to parse them for special characters.  Adding it will not make the data incorrect, but is not necessary.  With more manipulation of the parsing, it is better we don't add more processing time if not needed. 

One of the MV, ofec_candidate_fulltext_mv, did not grant permission to fec_read like other MVs.  Would you please add it in your migration file?

jason-upchurch:
> See comments ^^ in PR

@fecjjeng I removed the `parse_fulltext` function on 

GRANT ALL ON TABLE ofec_candidate_fulltext_mv TO fec;
GRANT SELECT ON TABLE ofec_candidate_fulltext_mv TO fec_read;

> @jason-upchurch
> I am reviewing the migration files in this PR.
> In trigger functions
> disclosure.fec_fitem_sched_b_insert()
> disclosure.fec_fitem_sched_b_insert()
> The clean_recipient_cmte_id/clean_contbr_id, that are part of the recipient_name_text/contributor_name_text, are actually cmte_id, not actually text fields. So there is no need to parse them for special characters. Adding it will not make the data incorrect, but is not necessary. With more manipulation of the parsing, it is better we don't add more processing time if not needed.
> 
> One of the MV, ofec_candidate_fulltext_mv, did not grant permission to fec_read like other MVs. Would you please add it in your migration file?

@fecjjeng I updated the `clean_recipient_cmte_id` and `clean_contbr_id` in `disclosure.fec_fitem_sched_b_insert()` and `disclosure.fec_fitem_sched_a_insert()` and I added the following lines to part d of the `ofec_candidate_fulltext_mv` definition:
```sql
GRANT ALL ON TABLE ofec_candidate_fulltext_mv TO fec;
GRANT SELECT ON TABLE ofec_candidate_fulltext_mv TO fec_read;
```

I ran `invoke create_sample_db` and migrations ran as well as `pytest`.

Thank you for finding these errors!

fecjjeng:
@jason-upchurch I took a closer look on the "syntax" on the migration file.  Noticed that most of these you did not use the _tmp approach.  IF the MV is being used by API, you still need to go through the _tmp approach.  Another point to note is that comments like "this is not currently used" from previous migration files, you need to check again.  Because whatever is true that time, may not be true anymore.

jason-upchurch:
> See comments ^^ in PR

@fecjjeng cc @hcaofec I have addressed requested changes and added. PR is again ready for review! Thank you! cc @lbeaufort @fec-jli

fecjjeng:
@jason-upchurch I had review the updated migration file V0152.  This is great work.  Just noticed some minor changes needed:
- ofec_dates_vw: it is mssing the grant statements.
- ofec_sched_e_vw: you probably find the drop statement (instead of create or replace) in a previous migration file.  In that particular migration, we added columns in the middle instead of at the end of the select statement so we have to drop it and recreate.  It is not the case here so you can follow the create or replace syntax and grant the proper permission.  Since you don't drop and recreate it, you don't need to include the comment "It is not used by API OR other MV/VW".
- The last several MVs, the MV name in comments are inconsistent with the actual MV name.  It is not going to affect the real script.  But if you can make it consistent while doing updates for above objects would be nice.

jason-upchurch:
> * ofec_sched_e_vw: you probably find the drop statement (instead of create or replace) in a previous migration file.  In that particular migration file, we added columns in the middle instead of at the end of the select statement so we have to drop it and recreate.  It is not the case here so you can follow the create or replace syntax and grant the proper permission.  Since you don't drop and recreate it, you don't need to include the comment "It is not used by API OR other MV/VW".

@hcaofec cc @fecjjeng about the comment above, I removed the "it is not used by API.." comment, but I wasn't sure if I was supposed to change any of the existing permissions or create syntax. Can you review the latest commit and let me know specifically on `ofec_sched_e_vw` if any other changes are needed? Thank you! (I incorporated the other requested changes).

webservices/utils.py
```diff
@@ -241,8 +242,24 @@ def extend(*dicts):
 
 def parse_fulltext(text):
     '''
-    split on and remove any nonword characters for converion to ts_vector search
+    The main purpose of parse_fulltext is to decompose user input into a safe form
```
lbeaufort: Super helpful documentation, thanks @jason-upchurch!

PR #3827, opened 2019-06-19
Remove credentials from migration url

lbeaufort:
## Summary (required)

Resolves https://github.com/fecgov/fec-accounts/issues/182

- Remove username/password from migration url, pass as username/password parameters
- Strip out username from error message

## How to test the changes locally

**Successful migration**
-  Set `FEC_MIGRATOR_SQLA_CONN_DEV`
- Target `dev` space
- Make sure you don't have any new migration files (you can check out this branch)
- Run `invoke deploy --space dev --migrate-database`
- Migrations should "run" even though there are new ones

**Unsuccessful migration**
- Change the `FEC_MIGRATOR_URL_DEV` to something that doesn't exist (you can test bad DB name, bad username, bad password)
- Note the error message (talk to @lbeaufort if you want to know more about how to test this properly)

## Impacted areas of the application
- DB migrations

## Related PRs
List related PRs against other branches:

Original implementation of Circle automated migrations: https://github.com/fecgov/openFEC/pull/2768

PR #3877, opened 2019-07-22
Add postgres@9.6 login errors and change default return

### LINKED ISSUE #3869: Only strip out login-related Flyway error messages
Follow up to #3827, we stripped out postgres error messages but this has made troubleshooting failed builds more difficult. In [remove credentials](https://github.com/fecgov/openFEC/blob/develop/jdbc_utils.py#L26) we should only strip out the messages for login-specific error codes and output the whole error message for other codes.

jason-upchurch:
## Summary

- Resolves #3869

add more `postgresql@9.6` error codes, change default return message.

- set `FEC_MIGRATOR_SQLA_CONN_DEV`
- Change the `FEC_MIGRATOR_SQLA_CONN_DEV` to something that doesn't exist (you can test bad DB name, bad username, bad password)
- no username should be passed in any error messages
- for cases where the error message is not one explicitly listed in `jdbc_utils.py`, the full non-stripped error message should be displayed.

-  `flyway` migration, `dev`, `stg`, `prd` spaces

## Related PRs

branch | PR
------ | ------
feature/other_pr | [https://github.com/fecgov/openFEC/pull/3827](https://github.com/fecgov/openFEC/pull/3827)

### DIFF
--- a/jdbc_utils.py
+++ b/jdbc_utils.py
@@ -25,18 +25,29 @@ def to_jdbc_url(dbi_url):
 
 def remove_credentials(error):
     """
-    For PostgreSQL errors, strip out everything except error code and description.
+    For PostgreSQL login errors, strip out everything except error code and description.
     This prevents usernames from being printed.
     For other Flyway-specific errors, go ahead leave them intact.
     """
-    code_lookup = {
-        '08001': 'Connection attempt failed',
-        '28P01': 'Password authentication failed',
-        '42501': 'Insufficient priveledge'
+    login_related_error_codes = {
+        '08000': 'connection_exception',
+        '08001': 'sqlclient_unable_to_establish_sqlconnection',
+        '08003': 'connection_does_not_exist',
+        '08004': 'sqlserver_rejected_establishment_of_sqlconnection',
+        '08006': 'connection_failure',
+        '08007': 'transaction_resolution_unknown',
+        '08P01': 'protocol_violation',
+        '28000': 'invalid_authorization_specificiation',
+        '28P01': 'invalid_password',
+        '3D000': 'invalid_catalog_name',
+        '42000': 'syntax_error_or_access_rule_violation',
+        '42501': 'insufficient_privilege',
+        '42601': 'syntax_error'
     }
     match = re.search(r'.*(SQL State  : )+(?P<error>[a-zA-Z0-9]{0,5})', error)
     if match:
         error_code = match.group('error')
-        return 'PostgreSQL error code {}: {}'.format(error_code, code_lookup.get(error_code, "Unrecognized code. See PostgreSQL docs."))
+        return 'PostgreSQL error code {}: {}'.format(
+            error_code, login_related_error_codes.get(error_code, error))
     else:
         return error


jdbc_utils.py
```diff
@@ -25,18 +25,28 @@ def to_jdbc_url(dbi_url):
...
+        '08P01': 'protocol_violation',
+        '28000': 'invalid_authorization_specificiation',
+        '28P01': 'invalid_password',
+        '42000': 'syntax_error_or_access_rule_violation',
+        '42501': 'insufficient_privilege',
+        '42601': 'syntax_error'
     }
     match = re.search(r'.*(SQL State  : )+(?P<error>[a-zA-Z0-9]{0,5})', error)
     if match:
         error_code = match.group('error')
```
lbeaufort: Since we’re using the fallback error for our .get, can we streamline our “if” logic here?

fec-jli:
I got error codes `42P07` and `28P01` before. Due to the dependency MV issue.
(ex: some temp  or testing MVs are not clean).

jdbc_utils.py
jason-upchurch: @lbeaufort I don't know a priori all the error messages flyway throws, so if there is no match, i still want to throw the error. There are also logically equivalent ways to handle that case as well, but let me know what you are thinking. Thank you!

jason-upchurch:
> I got error codes `42P07` and `28P01` before. Due to the dependency MV issue.
> (ex: some temp or testing MVs are not clean).

Thank you @fec-jli ! In the case of the dependent issue, we want to preserve the whole error message for debugging. The `28P01` is included in the list to check. `42P07` appears to be `duplicate_table` so I believe we want to throw the whole error as well for help in debugging. Let me know if any questions / concerns! Thank you!!

jdbc_utils.py
jason-upchurch: Below is a short discussion via comments. 
```python
match = re.search(r'.*(SQL State  : )+(?P<error>[a-zA-Z0-9]{0,5})', error)
"""
regex match successful indicating we know how to extract error code (do ALL errors satisfy the above regular expression?
"""
if match: 
    error_code = match.group('error')
    return 'PostgreSQL error code {}: {}'.format(
        error_code, code_lookup.get(error_code, error)) #get the error code, but if we didn't account for it in advance, throw the whole error, but still requires the match
"""
if the error is something not expected or in a format not known in advance, then return that error
"""
else:
    return error
```

lbeaufort: @jason-upchurch looks good, thanks for taking a look!

lbeaufort:
> > I got error codes `42P07` and `28P01` before. Due to the dependency MV issue.
> > (ex: some temp or testing MVs are not clean).
> 
> Thank you @fec-jli ! In the case of the dependent issue, we want to preserve the whole error message for debugging. The `28P01` is included in the list to check. `42P07` appears to be `duplicate_table` so I believe we want to throw the whole error as well for help in debugging. Let me know if any questions / concerns! Thank you!!

@jason-upchurch @fec-jli maybe to clear up future confusion we could rename the lookup `login_related_codes` or something like that?

jason-upchurch:
> > > I got error codes `42P07` and `28P01` before. Due to the dependency MV issue.
> > > (ex: some temp or testing MVs are not clean).
> > 
> > 
> > Thank you @fec-jli ! In the case of the dependent issue, we want to preserve the whole error message for debugging. The `28P01` is included in the list to check. `42P07` appears to be `duplicate_table` so I believe we want to throw the whole error as well for help in debugging. Let me know if any questions / concerns! Thank you!!
> 
> @jason-upchurch @fec-jli maybe to clear up future confusion we could rename the lookup `login_related_codes` or something like that?

👍 I will do that @lbeaufort and @fec-jli

PR #3891, opened 2019-08-01
Add security-related headers

### LINKED ISSUE #3876: [Tenable] API missing/inadequate response headers (due 8/31/19)
### 1. Insecure 'Access-Control-Allow-Origin' header: 
_Cross Origin Resource Sharing (CORS)_ is an HTML5 technology which gives modern web browsers the ability to bypass restrictions implemented by the _Same Origin Policy_. The _Same Origin Policy_ requires that both the JavaScript and the page are loaded from the same domain in order to allow JavaScript to interact with the page. This in turn prevents malicious JavaScript being executed when loaded from external domains. The CORS policy allows the application to specify exceptions to the protections implemented by the browser, and allows the developer to whitelist domains for which external JavaScript is permitted to execute and interact with the page. A weak CORS policy is one which whitelists all domains using a wildcard (`*`), which will allow any externally loaded JavaScript resource to interact with the affected page. This can severely increase the risk of attacks such as Cross Site Scripting etc. Scanner detected that the CORS policy being set by the server was weak, and used a wildcard value. This is evident by the `Access-Control-Allow-Origin` header being set to `*`.
It is important that weak CORS policies are not used. Policies can be hardened by removing the wildcard and individually specifying the domains where the trusted JavaScript resources are located.
Remediation path: If the list of hosts for externally hosted JavaScript resources is excessive, then a whole top level domain can be whitelisted by using a combination of the wildcard and the domain (example: `*.tenable.com`).

References:
https://www.owasp.org/index.php/CORS_OriginHeaderScrutiny
https://developer.mozilla.org/en-US/docs/Waeb/HTTP/Access_control_CORS

### 2. Missing 'X-Content-Type-Options' Header:

The HTTP 'X-Content-Type-Options' response header prevents the browser from MIME-sniffing a response away from the declared content-type. The server did not return a correct 'X-Content-Type-Options' header, which means that this website could be at risk of a Cross-Site Scripting (XSS) attack. 

Remediation path: Configure your web server to include an 'X-Content-Type-Options' header with a value of 'nosniff'.

References:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#xcto

### 3. Missing 'X-Frame-Options' header

Clickjacking (User Interface redress attack, UI redress attack, UI redressing) is a malicious technique of tricking a Web user into clicking on something different from what the user perceives they are clicking on, thus potentially revealing confidential information or taking control of their computer while clicking on seemingly innocuous web pages. The server didn't return an `X-Frame-Options` header which means that this website could be at risk of a clickjacking attack.  The `X-Frame-Options` HTTP response header can be used to indicate whether or not a browser should be allowed to render a page inside a frame or iframe. Sites can use this to avoid clickjacking attacks, by ensuring that their content is not embedded into other sites. 
Remediation path: Configure your web server to include an `X-Frame-Options` header. 
References:
https://www.owasp.org/index.php/Clickjacking
http://tools.ietf.org/html/rfc7034
https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options

### 4. Missing 'X-XSS-Protection' Header

The HTTP 'X-XSS-Protection' response header is a feature of modern browsers that allows websites to control their XSS auditors. The server is not configured to return a 'X-XSS-Protection' header which means that any pages on this website could be at risk of a Cross-Site Scripting (XSS) attack. This URL is flagged as an specific example. 
Remediation path: Configure your web server to include an 'X-XSS-Protection' header with a value of '1; mode=block' on all pages.
References:
https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#xxxsp
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

### 5. Missing Content Security Policy

Content Security Policy (CSP) is a web security standard that helps to mitigate attacks like cross-site scripting (XSS), clickjacking or mixed content issues. CSP provides mechanisms to websites to restrict content that browsers will be allowed to load. No CSP header has been detected on this host. This URL is flagged as a specific example. 
Remediation path: Configure Content Security Policy on your website by adding 'Content-Security-Policy' HTTP header or meta tag http-equiv='Content-Security-Policy'. 
References: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
https://csp-evaluator.withgoogle.com/
https://developers.google.com/web/fundamentals/security/csp/
https://content-security-policy.com/

lbeaufort:
## Summary (required)

Resolves #3876 

- Add security-related headers to all responses (keep `Access-Control-Allow-Origin: *` header unchanged because this is a public API). See original issue for background information.
- Add endpoint for Content Security Policy (CSP) violation reporting
- Currently deployed to stage: https://api-stage.open.fec.gov/developers/

- run the API locally, check swagger UI browser errors
- `curl -I "http://localhost:5000/v1/committees"` to check headers

## How to test the changes on stage

- I pushed a deploy to `stage` with Circle (as of 8/28)
- Test API for console errors
- Curl API to check headers
- Run scan like https://securityheaders.com/ and https://csp-evaluator.withgoogle.com/ results
- Make sure CMS can still access what it needs
- For developers, open up your inspector console and look for errors like this: `Refused to load the script 'https://example.gov' because it violates the following Content Security Policy directive:...`

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -202,7 +202,7 @@ def add_caching_headers(response):
 
 @app.after_request
 def add_secure_headers(response):
-    """Add secure headers """
+    """Add secure headers to each response"""
 
     headers = {
         "X-Content-Type-Options": "nosniff",
@@ -210,10 +210,24 @@ def add_secure_headers(response):
         "X-XSS-Protection": "1; mode=block",
     }
     if env.get_credential('PRODUCTION'):
-        headers["Content-Security-Policy"] = "default-src 'self' data: *.fec.gov *.app.cloud.gov; img-src 'self' data:; script-src 'self' 'unsafe-inline'; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; font-src 'self' https://fonts.gstatic.com data:; connect-src *.fec.gov *.cloud.gov"
-    # Add localhost options
+        headers["Content-Security-Policy"] = \
+            "default-src 'self' " \
+            "data: *.fec.gov *.app.cloud.gov; " \
+            "img-src 'self' data:; " \
+            "script-src 'self' 'unsafe-inline'; " \
+            "style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; " \
+            "font-src 'self' https://fonts.gstatic.com data:; " \
+            "connect-src *.fec.gov *.cloud.gov"
+    # local development options
     else:
-        headers["Content-Security-Policy"] = "default-src 'self' data: *.fec.gov *.app.cloud.gov localhost:* http://127.0.0.1:*; img-src 'self' data:; script-src 'self' 'unsafe-inline'; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; font-src 'self' https://fonts.gstatic.com data:; connect-src *.fec.gov *.cloud.gov localhost:* http://127.0.0.1:*"
+        headers["Content-Security-Policy"] = \
+            "default-src 'self' " \
+            "data: *.fec.gov *.app.cloud.gov localhost:* http://127.0.0.1:*; " \
+            "img-src 'self' data:; " \
+            "script-src 'self' 'unsafe-inline'; " \
+            "style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; " \
+            "font-src 'self' https://fonts.gstatic.com data:; " \
+            "connect-src *.fec.gov *.cloud.gov localhost:* http://127.0.0.1:*"
 
     for header, value in headers.items():
         response.headers.add(header, value)
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -204,7 +204,11 @@ def add_caching_headers(response):
 
 @app.after_request
 def add_secure_headers(response):
-    """Add secure headers to each response"""
+    """
+    Add secure headers to each response.
+    The 'unsafe-inline' Content Security Policy (CSP) setting is
+    needed for Swagger docs (see https://github.com/swagger-api/swagger-ui/issues/3370)
+    """
 
     headers = {
         "X-Content-Type-Options": "nosniff",
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -208,8 +208,13 @@ def add_secure_headers(response):
         "X-Content-Type-Options": "nosniff",
         "X-Frame-Options": "Deny",
         "X-XSS-Protection": "1; mode=block",
-        "Content-Security-Policy": "default-src 'none'; img-src 'self' data:; script-src 'self' 'unsafe-inline'; style-src 'self' https://fonts.googleapis.com 'unsafe-inline'; font-src 'self' https://fonts.gstatic.com data:; connect-src localhost:5000 *.fec.gov *.cloud.gov",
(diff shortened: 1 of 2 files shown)


webservices/rest.py
```diff
@@ -476,6 +514,17 @@ def api_ui():
...
 
+@app.route('/report-csp-violation/', methods=['POST'])
+def report():
+    """
+    Log Content Security Policy (CSP) violations from the browser
+    for both API and CMS.
+    Reports come in with tag 'csp-report'
+    """
+    app.logger.info(ujson.loads(str(request.data, 'utf-8')))
+    return util.output_json("CSP violation reported", 200)
```
pkfec: without the `unsafe-inline` tag the swagger docs are not loading on the browser. And the console error prevents the user from further accessing the swagger docs, which looks right.  But on the API,  we are logging this as a successful POST request with status_code 200.  I think we should change the log from INFO to ERROR and pass a different  status_code 405.
More on status codes :https://www.restapitutorial.com/httpstatuscodes.html

webservices/rest.py
```diff
@@ -201,6 +202,43 @@ def add_caching_headers(response):
...
+    }
+    content_security_policy = {
+        "default-src": "'self' *.fec.gov *.app.cloud.gov",
+        "img-src": "'self' data:",
+        "script-src": "'self' 'unsafe-inline'",
+        "style-src": "'self' https://fonts.googleapis.com 'unsafe-inline'",
+        "font-src": "'self' https://fonts.gstatic.com data:",
+        "connect-src": "*.fec.gov *.cloud.gov",
+        "object-src": "'none'",
+        "report-uri": "/report-csp-violation/",
```
pkfec: It appears that `report-uri` directive is deprecated. Only few browser support this directive. Could you replace this directive with the new `report-to` or add a forward compatability to redirect all `report-uri` to `report-to` as described here:
 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

lbeaufort: @pkfec thanks for taking a look! I implemented `report-uri` because it's still supported by more browsers than `report-to`, but I'll go ahead and add `report-to` now.

webservices/rest.py
lbeaufort: @pkfec the 200 status code response is for the reporting URL endpoint, so it makes sense to keep it as 200 to indicate the report was properly received. The `error` log level usually indicates an application error, and in this case the application is working as expected, so it makes more sense to me to keep it at `info`.

lbeaufort: @pkfec I can't get `report-to` to work and it's only supported in Chroms, so I think I'd like to keep `report-uri` for now, since it's still supported.

pkfec: `report-uri` is supported for now on all the browsers `except` IE.

pkfec: @lbeaufort Would it add more value, if we send a slack msg (alerts channel) or send an email to the team  that the CSP violation has occurred than just logging with 200 status code?

Issue #3890, opened 2019-08-01
Refactor ofec_totals_candidate_committees_mv based on ofec_totals_combined_mv


fec-jli:
Summary:
This ticket is partial of the cms issue: Candidate profile page wrong financial data [https://github.com/fecgov/fec-cms/issues/3076](https://github.com/fecgov/fec-cms/issues/3076)

1) Candidate financial data is not correct on candidate profile financial total summary.
in order to fix this, we will re-write ofec_totals_candidate_committees_mv based on these MVs:
ofec_cand_cmte_linkage_mv
**ofec_totals_combined_mv**
ofec_agg_coverage_date_vw

2) In current ofec_totals_candidate_committees_mv, the logic to calculate the totals for each committee per each fec_election_year already exists in **ofec_totals_combined_mv**,  Instead of repeating the logic here, we can utilize **ofec_totals_combined_mv**, thus to keep the logic in one place and easy to maintain in the future.

- [x] create new mv similar to ofec_totals_candidate_committees_mv using different logic
- [x] make sure finance total data correct for candidate by candidate two-election-year and full-election=true.
- [x] update flow.py /manage.py
- [x] check pytest
- [x] create ticket to remove ofec_totals_candidate_committees_mv.

fec-jli:
1) created a new **ofec_candidate_total_all_mv** based on ofec_totals_combined_mv.
2) test the endpoint, basic function works well.
3) need finalize the **ofec_candidate_total_all_mv** and do more test on new MV
4) also need add test case.

PR #3914, opened 2019-08-16
Add ofec_candidate_totals_detail_mv

fec-jli:
## Summary (required)
Candidate financial data is not correct on candidate profile financial total summary. To fix this,
we create this new **ofec_candidate_totals_detail_mv** to replace existing **ofec_totals_candidate_committees_mv** 

ofec_candidate_totals_detail_mv is based on these MVs:
ofec_cand_cmte_linkage_mv
ofec_totals_combined_mv
ofec_agg_coverage_date_vw

- Resolves [#[_3890_]](https://github.com/fecgov/openFEC/issues/3890)

check new ofec_candidate_totals_detail_mv
1)make sure finance total data correct for candidate by candidate 
both on 
two_election_year(election_full=false)
candidate_election_year (election_full=true).

2)check for 'P', 'H', 'S' and special election(odd number or even number) candidates.

3)check for Puerto Rico candidate

ex: P00003392
S0NY00188
H8GA06195
C00674580
H6PR00082 (Puerto Rico)


fecjjeng:
I had downloaded the branch to local machine, run create_sample_db and pytest.  Both completed successfully.  Materialized View refresh output at the end of create_sample_db shows this new MV had been included in the MV refresh and is refreshed AFTER all MVs that it depends on.

fecjjeng:
I run some data comparison between ofec_candidate_totals_mv (totals) and ofec_candidate_totals_detail_mv_tmp_jl (details).
After refresh both of the MVs, only compared receipts, disbursements, cash_on_hand_end_period, debts_owed_by_committee, coverage_start_date, coverage_end_date since details has much more details.  (also do a coalesce on the corresponding columns in details table since it has null if cmte does not submit data).  
1. Notice new details mv does not have entries for candidate that has no committees.  It is probably ok since this new MV is designed to replace ofec_totals_candidate_committees_mv, which also does not have entries for candidate that has no committees.
2. The cycle column in the new details MV is null when election_full is true.  While both the old ofec_totals_candidate_committees_mv, and the ofec_candidate_totals_mv has value (the same as the candidate_election_year).  I actually like the new way since it is more logically sound :+1:.  We just need to make sure this change won't break code somewhere.
3. Some of the candidates seems still have doubles compared to ofec_candidate_totals_mv in the columns I compared.  Some examples are: 'H0OK04117', 'S2LA00069', 'H0OK06054','H0TX01058'.

hcaofec:
For item 3, the doubles is solved by de-dup ofec_agg_coverage_date_mv in this PR:
https://github.com/fecgov/openFEC/pull/3916

Issue #3946, opened 2019-09-09
Create materialized view to get current_version for each filing


hcaofec:
This is a following up ticket of #[3455](https://github.com/fecgov/openFEC/issues/3455). 

Try a different approach to calculate most_recent_file_number and use this data to get the version of each filing:

We can add image_number to window function to handle the scenario where more than one amendment are received on the same day.

CREATE MATERIALIZED VIEW public.ofec_filings_current_version_mv_tmp_hc AS
SELECT f_rpt_or_form_sub.cand_cmte_id,
f_rpt_or_form_sub.rpt_yr,
f_rpt_or_form_sub.rpt_tp,
f_rpt_or_form_sub.file_num,
f_rpt_or_form_sub.receipt_dt,
first_value(f_rpt_or_form_sub.file_num) OVER (PARTITION BY f_rpt_or_form_sub.cand_cmte_id, f_rpt_or_form_sub.rpt_yr, f_rpt_or_form_sub.rpt_tp ORDER BY f_rpt_or_form_sub.receipt_dt DESC NULLS LAST) AS most_recent_file_num
FROM disclosure.f_rpt_or_form_sub
WHERE f_rpt_or_form_sub.file_num IS NOT NULL AND f_rpt_or_form_sub.rpt_yr >= 1979::numeric AND f_rpt_or_form_sub.form_tp::text <> 'SL'::text
WITH DATA;

## Completion criteria ##
- [ ] complete reference implementation
- [ ] decide if approach should be adopted (chat with other devs)
- [ ] if accepted, use reference implementation, i.e., put in PR for this issue.

hcaofec:
A filing's version is decided by "most_recent" from endpoint: /filings/.  Instead of calling the function to calculate it,  another way is to left join to v_sum table.  If this filling exists in v_sum, then it is the most recent filing.  This applies to F3/3X/3P and all other form types that are pulled into v_sum.  The similar logic can be seen in fec_vsum_f3_vw.  The logic in ofec_processed_non_financial_amendment_chain_vw will be used for F1 and F2 with some modification.  The partition will be order by receipt_date. 

Calculation of amendment_chain seems complicated because of the nature of the data in f_rpt table, and could be fixed in a separate ticket.

patphongs:
Moving this to 10.2, since it looks like more work needs to be done

lbeaufort:
PR has been merged

PR #3975, opened 2019-09-27
Recalculate most_recent flag of filings by using v_sum table

hcaofec:
## Summary (required)

- Resolves #[_3946_](https://github.com/fecgov/openFEC/issues/3946)

Currently most_recent flag is obtained from ofec_filings_amendments_all_mv which collects data by two ways: 
1) using the relationship between file_num and previous_file_num 
2) partition over cmte_id,form_tp order by file_num

Because previous_file_num is not updated sometimes, especially for paper filing, this PR change the logic to use v_sum table to check if a filling is the most recent. And for F1, F1M, and F2 which are not in v_sum, calculation is partitioned over cmte_id,form_tp order by receipt_date, begin_image_number

1)   Download the branch to local, run flyway to make sure migration run smoothly
2)  To test endpoint:  
      endpoint:  /filings/ 
      SQLA_CONN: DEV
      webservices/common/models/filings.py:   
                   class Filings(FecFileNumberMixin, CsvMixin, db.Model):
                   __tablename__ = 'ofec_filings_all_mv_tmp_hc'
      
     2.1)  committee_id=C00334029, rpt_yr=2018, 'current version' should be seen for quarterly reports
 http://127.0.0.1:5000/v1/filings/?page=1&sort_hide_null=false&sort=-receipt_date&per_page=20&sort_null_only=false&sort_nulls_last=false&committee_id=C00334029&report_year=2018
   2.2)  committee_id=C00665554, rpt_yr=2018, rpt_tp=Q2, Current version is '1350957' 
http://127.0.0.1:5000/v1/filings/?page=1&sort_hide_null=false&sort=-receipt_date&per_page=20&report_type=Q2&sort_null_only=false&sort_nulls_last=false&committee_id=C00665554&report_year=2018

Current Production:

Screen Shot 2019-09-30 at 4 04 23 PM
====================

Local CMS after:

Screen Shot 2019-09-30 at 4 06 55 PM


fecjjeng:
- branch had been downloaded to local server and migration files executed successfully.
- pytest pass successfully
- no added dependency on other MVs so no changes needed to manage.py and/or flow.py
- refresh both public.ofec_filings_all_mv_tmp_hc and public.ofec_filings_all_mv in DEV and compare data.  Row count and data except most_recent_file_number, most_recent are the same, as expected.
- the updated logic to calculate most_recent_file_number, most_recent is more reliable to return results for regular reports and non-financial filings
- For F5 24/48 hours reports, the original way of calculation produce more answers.  However, since the amendment chain is not consistent, especially for paper and earlier e-filing.  Leave the most_recent as null as in the new calculation is acceptable.
- Approve the PR.

tests/integration/test_amendment_chain.py
```diff
@@ -294,12 +308,21 @@ def test_negative_filing_chain(self):
         assert len(results) == 1
 
         result = results[0]
```
lbeaufort: @hcaofec can you remove this print statement?

tests/integration/test_amendment_chain.py
```diff
@@ -294,12 +308,21 @@ def test_negative_filing_chain(self):
...
-        self.assert_filings_equal(result, non_negative_f1)
+        print(result)
+        # self.assert_filings_equal(result, non_negative_f1)
```
lbeaufort: Do we still need this commented code?

hcaofec: Thanks for catching this.

tests/integration/test_amendment_chain.py
hcaofec: Thanks for catching this.

hcaofec:
For the latest paper filling that still has negative file number,  the CMS will show 'current version' without a clickable file number

Screen Shot 2019-10-06 at 6 42 39 PM

http://localhost:8000/data/filings/?data_type=processed&committee_id=C00001305&form_type=F3X

PR #3421, opened 2018-10-05
Set 500k threshold estimate for all endpoints

lbeaufort:
## Summary (required)

- Resolves #3104: **Count estimate for receipts search approximately doubles the amount returned**

Set 500k default threshold estimate for all endpoints

Tested performance and results aren't wildly different, but we'll want to keep testing. Locust test results: https://docs.google.com/spreadsheets/d/11y9OcjiK5ZHYkyPXeN-Toerrtvh-J0mJNAEccRQDHTE/edit#gid=1816955716

## How to test the changes locally

- run this branch of the api, go to http://127.0.0.1:5000/v1/schedules/schedule_a/?api_key=XE9ZYUzWKKe8SRLfEftx9vUndkWYnIXIQw3cOQOS&sort_hide_null=true&data_type=processed&committee_id=C00000935&two_year_transaction_period=2018&min_date=01%2F01%2F2017&max_date=05%2F01%2F2018&contributor_state=IL&sort=-contribution_receipt_date&per_page=30 - counts are no longer doubled
- you can also test `feature/add-datatable-counts` branch of cms, point to your local API and go to  http://localhost:8000/data/receipts/?two_year_transaction_period=2018&data_type=processed&committee_id=C00000935&min_date=01%2F01%2F2017&max_date=05%2F01%2F2018&contributor_state=IL

## Impacted areas of the application
List general components of the application that this PR will affect:

-  Endpoint counts

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
'feature/add-datatable-counts' | https://github.com/fecgov/fec-cms/pull/2430

PR #4002, opened 2019-10-07
Use actual counts for Schedule E efile endpoint

### LINKED ISSUE #3948: Counts estimate significantly off from actual count on raw IE datatable
**What we're after:**
A bug was discovered with the estimated counts on the raw IE datatable. As of 9/10/19 it was displaying `9,919,673,000` when the actual count was `545,952`. We have it set that an estimate is sufficient when the count is over 500,000. It seems that because the actual count is over 500,000, it's not taken into account, so it's spitting out something that is severely inaccurate. 

We need to determine how to fix this in a way that provides a more accurate count of what actually exists in the data. Without a solution, this is a problem that will perpetuate as we continue to receive new data. 

Screen Shot 2019-09-10 at 12 29 05 PM

**Example:** 
https://api.open.fec.gov/v1/schedules/schedule_e/efile/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=true&data_type=efiling&is_notice=true&sort=-expenditure_date&per_page=30&page=1

**Completion criteria:**
- [x] Raw filing data is being actively being reduced to 4 months. We are testing in staging and will be moving to apply to production soon. See ticket: https://github.com/fecgov/openFEC/issues/3789
- [x] After stage and production data is reduced, check the actual database counts of raw IE data vs what's displayed on the page and in the API call. The assumption is the totals should be more accurate.

lbeaufort:
## Summary (required)

- Resolves #3948: Use actual counts for Schedule E efile endpoint. 

The API will use an estimated count if the estimate is greater than the threshold (defaults to 500K.) Because the estimate in this case is way off (in our example, 341,452,204 when it should be 13,227) and the Schedule E efile table isn't very big (currently 13,227 rows), we're telling the API using the **exact** count of the query. 

- Check out this branch, query http://localhost:5000/v1/schedules/schedule_e/efile/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=true&data_type=efiling&is_notice=true&sort=-expenditure_date&per_page=30&page=1
- Counts are correct 

-  Schedule E raw datatable

## Related PRs
List related PRs against other branches: https://github.com/fecgov/openFEC/pull/3421 Set threshold to 500K for all endpoints

### DIFF
--- a/webservices/resources/sched_e.py
+++ b/webservices/resources/sched_e.py
@@ -8,6 +8,7 @@ from webservices import schemas
 from sqlalchemy.orm import aliased, contains_eager
 from webservices.common import models
 from webservices.common import views
+from webservices.common import counts
 from webservices.common.views import ItemizedResource
 from webservices.common.models import (
     EFilings,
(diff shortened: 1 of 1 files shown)


pkfec:
great work @lbeaufort. SE raw filings counts are the same on the CMS and the API endpoints. 

Screen Shot 2019-10-08 at 9 58 29 AM

Issue #4003, opened 2019-10-07
No indication of whether counts are exact or estimated


lbeaufort:
It would be helpful to API users to indicate whether counts are exact or estimated. We could look at the feasibility of adding a new `pagination.count_type` field to the API that would be `exact` or `estimate`.

![Screen Shot 2019-10-07 at 4 56 41 PM](https://user-images.githubusercontent.com/31420082/66347986-7fdfa080-e923-11e9-85bc-ceaaa0ca01b3.png)

Related issue: https://github.com/fecgov/openFEC/issues/3948 Counts estimate way off

PR #4087, opened 2019-11-13
Update downloads to check for exact counts, refactor count logic

### LINKED ISSUE #4071: Downloads using incorrect estimates for schedule E, causing downloads to fail
Related issue:  #3948 

Downloads using incorrect estimates for schedule E, causing downloads to fail

https://www.fec.gov/data/independent-expenditures/?data_type=efiling&is_notice=true, click "Export"
Results are less than 500,000. Check console, it says:
```
message: "Cannot request downloads with more than 500000 records"
status: 403
```

lbeaufort:
## Summary (required)

Resolves #4071

- Update downloads to check for exact counts
- Clean up tests
- Modify count logic to handle exact or estimated counts. Indicate whether counts are estimates for better testing and for future work: https://github.com/fecgov/openFEC/issues/4003
- Add test coverage for count logic

- Check out this branch
- Get downloads working locally: https://github.com/fecgov/openFEC/wiki/Set-up-Redis,--Celery-on-local-and-test-the-downloads
- Run your local CMS and point to your local API
- Test downloading all raw IEs: http://localhost:8000/data/independent-expenditures/?data_type=efiling&is_notice=true . You shouldn't get an error message like you do on production.
- Test other downloads
- Test estimated and exact counts (under and over 500K)

- Downloads
- Counts for all endpoints

## Related PRs

branch | PR
------ | ------
Use actual counts for Schedule E efile endpoint | https://github.com/fecgov/openFEC/pull/4002

### DIFF
--- a/webservices/common/views.py
+++ b/webservices/common/views.py
@@ -32,11 +32,7 @@ class ApiResource(utils.Resource):
     @marshal_with(Ref('page_schema'))
     def get(self, *args, **kwargs):
         query = self.build_query(*args, **kwargs)
-        count = (
-            counts.count_estimate(query, models.db.session)
-            if self.use_estimated_counts
-            else query.count()
-        )
+        count, _ = counts.get_count(query, models.db.session, self.use_estimated_counts)
         multi = False
         if isinstance(kwargs['sort'], (list, tuple)):
             multi = True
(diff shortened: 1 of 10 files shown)


webservices/common/views.py
```diff
@@ -25,12 +26,17 @@ class ApiResource(utils.Resource):
     join_columns = {}
     aliases = {}
     cap = 100
+    use_estimated_counts = True
 
     @use_kwargs(Ref('args'))
     @marshal_with(Ref('page_schema'))
     def get(self, *args, **kwargs):
         query = self.build_query(*args, **kwargs)
```
lbeaufort: It’s probably safer to fall back to estimates rather than exact. To avoid double negative, parameter name should be “use exact counts” defaulted to false or unset

lbeaufort: Double check more utils.resource endpoints

lbeaufort: Maybe consider putting the count logic all together in a function so we can test it more easily

Issue #4066, opened 2019-11-06
Electioneering communications filer ECs need to be specific to each filer


JonellaCulmer:
@JonellaCulmer commented on [Mon Nov 04 2019](https://github.com/fecgov/fec-cms/issues/3320)

**What we're after:** When viewing the ECs on EC filer profile pages (disbursements tab), those ECs need to be specific to that filer only. We're seeing transactions from other filers on the wrong EC filer profiles pages. 

**Completion criteria:**
- [ ] Ensure all EC transactions displayed are specific to the relevant filer only

Examples: 
Nathan, Joseph: https://www.fec.gov/data/committee/C30002562/?cycle=2018&tab=spending
Majority Forward: https://www.fec.gov/data/committee/C30002802/?tab=spending&cycle=2018

Future work:

- Open separate ticket to check the business logic for the disbursements table on the disbursements tab. Possible problem on EC filer profiles pages. Dates are in 2016, but could still be within the applicable coverage dates. Need to check others. Example: https://www.fec.gov/data/committee/C30002562/?cycle=2018&tab=spending

PR #4098, opened 2019-11-22
Add new EC endpoint /electioneering/aggregates/.

fec-jli:
## Summary (required)
The current endpoint:  /committee/{committee_id}/electioneering/by_candidate/ filter by committee_id not work.

- Resolves [#[_4066_]](https://github.com/fecgov/openFEC/issues/4066)

1. we create a new endpoint: **/electioneering/aggregates/**
based on same **ofec_electioneering_aggregate_candidate_mv**
and make all filters work as expected.

2. to do list: create a separated issue to remove (or deprecated ? don’t think any api user is using current one, because filter not work properly)current endpoint.
/committee/{committee_id}/electioneering/by_candidate/

3. change cms code to use /electioneering/aggregates/?committee_id={committee_id}

1) checkout branch
2) dev db
3) pytest
4) ./manage.py runserver

Test 1: compare current and new endpoint filter by committee_id
prod: **/committee/{committee_id}/electioneering/by_candidate/**
return all committee EC spending: 342 rows (filter by committee_id not work)
https://api.open.fec.gov/v1/committee/C30000533/electioneering/by_candidate/?api_key=DEMO_KEY&cycle=2008&election_full=false

local: **/electioneering/aggregates/**
return 4 rows (filter by C30000533 only)
http://127.0.0.1:5000/v1/electioneering/aggregates/?api_key=DEMO_KEY&cycle=2008&committee_id=C30000533

Test2: other filters for new endpoint: /electioneering/aggregates/
ex1: S8NC00239  (9 results) filter by candidate_id only
http://127.0.0.1:5000/v1/electioneering/aggregates/?api_key=DEMO_KEY&candidate_id=S8NC00239

ex2: S8NC00239 / 2014 (3 results) filter by candidate_id and cycle
http://127.0.0.1:5000/v1/electioneering/aggregates/?api_key=DEMO_KEY&candidate_id=S8NC00239&cycle=2014

### DIFF
--- a/tests/test_spending_by_others.py
+++ b/tests/test_spending_by_others.py
@@ -1,10 +1,18 @@
 from tests import factories
 from tests.common import ApiBaseTest
 
-from webservices.rest import api
-from webservices.resources.spending_by_others import ECTotalsByCandidateView
+from webservices.rest import db, api
+from webservices.resources.spending_by_others import (
+    ECTotalsByCandidateView,
+)
+from webservices.resources.aggregates import (
+    ECAggregatesView,
+)
+from webservices.schemas import (
+    ECAggregatesSchema,
+)
 
-
+# test endpoint: /electioneering/totals/by_candidate/ under tag:electioneering
 class TestTotalElectioneering(ApiBaseTest):
 
     def test_fields(self):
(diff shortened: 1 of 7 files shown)


pkfec:
@fec-jli  In the candidate electioneering aggregates result set,  could you exclude `counts`? 
Not sure what the counts mean and how useful is to display in the output.

Screen Shot 2019-11-25 at 9 49 50 AM

fec-jli:
@pkfec Thank you for your review. 
that count is kind of confuse. that means how many transaction to aggravate.
Because current endpoint have that, so it is better to keep that for now in case some api users need that. ex:  https://api.open.fec.gov/v1/electioneering/by_candidate/?cycle=2016&candidate_id=H0KS04051&district=4&state=KS&api_key=DEMO_KEY&per_page=20&election_full=true&sort_null_only=false&sort_nulls_last=false&office=house&sort_hide_null=false&page=1

pkfec:
@fec-jli i believe the `counts` in this endpoint are for internal purpose to get an idea on how many records we have aggregated. The actual count vs count in each result set  might add more confusion to the end user. We can always refactor and make our endpoints better than before. 
I would suggest to exclude `count` from the output.

fec-jli:
@pkfec  Thanks for your suggestion. We can create a separated issue for that to modify all 2 **spending by others** endpoint together.

PR #4104, opened 2019-11-27
Feature/4063 add endpoint caluclate total cc

### LINKED ISSUE #4063: Make an endpoint calculate total communication costs
Make an endpoint calculate total communication costs for  "other spending"

While investigating  [this bug](https://github.com/fecgov/fec-cms/pull/2702
), @lbeaufort discovered that we calculate "other spending" totals on the front-end side, using JavaScript. We should make endpoints to calculate these totals rather than calculating them on the front-end, for consistency and performance improvement.

Example: https://www.fec.gov/data/candidate/P80001571/?tab=other-spending

- [ ] Research data needs for CC totals
- [ ] Make endpoint for CC totals
- [ ] Add automated tests
- [ ]  Make front-end ticket to use these endpoints

jason-upchurch:
## Summary

- Resolves #4063 

- add communication costs totals endpoint to sum over committees on support or oppose

- checkout this branch
- `pytest`
- run local api and see new endpoint in swagger ui--see test it out to see that totals are given as expected.

-  `/communication_cost/by_candidate/totals/` endpoint

## Related PRs

branch | PR
------ | ------
| | #4097 |
| | #4096 |


fecjjeng:
@jason-upchurch 
- The source of the three spending by others are completely different.  Only IE is from schedule_e.  So please do not use the name schedule_e_totals_by_candidate_other_costs_XX in the class name for EC and CC in the args.py file since it is not correct.
- ScheduleEByCandidate is not the source for CC.  Please replace it with correct model.  Currently the API for CC returns the values for IE, not CC.
- The IE endpoint currently does use support_oppose_indicator as a filter.  The design is to return the totals for all possible value (currently we have S, O, and null in the table) for given cand_id/cycle.  - If the front end prefer to call the API each time for each support_oppose_indicator value for both IE and CC, we can add this one as a filter.  But the resource file need to handle it appropriately.  Currently, no matter what value pass in, the API returns all values.

jason-upchurch:
@fecjjeng cc @fec-jli @hcaofec 
> * The source of the three spending by others are completely different.  Only IE is from schedule_e.  So please do not use the name schedule_e_totals_by_candidate_other_costs_XX in the class name for EC and CC in the args.py file since it is not misleading.

fixed

> * ScheduleEByCandidate is not the source for CC.  Please replace it with correct model.  Currently the API for CC returns the values for IE, not CC.

fixed

> * The IE endpoint currently does use support_oppose_indicator as a filter.  The design is to return the totals for all possible value (currently we have S, O, and null in the table) for given cand_id/cycle.

It looks like we were using "support_oppose_indicator" as opposed to "support_oppose".

both of our underlying endpoints use this "support_oppose" so I changed both of our totals endpoints to use this one instead. If IE should NOT have this please let me know.

> * If the front end prefer to call the API each time for each support_oppose_indicator value for both IE and CC, we can add this one as a filter.  But the resource file need to handle it appropriately.  Currently, no matter what value pass in, the API returns all values.

Please see previous comment.

fecjjeng:
@jason-upchurch, good job!  
I had downloaded the updated branch and test out the function.  It works well.
pytest passed.

Just couple minor points:  
- The other 2 end points in this sets (and some other endpoints in the API) are named as xxxx/totals/by_candidate/
this endpoint is named as xxxx/by_candidate/totals/
So there is inconsistency here.  
Sorry I did not notice it earlier.  

/electioneering/totals/by_candidate/
/schedules/schedule_e/totals/by_candidate/
/communication_costs/by_candidate/totals/

- EC/IE/CC are "Spending by Others", not other costs.  That is, these money spend by other committees on the candidates.  These names only be used internally, so it is ok not to change them.  Just noted it here.

PR #4179, opened 2020-02-03
Add file structure for presidential data endpoints

lbeaufort:
## Summary (required)

- Add file structure for presidential data endpoints, based on RAD Analyst (simple endpoint)

- Run this branch without `FEC_FEATURE_PRESIDENTIAL` env var set, or with it set to `FALSE` and there should be no changes
- Run this branch with `export FEC_FEATURE_PRESIDENTIAL=True` and you should see 5 presidential endpoints, based on RAD analyst. 

- Sets groundwork for presidential map work
- Still needs tests but we can coordinate


fec-jli:
It looks good.
Is it better change the file name to **resources/presidential_map.py**  and **model/presidential_map.py**?
then we know all those endpoints will support the presidential map. We have many class name called xxxxPresidentialxxxx something, such as,   CommitteeTotalsPresidential , CandidateCommitteeTotalsPresidential... those are for other dataset of presidential.

lbeaufort:
Thanks for reviewing, Jun! I think the map layer is the presentation layer - it might not always be a map that we use to present the data so I wouldn’t want to specify  “map” in the API.

I’ll take a look at the class names to try to understand the patterns better.

________________________________

fec-jli:
Sure. we can keep the name: presidential.py now. I just want to distinguish these 5 endpoints which are based on those 8 Pre-xxxxxx tables from classic. I am not sure how long we will keep those 8 tables which logic heavy rely on informatic system. Thanks @lbeaufort

PR #4181, opened 2020-02-04
[V0182] Add /presidential/contributions/by_candidate/

lbeaufort:
## Summary (required)

- Resolves #4170 

Add `/presidential/contributions/by_candidate/`
Future work: Will check with @fecjjeng on indexes as part of #4186 performance testing

## How to test the changes locally

- `export FEC_FEATURE_PRESIDENTIAL=True`
- Change `PresidentialByCandidate` tablename to `'ofec_presidential_by_candidate_vw_lb'`
- Test a query with Swagger
- http://localhost:5000/v1/presidential/contributions/by_candidate/?contributor_state=US&election_year=2020&per_page=100&page=1&sort_hide_null=false&sort_null_only=false&sort_nulls_last=false
- Test migration with 
```
dropdb cfdm_test
createdb cfdm_test
invoke create_sample_db
flyway migrate
```

## Impacted areas of the application
List general components of the application that this PR will affect:

-  Future presidential map


**Query plan:**
```
explain(SELECT public.ofec_presidential_by_candidate_vw_lb.idx, public.ofec_presidential_by_candidate_vw_lb.candidate_id, public.ofec_presidential_by_candidate_vw_lb.candidate_last_name, public.ofec_presidential_by_candidate_vw_lb.candidate_party_affiliation, public.ofec_presidential_by_candidate_vw_lb.net_receipts, public.ofec_presidential_by_candidate_vw_lb.rounded_net_receipts, public.ofec_presidential_by_candidate_vw_lb.contributor_state, public.ofec_presidential_by_candidate_vw_lb.election_year  
 FROM public.ofec_presidential_by_candidate_vw_lb  
 WHERE public.ofec_presidential_by_candidate_vw_lb.contributor_state = 'US' AND public.ofec_presidential_by_candidate_vw_lb.election_year = 2020 ORDER BY public.ofec_presidential_by_candidate_vw_lb.net_receipts DESC)    
```
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
[...]

Issue #4178, opened 2020-02-03
Make endpoint for coverage dates and summaries (financial_summary, coverage_date)


lbeaufort:
### Summary
Design ticket: https://github.com/fecgov/openFEC/issues/4175
![Copy of Endpoints for presidential bubble map-Presidential bubble map](https://user-images.githubusercontent.com/31420082/73679267-0f098480-4688-11ea-87cb-cd3ed6e1276c.jpg)

## 2) /presidential/financial_summary/

(right side summary; Expenditure Tab ): 
(view on DISCLOSURE.PRES_F3P_TOTALS_CA_CM_LINK_20D / _16)
Filter: candidate_id, election_year

-- TODO: Determine financial summary fields, use readable column names. Maybe bring them all over as readable columns in case we need others later
 ```
SELECT distinct 
cand_id AS candidate_id, 
cand_nm AS candidate_name,
SUBSTR(cand_nm, 0, strpos(cand_nm,',')) AS candidate_last_name, 
(CASE WHEN cand_pty_affiliation IS NULL THEN NULL ELSE '('||SUBSTR(CAND_PTY_AFFILIATION,1,1)||')' END) AS candidate_party_affiliation, 
*
FROM DISCLOSURE.PRES_F3P_TOTALS_CA_CM_LINK_20D
ORDER BY 4 DESC
 ```
 
## 5) /presidential/coverage_date/

(dateThrough) (DISCLOSURE.PRES_NML_FORM_3P_20D / _16)
Filter: candidate_id, election_year

--TODO: Bring in 2016 data

```                  
select cand_id AS candidate_id,
to_char(max(cvg_end_dt), 'mm/dd/yyyy') as coverage_date,
'2020' as election_year
from DISCLOSURE.PRES_NML_FORM_3P_20D
```

### Completion criteria
- [ ] Make views in database with migration (pair w/DB team)
- [ ] Make endpoints based on design
- [x] Use a feature flag
- [ ] Add test coverage for new endpoints

fec-jli:
query for expenditure tab:
		SELECT 
		ROUND(SUM(NVL(OP_EXP_PER,0))) as oper_exp,
		ROUND(SUM(NVL(TRANF_TO_OTHER_AUTH_CMTE_PER,0))) as transfer_to_auth_cmte, 
		ROUND(SUM(NVL(FNDRSG_DISB_PER,0)))as fundraising_exp,
[...]

PR #4189, opened 2020-02-05
[V0184] Make endpoint /presidential/financial_summary/

lbeaufort:
## Summary (required)

- Partial resolution for #4178 

 Make endpoint `/presidential/financial_summary/`

- `export FEC_FEATURE_PRESIDENTIAL=True`
- Change `PresidentialSummary` tablename to `'ofec_presidential_financial_summary_vw_lb'`
- Test a query with Swagger

**All:** 
- http://localhost:5000/v1/presidential/financial_summary

**Candidate ID:** 
- http://localhost:5000/v1/presidential/financial_summary/?candidate_id=P00000001
- http://localhost:5000/v1/presidential/financial_summary/?candidate_id=P00009621

**Election_year:**
- http://localhost:5000/v1/presidential/financial_summary/?election_year=2020
- Test migration with 
```
dropdb cfdm_test
createdb cfdm_test
invoke create_sample_db
flyway migrate
```

- Presidential map data, right-hand chart  "Summary" and "Expenditures"

Screen Shot 2020-02-06 at 2 32 02 PM

Screen Shot 2020-02-06 at 2 31 44 PM

## Related PRs
https://github.com/fecgov/openFEC/pull/4190
https://github.com/fecgov/openFEC/pull/4181

### DIFF
--- a/tests/test_presidential.py
+++ b/tests/test_presidential.py
@@ -96,9 +96,6 @@ class PresidentialByState(ApiBaseTest):
             self.assertGreater(len(results), 0)
             # doesn't return all results
             response = self._response(page)
-            print("field=" + field)
-            print("original_count=" + str(original_count))
-            print("count=" + str(response['pagination']['count']))
             self.assertGreater(original_count, response['pagination']['count'])
 
     def test_filters_candidate_id(self):
@@ -176,9 +173,6 @@ class PresidentialSummary(ApiBaseTest):
             self.assertGreater(len(results), 0)
             # doesn't return all results
             response = self._response(page)
-            print("field=" + field)
-            print("original_count=" + str(original_count))
-            print("count=" + str(response['pagination']['count']))
             self.assertGreater(original_count, response['pagination']['count'])
 
     def test_sort(self):
--- a/tests/test_presidential.py
+++ b/tests/test_presidential.py
@@ -5,6 +5,7 @@ from webservices.rest import api
 from webservices.resources.presidential import(
     PresidentialByCandidateView,
     PresidentialByStateView,
+    PresidentialSummaryView,
 )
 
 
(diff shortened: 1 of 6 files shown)


fec-jli:
Great. this sql is a hard work. @lbeaufort 
just think about, two cents here:
1) if we can select one or two more columns from this query, it can feed the data for **left side candidate list** on national map page.  
2) same as other endpoint, need union multi-election_year and add one more column: election_year. Thanks

lbeaufort:
@fec-jli thanks for taking a look at my WIP PR!
>if we can select one or two more columns from this query, it can feed the data for left side candidate list on national map page.

I will add them to SQL so if we make the change it can just be on the python change

>same as other endpoint, need union multi-election_year and add one more column: election_year. Thanks
😁 thank you! Definitely hadn't caught that yet.

tests/test_presidential.py
```diff
@@ -137,3 +138,57 @@ def test_sort(self):
...
+        for field, example in filter_fields:
+            page = api.url_for(PresidentialSummaryView, **{field: example})
+            # returns at least one result
+            results = self._results(page)
+            self.assertGreater(len(results), 0)
+            # doesn't return all results
+            response = self._response(page)
+            print("field=" + field)
+            print("original_count=" + str(original_count))
+            print("count=" + str(response['pagination']['count']))
```
fec-jli: Can you remove these print ? Thanks

fec-jli:
Great job. work as expected. I will leave Jean to review sql file.
one minor comment to clean the debug print lines in test_presidential.py.
Thanks

lbeaufort:
@fec-jli good catch, thanks! I pushed a commit: https://github.com/fecgov/openFEC/pull/4189/commits/77445751a343bac58878a0a76cbc4c79fe1969cf

fecjjeng:
In general, specify the columns you want instead of * is a better practice.  But in this particular case, since we are very sure these two tables are of exactly the same structure, not listing all the columns (which is a lot) improve the readability.  Just be cautious while doing so.

Issue #3947, opened 2019-09-09
Incorporate API key signup elements from related API key work


lbeaufort:
### Summary ###

Related issue #4323 is to include vanilla api signup key to `swagger-ui.html`. This issue should follow that work and also #4305 to incorporate validation elements required for our signup process.

### Related issues ###
- [x] https://github.com/fecgov/fec-accounts/issues/254 (validation)
- [x] https://github.com/fecgov/openFEC/issues/4305 (aup)
- [x] https://github.com/fecgov/openFEC/issues/4323 (vanilla signup)

lbeaufort:
Another advantage is we could make purpose of use and reading/acknowledging the acceptable use criteria required.

lbeaufort:
We could also make a separate signup page rather than imbedding it in the docs

JonellaCulmer:
Closing in favor of https://github.com/fecgov/fec-cms/issues/3867

PR #4349, opened 2020-05-12
make basic api key signup page

jason-upchurch:
## Summary (required)

- Resolves #4323
- add plain vanilla API Key Signup page to our swagger-ui in advance of heavier lift https://github.com/fecgov/openFEC/issues/3947

## How to test the changes locally
Reviewers can either do a manual deploy to dev if desired, or run locally. The following are the steps for running locally:
- checkout this branch
- locally `export` the `API_UMBRELLA_SIGNUP_KEY` envar value located in `cf env api` (`cf target -s dev`)
- run local api
- (please test functionality/correctness. Commits can be cleaned up post review)

## Impacted areas of the application
List general components of the application that this PR will affect:
-  swagger-ui, rest.py, docs.py, spec.py

PR #4359, opened 2020-05-15
API key signup page

### LINKED ISSUE #4323: Make API Key signup page (vanilla prototype)
### Summary
From https://github.com/fecgov/fec-accounts/issues/254, we would like to make our own API key signup page. This ticket is to create a bare bones (vanilla) prototype for a simple signup. Additional functional requirements can be met with additional tickets in iterative fashion.

### Completion criteria
- [x] a working API key signup page linked through the swagger-ui

jason-upchurch:
## Summary (required)

- Resolves #4323
- add plain vanilla API Key Signup page to our swagger-ui in advance of heavier lift https://github.com/fecgov/openFEC/issues/3947

Reviewers can either do a manual deploy to dev if desired, or run locally. The following are the steps for running locally:
- checkout this branch
- locally `export` the `API_UMBRELLA_SIGNUP_KEY` envar value located in `cf env api` (`cf target -s dev`)
- locally `export API_UMBRELLA_SIGNUP_KEY_FEATURE_FLAG="True"`
- run local api
- (please test functionality/correctness)

-  swagger-ui, rest.py, docs.py, spec.py

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -236,10 +236,10 @@ def add_secure_headers(response):
     content_security_policy = {
         "default-src": "'self' *.fec.gov *.app.cloud.gov",
         "img-src": "'self' data:",
-        "script-src": "'self' 'unsafe-inline'",
-        "style-src": "'self' https://fonts.googleapis.com 'unsafe-inline'",
-        "font-src": "'self' https://fonts.gstatic.com data:",
-        "connect-src": "*.fec.gov *.cloud.gov",
+        "script-src": "'self' https://api.data.gov 'unsafe-inline'",
+        "style-src": "'self' https://fonts.googleapis.com https://api.data.gov 'unsafe-inline'",
+        "font-src": "'self' https://fonts.gstatic.com data: https://api.data.gov",
+        "connect-src": "*.fec.gov *.cloud.gov https://api.data.gov",
         "object-src": "'none'",
         "report-uri": "/report-csp-violation/",
     }
@@ -563,6 +563,7 @@ def api_ui():
         'swagger-ui.html',
         specs_url=url_for('docs.api_spec'),
         PRODUCTION=env.get_credential('PRODUCTION'),
+        api_key_signup_key=env.get_credential('API_UMBRELLA_SIGNUP_KEY'),
     )
 
 
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -564,6 +564,7 @@ def api_ui():
         specs_url=url_for('docs.api_spec'),
         PRODUCTION=env.get_credential('PRODUCTION'),
         api_key_signup_key=env.get_credential('API_UMBRELLA_SIGNUP_KEY'),
+        api_key_signup_feature_flag=bool(env.get_credential('API_UMBRELLA_SIGNUP_KEY_FEATURE_FLAG', '')),
     )
 
 
--- a/webservices/templates/swagger-ui.html
+++ b/webservices/templates/swagger-ui.html
@@ -2,7 +2,6 @@
 <html>
 <head>
   <title>OpenFEC API Documentation</title>
-
   <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700" rel="stylesheet">
     <link rel="stylesheet" type="text/css" href="{{ swagger_static('dist/main.css') }}" >
     <link rel="icon" type="image/png" href="{{ swagger_static('favicon.ico') }}" />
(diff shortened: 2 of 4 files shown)


jason-upchurch:
replaces #4349 to clean up commits (incorporates review from @lbeaufort )
@fec-jli @pkfec this is ready for your review as well.

webservices/rest.py
```diff
@@ -563,6 +563,7 @@ def api_ui():
         'swagger-ui.html',
         specs_url=url_for('docs.api_spec'),
         PRODUCTION=env.get_credential('PRODUCTION'),
+        api_key=env.get_credential('API_UMBRELLA_SIGNUP_KEY'),
```
pkfec, line 566: the `api_key` variable is very generic. can you rename this variable to be more specific?example api_signup_key

webservices/templates/swagger-ui.html
```diff
@@ -30,14 +29,85 @@
...
+     // actual address will be "noreply@api.data.gov", but this will
+     // change the name of the displayed sender in this fashion:
+     // "{{emailFromName}} <noreply@api.data.gov>".
+     emailFromName: 'openFEC',
+
+     // Provide an example URL you want to show to users after they signup.
+     // This can be any API endpoint on your server, and you can use the
+     // special {{api_key_signup_key}} variable to automatically substitute in the API
+     // key the user just signed up for.
+     // exampleApiUrl: 'https://api.open.fec.gov/v1/legal/search/?api_key_signup_key={{api_key_signup_key}}',
```
lbeaufort, line 65: It looks like the example API call that goes out is still NREL - can you uncomment this line please?

jason-upchurch:
@JonellaCulmer following are the first and second screenshots of the api key signup feature (the first shot was there by default (did not touch).

Screen Shot 2020-05-18 at 10 31 52 AM

Screen Shot 2020-05-18 at 10 31 42 AM

jason-upchurch:
@JonellaCulmer following is a copy of the welcome email:

Hi <user>,

Your API key for <email REDACTED> is:

<api_key REDACTED>
You can start using this key to make web service requests. Simply pass your key in the URL when making a web request. Here's an example:

https://developer.nrel.gov/api/alt-fuel-stations/v1/nearest.json?api_key=REDACTED&location=Denver+CO
For additional support, please contact us. When contacting us, please tell us what API you're accessing and provide the following account details so we can quickly find you:

Account Email: REDACTED
Account ID: REDACTED

JonellaCulmer:
> @JonellaCulmer following are the first and second screenshots of the api key signup feature (the first shot was there by default (did not touch).
> 
> Screen Shot 2020-05-18 at 10 31 42 AM
> 
> Screen Shot 2020-05-18 at 10 31 52 AM

@jason-upchurch Thanks so much for the screenshots. Three quick edits for the form.
1. We include help text that says to include information where "* Required", but none of the fields include an asterisk. Can you include one where appropriate? Or add more "(optional)" text everywhere else.
2. We need to reduce the size of the header for the sign-up text. Currently, it's the same size and prominence as the page title. Can we reduct the text "Sign up for an API key" to 24px?
3. Can we add a second sentence that explains that users can still use the demo key lower on the page? Not sure how to phrase it. Example: Signing up here does not prevent you from trying out the API below using a DEMO_KEY. For those interested in doing XXX, please sign up for an API key. 
cc: @dorothyyeager @jason-upchurch @johnnyporkchops

dorothyyeager:
@JonellaCulmer I'm confused; will the two screenshots eventually end up on the same page. Will  clicking the "sign up for a key" link in the first screenshot lead them to the second screenshot? I think I don't understand enough to make a good recommendation for wording, but here's my shot at it:

"If you just want to try out using the API, you may sign up for a DEMO_KEY instead (link where they'd do that)."

JonellaCulmer:
@dorothyyeager No, they are on the same page in the order Jason uploaded the screenshots. Just imagine the two images are above and below the fold.

dorothyyeager:
I think I'll stick with my suggestion in the last comment, but with the understanding that I am not sure what the difference is that we are trying to explain. Is there a difference between "get an API key here" and the form?

In the meantime, I caught something else that needs fixing on the API page while trying to learn more about it. Maybe it can be worked into this PR? In this screencap:

image

The URL is actually https://www.fec.gov/campaign-finance-data/presidential-map-data/. A redirect is in place so the link works but we should update the link title and URL to be current.

Let me know if you'd rather I put in a separate ticket for this.

JonellaCulmer:
@dorothyyeager If you could put in a separate ticket for link issue you found, that would be perfect.

dorothyyeager:
> @dorothyyeager If you could put in a separate ticket for the link issue you found, that would be perfect.

Done: https://github.com/fecgov/openFEC/issues/4362

webservices/templates/swagger-ui.html
jason-upchurch: There's an issue with the variable passing the api umbrella  key rather than inserting the NEW key. We will work it out in the next iteration!

lbeaufort: 👍 From this page: https://api.data.gov/docs/agency-manual/ it looks like we need to use `{{api_key}}`

```
     // Provide an example URL you want to show to users after they signup.
     // This can be any API endpoint on your server, and you can use the
     // special {{api_key}} variable to automatically substitute in the API
     // key the user just signed up for.
     exampleApiUrl: 'https://api.data.gov/gsa/auctions?api_key={{api_key}}&format=JSON'
```

lbeaufort: Hm, it looks like that's just a different version of above, sorry if that doesn't work.

webservices/templates/swagger-ui.html
```diff
@@ -30,16 +29,102 @@
...
+
+     // Enter the API key you signed up for and specially configured for this
+     // API key signup embed form.
+     apiKey: "{{ api_key_signup_key }}",
+     // document.write(apiKey.toLocaleString( ));
+     // Provide a URL or e-mail address to be used for customer support.
+     //
+     // The format for e-mail addresses can be given as either
+     // 'example@example.com' or 'mailto:example@example.com'.
+     contactUrl: 'APIinfo@fec.gov',
```
pkfec, line 49: The email sent out to the user after successfully signing up for a new API key has contactus email still pointing to `api.data.gov`. It should be apiinfo@fec.gov. if you change the contractUrl value to `mailto:APIinfo@fec.gov`, then the response email shows  fec email email address.

webservices/templates/swagger-ui.html
```diff
@@ -30,16 +29,102 @@
...
+
+     // Provide the name of your developer site. This will appear in the
+     // subject of the welcome e-mail as "Your {{siteName}} API key".
+     siteName: 'openFEC',
+
+     // Provide a sender name for who the welcome email appears from. The
+     // actual address will be "noreply@api.data.gov", but this will
+     // change the name of the displayed sender in this fashion:
+     // "{{emailFromName}} <noreply@api.data.gov>".
+     emailFromName: 'openFEC',
```
pkfec, line 59: emailFromName still shows as `noreply@api.data.gov`. It should be replaced with fec email address. 
Example:"{{fec.gov}} <noreply@apiinfo@fec.gov>"

jason-upchurch: `emailFromName` shows above as `'openFEC`'. This is not the actual email address, this is the "sender name" that the email actually comes from. If we want a different name from openFEC we can add that instead. But the contact email address is given earlier in the code as `'APIinfo@fec.gov'`. Please let me know if you still think we need to immediately change any values.

webservices/templates/swagger-ui.html
jason-upchurch: Thank you @pkfec!! This is GREAT! I will fix it.

webservices/templates/swagger-ui.html
pkfec: If user is coming to FEC API to signup for a key then i think we should have FEC email in sender section. 
Screenshot for your reference.

Screen Shot 2020-05-28 at 2 15 18 PM

jason-upchurch: @pkfec thank you! Yes, I totally agree and believe the concern is already being addressed under this comment: https://github.com/fecgov/openFEC/pull/4359#discussion_r431458954. The `emailFromName` is openFEC however. I think the end result will be what you are expecting (just check that other comment)

jason-upchurch:
@pkfec @lbeaufort @fec-jli @johnnyporkchops this work was blocked by the feature flag component of #4363. With the feature flag enabled and envars now stored in cf, can we merge this and continue addressing the refinement under that issue? Right now this should only `dev`.

pkfec:
> @pkfec @lbeaufort @fec-jli @johnnyporkchops

Thanks @jason-upchurch for working on the basic vanilla version of the APIsignup page. Works great.

webservices/templates/swagger-ui.html
```diff
@@ -30,16 +30,102 @@
...
         SwaggerLayout(
           "{{ specs_url }}",
           document.getElementById("swagger-ui")
         );
+        
+         var apiUmbrella = document.createElement('script'); 
+         apiUmbrella.type = 'text/javascript'; 
+         apiUmbrella.async = true;
+         apiUmbrella.src = 'https://api.data.gov/static/javascripts/signup_embed.js';
+         (document.getElementsByTagName('head')[0] ||
```
johnnyporkchops, line 113: We can remove 
`|| document.getElementsByTagName('body')[0]`
Since we know we have a head in this document

lbeaufort:
We should remove `Get an [API key here](https://api.data.gov/signup/).` when the feature flag is turned on - I'm having trouble finding the UX issues ticket, I can add it there if you like.

webservices/templates/swagger-ui.html
jason-upchurch: removed under issue https://github.com/fecgov/openFEC/pull/4359#pullrequestreview-421086744

webservices/templates/swagger-ui.html
jason-upchurch: Correct syntax to resolve for the FEC's architecture:

```
exampleApiUrl: 'https://api.open.fec.gov/v1/legal/search/?api_key={{ '{{api_key}}' }}'
```

(need to escape the already-used `{{api_key}}` variable)

PR #4400, opened 2020-06-04
[WIP] Feature/4370 change filter fulltext logic

lbeaufort:
## Summary (required)

- Resolves #[_Issue number_]

-

-  

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

### DIFF
--- a/webservices/resources/sched_a.py
+++ b/webservices/resources/sched_a.py
@@ -69,10 +69,6 @@ class ScheduleAView(ItemizedResource):
     filter_multi_start_with_fields = [
         ('contributor_zip', models.ScheduleA.contributor_zip),
     ]
-    query_options = [
-        sa.orm.joinedload(models.ScheduleA.committee),
-        sa.orm.joinedload(models.ScheduleA.contributor),
-    ]
     sort_options = [
         'contribution_receipt_date',
         'contribution_receipt_amount',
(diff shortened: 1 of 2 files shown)


lbeaufort:
Prototyping

webservices/resources/sched_a.py
```diff
@@ -115,7 +111,10 @@ def build_query(self, **kwargs):
                     ),
                     status_code=400,
                 )
-        query = super().build_query(**kwargs)
+        # Do these left joins first for better performance
+        query = self.model.query.\
+            outerjoin(models.ScheduleA.committee, aliased=True).\
```
lbeaufort, line 116: `aliased=True` is not recommended per https://docs.sqlalchemy.org/en/13/orm/query.html

Legacy Features of Query.join()

The Query.join() method currently supports several usage patterns and arguments that are considered to be legacy as of SQLAlchemy 1.3. A deprecation path will follow in the 1.4 series for the following features:


**Automatic aliasing with the aliased=True flag:**
```
session.query(Node).join(Node.children, aliased=True).\
    filter(Node.name == 'some name')
```
Why it’s legacy: the automatic aliasing feature of Query is intensely complicated, both in its internal implementation as well as in its observed behavior, and is almost never used. It is difficult to know upon inspection where and when its aliasing of a target entity, Node in the above case, will be applied and when it won’t, and additionally the feature has to use very elaborate heuristics to achieve this implicit behavior.

Modern calling pattern: Use the aliased() construct explicitly:
```
from sqlalchemy.orm import aliased

n1 = aliased(Node)

session.query(Node).join(Node.children.of_type(n1)).\
    filter(n1.name == 'some name')
```

PR #4493, opened 2020-07-21
Increase number of gunicorn workers

### LINKED ISSUE #4489: Test more gevent workers per app instance
1G app = 4 CPU cores
(`cf ssh api -i 0` to log into one app instance, `cat /proc/cpuinfo` to see cores)
https://docs.gunicorn.org/en/stable/run.html#commonly-used-arguments

Recommended workers is 2xCPUs+1 = 9 workers (currently 4 workers)

**Completion criteria**
- [x] Take a baseline locust test of `dev` at 10 instances, note database CPU and connections start/end
- [x] Update `gunicorn --access-logfile - --error-logfile - --log-level info --timeout 300 -k gevent -w 9 webservices.rest:app` to `-w 9`
- [x] Take an "after" locust test of `dev` , note database and connections CPU start/end

-- Check the number of workers with `ssh api` and `ps aux | grep gunicorn | grep -v grep | wc -l` (subtract one) per https://stackoverflow.com/a/33107962

I couldn't get gunicorn to recognize the `WEB_CONCURRENCY` or `GUNICORN_CMD_ARGS` env vars.

lbeaufort:
## Summary (required)

Resolves #4489

- Update number of `gunicorn` workers to `9` (2*CPU+1) per https://docs.gunicorn.org/en/stable/design.html#how-many-workers
- Remove unused env var `WEB_CONCURRENCY` (gunicorn ignores this - can't figure out why we can't use this env var but it's not being used, so removed it)

Load testing results show improved speed: https://docs.google.com/spreadsheets/d/1DNs0O1YiL7AgHptYd9de-iuUv1c6tZ8ZtFA4Soym9cQ/edit#gid=0

- Manually deploy this branch with `invoke deploy --space dev` (please don't commit to this branch)
- ssh into instance 0 of the api app with `cf ssh api -0` 
- Check number of CPU cores with `cat /proc/cpuinfo`
- Check number of workers with `ps aux | grep gunicorn | grep -v grep | wc -l` (subtract one) per https://stackoverflow.com/a/33107962

-  Concurrency/speed should be less impacted when the database is slow

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,2 +1,2 @@
 python manage.py cf_startup
-gunicorn --access-logfile - --error-logfile - --log-level info --timeout 300 -k gevent -w 4 webservices.rest:app
+gunicorn --access-logfile - --error-logfile - --log-level info --timeout 300 -k gevent -w 9 webservices.rest:app
--- a/manifests/manifest_api_dev.yml
+++ b/manifests/manifest_api_dev.yml
@@ -5,7 +5,6 @@ stack: cflinuxfs3
 buildpack: python_buildpack
 env:
   APP_NAME: fec | api | dev
-  WEB_CONCURRENCY: 4
 services:
   - fec-api-search56
   - fec-redis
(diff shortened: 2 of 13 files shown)


pkfec:
@lbeaufort - Could you update the new worker and instances numbers (to the right) that you mentioned  in your `Increase workers locust testing spreadsheet`?

lbeaufort:
Thanks @pkfec that was a copy/paste oversight. Updated!

PR #4540, opened 2020-08-05
Add user_downloads prefix to downloads bucket

### LINKED ISSUE #4502: Prefix the generated data downloads/exports
**What we're after:**
To better organize and target objects in the bulk downloads bucket, we should put the user generated data downloads/exports into a prefix. 

This would also allow us to set a lifecycle policy to not retain versions in s3 of the deleted files.

Prefix name: user-downloads

**Completion criteria:**
- [x] Prefix the files, without affecting the file name the user sees when downloading
- [x] Ask cloud.gov to set a one or zero day version lifecycle policy for this prefix. 
- [x] Ask cloud.gov to remove the blanket lifecycle policy on the downloads bucket

patphongs:
## Summary

- Resolves #4502 

_Adds the `user_downloads` prefix to the API's user generated downloads path. The name is unaffected.

## How to test the changes

- Deploy to the dev or stage environment to test that downloads are writing into the stage downloads bucket under the `user_downloads` prefix.

-  /downloads/ endpoint

### DIFF
--- a/webservices/tasks/download.py
+++ b/webservices/tasks/download.py
@@ -119,7 +119,8 @@ def get_s3_name(path, qs):
     # TODO: consider base64 vs hash
     raw = '{}{}'.format(path, qs)
     hashed = hashlib.sha224(raw.encode('utf-8')).hexdigest()
-    return '{}.csv'.format(hashed)
+    prefix = 'user-downloads/'
+    return '{}{}.csv'.format(prefix, hashed)
 
 
 def make_bundle(resource):
--- a/tests/test_downloads.py
+++ b/tests/test_downloads.py
@@ -33,7 +33,8 @@ class TestDownloadTask(ApiBaseTest):
     def test_get_filename(self):
         path = '/v1/candidates/'
         qs = '?office=H&sort=name'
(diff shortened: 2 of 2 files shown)


webservices/tasks/download.py
```diff
@@ -119,8 +119,9 @@ def get_s3_name(path, qs):
     # TODO: consider base64 vs hash
     raw = '{}{}'.format(path, qs)
     hashed = hashlib.sha224(raw.encode('utf-8')).hexdigest()
-    return '{}.csv'.format(hashed)
-
+    prefix = 'user_downloads/'
```
lbeaufort, line 122: Can we make the prefix `user-downloads` instead of `user_downloads` because we're using `bulk-downloads`?

✅ Verified files exist with `aws s3api list-objects-v2 --max-items 10 --bucket "<bucket-name>" --prefix=user_downloads`

lbeaufort: If you agree, could you also squash these two commits?

patphongs: Thanks @lbeaufort, I'm all for consistency. Commits have been squashed into 1 as well.

lbeaufort: Thanks @patphongs!

PR #4556, opened 2020-08-13
[MERGE WITH GIT FLOW] Hotfix/4554 fix swagger math error

### LINKED ISSUE #4554: Fix math error in swagger
###  Summary
We introduced a math error when we converted calls per minute to calls per hour. This removes any reference to calls per hour and restores the language as it was previously.

jason-upchurch:
## Summary

- Resolves #4554 

- returns language to indicate 120 calls per minute and removes conversion to other units
- fixes syntax for swagger-ui.html (Thanks @johnnyporkchops )

- check out branch
- run local api
- verify upgraded api key is 120 calls per minute and no longer has the incorrect "7,600 calls per hour"
- I signed up for an api key to test the swagger-ui.html change--you can as well, or check linting (@johnnyporkchops )

-  /webservices/templates/swagger-ui.html
- /webservices/docs.py

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -27,7 +27,7 @@ and [Acceptable Use policy](https://github.com/fecgov/FEC/blob/master/ACCEPTABLE
 
 Signing up for an API key will enable you to place up to 1,000 calls an hour. Each call \
 is limited to 100 results per page. You can email questions, comments or a request to get \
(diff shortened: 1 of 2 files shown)


jason-upchurch:
@pkfec thank you for your review--I spoke with @patphongs and since we begin talking in units of hours, I think the intent to keep talking in the same units is a good one, so I just fixed the math and put the hours back in again (I amended the hotfix you approved--just in case you want to "re-review"). Thank you! cc @fec-jli

PR #4603, opened 2020-09-02
[Please don't merge] Increase API memory in production and stage, remove extra celery-worker instance

### LINKED ISSUE #4578: API memory usage too high
Since we increased gunicorn workers to `9` on 8/11/20 we've seen faster response times ([Pingdom - compare 7/20 to 8/20 and look at peaks since 8/11](https://my.pingdom.com/reports/uptime#check=5315690&daterange=6months&tab=uptime_tab&checkName=PROD-/schedules/schedule_a/)), but also a [sustained higher memory usage in production](https://logs.fr.cloud.gov/app/kibana#/dashboard/App-Metrics?_g=(refreshInterval%3A(pause%3A!t%2Cvalue%3A0)%2Ctime%3A(from%3Anow%2Fy%2Cmode%3Aquick%2Cto%3Anow))). We should investigate application performance if we increase application memory. Previously we were at about 50% memory usage (500 MB of 1GB) and now we're averaging above 100% (cloud foundry lets you go above 100%  - I think it's 150% but I need to double check).

Screen Shot 2020-08-25 at 1 43 24 PM

We should test performance at higher memory, and possibly test fewer instances. Would 8 instances at 2GB perform better than 10 at 1GB? We should also be aware of org memory caps (need to double the memory for zero-downtime deploys). I think our cap is 60GB. I set up this [tracking chart](https://docs.google.com/spreadsheets/d/1Y7j5Qd0h7PCEpdfTbnmrpk2-eILsLhqT7GkiiC-86cs/edit) a while back but it's out of date. We can probably leave disk as-is. Disk is for how much space the app takes up (each one gets up to 6GB but there's no org-level limit) and shouldn't change too much over time. 

**Action items:**

**Before testing day**
- [X] Double-check locust tests to see if any should be added (I think I've modified them in the recent past but didn't commit them - should we consider making permanent changes?)
- [X] Locust tests need to be updated for the download key. Put in a new PR: https://github.com/fecgov/openFEC/pull/4604
- [x] Pick day to do testing, let cloud.gov know - planned for 9/2

**Testing day**
Locust results: https://docs.google.com/spreadsheets/d/1RDghse7ILJDbizCikNSeVPfTd8XYWmmWosjYFDSR14w/edit#gid=0
- [x] Scale up `stage` db using these instructions: https://github.com/fecgov/openFEC/issues/4327#issuecomment-634056901
- [x] Do a manual deploy to `stage` with memory to match production
- [x] Test "before" performance using locust, see if you can get memory up to 100% (track with Kibana), note requests/second, users/hatch rate
- [x] Test 1.5G memory
- [x] Check on CPU cores (still 4 at 1.5G, was 4 at 1G) to see if `gunicorn` workers are still at the appropriate amount. If we can up the gunicorn workers, we could reduce the number of app instances ~Test again if updating workers~
- [x] Test 2G
- [x] Check on CPU cores (still 4 at 2G, was 4 at 1Gto see if `gunicorn` workers are still at the appropriate amount. If we can up the gunicorn workers, we could reduce the number of app instances ~Test again if updating workers~
~Test fewer application instances~ will reduce concurrency capability
- [x] Determine memory needs and ideal number of instances. Do we need to keep at 50% or would 75% do? 
➡️  Settling at 60% with 1.5G
- [x] Scale `stage` back down
- [x] Update the [tracking chart](https://docs.google.com/spreadsheets/d/1Y7j5Qd0h7PCEpdfTbnmrpk2-eILsLhqT7GkiiC-86cs/edit), don't go over org limit 
- [x] Update memory allocation in the manifest files

**Completion criteria:**
- [x] API memory usage around 50%
➡️  Settling at 60% with 1.5G. Setting memory to 2G didn't have noticeable performance improvements or additional CPUs, and we're running low on org memory.

lbeaufort:
## Summary (required)

- Resolves #4578

Increase API memory in production and stage to 1.5G, remove extra celery-worker instance in `dev`.

Note: Memory is in MB (1500M instead of 1.5G) because the `cf cli` didn't take decimals until version 6.33.0 and we're on 6.32.0. When I tested 6.52.0 there were still issues - might be with the `zero-downtime-push` but it didn't seem worth it to muck around with things in there when a comment can address any confusion.

## Reviewers
- One developer (required) additional developers optional

- Test in stage with a manual deploy `invoke deploy --space stage`. Please don't commit to this branch, thanks!

-  API available memory

### DIFF
--- a/manifests/manifest_api_prod.yml
+++ b/manifests/manifest_api_prod.yml
@@ -1,6 +1,6 @@
 ---
 path: ../
-memory: 1G
+memory: 1500M # 1.5G
 stack: cflinuxfs3
 buildpack: python_buildpack
 env:
--- a/manifests/manifest_api_stage.yml
+++ b/manifests/manifest_api_stage.yml
@@ -1,6 +1,6 @@
 ---
 path: ../
-memory: 1G
+memory: 1500M # 1.5G
 stack: cflinuxfs3
 buildpack: python_buildpack
 env:
--- a/manifests/manifest_celery_worker_dev.yml
+++ b/manifests/manifest_celery_worker_dev.yml
@@ -12,7 +12,7 @@ services:
   - fec-s3-dev
 applications:
   - name: celery-worker
-    instances: 2
+    instances: 1
     memory: 1G
     disk_quota: 1G
     no-route: true


pkfec (approved):

lbeaufort:
Closing in favor of https://github.com/fecgov/openFEC/pull/4605 which targets the release

PR #4620, opened 2020-09-16
[WIP] Feature/4414 use union for 2 year period v2

### LINKED ISSUE #4414: Use "UNION ALL+limit" for full 45 years SA search  by using "two_year_transaction_period"
**Summary:**
Code change for "All year Search" to "use partitions cluster index  and "limit"" , in case of system crash by accumulated long sqls and even no any result to clients of 19 mins sql run. new sql 36 secs

**Completion criteria:**
- [ ] Schedule A uses "union all" for two-year transaction periods instead of "or"

**Problem:** 
Why we need to change this?
1) run 19 mins that endUser got nothing and
2) high risk to  accumulate to cause system crash 
This is the top 1 slow sqls at production:
image
and we saw cpu spikes.

**The exact sql code is** 
SELECT * FROM  disclosure.fec_fitem_sched_a LEFT OUTER JOIN ofec_committee_history_mv AS ofec_committee_history_mv_1 ON disclosure.fec_fitem_sched_a.cmte_id = ofec_committee_history_mv_1.committee_id AND disclosure.fec_fitem_sched_a.two_year_transaction_period = ofec_committee_history_mv_1.cycle LEFT OUTER JOIN ofec_committee_history_mv AS ofec_committee_history_mv_2 ON disclosure.fec_fitem_sched_a.clean_contbr_id = ofec_committee_history_mv_2.committee_id AND disclosure.fec_fitem_sched_a.two_year_transaction_period = ofec_committee_history_mv_2.cycle
WHERE  disclosure.fec_fitem_sched_a.is_individual = true 
AND disclosure.fec_fitem_sched_a.contributor_name_text @@ to_tsquery('ActBlue:*') 
ORDER BY disclosure.fec_fitem_sched_a.contb_receipt_dt DESC, disclosure.fec_fitem_sched_a.sub_id DESC
LIMIT 30

**Analysis and Suggestions:
Analysis:**
The issue is searching for "all" yeas from 1975 to 2020 of schedule_a  and the sql is not using our  index at all. full table scan 332,757,433 rows. We have the partitions from 1975 (23 partitions now). I will be much better to use the fec_fitem_sched_a.two_year_transaction_period  index as we are using for others sqls.
There is only 30 rows return but current it has to scan 332millions rows.

There is some ideas on this, but it may take more time before the election of this year.

Suggestions:
1) For this kind of "ALL Years' search, add this in the sql disclosure.fec_fitem_sched_a.two_year_transaction_period IN (2020) 
This is to use the cluster index of the partitions we have now.

2)use "UNION all" to union all of the years, to keep the same sql return.

3)Code suggestion as below (but it needs code change).

Thanks for the review.

lbeaufort:
## Summary (required)

- Resolves #4414

-

-  

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

### DIFF
--- a/tests/integration/test_tsvector_generation.py
+++ b/tests/integration/test_tsvector_generation.py
@@ -425,11 +425,12 @@ class TriggerTestCase(common.BaseTestCase):
                 'contbr_employer': n,
                 'sub_id': 9999999959999999980 + i,
                 'filing_form': 'F3',
+                'two_year_transaction_period': 2020,
             }
             insert = (
                 "INSERT INTO disclosure.fec_fitem_sched_a "
-                + "(contbr_employer, sub_id, filing_form) "
-                + " VALUES (%(contbr_employer)s, %(sub_id)s, %(filing_form)s)"
+                + "(contbr_employer, sub_id, filing_form, two_year_transaction_period) "
+                + " VALUES (%(contbr_employer)s, %(sub_id)s, %(filing_form)s, %(two_year_transaction_period)s)"
             )
             connection.execute(insert, data)
         manage.refresh_materialized(concurrent=False)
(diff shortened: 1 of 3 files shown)


lbeaufort:
Using this to show incremental work, but https://github.com/fecgov/openFEC/pull/4621 has cleaner commits

PR #4622, opened 2020-09-17
Upgraded flask-cors to v3.0.9

patphongs:
## Summary

- Resolves #4611 

_Upgraded flask-cors to v3.0.9._

flask_cors is imported into our rest.py. Tested the `@app.route` for swagger docs and it seems to still function ok.

## How to test the changes locally

- checkout this branch
- `pip install -r requirements.txt` 
- `./manage.py runserver`

PR #4631, opened 2020-09-23
Revert "Upgraded flask-cors to v3.0.9"

lbeaufort:
Reverts fecgov/openFEC#4622 until we can figure out what's going on

```
/home/vcap/deps/0/python/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
```


jason-upchurch:
problem in greenlet dependency of gevent (from locustio). Closing reversion.

PR #4666, opened 2020-10-20
Add sponsor_candidate_id to endpoint: /committees/

hcaofec:
## Summary (required)

- Resolves #4613

_Add filter sponsor_candidate_id to endpoint: /committees/_

## Reviewers
- Two developers (required) additional developer optional

## How to test the changes locally

- Download feature branch
- Run `pytest`
- Set SQLA_CONN to dev database and start endpoint on local
- Compare these endpoints, and sponsor_candidate_ids can be found in endpoint output.

Before change:
https://api.open.fec.gov/v1/committees/?page=1&per_page=20&sort=name&sort_nulls_last=false&api_key=DEMO_KEY&committee_id=C00692715&sort_hide_null=false&sort_null_only=false

After change:
http://127.0.0.1:5000/v1/committees/?per_page=20&page=1&sort_null_only=false&committee_id=C00692715&sort_nulls_last=false&sort=name&sort_hide_null=false

Screen Shot 2020-10-20 at 11 32 04 AM

- Run this endpoint to test filter

http://127.0.0.1:5000/v1/committees/?sort_null_only=false&page=1&sort_nulls_last=false&sort_hide_null=false&sort=name&sponsor_candidate_id=H2NV02395&per_page=20

http://127.0.0.1:5000/v1/committees/?sort_null_only=false&page=1&sort_nulls_last=false&sort_hide_null=false&sort=name&sponsor_candidate_id=H8OK01157&per_page=20

## Impacted areas of the application
These endpoints will have sponsor_candidate_ids in the output:
/committee/{committee_id}/
/committees/
/candidate/{candidate_id}/committees/

PR #4680, opened 2020-11-03
Add sponsor id to committee history endpoint

### LINKED ISSUE #4612: Update committee/{committee_id}/history and related endpoints  to add sponsor ID
**What we're after:** 
Update these endpoints to return with sponsor candidate ID.
'/committee/<committee_id>/history/',
'/committee/<committee_id>/history/<int:cycle>/',
'/candidate/<candidate_id>/committees/history/',
'/candidate/<candidate_id>/committees/history/<int:cycle>/'

**Completion criteria:**
- [x] Endpoints return with sponsor ID

hcaofec:
## Summary (required)

- Resolves #4612 

_Updated these endpoints to include sponsor_candidate_id_

'/committee/<committee_id>/history/',
'/committee/<committee_id>/history/cycle/',
'/candidate/<candidate_id>/committees/history/',
'/candidate/<candidate_id>/committees/history/cycle/'

## Reviewers
Two developers (required), additional developer optional

- Download feature branch
- run `pytest` to make sure all tests pass
- Set SQLA_CONN to dev database, and run `./manage.py runserver`
- Check these endpoints and make sure sponsor_candidate_ids is in the output

http://127.0.0.1:5000/v1/committee/C00432104/history/?election_full=true&sort_null_only=false&sort_hide_null=false&sort=-cycle&per_page=20&page=1&sort_nulls_last=false

http://127.0.0.1:5000/v1/committee/C00497677/history/2016/?election_full=true&sort_null_only=false&sort_hide_null=false&sort=-cycle&per_page=20&page=1&sort_nulls_last=false

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -278,3 +278,35 @@ class TestFilterOther(ApiBaseTest):
                 if 'Report' in each.summary and 'IE' not in each.summary
             ),
         )
+
+
+class TestFilterOverlap(ApiBaseTest):
+    def setUp(self):
+        super(TestFilterOverlap, self).setUp()
+        self.committees = [
+            factories.CommitteeFactory(committee_id='C001', sponsor_candidate_ids=['S001']),
+            factories.CommitteeFactory(committee_id='C002', sponsor_candidate_ids=['H001']),
+        ]
+
+    def test_filter_overlap(self):
+        """Test the filter that compares whether two arrays have elements in common."""
+        query_committees = filters.filter_overlap(
+            models.Committee.query,
+            {'sponsor_candidate_id': ['H001']},
+            CommitteeList.filter_overlap_fields,
+        )
+        self.assertEqual(
+            set(query_committees.all()),
+            set(each for each in self.committees if each.sponsor_candidate_ids == ['H001']),
+        )
+
+    def test_filter_overlap_exclude(self):
+        query_committees = filters.filter_overlap(
+            models.Committee.query,
+            {'sponsor_candidate_id': ['-H001']},
+            CommitteeList.filter_overlap_fields,
+        )
+        self.assertEqual(
+            set(query_committees.all()),
+            set(each for each in self.committees if each.sponsor_candidate_ids == ['S001']),
+        )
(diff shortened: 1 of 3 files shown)


tests/test_filters.py
```diff
@@ -278,3 +278,35 @@ def test_filter_fulltext_exclude(self):
...
+        )
+        self.assertEqual(
+            set(query_committees.all()),
+            set(each for each in self.committees if each.sponsor_candidate_ids == ['H001']),
+        )
+
+    def test_filter_overlap_exclude(self):
+        query_committees = filters.filter_overlap(
+            models.Committee.query,
+            {'sponsor_candidate_id': ['-H001']},
```
lbeaufort, line 306: Should this be `sponsor_candidate_ids`?

lbeaufort: I see now that the field in the db/model is `sponsor_candidate_ids` but the filter is `sponsor_candidate_id`, so disregard please :)

lbeaufort (approved):
Looks great, @hcaofec! Thank you for adding tests for `filter_overlap`! Confirming that you intended not to add `sponsor_candidate_id` as a filter for these endpoints.

Side note: I looked back at https://github.com/fecgov/openFEC/pull/4666 and was wondering how the `filter_overlap` would handle multiple arguments, i.e. `sponsor_candidate_id=1` AND `sponsor_candidate_id=2`  -  would it find the row with `sponsor_candidate_ids=[1,2]`? It looks like that doesn't matter for `ofec_committee_detail_mv` because there isn't any relevant data, but we might want to keep that in mind for future usage, or add test coverage for that case in the future. 

```
select sponsor_candidate_ids, * from ofec_committee_detail_mv 
where array_length(sponsor_candidate_ids,1) >1 
limit 10  
```

Thanks!

jason-upchurch (approved):
Looks great @hcaofec ! Thanks for your work on this, and for explaining the behavior of overlap in the theoretical case when multiple elements of the array are present! 👍

fec-jli (approved):
great work!

PR #4683, opened 2020-11-04
Fix incorrect counts when sort_hide_null=true

### LINKED ISSUE #4626: Counts for all endpoints are incorrect when 'sort_hide_null=True'
### User story
As a user, I want the endpoints to return correct counts when I use `sort_hide_null=True`  because  I want accurate  results.

### Expected behavior
When we use  `sort_hide_null=True`, it's supposed to hide rows  that have  null values for whatever we are sorting on (e.g., `first_file_date`
(We want count to reflect the number of rows that are shown)

### Actual behavior
- correctly hides the row
- count reflects rows that hidden as well 

Examples  of incorrect behavior:

https://api.open.fec.gov/v1/candidates/?api_key=DEMO_KEY&sort_hide_null=true&sort_nulls_last=true&is_active_candidate=true&sort=-first_file_date 
is the same count as 
https://api.open.fec.gov/v1/candidates/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=true&is_active_candidate=true&sort=-first_file_date&per_page=100 ➡️ when I confirmed there are records with null `first_file_date` values. 

https://api.open.fec.gov/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=true&sort_nulls_last=false&data_type=processed&committee_id=C00671339&line_number=F3-11AI&sort=-contribution_receipt_date&per_page=100&two_year_transaction_period=2020 
➡️ shows count of 3 and no results

This applies to ItemizedResource and ApiResource endpoints. This means the counts are wrong, and makes pagination really tough because it looks like there are more records than there are. I'm not sure this impacts front-end users. 
I put some tests in the `4626-tests` branch: https://github.com/fecgov/openFEC/pull/4627/commits/ff8ab0ac117ced0444187fd98f75c936f5c174b5

**Completion criteria**
- [ ] Counts update for `sort_hide_null=True` 
- [ ] Double-check counts update for `sort_nulls_only=True` 
- [ ] Need to add test cases to check count behavior

### Helpful additions:
This is an example of an API resource being ok (updating counts correctly):
https://api.open.fec.gov/v1/candidates/?api_key=DEMO_KEY&sort_nulls_only=true&sort_nulls_last=true&is_active_candidate=true&sort=-first_file_date&per_page=100

jason-upchurch:
## Summary (required)

- Resolves #4626 

- adds function to determine whether count will be estimated or not (this avoids using `get_count` for the `is_estimate` return value which causes a count)
- use this function to:
  1. set `count=None` when the count is not going to be estimated--this lets the internal marshmallow pagination determine the count
  2. pass the estimated count along when `is_estimate=False`
- modifies signature of `join_committee_queries` so that total is not computed as an intermediate value

- two reviewers please.
- check out branch
- pytest
- please feel free to add additional relevant pytests for correct behavior
- performance tests are included below (the fix corrects the bug without adding performance overhead for the indicated test cases). The method was:
  1. day 0: determine a relevant subset of queries that would not be  estimates
  2. day 1: run queries on `dev`
  3. day 2: deploy PR branch to `dev` space. Rerun queries. This lag is necessary because caching preserves the original queries until an overnight refresh per @fec-jli 

```
# Schedule A
START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/schedules/schedule_a/?contributor_occupation=attorney&contributor_type=individual&api_key=DEMO_KEY&two_year_transaction_period=2020&recipient_committee_type=P&sort=-contribution_receipt_date&is_individual=true&sort_null_only=false&sort_hide_null=true&per_page=20" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":325649 at 247 seconds
# fix: "count":325649 at 248 seconds

START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/schedules/schedule_a/efile/?per_page=20&sort=-contribution_receipt_date&api_key=DEMO_KEY&sort_nulls_last=false&contributor_state=NJ&contributor_state=CA&sort_null_only=false&page=1&contributor_occupation=attorney&sort_hide_null=false" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":420037 at 125 seconds
# fix: "count":420542 at 127 seconds

# Candidates
START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/candidates/?api_key=DEMO_KEY&sort=name&page=1&sort_null_only=false&sort_nulls_last=false&sort_hide_null=false&per_page=20" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":43213 at 1 second
# fix: "count":43213 at 0 seconds

# Schedule B
START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/schedules/schedule_b/?min_amount=2000&per_page=20&sort=-disbursement_date&api_key=DEMO_KEY&max_amount=5000&sort_null_only=false&spender_committee_designation=J&spender_committee_designation=P&spender_committee_designation=U&spender_committee_org_type=C" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":309298 at 21 seconds
# fix: "count":309298 at 28 seconds

# Committees
START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/committees/?per_page=20&sort=name&api_key=DEMO_KEY&sort_nulls_last=false&sort_null_only=false&page=1&year=2016" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":14435 at 1 second
# fix: "count":14435 at 1 second

# Schedule E
START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/schedules/schedule_e/?per_page=20&sort_nulls_last=false&sort_hide_null=false&sort=-expenditure_date&api_key=DEMO_KEY&cycle=2020&sort_null_only=false" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":102198 at 12 seconds
# fix: "count":102380 at 9 seconds

START=$(date +%s)
curl -X GET "https://fec-dev-api.app.cloud.gov/v1/schedules/schedule_e/efile/?support_oppose_indicator=S&per_page=20&sort=-expenditure_date&api_key=DEMO_KEY&sort_nulls_last=false&sort_null_only=false&page=1&sort_hide_null=false" -H "accept: application/json"
END=$(date +%s)
DIFF=$(( $END - $START ))
echo "It took $DIFF seconds"
# "count":59299 at 5 seconds
# fix: "count":59463 at 5 seconds
```


lbeaufort (commented):
It looks like there are some scenarios where counts are off: compare these two:

http://127.0.0.1:5000/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=false&committee_id=C00589309&committee_id=C00660795&sort=-contribution_receipt_date&per_page=30&is_individual=true

Count: 25427605080

https://api.open.fec.gov/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=false&committee_id=C00589309&committee_id=C00660795&sort=-contribution_receipt_date&per_page=30&is_individual=true

Count: 14602

Here are some examples I checked: https://docs.google.com/spreadsheets/d/1rkVnmuMUzlBCgOE3ZsCDJC7AqXWI2VOhC57RvJcdasI/edit#gid=0

jason-upchurch:
> It looks like there are some scenarios where counts are off: compare these two:
> 
> http://127.0.0.1:5000/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=false&committee_id=C00589309&committee_id=C00660795&sort=-contribution_receipt_date&per_page=30&is_individual=true
> 
> Count: 25427605080
> 
> https://api.open.fec.gov/v1/schedules/schedule_a/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=false&committee_id=C00589309&committee_id=C00660795&sort=-contribution_receipt_date&per_page=30&is_individual=true
> 
> Count: 14602
> 
> Here are some examples I checked: https://docs.google.com/spreadsheets/d/1rkVnmuMUzlBCgOE3ZsCDJC7AqXWI2VOhC57RvJcdasI/edit#gid=0

@lbeaufort thank you for discovering this--yes, the counts as reported under my "fix" are wildly off because I was enforcing an estimated count, however, the join_committee_queries seems to "build up" the count.

I fixed this by assuming the counts in `build_committee_query` and `join_committee_queries` are correct as written, so that when this branch is hit in the `get` method, I leave the counts however they were computed. But if we don't hit that branch (conditional), then I stick with checking if the count is going to be an estimate or not and handling accordingly.

Part of the challenge on this was keeping a simple solution while knowing that the `query` is modified all the way through to the `utils.fetch_page` method (additional filters, such as `sort_hide_null` potentially get added on at this stage, so the trick was to ensure that the count that accompanied the call was `None` where appropriate (exact count, or, in this case, when we build up the query through the join methods above).

That being said, I think we could stand to reduce some technical debt with counting in general, but mainly I'd like to consider some additional test cases that might tease out the examples you came up with until then. If you want to talk about adding some specific cases I'd appreciate it, because passing  `pytest` is nice, but the automated tests should also be able to catch the types of cases you found (I could use your help here).

Thanks again! cc @pkfec @fec-jli

lbeaufort (commented):
@jason-upchurch using the built-in `marshmallow_pagination` count to fix this issue is a great idea! Noting though, that I think we lose primary-key only count functionality for exact counts when make this change, but since these exact-count result sets will be smaller I think that's ok.

Your changes look good to me, thanks!

A comment on style:
If I'm understanding your changes, I think we can use the existing `get_count` function instead of creating the `is_estimated_count` which repeats much of its logic. Another disadvantage of the new function seems to be that for estimated counts, we're calling both `is_estimated_count` and later `get_count` for the same (estimated) query, which causes us to run the `EXTRACT ANALYZE` SQL query twice. Maybe something like this: https://github.com/fecgov/openFEC/pull/4706/files?

By returning out of `get_count` when we've determined whether the count is an estimate, we don't run the actual query. Let me know if you'd like to discuss!

jason-upchurch:
> @jason-upchurch using the built-in `marshmallow_pagination` count to fix this issue is a great idea! Noting though, that I think we lose primary-key only count functionality for exact counts when make this change, but since these exact-count result sets will be smaller I think that's ok.
> 
> Your changes look good to me, thanks!
> 
> A comment on style:
> If I'm understanding your changes, I think we can use the existing `get_count` function instead of creating the `is_estimated_count` which repeats much of its logic. Another disadvantage of the new function seems to be that for estimated counts, we're calling both `is_estimated_count` and later `get_count` for the same (estimated) query, which causes us to run the `EXTRACT ANALYZE` SQL query twice. Maybe something like this: https://github.com/fecgov/openFEC/pull/4706/files would allow us to not repeat the code or the SQL call?
> 
> By returning out of `get_count` when we've determined whether the count is an estimate, we don't run the actual counts query, just the `EXTRACT ANALYZE`. Let me know if you'd like to discuss!

Thanks for your review @lbeaufort

regarding the style comment and counting:
using `is_estimated_count` avoids counting when the count is **not** an estimate, which would lead to O(2n)  (still=O(n)  though!). Were we to use `get_counts` and make use of the `is_estimate` parameter, it would perform the count when `is_estimate` is `False`,  and then  again later when we declare `count=None`. My thinking here is that the estimated count (extract analyze) was less expensive than the actual count, but we can discuss if that is not the case.

regarding the performance for primary key: 
I ran tests for cases where 300,000 (not estimate) or so results were returned and saw  no performance difference in the pk_for_count vs  not, but I didn't exhaustively test every combination (this would be a good use case for https://github.com/fecgov/openFEC/issues/4700, so that we can understand performance issues with concrete metrics)

Thanks again for your review of this work

lbeaufort:
@jason-upchurch thanks for explaining, that makes sense to me. I see how the double estimate is preferable to double-counting.

To address the duplicated code between `is_estimate` and `get_count`, we could pull out the shared and duplicated code into functions, something like this: https://github.com/fecgov/openFEC/pull/4706/files. With this change, both functions use the `get_estimated_count` function and `get_count` can use the functionality from `is_estimate`.

Let me know if you'd like to discuss, otherwise changes look good to me!

jason-upchurch:
> @jason-upchurch thanks for explaining, that makes sense to me. I see how the double estimate is preferable to double-counting.
> 
> To address the duplicated code between `is_estimate` and `get_count`, we could pull out the shared and duplicated code into functions, something like this: https://github.com/fecgov/openFEC/pull/4706/files. With this change, both functions use the `get_estimated_count` function and `get_count` can use the functionality from `is_estimate`.
> 
> Let me know if you'd like to discuss, otherwise changes look good to me!

I support using this bug fix to refactor and will work on incorporating your suggested changes--thanks for helping me pay down the technical debt!

webservices/common/counts.py
```diff
@@ -28,13 +27,46 @@ def get_count(resource, query):
         estimated_count = extract_analyze_count(rows)
         if estimated_count > resource.estimated_count_threshold:
             is_estimate = True
-            return estimated_count, is_estimate
+            return is_estimate
+    return False
+
+def is_estimated_count(resource, query):
```
lbeaufort, line 33: Is `is_estimated_count` defined twice?

lbeaufort (approved):
Thanks for your work on this, @jason-upchurch!

fec-jli (approved):
test it locally, works as expected.
It is fine to call get_estimated_count() twice instead of get_count() twice.
Thanks the performance test. We are good to go.
Great job!

Issue #4693, opened 2020-11-23
Update ReadMe to modify ES 7 task commands 


fec-jli:
After upgrading ES7, update the ReadMe documentation.
Ref: [https://github.com/fecgov/openFEC/pull/4657](https://github.com/fecgov/openFEC/pull/4657)

**Completion criteria:**
- [ ] Readme documentation is updated to reflect ES 7 upgrade

fec-jli:
done, close

PR #4732, opened 2020-12-29
Update README for ES7 upgrade

fec-jli:
## Summary (required)
After upgrading ES7, we need update the README file.
Ref ticket #4693 

Reviewer:
One developer is good.


pkfec (approved):
@fec-jli Thanks for updating the readme file with the latest elasticsearch service management commands.

PR #4760, opened 2021-02-08
4744-show-entity-names-for-individual-types

pkfec:
## Summary (required)

In entity table, `name` for a individual  type could be empty . In such scenario, the `name`  is populated by concatenating the prefix, first & last names and suffix. 

- Resolves #4744 

## Reviewers
One developer required, @patphongs  is optional. 
(@patphongs i am not sure if you have ES setup on your local, that's why i put you as optional. if you have the setup already, please feel free to test this PR)

_Include a summary of proposed changes._
**In advisory_opinions.py:** 
- Update the AO_ENTITIES SQL to include the prefix, firstname, lastname and suffix columns. 
- Add logic to populate the name column for individual entity type. 

**Add  test_advisory_opinions.py :**
- Add a test case for individual entity type.
## Screenshots:

**Production (Before fix):**
Screen Shot 2021-02-08 at 9 59 26 AM

**Local (After fix):**
Screen Shot 2021-02-08 at 9 59 15 AM

## How to test the changes locally

- checkout branch:  `feature/4744-show-entity-names-for-individual-types`
- run local elasticsearch service: [follow local ES service setup instructions ](https://github.com/fecgov/openFEC/wiki/Elasticsearch-7.x.0-setup-and-test-locally)
- load ao 2020-03 onto your local elasticsearch service: run `python manage.py load_advisory_opinions -f 2020-03`
- Under entities section, `name` for the individual type shows now:  `http://localhost:9200/docs/_doc/2020-03` 

Screen Shot 2021-02-08 at 9 26 57 AM

-  **Before fix (In Production)**:
Screen Shot 2021-02-08 at 9 22 45 AM

## Impacted areas of the application
List general components of the application that this PR will affect:

-  Legal Search/ Advisory Opinions

Issue #4443, opened 2020-07-01
Entities missing from some AOs


lbeaufort:
The contractor is working on this. 

Here’s the query:
```
SELECT 
     ao_parsed.ao_id, 
     ao_parsed.ao_no, 
     ao_parsed.name, 
     ao_parsed.summary, 
     ao_parsed.req_date, 
     ao_parsed.issue_date, 
     ao.stage, 
     entities.*
    FROM aouser.aos_with_parsed_numbers ao_parsed 
    INNER JOIN aouser.ao ao 
     ON ao_parsed.ao_id = ao.ao_id 
    LEFT JOIN (SELECT
        p.ao_id,
        e.name,
        et.description AS entity_type_description,
        r.description AS role_description
    FROM aouser.players p
    INNER JOIN aouser.entity e USING (entity_id)
    INNER JOIN aouser.entity_type et ON et.entity_type_id = e.type
    INNER JOIN aouser.role r USING (role_id)
    ) as entities ON ao_parsed.ao_id = entities.ao_id 
    WHERE entities.ao_id IS NULL
    ORDER BY ao_parsed.ao_year, ao_parsed.ao_serial limit 100  
```

| ao_no  |
|---------|
| [1975-76](https://www.fec.gov/data/legal/advisory-opinions/1975-76/#entities) |
| [1976-64](https://www.fec.gov/data/legal/advisory-opinions/1976-64/#entities) |
| [1977-36](https://www.fec.gov/data/legal/advisory-opinions/1977-36/#entities) |
| [1978-27](https://www.fec.gov/data/legal/advisory-opinions/1978-27/#entities) |
| [1987-13](https://www.fec.gov/data/legal/advisory-opinions/1987-13/#entities) |
| [2016-21](https://www.fec.gov/data/legal/advisory-opinions/2016-21/#entities) |
| [2018-11](https://www.fec.gov/data/legal/advisory-opinions/2018-11/#entities) |
| [2018-12](https://www.fec.gov/data/legal/advisory-opinions/2018-12/#entities) |
| [2018-15](https://www.fec.gov/data/legal/advisory-opinions/2018-15/#entities) |
| [2019-07](https://www.fec.gov/data/legal/advisory-opinions/2019-07/#entities) |
| [2019-10](https://www.fec.gov/data/legal/advisory-opinions/2019-10/#entities) |
| [2019-14](https://www.fec.gov/data/legal/advisory-opinions/2019-14/#entities) |
[...]

PR #4762, opened 2021-02-12
fix-ao-data-load-issue

pkfec:
## Summary (required)

Data reload tasks are failing - `python manage.py load_advisory_opinions` or `cf run-task api --command "python manage.py refresh_current_legal_docs_zero_downtime" -m 4G --name refresh_data` due to NULL values in prefix, suffix and firstname columns in the entity table. 
Use the COALESCE function and check for empty or null values on prefix, suffix, first and lastname columns on the entity table and replace with an empty string if found any. 

- ~Resolves~ Related #4443 

- In advisory_opinons.py,  AO_ENTITIES SQL:  check for empty or null values on prefix, suffix, first and lastname columns and replace with an empty string if found any.
- Add a test case with empty/null values in the prefix,suffix columns for individual enitity.

- checkout branch `feature/fix-ao-individual-entity-data-load-issue`
- export `$SQLA_CONN` to point to Dev DB
-  load AO's onto local elasticsearch service : run `python manage.py load_advisory_opinions` or `python manage.py load_advisory_opinions -f 1975-05` (Dont have to load all the AO's. Quit the process in the middle)
- Ex 1: 1975-05 (suffix is NULL)
- Ex 2: 2014-02 (prefix, suffix and firstname is NULL)
- What to expect: AO's should load onto local elasticsearch service without any errors in the terminal
- **Before the fix:** AO's reload task fails with an error. Screenshot attached below:
Screen Shot 2021-02-12 at 5 47 00 AM

- **After the fix**: AO's reload task run OK without any errors. Screenshot attached below:
Screen Shot 2021-02-12 at 5 47 33 AM

-  Legal Search/Advisory_Opinions
-  Data reload tasks are failing - `python manage.py load_advisory_opinions` or `cf run-task api --command "python manage.py refresh_current_legal_docs_zero_downtime" -m 4G --name refresh_data`
## Related PRs
#4760

### DIFF
--- a/tests/integration/test_advisory_opinions.py
+++ b/tests/integration/test_advisory_opinions.py
@@ -129,6 +129,35 @@ class TestLoadAdvisoryOpinions(BaseTestCase):
             "type": "Individual"}]
 
     @patch("webservices.legal_docs.advisory_opinions.get_bucket")
+    def test_ao_with_null_values_entity_individual(self, get_bucket):
+        expected_entity = {
+            "role": "Commenter",
+            "name": "Tom Dolan",
+            "type": "Individual",
+        }
+        expected_ao = {
+            "type": "advisory_opinions",
+            "no": "2020-01",
+            "name": "An AO name",
+            "summary": "An AO summary",
+            "request_date": datetime.date(2016, 6, 10),
+            "issue_date": datetime.date(2016, 12, 15),
+            "documents": [],
+            "requestor_names": [],
+            "requestor_types": [],
+            "entities": [expected_entity],
+        }
+        self.create_ao(2, expected_ao)
+        # create an individual entity by passing None in prefix and suffix columns
+        self.create_entity_individual(2, 456, "", 15, 2, None, "Tom", "Dolan", None)
+
+        actual_ao = next(get_advisory_opinions(None))
+        assert actual_ao["entities"] == [
+            {"role": "Commenter",
+            "name": " Tom Dolan ",
+            "type": "Individual"}]
+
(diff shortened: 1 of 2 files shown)


fec-jli:
test on local, upload to ES work without error. good.

have a question here:
` "name": row["prefix"] + " " + row["firstname"] + " " + row["lastname"] + " " + row["suffix"],
`
are these four columns(prefix, firstname, lastname, suffix) mandatory?
can these be all null?
if firstname=null and lastname=null, then name=prefix+ "two white spaces" +suffix ?

you can think about use string.join() to concatenate these 4 value.

pkfec:
> test on local, upload to ES work without error. good.
> 
> have a question here:
> `"name": row["prefix"] + " " + row["firstname"] + " " + row["lastname"] + " " + row["suffix"],`
> are these four columns(prefix, firstname, lastname, suffix) mandatory?
> can these be all null?
> if firstname=null and lastname=null, then name=prefix+ "two white spaces" +suffix ?
> 
> string.join() maybe do something?

@fec-jli thanks for taking time to review this pr. Yes, the four columns may or may not be NULL and on the DB. I have came across AO with just firstname column with value and NULL values on other columns. And in others prefix, suffix are NULL and  first and lastname column with some values. And in the DB, the entity table is setup such that none of these fields are REQUIRED.

pkfec:
Adding spaces btw the values gives more visibility than squishing all the columns using the join() function.

Screen Shot 2021-02-16 at 11 20 03 AM

lbeaufort (approved):
Looks great!

fec-jli (approved):
good. Thanks

PR #5022, opened 2022-01-10
Change idx to row number in ofec_sched_b_aggregate_recipient_mv.

### LINKED ISSUE #5016: Change column idx in ofec_sched_b_aggregate_recipient_mv to use row_number
### What we’re after

To make sure the value of idx is unique, we need to change this column to use row_number

### Action item(s)
- [ ] create migration file

### Completion criteria
- [ ] idx column is using row_number to populate its value

fec-jli:
## Summary (required)
To make sure the value of idx is unique, This PR will change this column to use row_number.

- Resolves #5016 

### Required reviewers
one developer

## How to test
- Checkout branch and point to aurora 13 dev db
- `flyway migrate`
- `pytest`
-  `./manage.py runserver`


hcaofec (approved):
Looks good. So approved.

PR #5224, opened 2022-08-24
pin setuptools to 65.1.1

cnlucas:
## Summary (required)

- Fix the current failing builds due to an issue with the most recent setuptools release and psycopg2

### Required reviewers

1-2 devs

I will deploy this today after it's been reviewed


patphongs (approved):
👍 Thanks @cnlucas!

pkfec (commented):
Can you add a small comment here on why we need to pin setuptools here?

pkfec (approved):
Thanks @cnlucas

PR #5248, opened 2022-10-03
Fix null value in committee name

### LINKED ISSUE #5247: Fix null value in committee name in ofec_sched_e_mv
### What we’re after
Null value in committee name column in ofec_sched_e_mv will return incorrect results when searching schedule E by committee ID/name text.

Because the committee name column is always null in SE base table,  cmte_name has to be pulled from other table.

### Related ticket(s)
- [CMS #5422](https://github.com/fecgov/fec-cms/issues/5422)

### Action item(s)

- [ ]  Add coalesce function
- [ ] Pull committee name from other table

### Completion criteria
- [ ] Add migration file

hcaofec:
## Summary (required)

- Resolves #5247 
This PR modifies ofec_sched_e_mv by adding coalesce to cmte_name, and pulling cmte_name from other tables.
### Required reviewers
1 developer

## How to test
- Download feature branch 
- Run `pytest`
- Run `flyway migrate`

### DIFF
--- a/data/migrations/V0262__ofec_sched_e_mv_add_coalesce.sql
+++ b/data/migrations/V0262__ofec_sched_e_mv_add_coalesce.sql
@@ -1,0 +1,551 @@
+/*
+This is migration is for issue ##5247
+This is the latest version of  public.ofec_sched_e_mv
+-- Change made in this file 
+Add coalesce to cmte_name; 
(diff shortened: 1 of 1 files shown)


fec-jli (approved):
looks good. Thanks

PR #5274, opened 2022-11-03
[MERGE WITH GITFLOW] Increase api memory to 2.5G

### LINKED ISSUE #5267: API memory usage
### What we’re after

Since July (after the 6/21 release) we have increased the amount of memory we are using and getting more and more 502 errors. This is also when we experienced a huge jump in worker boot errors.
image

We should increase the API memory to 2G on dev, stage, and prod. Hopefully this should take care of the 502 errors, but we should make this increase either way. We should also look into why our memory usage jumped after the release on 6/21 and if there's perhaps a way to decrease the memory usage. 

**We need to look at how we structured the joining of JFC committees as a potential cause of this increase in memory:** 
Shows the null JFC fields that might be able to be removed/changed:
https://fec-dev-api.app.cloud.gov/v1/committee/C00075820/history/?api_key=DEMO_KEY  

https://github.com/fecgov/openFEC/blob/ec7105497d3386f994e468835045dbd0cd2e2b14/webservices/common/models/committees.py#L89

 
### Related ticket(s)

- [x] https://github.com/fecgov/openFEC/issues/5184

### Action item(s)

- [x] Increase API memory on dev, stage, and prod to 2G
- [x] Investigate memory usage jump
- [x] Investigate costs around increasing memory
- [x] Investigate implementing testing which would warn us before we hit memory limits again (perhaps also looking into this for other limits like disk quota as well) (Automated testing for how much a change in our code with cost us from a memory standpoint?)
- [x] Develop a plan to mitigate usage jump (or decide this was a necessary jump) and open follow-up ticket
- [x] Create a separate ticket to address JFC join issue. (@fec-jli )[https://github.com/fecgov/openFEC/issues/5270](https://github.com/fecgov/openFEC/issues/5270)

fec-jli:
## Summary (required)
App metrics: https://logs.fr.cloud.gov/goto/03634cc2122630e2a86cad813ddefe52
Booting workers still a bit high: https://logs.fr.cloud.gov/goto/67b5912713e2588c523b63597baa54e8

- Partially resolves #5267

Increases api memory to 2.5G in dev, stage, and prod

### Required reviewers

1 dev

## How to test
Make sure api work well.

### DIFF
--- a/manifests/manifest_api_dev.yml
+++ b/manifests/manifest_api_dev.yml
@@ -2,7 +2,7 @@
 applications:
   - name: api
     instances: 1
-    memory: 2G
+    memory: 2.5G
     disk_quota: 1G
     stack: cflinuxfs3
     buildpacks:
--- a/manifests/manifest_api_dev.yml
+++ b/manifests/manifest_api_dev.yml
@@ -2,7 +2,7 @@
 applications:
   - name: api
     instances: 1
-    memory: 2.5G
+    memory: 1G
     disk_quota: 1G
     stack: cflinuxfs3
     buildpacks:
--- a/manifests/manifest_api_prod.yml
+++ b/manifests/manifest_api_prod.yml
@@ -2,7 +2,7 @@
 applications:
   - name: api
     instances: 10
-    memory: 2G
+    memory: 2.5G
     disk_quota: 1G
     stack: cflinuxfs3
     buildpacks:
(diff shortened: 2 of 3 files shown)


cnlucas (approved):
LGTM, @fec-jli

cnlucas (approved):

PR #5284, opened 2022-11-14
Adjust API memory for all environments to 2G

patphongs:
## Summary

Our memory usage has decreased significantly to about 1.5GB. We can now adjust to 2GB for all environments.

### Required reviewers

1 backend developer

-  API app


fec-jli (approved):
Looks good. Thanks

Issue #5327, opened 2023-01-20
Footer Button flagged by eslint


cnlucas:
### What we’re after

The footer button is breaking react's rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md

This is blocking us upgrading babel/es packages. When we upgrade eslint it will break builds due to the above rule. 
![Screen Shot 2023-01-20 at 2 58 55 PM](https://user-images.githubusercontent.com/66386084/213794356-f81a9a8d-3944-4a36-9ed8-f735e3972710.png)

<img width="899" alt="Screen Shot 2023-01-20 at 3 01 16 PM" src="https://user-images.githubusercontent.com/66386084/213794887-9025dc2f-a998-46aa-92f8-692d1f71637b.png">

### Related ticket(s)

(Include the tickets that either came before, after, or are happening in tandem with this new ticket)
- [ ] 

### Action item(s)

(These are the smaller tasks that should happen in order to complete this work)
- [ ] 

### Completion criteria

(What does the end state look like - as long as this task(s) is done, this work is complete)
- [ ] 

### References/resources/technical considerations

(Is there sample code or a screenshot you can include to highlight a particular issue? Here is where you reinforce why this work is important)

Issue #5325, opened 2023-01-20
Upgrade Webpack to 5


cnlucas:
### What we’re after

We will eventually need to upgrade to webpack 5. Creating this issue to keep track of edits that can be pulled out when we transition. We should wait until the CMS upgrade to 4 is complete. 

We can remove "export NODE_OPTIONS=--openssl-legacy-provider &&" from our package.json file for build and develop. We will need to make either make sure that node has fixed this issue (which they are planning to do) or we update the hashing algorithm https://stackoverflow.com/questions/69394632/webpack-build-failing-with-err-ossl-evp-unsupported
 

### Related ticket(s)

(Include the tickets that either came before, after, or are happening in tandem with this new ticket)
- [ ] 

### Action item(s)

(These are the smaller tasks that should happen in order to complete this work)
- [ ] 

### Completion criteria

(What does the end state look like - as long as this task(s) is done, this work is complete)
- [ ] 

### References/resources/technical considerations

(Is there sample code or a screenshot you can include to highlight a particular issue? Here is where you reinforce why this work is important)

PR #5328, opened 2023-01-20
5314-upgrade node 

### LINKED ISSUE #5314: Upgrade Node
### What we’re after

We have been getting warnings in builds that the latest versions of npm do not support 14.15.5. We are getting pretty far behind (current version is 19.4.0) 

Upgrading would ensure we aren't getting issues with compatibility and also bypass the issues we've been having of developers using the shipped-with-nvm npm version vs LTS version of NPM (which has caused weird lockfile issues) 

It would also be a good idea to take a look at some of the current npm warnings as there are quite a few deprecation messages 

We will need to either [export SET NODE_OPTIONS=--openssl-legacy-provider ](https://stackoverflow.com/questions/69962209/what-is-openssl-legacy-provider-in-node-js-v17) or figure out a workaround unless we want to just upgrade to 16

### Related ticket(s)

- [ ] 

### Action item(s)

- [ ] 

### Completion criteria

- [ ] 

cnlucas:
## Summary

- Resolves #5314 

We are getting warnings on our builds as the latest npm is no longer compatible with versions of node <14.17. We were on 14.15.5 and this upgrades us to the LTS 18.13.0

We will be able to remove the legacy node_option once we either upgrade webpack to 5 or node fixes this issue. I created a [ticket for the webpack upgrade](https://github.com/fecgov/openFEC/issues/5325) (which won't happen until after CMS is upgraded to 4) just to include a reminder about removing this language. 

I did some research on upgrading packages which are causing NPM vunerabilities. We could upgrade babel and es packages (up several years) but we need to make our footer button compatible with [eslint rule here.](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.mdhttps://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md) The ticket for that work is [here](https://github.com/fecgov/openFEC/issues/5327).  [EDIT: Thanks @tmpayton for putting in this fix!]

### Required reviewers

2 devs

-  node packages

## Related PRs
In the above

CMS: https://github.com/fecgov/fec-cms/pull/5566
EREGS:

## How to test

-  `gh pr checkout`
- `nvm install 18.13.0`
-  you can use the nvm package npm or LTS
- `rm -rf node_modules`
- `npm i && npm run build`
-  `pytest`
- `flask run`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,9 +55,9 @@ jobs:
             export NVM_DIR="$HOME/.nvm"
             [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
             [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
-            nvm install 14.15.5
-            nvm use 14.15.5
-            nvm alias default 14.15.5
+            nvm install 18.13.0
+            nvm use 18.13.0
+            nvm alias default 18.13.0
             npm install -g npm
             npm install
             npm run build
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,51 +47,32 @@
         "webpack-dev-server": "^3.11.0"
       },
       "engines": {
-        "node": "14.15.5",
-        "npm": "8.3.1"
+        "node": "18.13.0",
+        "npm": "8.19.3"
       }
     },
     "node_modules/@babel/runtime": {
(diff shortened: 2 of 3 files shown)


tmpayton (approved):
This works well for me! Awesome job 😀

pkfec (approved):
@cnlucas  Works as expected!

fec-jli (approved):
works as expected. Thanks

PR #5354, opened 2023-02-23
Add interpretation for candidate_id

### LINKED ISSUE #5296: HELP: Interpretation of candidate_id
Hi there!

The docs explaining candidate_id: https://github.com/fecgov/openFEC/blob/09808c10f25aae3efa352b574c64ad2103a6ac88/webservices/docs.py#L67-L70

are a start to me understanding candidate_id, but I think there is some info encoded in there that I don't understand.

For example given candidate ID:
H0AL02087
H = House
0 = ?????
AL = Alabama
02 = 2nd District
087 = 87th candidate, a unique ID

What does the 2nd number mean? It seems to almost always be a 0, 2, 4, or 6 (if it's not one of these then I'm interpreting it as messy data). For instance here are the value counts for a sample of data I pulled:
0    8587
2    7505
8    5847
6    5524
4    3699
1       5
7       3
3       2
5       1
9       1

Does this second number have something to do with the election year? They seem to be somewhat correlated with the last digit of the year, eg `0` most often occurs in 2010 and 2020, though `0` also appears in other years as well. Same with `2`, it most often appears in 2012 and 2022, but in other years as well.

Thank you for the assistance!

hcaofec:
## Summary (required)

- Resolves #5296 

Add interpretation for candidate_id to help users to understand more of its composition.

### Required reviewers

1 developer

## How to test
Download feature branch and run pytest.
Review the text next to candidate_id.

http://127.0.0.1:5000/developers/#/candidate/get_candidate__candidate_id__

Screen Shot 2023-02-23 at 2 13 56 PM

### DIFF
--- a/webservices/docs.py
+++ b/webservices/docs.py
@@ -71,6 +71,12 @@ The candidate endpoints primarily use data from FEC registration
 CANDIDATE_ID = '''
 A unique identifier assigned to each candidate registered with the FEC.
 If a person runs for several offices, that person will have separate candidate IDs for each office.
+First character indicates office - [P]residential, [H]ouse, [S]enate].
+Second character is the last digit of the two-year period the ID was created.
+Third and fourth is the candidate state. Presidential IDs don't have state.
+Fifth and sixth is the district when the candidate first ran. This does not change if the
+candidate/member's district changes during re-districting. Presidential IDs don't have districts.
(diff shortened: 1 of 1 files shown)


pkfec (approved):
Looks good @hcaofec

PR #5390, opened 2023-04-05
Fix es snapshot bug

fec-jli:
## Summary (required)
Fix Elasticsearch snapshot backup bug.

## How to test
deploy on dev.


pkfec (approved):
Changes looks good, thanks for fixing this bug @fec-jli

PR #5413, opened 2023-04-18
Increase snapshot sleep time to 60s

fec-jli:
Increase snapshot sleep time to 60s


pkfec (approved):
@fec-jli  Increasing the time interval from 20 to 60s should give enough time to create snapshots for case and ao indexes. Dev and Stage snapshots seems to be OK with 20s. Lets monitor this change in Prod space next week.



Thanks for making this change!

PR #5452, opened 2023-05-20
Upgrade webpack and other dependencies

### LINKED ISSUE #5325: Upgrade Webpack to 5

jmetev1:
## Summary (required)

- Resolves #5325 

(Include a summary of proposed changes and connect issue below)
upgrading webpack requires a chain of dependency updates so this grew a bit. In particular, updated lint rule packages had to be added which added new rules. To address that without changing code, I had to disable some of the new rules in eslintrc. Let me know if there's any changes you'd like to see.

### Required reviewers

(Include who is required to review prior to merge. For example: One designer and two front end developer reviews are required prior to merge)

build
-  

## Screenshots

(Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface change, include both before and after screenshots)

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test

- 

## System architecture updates (if applicable)

(If this pull request changes our [current system diagram](https://github.com/fecgov/FEC/wiki/2.-FEC-system-diagram), include a description of those changes here and create a new ticket to update the system diagram)


package.json
```diff
@@ -7,49 +7,46 @@
...
-    "develop": "export NODE_OPTIONS=--openssl-legacy-provider && webpack-dev-server --config webpack.config.js --mode development",
+    "build": "npm run lint && webpack --config webpack.production.config.js",
+    "develop": "webpack-dev-server --config webpack.config.js",
     "format": "prettier --write 'src/**/*.js'",
     "lint": "eslint ./src/**/*.js",
     "format:lint": "npm run format && npm run lint"
   },
   "dependencies": {
-    "@braintree/sanitize-url": "^6.0.0",
-    "immutable": "^3.8.2",
```
jmetev1, line 18: immutable was added here to upgrade it. at the time it was used by another package. but now it is not needed at all so we can remove it here. same for braintree.

PR #5451, opened 2023-05-19
Fix build script

jmetev1:
## Summary (required)

- Resolves none

Right now the build script runs format before lint. So things that should be fixed get formatted in CI. But that does not get committed. So this just covers up peoples mistakes. I don't think this is what was intended but I could be wrong. When I open the repo and run npm run develop, it immediately fails on the footer formatting change. 

### Required reviewers

(Include who is required to review prior to merge. For example: One designer and two front end developer reviews are required prior to merge)

## Impacted areas of the application

General components of the application that this PR will affect:

-  

## Screenshots

(Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface change, include both before and after screenshots)

## Related PRs

Related PRs against other branches:

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test

(Include any information that may be helpful to the reviewer(s). This might include links to sample pages to test or any local environmental setup that is unusual such as environment variable (never credentials), API version to point to, etc)

- 

## System architecture updates (if applicable)

(If this pull request changes our [current system diagram](https://github.com/fecgov/FEC/wiki/2.-FEC-system-diagram), include a description of those changes here and create a new ticket to update the system diagram)

PR #5454, opened 2023-05-22
5453-Add jmetev1's build script fix and add prettier to pre-commit

### LINKED ISSUE #5453: Remove format from npm build, add to pre-commit
### What we’re after

https://github.com/fecgov/openFEC/pull/5451
@jmetev1 noticed that we run our linter after fomat when we run npm run build. This is causing us to not fix formatting changes as they are being re-formatted in production, but never committed. We should add his changes (thank you!!) and also add prettier to our pre-commit hooks, so that we are catching these issues in development more often. We need to upgrade prettier to do this as well.

cnlucas:
## Summary (required)

- Resolves #5453 
@jmetev1 found an issue in our build scripts. He noticed we were running format before lint during npm run build which allows us to ignore issues as they are being fixed, but not committed. I used his code changes and then also added prettier to our pre-commit hooks and upgraded prettier. 

Thank you @jmetev1, for your contribution and for catching this issue for us!

### Required reviewers
2 developers

-  npm run build, pre-commit hooks

## Related PRs

(https://github.com/fecgov/openFEC/pull/5451)

## How to test

- start your virtualenv
- pull recent develop
- rm -rf node_modules
- npm i && npm run develop
-  (should see footer issue) 
- npm run build 
- (should see footer change is made but not committed) 
- pull this branch
- rm -rf node_modules
- npm i && npm run develop
- npm run build

Then try testing out committing a change that would break prettier's  rules like removing the new line in the footer

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,16 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
-    -   id: end-of-file-fixer
+      - id: end-of-file-fixer
 
--   repo: https://github.com/PyCQA/flake8
+  - repo: https://github.com/PyCQA/flake8
     rev: 6.0.0
     hooks:
-    -   id: flake8
+      - id: flake8
         args: [--max-line-length=120]
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: 'v3.0.0-alpha.9-for-vscode' # Use the sha / tag you want to point at
+    hooks:
+      - id: prettier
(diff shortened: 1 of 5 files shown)


rfultz:
Do we _want_ to use Prettier? We specifically removed it from CMS. I'm not familiar enough with openFEC; did we remove it from the build and forget to remove it from package? Or was it mistakingly removed from the build process, or maybe just never fully implemented?

We removed it from CMS because Prettier is proudly opinionated and enforcing rules we don't like and each version makes it harder for us to have our own formatting rules

cnlucas:
> 

@rfultz, we are currently running prettier right now when we run npm run build. (format:lint runs both lint and format) Since we are currently compatible with prettier I would recommend keeping it? I don't think we should remove it without replacing it with something similar.

johnnyporkchops (approved):
Thanks @jmetev1  and @cnlucas  for fixing  this. It was always confusing why the footer would show up as being formatted on the rare occasions I made edits to open_fec javascript.

tmpayton (approved):
Looks good, great job @cnlucas and thank you for contributing @jmetev1!

Issue #242, opened 2015-01-05
Begin looking at documents endpoint


micahsaul:
Begin looking at what will be required in an endpoint for the documents.

- [ ] Populate the rest of this list
- [X] Examine current state of the world on existing FEC page.
- [x] Contact FEC and ask about where the document metadata lives.
- [ ] Write up plan for endpoint, including mapping, with intent to begin implementing next sprint.

rjmajma:
Moved to the backlog. Awaiting a response from FEC.

PR #5463, opened 2023-05-26
 Flask upgrade to 2.2.5

### LINKED ISSUE #5440: [Snyk:High] flask Information Exposure (due by 06/09/2023)
Introduced through
    flask@2.1.2, flask-apispec@0.7.0 and others
    Fixed in
    flask@2.2.5, @2.3.2

    Exploit maturity
    No known exploit

Detailed paths and remediation

    Introduced through: project@0.0.0 › flask@2.1.2
    Fix: Upgrade flask to version 2.2.5 or 2.3.2

Introduced through: project@0.0.0 › flask-apispec@0.7.0 › flask@2.1.2
Fix: Pin flask to version 2.2.5 or 2.3.2
Introduced through: project@0.0.0 › flask-cors@3.0.10 › flask@2.1.2
Fix: Pin flask to version 2.2.5 or 2.3.2

…and 2 more
Security information
Factors contributing to the scoring:

    Snyk: [CVSS 7.5](https://security.snyk.io/vuln/SNYK-PYTHON-FLASK-5490129) - High Severity
    NVD: Not available. NVD has not yet published its analysis.

Why are the scores different? Learn how Snyk evaluates vulnerability scores
Overview

Affected versions of this package are vulnerable to Information Exposure in the form of exposing the permanent session cookie, when all of the following conditions are met:

    The application is hosted behind a caching proxy that does not strip cookies or ignore responses with cookies.

    The application sets session.permanent = True.

    The application does not access or modify the session at any point during a request.

    SESSION_REFRESH_EACH_REQUEST is enabled (the default).

    The application does not set a Cache-Control header to indicate that a page is private or should not be cached.

A response containing data intended for one client may be cached and sent to other clients. If the proxy also caches Set-Cookie headers, it may send one client's session cookie to other clients. Under these conditions, the Vary: Cookie header is not set when a session is refreshed (re-sent to update the expiration) without being accessed or modified.

**Action items:**
- [ ] Pin flask to version 2.2.5 or 2.3.2

**Completion criteria:**
- [ ] Vulnerability is no longer flagged by snyk

tmpayton:
## Summary (required)

- Resolves #5440 

This ticket upgrades Flask to version 2.2.5. It also upgrades the following dependencies: 
API spec 0.39 -> 4.0
webargs 5.5.3 -> 7.0.0
flask-apispec 0.7.0 -> 0.11.4
marshmallow 2.16.3 -> 3.0.0 
marshmallow-sqlalchemy 0.15.0 -> 0.23.1

-  webservices/Schemas
- webservices/Resources
- webservices/args
- automated tests 

## Related PRs

[Marshmallow Pagination](https://github.com/fecgov/marshmallow-pagination/pull/6) PR - upgrade marshmallow to v3 

## How to test

1. Checkout this branch 
2. `pip install -r requirements.txt`
3.  `snyk test --file=requirements.txt --package-manager=pip`
4. `pytest`
5. `flask run`
6. Test all endpoints

### DIFF
--- a/tests/test_aggregates.py
+++ b/tests/test_aggregates.py
@@ -250,7 +250,7 @@ class TestAggregates(ApiBaseTest):
             results = self._results(
                 api.url_for(
                     resource,
-                    candidate_id=self.candidate.candidate_id,
+                    committee_id=self.committee.committee_id,
                     cycle=2012,
                     office='president',
                     election_full=False,
--- a/tests/test_aggregates.py
+++ b/tests/test_aggregates.py
@@ -250,14 +250,14 @@ class TestAggregates(ApiBaseTest):
             results = self._results(
                 api.url_for(
                     resource,
-                    committee_id=self.committee.committee_id,
+                    candidate_id=self.candidate.candidate_id,
                     cycle=2012,
                     office='president',
                     election_full=False,
                 )
             )
             assert len(results) == 1
-            serialized = schema().dump(aggregates[0]).data
+            serialized = schema().dump(aggregates[0])
             serialized.update(
                 {
                     'committee_name': self.committee.name,
@@ -278,14 +278,13 @@ class TestAggregates(ApiBaseTest):
                 api.url_for(
                     resource,
                     candidate_id=self.candidate.candidate_id,
-                    committee_id=self.committee.committee_id,
                     cycle=2012,
                     office='president',
                     election_full='true',
                 )
             )
             assert len(results) == 1
-            serialized = schema().dump(aggregates[0]).data
+            serialized = schema().dump(aggregates[0])
             serialized.update(
                 {
                     'committee_name': self.committee.name,
--- a/tests/test_committees.py
+++ b/tests/test_committees.py
@@ -386,13 +386,12 @@ class CommitteeFormatTest(ApiBaseTest):
                 for each in results
             )
         )
-        results = self._results(
-            api.url_for(
-                CommitteeList,
-                **{min_date_field: datetime.date.fromisoformat(values[1]),
-                max_date_field: datetime.date.fromisoformat(values[2])}
-            )
-        )
+        results = self._results(api.url_for(
+            CommitteeList,
+            **{min_date_field: datetime.date.fromisoformat(values[1]),
+               max_date_field: datetime.date.fromisoformat(values[2])}
+                                            )
+                                )
         self.assertTrue(
             all(
                 datetime.date.fromisoformat(values[1]).isoformat()
@@ -618,6 +617,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
             )
         )
         assert len(results) == 1
+        self.assertFalse(results[0]["is_active"])
 
         results = self._results(
             api.url_for(
@@ -628,6 +628,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
             )
         )
         assert len(results) == 1
+        self.assertTrue(results[0]["is_active"])
 
     def test_candidate_cycle(self):
         results = self._results(
--- a/tests/test_committees.py
+++ b/tests/test_committees.py
@@ -297,8 +297,8 @@ class CommitteeFormatTest(ApiBaseTest):
 
     def test_first_f1_date_sort(self):
         committees = [
-            factories.CommitteeFactory(first_f1_date="2003-10-12"),
-            factories.CommitteeFactory(first_f1_date="2017-05-14"),
+            factories.CommitteeFactory(first_f1_date=datetime.date.fromisoformat("2003-10-12")),
+            factories.CommitteeFactory(first_f1_date=datetime.date.fromisoformat("2017-05-14")),
         ]
         committee_ids = [each.committee_id for each in committees]
         results = self._results(api.url_for(CommitteeList, sort="first_f1_date"))
@@ -614,8 +614,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
                 CommitteeHistoryProfileView,
                 committee_id=self.committees[4].committee_id,
                 cycle=2014,
-                election_full=False,
-                is_active=False,
+                election_full=False
             )
         )
         assert len(results) == 1
@@ -625,8 +624,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
                 CommitteeHistoryProfileView,
                 committee_id=self.committees[2].committee_id,
                 cycle=2014,
-                election_full=False,
-                is_active=True,
+                election_full=False
             )
         )
         assert len(results) == 1
@@ -739,8 +737,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
                 CommitteeHistoryProfileView,
                 committee_id="id01",
                 cycle=2014,
-                election_full=False,
-                is_active=False,
+                election_full=False
             )
         )
         assert len(results) == 1
@@ -750,8 +747,7 @@ class TestCommitteeHistoryProfile(ApiBaseTest):
                 CommitteeHistoryProfileView,
                 candidate_id="h01",
                 cycle=2014,
-                election_full=False,
-                is_active=False,
+                election_full=False
             )
         )
         assert len(results) == 1
--- a/tests/test_dates.py
+++ b/tests/test_dates.py
@@ -30,7 +30,7 @@ class TestReportingDates(ApiBaseTest):
         factories.ReportDateFactory(update_date=datetime.datetime(2014, 4, 2))
 
         filter_fields = (
-            ('due_date', '2014-01-02'),
+            ('min_due_date', '2014-01-02'),
             ('report_year', 2015),
             ('report_type', 'YE'),
             ('min_create_date', '2014-03-02'),
--- a/webservices/schemas.py
+++ b/webservices/schemas.py
@@ -4,7 +4,7 @@ import functools
 from collections import namedtuple
 
 import marshmallow as ma
-from marshmallow_sqlalchemy import ModelSchema
+from marshmallow_sqlalchemy import SQLAlchemySchema, SQLAlchemyAutoSchema
 from marshmallow_pagination import schemas as paging_schemas
 
 from webservices import utils, decoders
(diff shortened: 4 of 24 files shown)


webservices/schemas.py
```diff
@@ -50,13 +49,13 @@ def __new__(cls, field, column, label, position=-1):
         )
 
 
-class BaseSchema(ModelSchema):
```
tmpayton, line 53: ModelSchema has been removed from marshmallow-sqlalchemy. We must either use SQLAlchemySchema or SQLAlchemyAutoSchema. 

Looking a the [documentation](https://marshmallow-sqlalchemy.readthedocs.io/en/latest/index.html#generate-marshmallow-schemas) we might have to explicitly declare all fields for a schema. We can try to use the  auto_field() method. But we would still need to declare the names of the fields

tmpayton:
marshamallow-sqlalchemy provided this [example](https://marshmallow-sqlalchemy.readthedocs.io/en/latest/recipes.html) for how to automatically generate schemas, but it looks like it creates all of them.

webservices/schemas.py
```diff
@@ -225,14 +226,46 @@ def make_schema(model, class_name=None, fields=None, options=None):
...
         else schema_map.get(model.__name__)
-
     )
     return type(
         class_name,
         (mapped_schema, ),
         utils.extend({'Meta': Meta}, fields or {}),
     )
 
+def generateClass(model, class_name=None):
```
tmpayton, line 236: I'm experimenting with generating the class automatically

tmpayton:
Any resource that uses flask-apispec "use_kwargs" is broken ([flask-apispec issue](Flask_apispec version of @use_kwargs not compatible with unknown parameter #242)). 

[This](https://webargs.readthedocs.io/en/latest/advanced.html) is the change in webargs.

webservices/resources/candidate_aggregates.py
```diff
@@ -102,9 +102,9 @@ def get(self, **kwargs):
     tags=['receipts'], description=docs.SCHEDULE_A_STATE_CANDIDATE_TOTAL_TAG,
 )
 class ScheduleAByStateCandidateTotalsView(utils.Resource):
-    @use_kwargs(args.paging)
-    @use_kwargs(args.make_sort_args())
-    @use_kwargs(args.schedule_a_candidate_aggregate)
```
tmpayton, line 107: Using the use_kwargs decorator multiple times causes it to break

tmpayton (commented):
Most changes in test/ are because schemas are now strict.  Link to Marshmallow [changes](https://marshmallow.readthedocs.io/en/stable/upgrading.html)

tests/test_committees.py
```diff
@@ -297,8 +297,8 @@ def test_committee_sort(self):
 
     def test_first_f1_date_sort(self):
         committees = [
-            factories.CommitteeFactory(first_f1_date="2003-10-12"),
```
tmpayton, line 300: User should be allowed to enter a date in this format

tmpayton: This one is just sorting

tests/test_dates.py
```diff
@@ -88,7 +88,7 @@ def test_filters(self):
         factories.CalendarDateFactory(end_date=datetime.datetime(2015, 1, 2))
 
         filter_fields = [
-            ('min_start_date', '01/01/2015'),
```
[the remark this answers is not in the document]
tmpayton: This format (MM/DD/YYYY ) does not work - looking into alternatives 
image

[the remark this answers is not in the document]
tmpayton: ma.fields.date only allows ISO8601-formatted date. I think we need to make our own custom class for mm/dd/yyyy.

tmpayton:
TO DO: 

- [ ] Review automated testing changes 
- [ ] figure out endpoint naming issue (added v1) 
- [ ] Make ticket to refactor views that inherit from utils.Resource to inherit from Apiresource instead
- [ ] Make list of breaking changes

tmpayton:
Breaking Changes: 
-  Location defaults to JSON/body and multiple locations are not supported in a single call. 
- Marshmallow schemas are strict meaning that errors throw an exception. 
- ma.fields.date only allows ISO8601-formatted dates 
- dump().data is depreciated 
- marshamallow.compat.text_type is depreciated 
- with marshmallow schemas, decorated methods, validate methods, and error methods must accept **kwargs          
- schemas raise validation errors when deserializing data with unknown keys
- get_attribute method signature has changed 
- We have to use the @post_dump decorator to add additional data on serialization 
- pre/post-processors must return modified data in marshmallow schema 
- custom field _deserialize methods should accept **kwargs 
- with webargs the custom error handle argument names and signature has changed 
- apispec requires openapi_version 
- using both missing and required=True are incompatible options for marshmallow fields
- apispec.definition is deprecated 
- marshmallow_sqlalchemy marshmallow_sqlalchemy is replaced with SQLAlchemySchema, SQLAlchemyAutoSchema
-

tmpayton:
image

The added v1/ is due to a change in apispec. It no longer strips the basePath from paths anymore. [Here](https://github.com/marshmallow-code/apispec/issues/78#issuecomment-431846368) is a thread discussing the change.

fec-jli:
We have to modify the function in views.py in cms repo.
as below in line number: 132

```
def format_receipt_date(receipt_date):
    # convert string to python datetime object
    receipt_date = datetime.strptime(receipt_date, "%Y-%m-%d")
    # parse for readable output
    return receipt_date.strftime("%m/%d/%Y")
```

tmpayton:
> We have to modify the function in views.py in cms repo. as below in line number: 132
> 
> ```
> def format_receipt_date(receipt_date):
>     # convert string to python datetime object
>     receipt_date = datetime.strptime(receipt_date, "%Y-%m-%d")
>     # parse for readable output
>     return receipt_date.strftime("%m/%d/%Y")
> ```

@fec-jli 
Another fix for this is to change the data type of receipt date to datetime instead of date. The automatic field generation with sqlalchemyautoschema seems to be slightly different, so I'm going through all the endpoints to make sure that the data types aren't off.

fec-jli:
line 298 in test_committees.py :
```
    def test_first_f1_date_sort(self):
        committees = [
            # factories.CommitteeFactory(first_f1_date=datetime.date(2003, 10, 12)),
            # factories.CommitteeFactory(first_f1_date=datetime.date(2017, 5, 14)),
            factories.CommitteeFactory(first_f1_date=datetime.date.fromisoformat("2003-10-12")),
            factories.CommitteeFactory(first_f1_date=datetime.date.fromisoformat("2017-05-14")),
        ]
        committee_ids = [each.committee_id for each in committees]
        results = self._results(api.url_for(CommitteeList, sort="first_f1_date"))
        self.assertEqual([each["committee_id"] for each in results], committee_ids)
        results = self._results(api.url_for(CommitteeList, sort="-first_f1_date"))
        self.assertEqual(
            [each["committee_id"] for each in results], committee_ids[::-1]
        )
```

fec-jli:
> > We have to modify the function in views.py in cms repo. as below in line number: 132
> > ```
> > def format_receipt_date(receipt_date):
> >     # convert string to python datetime object
> >     receipt_date = datetime.strptime(receipt_date, "%Y-%m-%d")
> >     # parse for readable output
> >     return receipt_date.strftime("%m/%d/%Y")
> > ```
> 
> @fec-jli Another fix for this is to change the data type of receipt date to datetime instead of date. The automatic field generation with sqlalchemyautoschema seems to be slightly different, so I'm going through all the endpoints to make sure that the data types aren't off.

good idea if we can use datetime type. it is better we create our own validate function: datetime that similar as function "Date" in args.py. we can do additional validation in case.

tmpayton:
> > > We have to modify the function in views.py in cms repo. as below in line number: 132
> > > ```
> > > def format_receipt_date(receipt_date):
> > >     # convert string to python datetime object
> > >     receipt_date = datetime.strptime(receipt_date, "%Y-%m-%d")
> > >     # parse for readable output
> > >     return receipt_date.strftime("%m/%d/%Y")
> > > ```
> > 
> > 
> > @fec-jli Another fix for this is to change the data type of receipt date to datetime instead of date. The automatic field generation with sqlalchemyautoschema seems to be slightly different, so I'm going through all the endpoints to make sure that the data types aren't off.
> 
> good idea if we can use datetime type. it is better we create our own validate function: datetime that similar as function "Date" in args.py. we can do additional validation in case.

Do we want to change the argument type for min_receipt_date and max_receipt_date to datetime as well ? I left the argument type as Date and it works even though receipt_date is type datetime. 

Example: https://fec-dev-api.app.cloud.gov/v1/filings/?page=1&per_page=20&min_receipt_date=06%2F01%2F2023&sort=-receipt_date&sort_hide_null=false&sort_null_only=false&sort_nulls_last=false

tests/test_committees.py
```diff
@@ -625,8 +624,7 @@ def test_is_active(self):
                 CommitteeHistoryProfileView,
                 committee_id=self.committees[2].committee_id,
                 cycle=2014,
-                election_full=False,
-                is_active=True,
```
cnlucas, line 629: So much great work with these tests! It makes sense why you removed is_active because it's not a filter. For the test_is_active test, however, do you think we should change the 2 assert statements to the is_active value like this:
 assert results[0].get("is_active") == False

tmpayton: That sounds like a good idea, I'll make the change!

tests/test_aggregates.py
```diff
@@ -250,14 +250,14 @@ def test_candidate_aggregates_by_committee(self):
             results = self._results(
                 api.url_for(
                     resource,
-                    committee_id=self.committee.committee_id,
```
cnlucas, line 253: The tests are passing for me with committee_id. Should we keep it committee_id since the test is by_committee?

cnlucas: That makes sense! I'll put in another ticket for this work. Thanks @tmpayton!

cnlucas (approved):
Great job @tmpayton! This was a ton of work and you made it as easy as possible to review. We are so lucky to have you on the team!

fec-jli (approved):
Finally, I go through all the Breaking Changes, learn and understand one by one. It is really helpful you make that list. It is so express you figure out how to implement those Breaking Changes into our project. It is really great job. Super!!!

hcaofec (approved):
Great work!

PR #5492, opened 2023-07-12
[WIP] Upgrade SQLAlchemy

tmpayton:
## Summary (required)

- Resolves #issue_number

(Include a summary of proposed changes and connect issue below)

### Required reviewers

(Include who is required to review prior to merge. For example: One designer and two front end developer reviews are required prior to merge)

-  

## Screenshots

(Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface change, include both before and after screenshots)

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test

- 

## System architecture updates (if applicable)

(If this pull request changes our [current system diagram](https://github.com/fecgov/FEC/wiki/2.-FEC-system-diagram), include a description of those changes here and create a new ticket to update the system diagram)

### DIFF
--- a/webservices/common/counts.py
+++ b/webservices/common/counts.py
@@ -58,7 +58,7 @@ def get_query_plan(query):
 
 
 def extract_analyze_count(rows):
-    for row in rows:
+    for row in reversed(rows):
         match = count_pattern.search(row[0])
         if match:
             return int(match.groups()[0])
--- a/webservices/common/counts.py
+++ b/webservices/common/counts.py
@@ -22,7 +22,7 @@ def is_estimated_count(resource, query):
     """
     if resource.use_pk_for_count and resource.model:
         primary_key = resource.model.__mapper__.primary_key[0]
-        query = query.with_entities(primary_key)
+        query = query.with_only_columns(primary_key)
     if resource.use_estimated_counts:
         estimated_count = get_estimated_count(query)
         if estimated_count > resource.estimated_count_threshold:
--- a/webservices/common/counts.py
+++ b/webservices/common/counts.py
@@ -9,7 +9,7 @@ import re
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.sql.expression import Executable, ClauseElement
 from webservices.common import models
-
+from sqlalchemy import select, func
 
 count_pattern = re.compile(r'rows=(\d+)')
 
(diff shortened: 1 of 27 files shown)


webservices/sorting.py
```diff
@@ -77,8 +77,8 @@ def sort(query, key, model, aliases=None, join_columns=None, clear=False,
     # In this case, use the string name of the sort key.
     sort_model = (
         model
-        if len(query._entities) == 1 and hasattr(query._entities[0], 'mapper')
-        else None
+        # if len(query._entities) == 1 and hasattr(query._entities[0], 'mapper')
+        # else None
```
tmpayton, line 81: I still need to figure out a solution for this. Select does not contain the attribute _entities.

webservices/utils.py
```diff
@@ -107,7 +108,7 @@ def fetch_page(
             index_column=index_column,
             nulls_last=nulls_last,
         )
-    paginator = paginators.OffsetPaginator(query, kwargs["per_page"], count=count)
+    paginator = paginators.OffsetPaginator(query, kwargs["per_page"], session, count=count)
```
tmpayton, line 111: I'm going to look into binding the session so hopefully we won't have to pass it in!

webservices/common/counts.py
cnlucas, line 1: This is very small and not helpful-- but I don't think you need to import sqlalchemy--both select and func are avail through the instance of sqlalchemy brought in through flask-sqlalchemy db

I think we import it quite a bit when we don't need to? Let me know if there's something I'm missing about that

webservices/sorting.py
cnlucas: I think we have to use column_descriptions.

cnlucas:
We are postponing this upgrade because it will require some major re-writing to how we are connecting to the database.

We discovered that we can no longer reset SQLALCHEMY_DATABASE_URI. Engines are now created when calling init_app rather than the first time they are accessed. The current recommended way of getting around this issue is to build an app factory. These changes would require us to re-create the app when we need to reset the configurations. This is extremely high risk for us.

Here's a helpful discussion of the issue:
https://github.com/pallets-eco/flask-sqlalchemy/issues/1172

We will be watching and continuing to work on solutions in case this upgrade becomes more urgent.

Issue #5513, opened 2023-07-27
Research the removal of extra columns from debts export


JonellaCulmer:
**What we're after:**
Following a review by RAD, we need to remove unnecessary columns Y-AM from the debts datatable export. This is not a quick fix and requires research to address. 

**Notes:**
The debts tables are using `/schedules/schedule_d/`, so the export/downloads would be `/download/schedules/schedule_d/`

**Action items:**
- [ ] See if front-end can remove columns
- [ ] Consider other exports that may have extra columns that may need to be removed
- [ ] Add other tickets as needed to handle next steps

**Completion criteria:**
- [ ] Document research and next steps to remove columns Y-AM from the debts datatable export

cnlucas:
I am recommending removing the columns from our debts output. They are all <1% non-null 
![image](https://github.com/fecgov/openFEC/assets/66386084/3f04cd74-9a2f-4c62-a59d-9c5b3e659a0c)


I have a PR ready for review by DB and PM's for the work.

PR #5543, opened 2023-08-10
5513/5522-add idxs and remove null fields

### LINKED ISSUE #5522: Debts --committee_type filter
### What we’re after

To support [this design ](https://github.com/fecgov/fec-cms/issues/5819#issuecomment-1662644550)we would like to be able to filter by committee_type. 

### Related ticket(s)

- [x] https://github.com/fecgov/openFEC/issues/5510

### Action item(s)

- [ ] Debts endpoint is filterable by committee_type (if we decide that's too much work we can add the filter for form_type

We had a sync between FE, Design and BE about whether we should filter by form_type and/or committee_type. Since committee type is the ultimate goal--just adding the filter for committee_type would be ideal. If we, for some reason, can't do that we should add the form_type filter instead.

cnlucas:
## Summary (required)

- Partially Resolves #5522 and #5513 

This migration file refactors our debts endpoint output for the two tickets above. I found two filters without indexes in the process of completing these tickets, so I'm adding those indexes as well. 

The null value fields are being removed in response to feedback from UAT testing of the debts datatable from RAD. They would prefer less null value fields in the exports. 

Adds indexes for new filters: cmte_tp, filing_form
Adds indexes for current filters without  indexes: cmte_id, image_num
Removes mostly null fields: cred_dbtr_id, cand_nm, cand_nm_first, cand_nm_last, cand_office, cand_office_st, cand_office_st_desc, cand_office_district, conduit_cmte_id, conduit_cmte_nm, conduit_cmte_st1, conduit_cmte_st2, conduit_cmte_city, conduit_cmte_st, conduit_cmte_zip
I am not removing cand_id from output so that we can follow the deprecation policy for that filter. 
Removes index for candidate_id as it is a mostly null field.

The highest amount is cand_office at 2%. 

I did some research looking at the API umbrella for current users of the debts endpoint. In the last 120 days there were 7 users, only 2 external.  If these changes are approved by PM, I will deprecate and remove the candidate_id filter according to the deprecation policy. This filter is not currently being used as it is <1% non-null values. 

Screen Shot 2023-08-14 at 9 39 10 AM

### Required reviewers
1-2 devs
1 PM

-  Debts endpoint

## How to test

- pull the branch 
- dropdb cfdm_test
- createdb cfdm_test
- invoke create_sample_db

### DIFF
--- a/data/migrations/V0283__sched_d_add_idxs_remove_nulls.sql
+++ b/data/migrations/V0283__sched_d_add_idxs_remove_nulls.sql
@@ -1,0 +1,119 @@
+/*
+This migration file will 
+add indexes: cmte_id, cmte_tp, image_num
+remove mostly null fields: cand_id, cand_nm, cand_nm_first, cand_nm_last, cand_office, cand_office_st, cand_office_st_desc,
+cand_office_district, conduit_cmte_id, conduit_cmte_nm, conduit_cmte_st1, conduit_cmte_st2, conduit_cmte_city, conduit_cmte_st,
+conduit_cmte_zip
+remove index: cand_id 
+issue #5522 and #5513 
+*/
+
+-- ---------------
+-- ofec_sched_d_mv
+-- ---------------
+DROP MATERIALIZED VIEW IF EXISTS public.ofec_sched_d_mv_tmp;
+
+CREATE MATERIALIZED VIEW public.ofec_sched_d_mv_tmp AS
+    SELECT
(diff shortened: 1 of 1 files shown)


PaulClark2 (approved):
This looks good to me. The list of fields you are removing makes sense as they are not on the Schedule D.

hcaofec (approved):
Test with pytest and flyway migrate. Looks good!

PR #5581, opened 2023-09-11
5570- remove re-install of npm (LTS)

### LINKED ISSUE #5570: NPM version issue
### What we’re after

We use `sudo npm install -g npm` in our CMS circle config.yml file and it seems to work correctly by installing globally the NPM version that is paired by NVM with the node version we have selected. For API, we seem to be pulling in the latest NPM version.  This may break our builds when they release a new NPM version and we are still on an older version. (this happened with the release of 10) 

Screen Shot 2023-09-05 at 1 01 08 PM

We should figure out a way to make this consistent across repos---and pull in the paired NPM version in API instead of the most recent version. 

### Action item(s)

- [ ] Research how our circle NPM version is being determined 
- [ ] Confirm this issue described above
- [ ] Implement a fix

### Completion criteria

- [ ] NPM is being pulled in correctly in circle for API

cnlucas:
## Summary (required)

- Resolves #5570

We have been installing the latest version of npm (`npm install -g npm`) after installing node with nvm in circleci.  Nvm already installs npm with node, so we are re-installing the latest npm version which can break the builds. This happened when npm 10 was released. We use the nvm recommended version (and specify it) in production, so this PR will remove the upgrade command and make the circleci config match production.

There is an argument to keeping the configuration the same and letting the circle builds fail to force us to upgrade, but I think we should just regularly upgrade without it. CMS runs the nvm recommended version in circleci, so that is why API deploy failed when CMS did not last week.

If we make this change, I will put in a ticket to remove the line from CMS-- it doesn't seem to be working and we should be able to remove the sudo version, but more research is necessary (we may need to add sudo to the nvm install) 

### Required reviewers

2 devs

-  circleci config

## How to test

-  nvm install 18.17.1 
-  npm -v
(should show 9.6.7)
- npm install -g npm
- npm -v 
(should show latest version 10)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,6 @@ jobs:
             nvm install 18.17.1
             nvm use 18.17.1
             nvm alias default 18.17.1
-            npm install -g npm
             npm install
             npm run build
 


tmpayton (approved):
Awesome job!

pkfec (approved):

fec-jli (approved):
Looks good for me.

PR #5622, opened 2023-10-18
5621-add deprecation notice for line_number filter

### LINKED ISSUE #5621: Line_number deprecation
### What we’re after

In order to fix the line_number filter across the schedule endpoints we need to create a hybrid_property for form_line_number. Currently our user input field which is a composite of form and line_number is named line_number. We need to rename it as we need to keep the short line_number field. This will require a notice to users and deprecation of the old filter name. 

### Related ticket(s)

- [ ] 

### Action item(s)

- [ ] 

### Completion criteria

- [ ] 

cnlucas:
## Summary (required)

- Resolves #5621 

Adds deprecation notice for line_number filter across all the schedule endpoints. This needs to be done as currently, line_number is being used for the user input composite form-line_number value and the short line_number field. Deprecation notice to users within 120 days was sent out Oct 18th,  2023. Second notice will go out November 17th, 2023. Filter access (renamed to form_line_number) will be shut down on November 21st, 2023. After Novermber 21st, 2023 users will have access to form_line_number filter with same functionality

### Required reviewers

2 devs

-  swagger docs

## Screenshots
Screen Shot 2023-10-18 at 1 20 26 PM

## Related PRs

## How to test

- flask run
- look at any schedule endpoints for notice

### DIFF
--- a/webservices/args.py
+++ b/webservices/args.py
@@ -564,9 +564,7 @@ itemized = {
     'max_amount': Currency(description='Filter for all amounts less than a value.'),
     'min_date': Date(description='Minimum date'),
     'max_date': Date(description='Maximum date'),
-    'line_number': fields.Str(description='Filter for form and line number using the following format: '
-                                          '`FORM-LINENUMBER`.  For example an argument such as `F3X-16` would filter'
-                                          ' down to all entries from form `F3X` line number `16`.')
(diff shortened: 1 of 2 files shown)


fec-jli (approved):
Looks good. thanks

tmpayton (approved):
Works for me!

tmpayton (approved):

PR #5650, opened 2023-11-21
[HOTFIX - MERGE WITH GITFLOW] 5648 - Add reCAPTCHA to CSP

### LINKED ISSUE #5648: API.data.gov CSP changes
### What we’re after

We need to make some changes to our CSP due to changes API.data.gov made last weekend. This was the cause of the outage of our API signup this past weekend 

### Action item(s)

- [ ] confirm with product managers we should make these changes
- [x] make CSP changes

### Completion criteria

- [ ] 

rfultz:
## Summary

- Resolves #5648 

Cloud.gov needs us to add the reCAPTCHA URLs to our CSP (content-security-policy).

Adding the reCAPTCHA paths to the `script-src` rule is pretty straightforward (tells the browser that we're cool with script requests made to those paths).

**Maybe tricky**: They requested that we add the reCAPTCHA to our `frame-src` rule but we didn't have one. Without a rule for `frame-src`, browsers would have used `default-src` as a fallback. Because of that, I copied our standard values for `default-src` into a new `frame-src` rule and added the new values for reCAPTCHA. We may not need the non-Google values in `frame-src`.

### Required reviewers

Best to review it on not-localhost
2 devs?

Theoretically every page, but only comes into play when requesting
- script files from
   - `https://www.google.com/recaptcha/` †
   - `https://www.gstatic.com/recaptcha/` †
   - `'self'` ‡
   - `https://api.data.gov https://dap.digitalgov.gov` ‡
   - `https://www.google-analytics.com` ‡
   - `https://www.googletagmanager.com` ‡
- frame content from
   - `https://www.google.com/recaptcha/` †
   - `https://recaptcha.google.com/recaptcha/` †
   - `'self'` (inherited from default-src, just in case) ‡
   - `*.fec.gov` (inherited from default-src, just in case) ‡
   - `*.app.cloud.gov` (inherited from default-src, just in case) ‡

† new addition, which should work now
‡ existing entry that shouldn't break

## Screenshots

No visible changes

## Related PRs

None

## How to test

1. on not-localhost, test that reCAPTCHA is working
2. if we were using frames anywhere, check that they're still working

### DIFF
--- a/webservices/rest.py
+++ b/webservices/rest.py
@@ -238,7 +238,8 @@ def add_secure_headers(response):
         "default-src": "'self' *.fec.gov *.app.cloud.gov",
         "img-src": "'self' data:",
         "script-src": "'self' https://api.data.gov https://dap.digitalgov.gov https://www.google-analytics.com \
-            https://www.googletagmanager.com 'unsafe-inline'",
+            https://www.googletagmanager.com https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ \
+            'unsafe-inline'",
         "style-src": "'self' https://fonts.googleapis.com https://api.data.gov 'unsafe-inline'",
         "font-src": "'self' https://fonts.gstatic.com data: https://api.data.gov",
         "connect-src": "*.fec.gov *.cloud.gov https://api.data.gov https://www.google-analytics.com",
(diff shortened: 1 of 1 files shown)


cnlucas (approved):
Works for me, thanks @rfultz

rfultz:
Closing in favor of https://github.com/fecgov/openFEC/pull/5655

PR #5660, opened 2023-11-29
Upgrade Webpack v5

### LINKED ISSUE #5325: Upgrade Webpack to 5

tmpayton:
## Summary (required)

- Resolves #5325 

This ticket upgrades Webpack to version 5. I'll include a list of breaking changes in the comments. [Here](https://webpack.js.org/migrate/5/) is a link to some documentation on changes to make when upgrading to webpack 5. 

Shoutout to @jmetev1, thank you for your contribution! 

-  Webpack Configs 

## How to test

1. `git checkout feature/5325-upgrade-webpack-v5`
2. `pyenv activate <your virtual environment>`
3. `rm -rf node_modules`
4. `npm install`
5.  `npm run build`
6.  `flask run` 
7. Compare UI to Prod


tmpayton:
# Changes 

-  Colors is enabled by default 
-  extract-text-webpack-plugin is deprecated need to use mini-css-extract-plugin instead
-  eslint-loader is deprecated need to use eslint-webpack-plugin
- We cannot use arrow functions for components (rule: react/function-component-definition)
   - examples: header.js and footer.js
- With Babel upgrade we cannot mix ES6 modules with Common.js

pkfec (approved):
Works as expected, thanks @tmpayton

PR #5669, opened 2023-12-09
5654- Fix locust test file

### LINKED ISSUE #5654: Locust Tests Update
### What we’re after

We have added quite a bit of validation since our locust tests have been written and need to re-write and add to our locust file. 

### Action item(s)

- [x] Fix all current queries to be valid (Most of our failures are from invalid queries currently) 
- [x] Add recent problematic queries

### Completion criteria

- [x] All queries in locustfile.py are either valid or testing an invalid query
- [ ] We have a new baseline that actually shows load errors

cnlucas:
## Summary (required)

- Resolves #5654 

We had several issues with the old load testing file. Most of our failures were due to 422 errors because we had added validation to many filters. These were not actual load failures and did not give us any information about performance.  I fixed all of the old queries and added new queries for A and B. I went through the API umbrella and researched our current slowest successful queries and queries that sometimes 502/504 and added a few of those as well. 

Our load tests recently have zero rps once we've hit max concurrency because of slow performance, too high wait time, and no timeout. I have added a timeout parameter which you can reset at the top of the file.  I set it to 300 sec as that is the max successful queries have taken via the API umbrella. You can set it lower for testing if you'd like a faster cutoff (Firefox network.http.connection-timeout is default 90 sec.) 

Our wait time was very high, which contributed to the low rps. I set it between 1 and 5 sec to improve rps. We didn't have additional parameters for our filings queries which is our third largest we were testing, so I added filings parameters. I also added small tasks for our C, D, E, F, and H4 endpoints. 

I added weighting of our tasks. This is the number beside the tasks decorator.  I gave a higher weight to our A and B queries and the highest to the smaller A and B's. Happy to adjust in the future if we can get a better, more realistic picture of our current usage.  

I also added logging as it will help us to keep the file updated in the future----much easier to diagnose issues. Also, easy to see if a specific query's result set has drastically change which could shift baseline. 

### Required reviewers

2-3 devs

-  Locust Test File

## How to test
Here's the [locust wiki ](https://docs.google.com/document/d/1ru6Qg0ErTQLD7a86qxW7kbaKbzkayt_X9PmGjZdC_nY/edit?usp=sharing)
- activate your virtual environment 
- (make sure you've run `pip install -r requirements-dev.txt`
- set pingdom maintenance window, make sure you have enough instances for your chosen load
- `locust --host=https://fec-dev-api.app.cloud.gov/v1/ --logfile=locustfile.log`

### DIFF
--- a/locustfile.py
+++ b/locustfile.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 """Load testing for the API and web app. Run from the root directory using the
 
-`locust --host=https://api-stage.open.fec.gov/v1/` (stage)
-`locust --host=https://api.open.fec.gov/v1/` (prod)
-`locust --host=https://fec-dev-api.app.cloud.gov/v1/` (dev)
+`locust --host=https://api-stage.open.fec.gov/v1/ --logfile=locustfile.log` (stage)
+`locust --host=https://api.open.fec.gov/v1/ --logfile=locustfile.log` (prod)
+`locust --host=https://fec-dev-api.app.cloud.gov/v1/ --logfile=locustfile.log` (dev)
 
 command, then open localhost:8089 to run tests.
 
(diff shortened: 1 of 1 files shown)


tmpayton (approved):
Thank you for working on this, you did a really great job!

fec-jli (approved):
It looks great, work as expected. Thanks working on this.

hcaofec (approved):
Great work!

PR #5838, opened 2024-05-20
Increase size for inner hits

### LINKED ISSUE #5817: MUR/ADRs: document_highlights limits to only 3 document hits (Bug)
### What we’re after
Currently the document_highlights only returns a max of 3 hits and a max of 5 fragments. We need to return all the hits so that the user can get the full dataset. We need to understand how we can remove the limit for hits from document_highlights.

Elasticsearch highlight documentation: https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html

**Example:**
When searching query “Reason to believe” and doc_category_id=1 for MUR 4530, it returns 3 matching document hits, but there should be more, at least 7. This is the API call: https://api.open.fec.gov/v1/legal/search?case_no=4530&case_doc_category_id=1&type=murs&hits_returned=20&from_hit=0&q=%22Reason+to+believe%22&api_key=NICAR16

### Action item(s)
- [x] Determine how these limits are being set
- [x] Remove all document limits, as appropriate
- [x] Ensure all relevant documents will appear, no documents with a given type will be excluded from the search results

### Completion criteria
- [x] All relevant documents are appearing in search results

- [x] Is there a performance issue when returning all results?
- [x] Do size parameters in elasticsearch settings help with returning results?

tmpayton:
## Summary (required)

- Resolves #5817 

This ticket increases the inner hit size so that it can return up to 100 hits. If we want more than 100 hits we'd have to change the [max_inner_result_window](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html) setting in elasticsearch. This is not recommended because it would have negative impacts on performance because inner_hits can be costly.

[Here](https://docs.google.com/spreadsheets/d/11UqvbnAJPxt9iPVFY9QdDPEXbGlf9MANhV4NjOjALXg/edit#gid=0) is a document showing the load testing results. I see a bit of an increase in the average response time. 

- legal search

## How to test

1. deploy this branch to dev
2. check that document highlights return more than three results: https://fec-dev-api.app.cloud.gov/v1/legal/search?case_no=4530&case_doc_category_id=1&type=murs&q=%22Reason+to+believe%22&api_key=DEMO_KEY
3. Go to legal search on dev.fec.gov and compare results to prod
example: https://dev.fec.gov/data/legal/search/murs/?search_type=murs&sort=&search=election&case_no=&case_respondents=&case_min_open_date=&case_max_open_date=&case_min_close_date=&case_max_close_date=

### DIFF
--- a/locustfile.py
+++ b/locustfile.py
@@ -787,6 +787,12 @@ class Tasks(TaskSet):
             )
         log_response(self, "ao202312", params, resp)
 
+    @task
+    def get_mur_inner_hits(self):
+        params = {"q": "reason", "type": "murs", "case_no": "4530", "case_doc_category_id": "1", "api_key": API_KEY}
+        resp = self.client.get("legal/search", name="get_mur_inner_hits", params=params, timeout=timeout)
+        log_response(self, "mur_inner_hits", params, resp)
+
 
 class Swarm(user.HttpUser):
     tasks = [Tasks]
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -35,6 +35,7 @@ INNER_HITS = {
         "require_field_match": False,
         "fields": {"documents.text": {}, "documents.description": {}},
     },
+    "size": 100,
 }
 
 ALL_DOCUMENT_TYPES = [


cnlucas (approved):
Great job!

pkfec (approved):
Thank you for debugging and solving the inner_hits issue quickly.  Adding the new locust test helps identify the response time especially with murs. Well done @tmpayton

Issue #5831, opened 2024-05-14
Research what fields document_highlights are searching


patphongs:
### What we’re after

Keyword search is searching more than just documents. In the [code](https://github.com/fecgov/openFEC/blob/ee4d36e38e4db605ebeeb075a0b1ad204dbba1d2/webservices/resources/legal.py#L166), it is searching for: Highlight searched: "text", "name", "no", "summary", "documents.text", "documents.description". We need to find out what each of these things are so we can better explain to the user.

Maybe we don't need to include no? Are all of the fields relevant? 

### Action item(s)

(These are the smaller tasks that should happen in order to complete this work)

- [x] List out and define all of the searched fields of keyword search

### Completion criteria

(What does the end state look like - as long as this task(s) is done, this work is complete)

- [x] We have the dictionary above

tmpayton:
Here is a list of definitions for all of the searched fields. Some of the definitions differ based on document type. 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


Document type | Text | Name | No | Summary | Documents.text | Documents.description
-- | -- | -- | -- | -- | -- | --
Statutes | Full text of a   section of a statute | Section subheading | Section number of   a statute | N/A | N/A | N/A
MUR | N/A | Primary respondent | MUR number | N/A | Full text from   documents related to a matter under review | Title of the   related document
ADR | N/A | Primary respondent | ADR number | N/A | Full text from   documents related to an alternative dispute resolution | Title of the   related document
AF | N/A | Committee_name | AF number | N/A | Full text from   documents related to an admin fine case | Title of the   related document
Archived MUR | N/A | N/A | MUR number | N/A | Full text of the   archived MUR | N/A
AO | N/A | Requestor | AO number | A short summary of   the advisory opinion subject | Full text from   documents related to the advisory opinion | Title of the   related document



</div>


</body>

</html>

[...]

PR #5886, opened 2024-06-28
[WIP] Feature/switch to simple query string

### LINKED ISSUE #5830: [Research] /legal/search endpoint: Respondents need capability for exact string match
### What we’re after

Currently the `respondents[ ]` field searches any word that matches. We want to add the ability to add exact string match in quotes for respondent as well. This is so results can be reduced, especially for common respondent names.

Ex: The search term "Salazar for Congress" currently returns many results because there are common respondent names with `for congress` in it. Limiting it to exactly that string would help narrow down the results.

**Action items:**
- [x] MURs
- [x] ADRs
- [ ] Create future work ticket to consider how this can be applied for admin fines
 This is not a hard change to make, but we need PM/design approval to search by name for AF's

https://api.open.fec.gov/v1/legal/search?case_respondents=%22Salazar+for+Congress%22&type=murs&hits_returned=20&from_hit=0&q=&api_key=NICAR16

### Completion criteria

- [x] `respondents` field has the capability for exact string match

cnlucas:
Draft PR adding the case_respondents query_string and then also switching everything over to simple_query_string as it doesn't seem like ES recommends using query_string in search boxes in production. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html

Questions for Monday: 
When a filter is added--the current behavior is to include all documents where that field is not present. Is that what we want? 
Because highlight and document highlight are attached to query, this will show highlights for any searched term. Is this what we want? 
If we switch to simple, we lose the ability to specify field in the search. Is this something we need? 
Would this improve downtime? 
[API umbrella 400's](https://api.data.gov/admin/#/stats/logs?end_at=2024-06-28&query=%7B%22condition%22%3A%22AND%22%2C%22rules%22%3A%5B%7B%22id%22%3A%22request_path%22%2C%22field%22%3A%22request_path%22%2C%22type%22%3A%22string%22%2C%22input%22%3A%22text%22%2C%22operator%22%3A%22contains%22%2C%22value%22%3A%22legal%22%7D%2C%7B%22id%22%3A%22response_status%22%2C%22field%22%3A%22response_status%22%2C%22type%22%3A%22integer%22%2C%22input%22%3A%22number%22%2C%22operator%22%3A%22equal%22%2C%22value%22%3A400%7D%5D%2C%22valid%22%3Atrue%7D&start_at=2024-01-01)
Legal search[ downtime comparison ](https://docs.google.com/document/d/1a4BjykL5v0tmnZzCOi9z1PKeeOprua4aPGfT5JQpfZU/edit)
Would the syntax changes be hard to deal with in the FE? AND/OR to +/|? 
      notices? 
Need to improve testing
if we stick with query_string are the 400's currently being handled correctly? 

Should we allow wildcard? 
 analyze_wildcard
    (Optional, Boolean) If true, the query attempts to analyze wildcard terms in the query string. Defaults to false. 

We should keep this in mind if there's issues with standard:
https://www.elastic.co/guide/en/elasticsearch/reference/current/mixing-exact-search-with-stemming.html

## Summary (required)

- Resolves #5830

Switches from query_string to simple_query_string for our keyword search and switches case_respondents to simple_query_string from match

### Required reviewers

2-3 devs

- legal search

## How to test

- You need the docs loaded for testing locally. Pat and I have been testing on dev with the standard analyzer. 
- activate your virtualenv
- pytest

http://127.0.0.1:5000/v1/legal/search/
http://127.0.0.1:5000/v1/legal/search/?case_respondents=jackson
http://127.0.0.1:5000/v1/legal/search/?case_respondents=%22Aaron%20Woolf%20for%20Congress%22
http://127.0.0.1:5000/v1/legal/search/?case_respondents=%22Friends%20of%20Jack%20Kingston%22

http://127.0.0.1:5000/v1/legal/search/?case_respondents=%22Casey%20|%20Michael%22
http://127.0.0.1:5000/v1/legal/search/?case_respondents=Fratto%20|%20Elissa

### DIFF
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -205,6 +205,9 @@ def case_query_builder(q, type_, from_hit, hits_returned, **kwargs):
         must_clauses = [
             Q("terms", documents__category=kwargs.get("case_document_category"))
         ]
+
+    if kwargs.get("case_respondents"):
+        must_clauses.append(Q("query_string", query=kwargs.get("case_respondents"), fields=["respondents"]))
     query = query.query("bool", must=must_clauses)
 
     logger.debug("case_query_builder =" + json.dumps(query.to_dict(), indent=3, cls=DateTimeEncoder))
@@ -307,8 +310,6 @@ def apply_mur_specific_query_params(query, **kwargs):
 
     if kwargs.get("mur_type"):
         must_clauses.append(Q("match", mur_type=kwargs.get("mur_type")))
-    if kwargs.get("case_respondents"):
-        must_clauses.append(Q("match", respondents=kwargs.get("case_respondents")))
     if kwargs.get("case_dispositions"):
         must_clauses.append(
             Q("term", disposition__data__disposition=kwargs.get("case_dispositions"))
@@ -443,8 +444,6 @@ def apply_adr_specific_query_params(query, **kwargs):
 
     if kwargs.get("mur_type"):
         must_clauses.append(Q("match", mur_type=kwargs.get("mur_type")))
-    if kwargs.get("case_respondents"):
-        must_clauses.append(Q("match", respondents=kwargs.get("case_respondents")))
     if kwargs.get("case_dispositions"):
         must_clauses.append(
             Q("term", disposition__data__disposition=kwargs.get("case_dispositions"))
--- a/webservices/resources/legal.py
+++ b/webservices/resources/legal.py
@@ -157,7 +157,7 @@ def generic_query_builder(q, type_, from_hit, hits_returned, **kwargs):
     must_query = [Q("term", type=type_)]
 
     if q:
-        must_query.append(Q("query_string", query=q))
+        must_query.append(Q("simple_query_string", query=q))
 
     query = (
         Search()
@@ -193,7 +193,7 @@ def case_query_builder(q, type_, from_hit, hits_returned, **kwargs):
 
     should_query = [
         get_case_document_query(q, **kwargs),
-        Q("query_string", query=q, fields=["no", "name"]),
+        Q("simple_query_string", query=q, fields=["no", "name"]),
     ]
     query = query.query("bool", should=should_query, minimum_should_match=1)
 
@@ -207,7 +207,7 @@ def case_query_builder(q, type_, from_hit, hits_returned, **kwargs):
         ]
 
     if kwargs.get("case_respondents"):
-        must_clauses.append(Q("query_string", query=kwargs.get("case_respondents"), fields=["respondents"]))
+        must_clauses.append(Q("simple_query_string", query=kwargs.get("case_respondents"), fields=["respondents"]))
     query = query.query("bool", must=must_clauses)
 
     logger.debug("case_query_builder =" + json.dumps(query.to_dict(), indent=3, cls=DateTimeEncoder))
@@ -253,7 +253,7 @@ def get_case_document_query(q, **kwargs):
     combined_query.append(Q("bool", should=category_queries, minimum_should_match=1))
 
     if q:
-        combined_query.append(Q("query_string", query=q, fields=["documents.text"]))
+        combined_query.append(Q("simple_query_string", query=q, fields=["documents.text"]))
 
     return Q(
         "nested",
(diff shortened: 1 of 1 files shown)


patphongs:
@cnlucas I answered as many questions as I could before we meet on Monday, thanks again for all of your poignant research:

> When a filter is added--the current behavior is to include all documents where that field is not present. Is that what we want?

Since this is the current default behavior, I wouldn't refactor this just yet. For example, on the front end, a normal user won't be able to filter by respondents generically anyways, they would type into the corresponding respondents field within their specific datasets. In this case, only [MURs](https://www.fec.gov/data/legal/search/murs/) and [ADRs](https://www.fec.gov/data/legal/search/adrs/) have respondents fields to use. Therefore, we would always pass `type=murs` or `type=adrs`, and the user would only see the applicable dataset returned for respondents.

> Because highlight and document highlight are attached to query, this will show highlights for any searched term. Is this what we want?

This is also the current default behavior. I'm actually ok with the fact that the document highlights will highlight more fields than just document text, see research done by @tmpayton [here](https://github.com/fecgov/openFEC/issues/5831#issuecomment-2139637289). I do have a [word of caution](https://github.com/fecgov/openFEC/issues/5831#issuecomment-2176501927) that if we were to remove any of the other things from the highlights, we would need to ensure there are other ways to search those fields such as AO summary, statutes subheading, etc.

> If we switch to simple, we lose the ability to specify field in the search. Is this something we need?

We did not fully understand what the universal keyword search (`q`) was searching before, so your question actually brings into light something that I didn't even know existed. That you can actually search for a specific field like `q=name:republican` to look within the name field specifically. Therefore, I think having this removed by switching over to simple query string is ok since we didn't even know this was a possibility to begin with. If we truly wanted to search those specific fields, we would build a specific separate field for that capability.

> Would the syntax changes be hard to deal with in the FE? AND/OR to +/- |?

The API syntax changes for booleans should be ok, we can do a transform so that when the front end passes the query string to the API, we can change `AND` to `+`, `OR` to `|`. I'm already doing something similar for the eregs API, where the `OR` syntax is `|` as well, so I transform it in the call to the API, here's an [example in the code](https://github.com/fecgov/fec-cms/blob/1cf33dcb34863fb7556b6b354fdb9857782f1081/fec/legal/views.py#L160-L181). This way, the users won't have to change their normal behavior. But the trick is to know what the syntax transformation will look like when building out the query. As long as it's not too complex, the simple replacements should work.

patphongs:
After doing some initial testing, I found an observation about case respondents when using simple query string exact match:

In the case of finding respondents where the `respondents` dictionary is built using `last_name, first_name middle_initial.`, it maybe challenging to find an exact string match with this syntax `case_respondents="karen grays"`. This will yield no results.

In order to find this match, you would need to utilize an `OR` statement such as this `case_respondents="grays%20|%20karen"`. Note that the syntax is last_name | first_name. 

Conversely, the reverse syntax will yield no results `case_respondents="karen%20|%20grays"`

This may be a difficult concept for users to understand.

pkfec:
The `query_string` is strict on syntax, and errors in the input are NOT ignored. For example, when searched for "agency resources\"   query_string throws an server error because of the parsing issues. 

Example url:  https://www.fec.gov/data/legal/search/murs/?search_type=murs&sort=&search=%22agency+resources%5C%22&case_no=&case_respondents=&case_min_open_date=&case_max_open_date=&case_min_close_date=&case_max_close_date=

`simple_query_string` query ignores any syntax error in the input criteria

Example url: https://dev.fec.gov/data/legal/search/murs/?search_type=murs&sort=&search=%22agency+resources%5C%22&case_no=&case_respondents=&case_min_open_date=&case_max_open_date=&case_min_close_date=&case_max_close_date=

cnlucas:
@patphongs I added minimum_should_match=2 which would mean if a user tried case_respondents=Karen Grays they would get only the MURs and ADRs with both Karen AND Grays. Unfortunately, this is still not perfect because sometimes there's a ton of respondents and maybe one is charles grays and another is karen smith. Then the user would be best served by using the exact string "Grays, Karen"

If this isn't sufficient we can consider: 
1. the old re-indexing solution---(I'm looking into seeing if any committee names have commas to see if we could use that as a delimiter. 
3. adding/switching to primary respondent search using name

cnlucas:
We don't need to add minimum_should_match=2. The bug with + was just due to URL encoding. Karen + Grays should work with proper encoding of + to %2B. 
http://127.0.0.1:5000/v1/legal/search/?case_respondents=Karen%2BGrays

cnlucas:
Closing as these changes are included in another PR

PR #5933, opened 2024-08-07
5931-upgrade node

### LINKED ISSUE #5931: Upgrade Node
### What we’re after

We should upgrade Node for API to be consistent with CMS which is now 
"node": "22.1.0",
"npm": "10.8.1"

### Related ticket(s)

- [ ]

### Action item(s)

- [ ]

### Completion criteria

- [ ] Node is upgraded to v. 22 (consistent with CMS)

cnlucas:
## Summary (required)

- Resolves #5931 

Upgrades node to match CMS

### Required reviewers

1 dev

- node modules

## How to test

- activate your venv
- nvm install 22.1.0
-npm i -g npm@10.7.0
- rm -rf node_modules
- npm i && npm run build
- flask run
- pytest


pkfec (commented):
@cnlucas Should we also consider upgrading the node version in circleci config?
https://github.com/fecgov/openFEC/blob/develop/.circleci/config.yml#L57-L59

pkfec (approved):
Thanks for keeping API node packages consistent with other apps.

fec-jli (approved):
It looks good. Thanks

PR #6083, opened 2024-12-17
6070-Fix pytest test_efile issue

### LINKED ISSUE #6070: Fix flyway migration file
### What we’re after

Migration file v0302 has a create table statement that is missing `IF NOT EXISTS`. Developers have to drop and recreate their local test database every time we run `pytest` or they will get failures. We need to figure out a way to fix this issue without editing the original migration file. 

Most likely will need to delete this table and recreate it using a different name. Will need to pair with FECfile to make sure they are ok with the tablename changes and because this would involve editing some of their code. 
### Action item(s)

- [ ] Determine a way to fix the migrations without editing v0302

### Completion criteria

- [ ] Developers can run `pytest` without having to drop the database first

fec-jli:
## Summary (required)
This PR fix issue that backend develop have to drop and recreate their local test database every time run pytest.

Note: 
When we add new schema, remember to add schema name into reset_schema() of tests/conftest.py

- Resolves #6070 

### Required reviewers
1 dev

pytest

## How to test
1) checkout branch
2) run `pytest` twice (don't need drop cfdm_unit_test)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,6 +42,7 @@ def reset_schema():
         "real_pfile",
         "rohan",
         "staging",
+        "test_efile",
     ]:
         rest.db.engine.execute('drop schema if exists %s cascade;' % schema)
     rest.db.engine.execute('create schema public;')


pkfec (approved):
Thanks for fixing this long lasting annoying bug while running the pytest. Works great @fec-jli

PR #6519, opened 2026-03-03
add has_eligible_documents

### LINKED ISSUE #6516: Handle rulemakings open for comment with no eligible documents
## What we're after

We need to implement a graceful way to handle rulemakings that are marked as open for comment (`is_open_for_comment = true`) but have no documents with `is_comment_eligible = true`. Currently, this creates an inconsistent state where users are told a rulemaking is open for comments, but there are no eligible documents to actually comment on.

The `is_open_for_comment` flag exists at the rulemaking level for performance reasons, to avoid looping through all documents just to check eligibility. However, this optimization can lead to scenarios where this rulemaking-level flag doesn't accurately reflect the document-level reality.

## Action item(s)

- [ ] Determine the business logic: Should we:
  - Automatically set `is_open_for_comment = false` when all documents have `is_comment_eligible = false`?
- [ ] Implement chosen solution in the codebase
- [ ] Add database constraints or model validation to prevent this state in the future
- [ ] Add unit tests to verify the fix prevents this scenario

## Completion criteria

- [ ] No rulemakings exist where `is_open_for_comment = true` AND all associated documents have `is_comment_eligible = false`
- [ ] The "Open for comments" column accurately reflects whether users can actually submit comments

### Testing examples and screenshot
The screenshot shows at least two rulemakings in this problematic state:
- **REG 2026-26** "test rulemaking round 2" - Shows "No comment eligible document specified"
- **REG 2025-99** "Hearing Date Known" - Shows "No comment eligible document specified"

#### Rulemakings table showing inconsistent comment eligibility states

Image

cnlucas:
## Summary (required)

- Resolves #6516

Adds a check for eligible documents and makes that the is_open_for_comment flag and then removing from schema/mapping current is_open_for_comment
Also removing  from schema is_doc_open_for_comment keep is_comment_eligible as is

### Required reviewers

1-2 devs

- Commenting boolean flags

## How to test

- gh pr checkout 6519
- pytest

- python cli.py delete_index rm_index        
- python cli.py create_index rm_index        
- python cli.py load_rulemaking 2026-01
- python cli.py load_large_rulemaking 2014-01
- flask run
- make sure the flags are showing correctly


pkfec (approved):
Looks good, thanks @cnlucas

PR #6580, opened 2026-04-28
add slow reindex

### LINKED ISSUE #6475: Clean up Opensearch documentation on zero downtime
### What we’re after

The zero downtime deployment documentation should be cleaned up. 

My current understanding is that we don't have memory to support the low downtime options because case_index is so large. We should confirm this and then consider breaking down this index so that we could use a low downtime solution. 

We should make the docs more clear and create tickets to start this work

### Related ticket(s)

- [ ]

### Action item(s)

- [ ] We have fixed/confirmed that the documentation is correct for the zero downtime reindexing
- [ ] We have determined how much extra memory this would require
- [ ] Create follow up tickets for breaking out indexes if we still can't support zero/low downtime solutions

### Completion criteria

- [ ] Research has been done on the above questions 

cnlucas:
## Summary (required)

- Resolves #6475 

Adds slow zero downtime option for reindexing large document

### Required reviewers

2-3 devs

## How to test

Rebuild this on dev:
https://app.circleci.com/pipelines/github/fecgov/openFEC/4215/workflows/1d8ad081-4bc3-4713-a380-c19b27a8e0d7/jobs/7301

Run this task on dev:
cf run-task api --command "python cli.py display_index_alias" -m 2G --name display_index_alias 

cf tasks api
Run slow reload of each legal type, make sure that the task is finished before running the next task
cf run-task api --command "python cli.py slow_reload_zero_downtime case_index" -m 4G --name slow_reload_zero_downtime_case

cf run-task api --command "python cli.py slow_reload_zero_downtime ao_index" -m 4G --name initialize_legal_data_ao

cf run-task api --command "python cli.py slow_reload_zero_downtime arch_mur_index" -m 4G --name initialize_legal_data_arch_mur

cf run-task api --command "python cli.py slow_reload_zero_downtime rm_index" -m 4G --name init_rm_data

cf run-task api --command "python cli.py display_index_alias" -m 2G --name display_index_alias 

Take a snapshot, run an intialize...
https://github.com/fecgov/openFEC/wiki/Opensearch-2.11.x-management-instructions

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -136,6 +136,12 @@ def initialize_legal_data_cli(index_name):
     load_utils.initialize_legal_data(index_name)
 
 
+@cli.command('slow_reload_zero_downtime')
+@click.argument('index_name', default=None, required=False)
+def slow_reload_zero_downtime_cli(index_name):
+    load_utils.slow_reload_zero_downtime(index_name)
+
+
 @cli.command('update_mapping_and_reload_legal_data')
 @click.argument('index_name', default=None, required=False)
 def update_mapping_and_reload_legal_data_cli(index_name):
--- a/webservices/legal/rulemaking_docs/rulemaking.py
+++ b/webservices/legal/rulemaking_docs/rulemaking.py
@@ -406,7 +406,7 @@ def load_large_rulemaking(specific_rm_no):
         doc.pop('text', None)
 
     # Index parent rulemaking
-    opensearch_client.index(index=constants.RM_ALIAS, body=rm_parent, id=rm["rm_id"])
+    opensearch_client.index(index=constants.RM_ALIAS, body=rm_parent, id=rm["rm_id"], routing=rm["rm_id"])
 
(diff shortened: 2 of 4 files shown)


tasks.py
```diff
@@ -143,7 +143,7 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    ('dev', lambda _, branch: branch == 'develop'),
+    ('dev', lambda _, branch: branch == 'feature/6475-add-zero-downtime'),
```
pkfec: @cnlucas Can you revert back to the develop branch?

cnlucas: Thanks for catching, fixed

tmpayton (approved):
I tested by re-indexing case_index on dev and everything worked really well, fantastic job @cnlucas!

PR #6647, opened 2026-06-29
change cloudfoundry package source

patphongs:
Update cf package source from `cli.run.pivotal.io` to `packages.cloudfoundry.org`


cnlucas (approved):
LGTM thanks!

PR #6650, opened 2026-07-01
increase disk quota

cnlucas:
increase disk quota


PaulClark2 (approved):
Looks good. Changing from 1G to 2G.

patphongs (approved):
👍

PR #6653, opened 2026-07-08
add statement_timeout to follower engine

cnlucas:
Adds back statement timeouts now that we have the slack alert turned off. 

## How to test

- pull this PR
- `pytest`
- Make sure you have SQLA_FOLLOWERS set to dev, stage, or your local db.
- set an artificially low statement timeout locally
`export SQLA_STATEMENT_TIMEOUT=1`

- `flask run`
- Run any query and you should see a 504 error, you can try a 404 and 422 as well:
504: http://127.0.0.1:5000/v1/committees/?page=1&per_page=20&filing_frequency=M&sort=name&sort_hide_null=false&sort_null_only=false&sort_nulls_last=false
404: http://127.0.0.1:5000/v1/committes/?page=1&per_page=20&filing_frequency=M&sort=name&sort_hide_null=false&sort_null_only=false&sort_nulls_last=false
422: http://127.0.0.1:5000/v1/committees/?page=1&per_page=20&filing_frequency=ZZ&sort=name&sort_hide_null=false&sort_null_only=false&sort_nulls_last=false

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -34,7 +34,7 @@ class TestBaseTestCase(BaseTestCase):
         assert response.headers["Access-Control-Allow-Origin"] == "*"
 
 
-class StatementTimeoutTest(unittest.TestCase):
+class TestStatementTimeout(unittest.TestCase):
     """Test statement_timeout configuration should only apply to reader"""
     def test_follower_engines_have_statement_timeout(self):
         with patch.dict('os.environ', {'SQLA_STATEMENT_TIMEOUT': '30000'}):
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -1,6 +1,11 @@
 from flask import current_app
 from tests.common import BaseTestCase
 from sqlalchemy import text
+from sqlalchemy.exc import OperationalError
+import unittest
+from unittest.mock import patch
+from webservices.rest import create_app
+from webservices.common.models import db
 
 
 class TestBaseTestCase(BaseTestCase):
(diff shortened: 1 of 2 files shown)


webservices/rest.py
```diff
@@ -117,25 +118,24 @@ def create_app(test_config=None):
         'webservices.tasks.download.export_query',]
     app.config['PROPAGATE_EXCEPTIONS'] = True
     query_cache_size = int(env.get_credential('QUERY_CACHE_SIZE', '100'))
-    pool_pre_ping = bool(env.get_credential('POOL_PRE_PING_BOOL', 'False'))
+    pool_pre_ping = bool(env.get_credential('POOL_PRE_PING_BOOL', False))
```
patphongs: We need to be careful here. I think it would be good to use the existing explicit `TRUE_VALUES` pattern to parse this. Otherwise, if the env value is the string `"False"`, Python treats it as `True`.

webservices/rest.py
```diff
@@ -486,6 +486,13 @@ def add_secure_headers(response):
             response.headers.add(header, value)
         return response
 
+    @app.errorhandler(OperationalError)
+    def handle_db_timeout(e):
+        if hasattr(e, 'orig') and 'canceling statement due to statement timeout' in str(e.orig):
+            app.logger.warning('Statement timeout on %s', request.path)
+            return jsonify({'message': 'Query timed out', 'status': 504}), 504
```
patphongs, line 493: This is new, it adds a new 504 response path. I don't think I see this in the tests anywhere, could you please add?

patphongs (changes requested):
@cnlucas I have a few request changes:

tests/test_app.py
```diff
@@ -27,3 +32,64 @@ def test_cors_headers(self):
         response = self.client.get('/swagger')
         assert "Access-Control-Allow-Origin" in response.headers
         assert response.headers["Access-Control-Allow-Origin"] == "*"
+
+
+class StatementTimeoutTest(unittest.TestCase):
```
patphongs: `StatementTimeoutTest` should be renamed because pytest won't use it otherwise. The default is `Test*` class naming convention like this `TestStatementTimeout`.

patphongs (approved):
@cnlucas Good to go ✈️

